PR #20, opened 2022-02-16
Grid class

rajeeja:
- Grid init, file type parser and some utility functions including placeholders.
- Includes first pass to exodus reader and writer
- Infra for other readers/writers
- open_database function with tests
- meshfiles for testing


uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+    from ._scrip import read_scrip
+
+    def __init__(self, *args, **kwargs):
+        """Initialize grid variables, decide if loading happens via file, verts
+        or gridspec If loading from file, initialization happens via the
+        specified file.
+
+        Raises:
+            RuntimeError: File not found
+        """
```
anissa111, line 38: What are the args for this function? The possible variables below should probably be included as kwargs if they aren't always applicable.

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+                print("File not found: ", args[0])
+                exit()
+            elif in_type is np.ndarray:
+                self.vertices = args[0]
+                self.from_vert()
+        except ValueError as e:
+            # initialize from vertices
+            if in_type is np.ndarray:
+                self.vertices = args[0]
+                self.from_vert()
```
anissa111, line 68: Should be able to rearrange with use of kwargs. Also, If an exception does need to be raised I suggest using:

```
raise FileNotFoundError("File not found: " + args[0])
```

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # initialize from file
+        if in_type is str and os.path.isfile(args[0]):
+            self.filepath = args[0]
+            self.from_file()
```
anissa111, line 73: could probably be combined with the above block.

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+
+        else:
+            # this may just be initialization for options other than above
+            pass
+
+    # vertices init
+    def from_vert(self):
+        """Create a grid with one face with vertices specified by the given
+        argument."""
+        print("Initializing with vertices")
```
anissa111, line 88:
```suggestion
```

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+        if self.meshFileType == "exo":
+            self.read_exodus(self.ext_ds)
+        elif self.meshFileType == "scrip":
+            self.read_scrip(self.ext_ds)
+        elif self.meshFileType == "ugrid":
+            self.read_ugrid(self.filepath)
+        elif self.meshFileType == "shp":
+            self.read_shpfile(self.filepath)
+        else:
+            raise RuntimeError("unknown file format.")
```
anissa111, line 140:
```suggestion
            raise RuntimeError("unknown file format:" + self.meshFileType)
```

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+            file_extension = path.suffix
+
+            # try to open file with xarray
+            self.ext_ds = xr.open_dataset(self.filepath, mask_and_scale=False)
+            #
+        except (TypeError, AttributeError) as e:
+            msg = str(e) + ': {}'.format(self.filepath)
+            print(msg)
+            raise RuntimeError(msg)
+            exit
```
anissa111, line 164: I'm not sure this is proper implementation of python try catch blocks with raise exceptions. At the very least the exit call at the bottom of the blocks should be a function call.

```suggestion
        try:
            # extract the file name and extension
            path = PurePath(self.filepath)
            file_extension = path.suffix

            # try to open file with xarray
            self.ext_ds = xr.open_dataset(self.filepath, mask_and_scale=False)
            #
        except (TypeError, AttributeError) as e:
            msg = str(e) + ': {}'.format(self.filepath)
            print(msg)
            raise RuntimeError(msg)
            exit()
```

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+                raise RuntimeError(msg)
+                exit
+
+        # Detect mesh file type, based on attributes of ext_ds:
+        # if ext_ds has coordx or coord - call it exo format
+        # if ext_ds has grid_size - call it SCRIP format
+        # if ext_ds has ? read as shape file populate_scrip_dataformat
+        # TODO: add detection of shpfile etc.
+        try:
+            self.ext_ds.coordx
```
anissa111, line 192: Statement has no effect, is this supposed to be a function call?

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+            self.meshFileType = "exo"
+        except AttributeError as e:
+            pass
+        try:
+            self.ext_ds.grid_center_lon
```
anissa111, line 197: Same as above

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+            self.meshFileType = "scrip"
+        except AttributeError as e:
+            pass
+        try:
+            self.ext_ds.coord
```
anissa111, line 202: same as above

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+            self.meshFileType = "exo"
+        except AttributeError as e:
+            pass
+        try:
+            self.ext_ds.Mesh2
```
anissa111, line 207: same as above

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+                "node_coordinates": "Mesh2_node_x Mesh2_node_y Mesh2_node_z",
+                "node_dimension": "nMesh2_node",
+                "face_node_connectivity": "Mesh2_face_nodes",
+                "face_dimension": "nMesh2_face"
+            })
+
+    # renames the grid file
+    def saveas_file(self, filename):
+        path = PurePath(self.filepath)
+        old_filename = path.name
```
anissa111, line 236:
```suggestion
```

variable never used

uxarray/dataset.py
```diff
@@ -0,0 +1,26 @@
...
+    import uxarray as ux
+    # open an exodus file with Grid object
+    mesh = ux.open_dataset("filename.g")
+    """
+    print("opening dataset: ", filename)
+    if type == "exo" or "ugrid" or "scrip" or "shp":
+        ux_grid = Grid(str(filename))
+    else:
+        print("Meshfile type is not supported")
+    return ux_grid.in_ds
```
anissa111, line 26: Should probably raise an error, like this `ux_grid.in_ds` can be returned before it is defined

uxarray/dataset.py
```diff
@@ -0,0 +1,26 @@
...
+def open_dataset(filename, type):
+    """A class for uxarray dataset object returns an xarray object with Mesh2
+    construct representing the grid.
+
+    Examples
+    ========
+    import uxarray as ux
+    # open an exodus file with Grid object
+    mesh = ux.open_dataset("filename.g")
+    """
```
anissa111, line 20: Docstring should include args and returns

anissa111 (changes requested):
The comments below are my preliminary review. I anticipate I'll have more comments once some simple things can be addressed. My main concern right now is how variable assignment is handled, which I think can lead to a lot of bugs in the current implementation. 

I also think we should prioritize adding expected inputs and outputs to the doc strings as we go for clarity for everybody reading the source code.

uxarray/grid.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # internal uxarray representation of mesh stored in internal object in_ds
+        self.in_ds = xr.Dataset()
+
+        # determine initialization type - string signifies a file, numpy array signifies a list of verts
+        in_type = type(args[0])
+
+        # check if this is a valid file:
+        try:
+            if os.path.isfile(args[0]) == False and in_type is not np.ndarray:
```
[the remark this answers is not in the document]
rajeeja: Just fixed.

uxarray/grid.py
rajeeja: This is developing, and there are not that many variables now.
I'll put a placeholder with a couple of parameters to guide it.

uxarray/grid.py
rajeeja: if "coordx" is not an available attribute, we get an exception.

uxarray/grid.py
rajeeja: some python versions raise ValueError, so this block is better safety.

uxarray/grid.py
rajeeja: same, this just checks for "grid_center_lon" presence in ext_ds

uxarray/_exodus.py
```diff
@@ -0,0 +1,224 @@
...
+# Exodus Number is one-based.
+def read_exodus(self, ext_ds):
+    print("Reading exodus file: ", self.filepath)
+    # populate self.in_ds
+
+    self._init_mesh2()
+
+    for key, value in ext_ds.variables.items():
+        if key == "qa_records":
+            # print(value)
```
michaelavs, line 16:
```suggestion

```

uxarray/dataset.py
```diff
@@ -0,0 +1,29 @@
+"""uxarray dataset module."""
+
+from logging import raiseExceptions
```
michaelavs, line 3: From line 3 through 7, the imported packages are unused. Should they be removed? Just wanted to double check that there isn't a technical reason for them to be here

uxarray/grid.py
```diff
@@ -0,0 +1,278 @@
+"""uxarray grid module."""
+import math
```
michaelavs, line 2: Same as above, but lines 2, 5, and 6

uxarray/grid.py
```diff
@@ -0,0 +1,278 @@
...
+        in_type = type(args[0])
+
+        # check if this is a valid file:
+        try:
+            if os.path.isfile(args[0]) is False and in_type is not np.ndarray:
+                raise FileNotFoundError("File not found: " + args[0])
+            elif in_type is np.ndarray:
+                self.vertices = args[0]
+                self.from_vert()
+        except ValueError as e:
```
michaelavs, line 72: Should this except block be 

```
except ValueError as e:

      if in_type is np.ndarray:

            ........
      
      else:
            
            print(e)
```
I ask because I think as it is written now, there wouldn't be a way to handle if the "if" statement fails but additionally, e isn't called in the block. So adding the elif, we could handle a False for the if statement while also using the e variable to return the ValueError in that case.

ci/environment.yml
```diff
@@ -7,5 +7,8 @@ dependencies:
   - pytest
   - pytest-cov
   - xarray
+  - netcdf4
+  - scipy
```
michaelavs, line 11:
```suggestion

```
package isn't used from what I can see

michaelavs (commented):
Just a couple of little comments/edits, but overall this is looking good!

erogluorhan (commented):
Thanks a lot for putting together this initial significant contribution! I made a first-pass review on it and am planning to add further comments (on grid.py, dataset.py etc.) tomorrow or Friday. (Hey @rajeeja there are a lot of review comments in this PR as this is the very first implementation of ours; if addressing all of them is too much load on your end, I am happy to help out with any level; e.g. writing unit tests etc, please let me know.)

Please find my initial comments below as well as inline through the changed files:

- We have good code coverage overall; however, we seem to lack important parts of the code not covered yet. For example, please see the blocks that are not covered in grid.py here: [gridp.py codecov report](https://app.codecov.io/gh/UXARRAY/uxarray/blob/b490aa3609fd4f9fa4b4f7e593b4a14a56d6340e/uxarray/grid.py). 
- Please be sure to activate your branch at readthedocs.org to get it built, so we can see how the [API reference](https://uxarray.readthedocs.io/en/latest/api.html) page is being populated with your implementations. 
- It looks like we need to thoroughly check our function, class, and variable names to make them comply with PEP8 standards (Examples to corrections needed for each of these types are given in my inline comments). We may consider utilizing a PEP8 pre-commit package to achieve this.

uxarray/grid.py
erogluorhan: I'd suggest putting a #todo comment for this type of developing parts and explaining what are to be discussed/improved in the future. That would give us the opportunity to comprehensively know our parts of the code to be developed as well as  the rest of the team  to give the same thing some thinking.

erogluorhan: >I also think we should prioritize adding expected inputs and outputs to the doc strings as we go for clarity for everybody reading the source code.
I second this. We should populate the doc strings along with implementation. For example, "latlon", "gridspec" that are already used in the implementation need to be described/explained in the doc strings.

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
+"""uxarray grid module."""
+import os
+import xarray as xr
+import numpy as np
+from pathlib import PurePath
+
+
+class Grid:
+    """A class for uxarray grid object.
```
erogluorhan, line 9: Maybe something like "The Uxarray Grid object class that describes an unstructured grid"

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+    ========
+    import uxarray as ux
+    # open an exodus file with Grid object
+    mesh = ux.Grid("filename.g")
+
+    # save as ugrid file
+    mesh.saveas("outfile.ug")
+    """
+
+    # Import read/write methods from files in this folder
```
erogluorhan, line 21: "file" sounds misleading with dataset files; use "modules" to refer to Python script files.

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+            RuntimeError: File not found
+        """
+        # initialize possible variables
+        for key, value in kwargs.items():
+            if key == "latlon":
+                self.islatlon = value
+            elif key == "gridspec":
+                self.gridspec = value
+
+        self.filepath = None
```
erogluorhan, line 45: To increase readability for the instance attributes for Grid class, I'd suggest moving these attributions to the very beginning of the `__init__` functions.

uxarray/grid.py
erogluorhan: This block of all the try-catch'es (and the others throughout this file) may be changed with if conditions to improve readability:

```
if self.ext_ds.coordx is not None or self.ext_ds.coord is not None:
   self.meshFileType = "exo"
elif self.ext_ds.grid_center_lon is not None:
   self.meshFileType = "scrip"
elif self.ext_ds.Mesh2 is not None:
   self.meshFileType = "ugrid"
else:
   print("mesh file not supported")
```

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+    # load mesh from a file
+    def from_file(self):
+        """Loads a mesh file Also, called by __init__ routine This routine will
+        automatically detect if it is a UGrid, SCRIP, Exodus, or shape file.
+
+        Raises:
+            RuntimeError: Invalid file type
+        """
+        # find the file type
+        self.meshFileType = self.find_type()
```
erogluorhan, line 136: Is it intentional that even though `self.find_type()` returns `self.meshFileType`, we are still making an assignment here?

ci/environment.yml
```diff
@@ -7,5 +7,7 @@ dependencies:
   - pytest
   - pytest-cov
   - xarray
+  - netcdf4
```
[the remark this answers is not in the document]
michaelavs: I think alphabetical is a good method and if we decide to use it now, it's easier to implement later on as well if/when we add more package dependencies

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+            self.read_scrip(self.ext_ds)
+        elif self.meshFileType == "ugrid":
+            self.read_ugrid(self.filepath)
+        elif self.meshFileType == "shp":
+            self.read_shpfile(self.filepath)
+        else:
+            raise RuntimeError("unknown file format:" + self.meshFileType)
+
+    # helper function to find file type
+    def find_type(self):
```
erogluorhan, line 151: I believe this helper function is intended for internal use only. If that is correct, please name it with a leading single underscore as [suggested by PEP8](https://pep8.org/#:~:text=internal%20interfaces%20(packages%2C%20modules%2C%20classes%2C%20functions%2C%20attributes%20or%20other%20names)%20should%20still%20be%20prefixed%20with%20a%20single%20leading%20underscore). This is also applicable to any other internal API function in this PR, if any.

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+    mesh.saveas("outfile.ug")
+    """
+
+    # Import read/write methods from files in this folder
+    from ._exodus import read_exodus, write_exodus
+    from ._ugrid import read_ugrid, write_ugrid
+    from ._shpfile import read_shpfile
+    from ._scrip import read_scrip
+
+    def __init__(self, *args, **kwargs):
```
erogluorhan, line 27: From the whole implementation of this function, I think that taking arguments as `**kwargs` only (i.e. not `*args` in addition) could help us simplify the flow. For instance, if "input_file" in kwargs, then initialize from a file, else if "vertices" in kwargs, then initialize from vertices, etc.

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # internal uxarray representation of mesh stored in internal object in_ds
+        self.in_ds = xr.Dataset()
+
+        # determine initialization type - string signifies a file, numpy array signifies a list of verts
+        in_type = type(args[0])
+
+        # check if initializing from verts:
+        try:
+            if os.path.isfile(args[0]) is False and in_type is not np.ndarray:
```
erogluorhan, line 63: `if not os.path.isfile(args[0])` instead of `if os.path.isfile(args[0]) is False`

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+        # this xarray variable holds an existing external netcdf file with loaded with xarray
+        self.ext_ds = None
+
+        # internal uxarray representation of mesh stored in internal object in_ds
+        self.in_ds = xr.Dataset()
+
+        # determine initialization type - string signifies a file, numpy array signifies a list of verts
+        in_type = type(args[0])
+
+        # check if initializing from verts:
```
erogluorhan, line 61: My personal thought is that following how this code determines if to initialize from verts, gridspec, etc. is not very easy, i.e. the code does not seem easily readable. We may consider simplifying it (maybe my recommendation of using only kwags helps here) as well as putting more explanation comments either here or into the function doc strings.

uxarray/grid.py
```diff
@@ -0,0 +1,279 @@
...
+        automatically detect if it is a UGrid, SCRIP, Exodus, or shape file.
+
+        Raises:
+            RuntimeError: Invalid file type
+        """
+        # find the file type
+        self.meshFileType = self.find_type()
+
+        # call reader as per meshFileType
+        if self.meshFileType == "exo":
```
erogluorhan, line 139: Variable naming with mixed cases and combined words (e.g. meshFileType) is not recommended in [PEP8](https://www.python.org/dev/peps/pep-0008/#:~:text=module%20non%2Dpublic%22).-,Function%20and%20Variable%20Names,Variable%20names%20follow%20the%20same%20convention%20as%20function%20names.,-mixedCase%20is%20allowed). Furthermore, I think we need to follow [PEP8 instance variable naming conventions](https://www.python.org/dev/peps/pep-0008/#:~:text=Use%20one%20leading%20underscore%20only%20for%20non%2Dpublic%20methods%20and%20instance%20variables.) for this variable as it is an instance variable.

uxarray/grid.py
rajeeja: I would like to have this option via args:

`mesh = ux.Grid("filename.g")`

There is logic to automatically identify several of the parameters specified in kwargs. How about we remove kwargs?

rljacob:
Yes all of our python should follow PEP8.

uxarray/grid.py
erogluorhan: Yes, that makes to keep `args` after seeing the Grid constructor you gave. Not sure about removing `kwargs` as it may make the implementation more difficult. I think after you address some of the inline comments in this same function, we may have a better flow of the function.

uxarray/grid.py
rajeeja: self.ext_ds.coordx doesn't work. Had to to "coordx" in self.ext_ds.data_vars

rajeeja:
> * We have good code coverage overall; however, we seem to lack important parts of the code not covered yet. For example, please see the blocks that are not covered in grid.py here: [gridp.py codecov report](https://app.codecov.io/gh/UXARRAY/uxarray/blob/b490aa3609fd4f9fa4b4f7e593b4a14a56d6340e/uxarray/grid.py).

Only the placeholders are causing the low code-coverage percentage.

uxarray/grid.py
rajeeja: fixed, thanks for pointing this out.

ci/environment.yml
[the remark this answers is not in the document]
rajeeja: Done. Alphabetical.

test/test_dataset.py
```diff
@@ -0,0 +1,27 @@
+import os
+import xarray as xr
```
erogluorhan, line 2: `xr` import not being used in this test

uxarray/grid.py
erogluorhan: Thanks for the commit to address this, but I think the above linked PEP8 doc suggests that the internal API functions use this convention instead: `self._find_type()` (i.e. prefixed with a single leading underscore). I might be mistaken though.

rajeeja:
> Thanks a lot for putting together this initial significant contribution! I made a first-pass review on it and am planning to add further comments (on grid.py, dataset.py etc.) tomorrow or Friday. (Hey @rajeeja there are a lot of review comments in this PR as this is the very first implementation of ours; if addressing all of them is too much load on your end, I am happy to help out with any level; e.g. writing unit tests etc, please let me know.)
> 
> Please find my initial comments below as well as inline through the changed files:
> 
> * We have good code coverage overall; however, we seem to lack important parts of the code not covered yet. For example, please see the blocks that are not covered in grid.py here: [gridp.py codecov report](https://app.codecov.io/gh/UXARRAY/uxarray/blob/b490aa3609fd4f9fa4b4f7e593b4a14a56d6340e/uxarray/grid.py).
> * Please be sure to activate your branch at readthedocs.org to get it built, so we can see how the [API reference](https://uxarray.readthedocs.io/en/latest/api.html) page is being populated with your implementations.
> * It looks like we need to thoroughly check our function, class, and variable names to make them comply with PEP8 standards (Examples to corrections needed for each of these types are given in my inline comments). We may consider utilizing a PEP8 pre-commit package to achieve this.

Feel free to modify the PR

erogluorhan:
@rajeeja Are you finished with your commits and will you ask for a re-review cycle?

rajeeja:
My commits are done for this initial PR

uxarray/grid.py
pilotchute: I would suggest using optionals set to `None` by default, rather than a `**kwargs` expecting a dictionary. 
it would be more 'pythonic' than the old 'c style' argments, and has the advantage of being cleaner to implement and far easier to maintain and document.

uxarray/_exodus.py
```diff
@@ -0,0 +1,223 @@
...
+    self.exo_ds = xr.Dataset()
+
+    path = PurePath(outfile)
+    out_filename = path.name
+
+    now = datetime.now()
+    date = now.strftime("%m/%d/%Y")
+    time = now.strftime("%H:%M:%S")
+
+    title = f"uxarray(" + str(out_filename) + ")" + date + ": " + time
```
pilotchute, line 112: Suggest using ISO 8601 time format: YYYY:MM:DDTHH:MM:SS as in my experience this allows for better file convenience and sorting when interacting with large numbers of files.

uxarray/_exodus.py
```diff
@@ -0,0 +1,223 @@
...
+    fp_word = np.int32(8)
+    version = np.float32(5.0)
+    api_version = np.float32(5.0)
```
pilotchute, line 115: Versioning information is usually pulled from a __version__.py file in the package, or some other packagewide source, am I missing the reason why this is hardcoded midway though a file?

uxarray/_exodus.py
```diff
@@ -0,0 +1,223 @@
...
+        "version": version,
+        "floating_point_word_size": fp_word,
+        "file_size": 0,
+        "title": title
+    }
+
+    self.exo_ds["time_whole"] = xr.DataArray(data=[], dims=["time_step"])
+
+    # qa_records
+    qa_records = [["uxarray"], ["1.0"], [date], [time]]
```
pilotchute, line 127: Curious about the hardcoded 1.0 value here.

uxarray/_exodus.py
```diff
@@ -0,0 +1,223 @@
...
+    elif num_nodes == 4:
+        element_type = "SHELL4"
+    elif num_nodes == 8:
+        element_type = "SHELL8"
+    elif num_nodes == 9:
+        element_type = "SHELL9"
+    elif num_nodes == 6:
+        element_type = "TRI6"
+    elif num_nodes == 7:
+        element_type = "TRI7"
```
pilotchute, line 221: please sort the elif tree by num_nodes.
there is no `elif num_nodes == 5:` which is a possible case.

uxarray/grid.py
```diff
@@ -0,0 +1,278 @@
...
+            file_extension = path.suffix
+
+            # try to open file with xarray
+            self.ext_ds = xr.open_dataset(self.filepath, mask_and_scale=False)
+            #
+        except (TypeError, AttributeError) as e:
+            msg = str(e) + ': {}'.format(self.filepath)
+            print(msg)
+            raise RuntimeError(msg)
+            exit()
```
pilotchute, line 180: This set of lines is duplicated three times, and could be moved to a `finally:` block.

uxarray/grid.py
```diff
@@ -0,0 +1,278 @@
...
+        except ValueError as e:
+            # check if this is a shp file
+            # we won't use xarray to load that file
+            if file_extension == ".shp":
+                self.mesh_filetype = "shp"
+            else:
+                msg = str(e) + ': {}'.format(self.filepath)
+                print(msg)
+                raise RuntimeError(msg)
+                exit
```
pilotchute, line 200: this needs to be `exit()`, and see above about handling this behavior in a `finally:` block

uxarray/grid.py
```diff
@@ -0,0 +1,278 @@
...
+    def build_edge_face_connectivity(self):
+        pass
+
+    # Build the array of latitude-longitude bounding boxes.
+    def buildlatlon_bounds(self):
+        pass
+
+    # Validate that the grid conforms to the UXGrid standards.
+    def validate(self):
+        pass
```
pilotchute, line 266: add exceptions to state that the behavior is not yet implemented, instead of quietly `pass`ing.

pilotchute (changes requested):

uxarray/_exodus.py
rajeeja: Good suggestions, will push several changes to this file soon. Thanks

uxarray/_exodus.py
dcherian: could consider making this a dict
``` python
ELEMENT_TYPE_DICT = {
    2: "BEAM",
    3: "TRI",
...
}
...
element_type = ELEMENT_TYPE_DICT[num_nodes]
````

**Comment by erogluorhan** [comment]
Hi @rajeeja , just an FYI: I got the RTD builds for PR autmotacillay triggering now within build checks. See below:

image

**Comment by rajeeja** [comment]
I'll work more on clean up and documentation and re-request a review soon

File: test/test_grid.py [hunk 38]
```diff
@@ -0,0 +1,85 @@
...
+from pathlib import Path
+
+import uxarray as ux
+
+
+class TestGrid(TestCase):
+
+    def test_saveas(self):
+        """Rename a file."""
+        current_path = Path(os.path.dirname(os.path.realpath(__file__)))
```
michaelavs, line 15: For the "current_path" variable definitions here and the one in test_dataset, I believe we can just move them outside of the testing functions and reduce to just one at the top of the file. I did this in my PR for test_scrip and it worked, but I haven't had the chance to try in your PR. If you want me to (and there isn't another reason not to do so), I'd gladly test and implement this change on my local of the branch and push the changes!

rajeeja: Sure, please test and push. Thanks!

uxarray/grid.py
```diff
@@ -0,0 +1,236 @@
...
+            # initialize from vertices
+            if in_type is np.ndarray:
+                self.vertices = args[0]
+                self.__from_vert__()
+            else:
+                raise RuntimeError(
+                    "Init called with args other than verts (ndarray) or filename (str)"
+                )
+        except TypeError as e:
+            raise RuntimeError("sting or np.ndarray supported as input")
```
anissa111, line 97: I'm still not 100% sure this is the best way to go about this. 

@erogluorhan @rajeeja Thoughts?

erogluorhan: @anissa111 I feel like it could be improved, too. Could you propose a different solution for comparison purposes?

erogluorhan (commented):
Thanks a lot for all the improvements! I will provide my review soon.

uxarray/grid.py
anissa111: @erogluorhan I'll look in to this!

anissa111:
Should we also expand the testing done in this PR or create a secondary PR for that purpose?

uxarray/grid.py
```diff
@@ -0,0 +1,244 @@
...
+        or gridspec If loading from file, initialization happens via the
+        specified file.
+
+        # TODO: Add or remove new Args/kwargs below as this develops further
+
+        Parameters
+        ----------
+        args : string, optional
+            - Input file name with extension or
+            - Vertex coordinates that form one face.
```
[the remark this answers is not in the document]
anissa111: Nevermind, I've addressed this.

michaelavs:
> Should we also expand the testing done in this PR or create a secondary PR for that purpose?

I am on board for expanding testing within this PR. There isn't a specific test for just the exodus function yet (which I think it could be argued that if current tests work, then _exodus.py is working), but I feel there should be more edge case testing for _exodus.py and grid.py overall. For me, I think when we have other file types added there will be overlaps with variable naming, or users may specify outfiles with unique suffixes (e.g. ".nc", "ugrid", no suffix, etc) and that may start to confuse the Grid class, so while we could wait to see on the tests, it may be better to just address things now instead of later. But, I can also, again, see arguments against addressing these things in this specific PR/at this point in the development process. So, if these aren't necessarily shared concerns to be addressed immediately, I'm fine with expanding tests in a new PR after this one is merged.

uxarray/grid.py
anissa111: @erogluorhan @rajeeja I have replaced this block with a more concise version. Let me know what your thoughts are! 

Just a note, I removed the gridspec initialization and replaced it with a todo for now since it doesn't appear to be implemented yet.

uxarray/grid.py
```diff
@@ -0,0 +1,244 @@
...
+        self.in_ds = xr.Dataset()
+
+        # Parse the keyword arguments and initialize class attributes
+        for key, value in kwargs.items():
+            if key == "latlon":
+                self.islatlon = value
+            elif key == "gridspec":
+                self.gridspec = value
+
+        # determine initialization type - string signifies a file, numpy array signifies a list of verts
```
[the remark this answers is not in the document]
rajeeja: > > Should we also expand the testing done in this PR or create a secondary PR for that purpose?
> 
> I am on board for expanding testing within this PR. There isn't a specific test for just the exodus function yet (which I think it could be argued that if current tests work, then _exodus.py is working), but I feel there should be more edge case testing for _exodus.py and grid.py overall. For me, I think when we have other file types added there will be overlaps with variable naming, or users may specify outfiles with unique suffixes (e.g. ".nc", "ugrid", no suffix, etc) and that may start to confuse the Grid class, so while we could wait to see on the tests, it may be better to just address things now instead of later. But, I can also, again, see arguments against addressing these things in this specific PR/at this point in the development process. So, if these aren't necessarily shared concerns to be addressed immediately, I'm fine with expanding tests in a new PR after this one is merged.

I'm envisioning test_grid.py contain all reader/writer tests. It'd be easier and more manageable if we have test_exodus.py and so on, is that in-accordance with the conventions? @erogluorhan

uxarray/_exodus.py
```diff
@@ -0,0 +1,336 @@
...
+                                      dims=["num_el_blk"],
+                                      attrs={"name": "ID"})
+    # eb_status
+    exo_ds["eb_status"] = xr.DataArray(data=xr.DataArray(
+        np.ones([num_blks], dtype="i4")),
+                                       dims=["num_el_blk"])
+
+    # eb_names
+    eb_names = np.empty(num_blks, dtype="S33")
+    eb_names.fill("")
```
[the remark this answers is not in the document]
rajeeja: Thanks for pointing this out. The other exodus files had a string of that length and initialized to "", but just using 'str' is cleaner. Fixed

uxarray/_ugrid.py
```diff
@@ -0,0 +1,32 @@
...
+    """UGRID file reader.
+
+    Parameters: string, required
+        Name of file to be read
+
+    Returns: the xarray Dataset loaded during init.
+    """
+
+    ext_ds = xr.open_dataset(filepath, mask_and_scale=False)
+    in_ds = xr.Dataset()
```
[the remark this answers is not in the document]
rajeeja: Removed. The C/C++ programmer still sometimes like to define the type for a variable :)

uxarray/grid.py
rajeeja: Thanks @anissa111 your changes improve the readability of the code.

test/meshfiles/README.md
```diff
@@ -0,0 +1,16 @@
...
+* Shape:
+  * grid_fire.shp
+* Exodus:
+  * outCSne8.g
+* SCRIP:
+  * outCSne30_vortex.nc
+  * outCSne8.nc
+  * outRLL1deg_vortex.nc
+  * ov_RLL10deg_CSne4_vortex.nc
+* UGIRD:
```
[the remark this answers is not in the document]
rajeeja: Ouch, fixed.

uxarray/helpers.py
```diff
@@ -0,0 +1,70 @@
...
+        # check if this is a shp file
+        # we won't use xarray to load that file
+        if file_extension == ".shp":
+            mesh_filetype = "shp"
+        else:
+            msg = str(e) + ': {}'.format(filepath)
+    finally:
+        if msg != "":
+            msg = str(e) + ': {}'.format(filepath)
+            print(msg)
```
erogluorhan, line 67: Should we remove this print as we already have an exception raised?

uxarray/grid.py
```diff
@@ -0,0 +1,217 @@
...
+    def __init__(self, *args, **kwargs):
+        """Initialize grid variables, decide if loading happens via file, verts
+        or gridspec If loading from file, initialization happens via the
+        specified file.
+
+        # TODO: Add or remove new Args/kwargs below as this develops further
+
+        Parameters
+        ----------
+        data_arg : string, ndarray, list, tuple, required
```
erogluorhan, line 39: Please put a line-space between doc string sections and arguments as it increases the readability.

docs/user_api/index.rst
```diff
@@ -3,19 +3,33 @@
...
+   _ugrid.write_ugrid
+   _ugrid.read_ugrid
 
+Helper functions
+----------------
 .. autosummary::
-   :toctree: ./generated/
+   :toctree: _autosummary
+
+   helpers.determine_file_type
```
erogluorhan, line 35: I think we intended to provide this helper function to be callable through `uxarray.grid.determine_file_type`, right? If so, I believe the documentation should show it through the same path to instruct the user correct. Can we please check if that's possible?

erogluorhan (commented):
I like the new code structure much; thanks a lot! We are getting fast to merge it soon. Please see my inline comments!

erogluorhan:
> > > Should we also expand the testing done in this PR or create a secondary PR for that purpose?
> > 
> > 
> > Unit testing should go along with the implementation.
> 
> Ok, I'll add another test for testing kwargs. Is it okay to create `test_exodus`? Will move some specific exodus tests to that file.

Yes

uxarray/_exodus.py
```diff
@@ -0,0 +1,340 @@
...
+                np.int32(
+                    0)  # NOTE: This might cause an error if numbering has holes
+        })
+    print("Finished reading exodus file.")
+    # done reading exodus file, close the external xarray ds object
+    ext_ds.close()
+    return in_ds
+
+
+def write_exodus(in_ds, outfile):
```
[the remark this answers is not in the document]
rajeeja: `write` and `saveas` are the same functions. Is there a different functionality that we are aiming for with these two functions?

Merge them into one - say keep only - `saveas`?

uxarray/grid.py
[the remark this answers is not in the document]
erogluorhan: Yeah, I see what you mean, and I can also be fine with that! Even though the convention is that we should unit-test each Python module in a separate test file dedicated to it, I don't have a strong feeling when it comes to such a case where grid.py module uses helper functions from other "_..." modules.

uxarray/grid.py
```diff
@@ -0,0 +1,244 @@
...
+            raise RuntimeError("sting or np.ndarray supported as input")
+
+        # check if initialize from file:
+        if in_type is str and os.path.isfile(args[0]):
+            self.filepath = args[0]
+            self.__from_file__()
+        # initialize for gridspec
+        elif in_type is str:
+            self.gridspec = args[0]
+            self.__from_gridspec__()
```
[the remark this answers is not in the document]
erogluorhan: Seems done.

uxarray/grid.py
```diff
@@ -0,0 +1,236 @@
...
+                "long_name": "Topology data of unstructured mesh",
+                "topology_dimension": -1,
+                "node_coordinates": "Mesh2_node_x Mesh2_node_y Mesh2_node_z",
+                "node_dimension": "nMesh2_node",
+                "face_node_connectivity": "Mesh2_face_nodes",
+                "face_dimension": "nMesh2_face"
+            })
+        self.in_ds.Mesh2.attrs['topology_dimension'] = self.vertices[0].size
+
+        self.in_ds["Mesh2"].topology_dimension
```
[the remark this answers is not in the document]
erogluorhan: Seems done.

uxarray/_exodus.py
```diff
@@ -0,0 +1,336 @@
...
+        if "num_nod_per_el" in dim:
+            if ext_ds.dims[dim] > max_face_nodes:
+                max_face_nodes = ext_ds.dims[dim]
+
+    # create an empty conn array for storing all blk face_nodes_data
+    conn = np.empty((0, max_face_nodes))
+
+    for key, value in ext_ds.variables.items():
+        if key == "qa_records":
+            pass
```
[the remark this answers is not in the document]
erogluorhan: Instead of silent passes, please put at least a warning, if not exception, in addition to a TODO comment.

uxarray/grid.py
erogluorhan: Thanks for doing this!

uxarray/_exodus.py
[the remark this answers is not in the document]
erogluorhan: The draft API listed `write` only; there was no `saveas`. I thought you were planning something different than write with saveas. Sticking to the API, we could remove `saveas` (as the word "write" is counterpart of "read" in which we have another function), or I am also fine with vice versa.

docs/user_api/index.rst
rajeeja: Grid doesn't have to be loaded to `determine_file_type`, so I made it accessible directly via uxarray as `uxarray.determine_file_type`

uxarray/_exodus.py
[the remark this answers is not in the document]
rajeeja: Sounds good. I will remove `saveas`.
My initial idea was a simple renaming function with `saveas`, but that is not required and confusing.

docs/user_api/index.rst
erogluorhan: Sounds good!

erogluorhan:
Hey @rajeeja , I'd like to approve this once the RTD build failure is fixed.

rajeeja:
> Hey @rajeeja , I'd like to approve this once the RTD build failure is fixed.

It seems like `dask` or something is causing the error: `https://readthedocs.org/projects/uxarray/builds/16389261/`
`Build exited due to excessive memory consumption` - we need the module installation to be quick.

Works locally for me.

rajeeja:
> Hey @rajeeja , I'd like to approve this once the RTD build failure is fixed.

It seems like `dask` or something is causing the error: `https://readthedocs.org/projects/uxarray/builds/16389261/`
`Build exited due to excessive memory consumption` - we need the module installation to ge quick.

Works locally for me.

erogluorhan:
> > Hey @rajeeja , I'd like to approve this once the RTD build failure is fixed.
> 
> It seems like `dask` or something is causing the error: `https://readthedocs.org/projects/uxarray/builds/16389261/` `Build exited due to excessive memory consumption` - we need the module installation to be quick.
> 
> Works locally for me.

When I attempted to install Dask after you added it as a dependency, I got some errors with the conda, and I ended up removing and creating my conda env from scratch. There might be something related to RTD. I am not sure for the time.

rajeeja:
Do you think dask should be lower down the order in the yml file?

erogluorhan:
> Do you think dask should be lower down the order in the yml file?

I wouldn't expect the order to be an issue as the conda package manager should be solving the entire environment requirements/dependecies when the yml fils is started to execute. However, it might be worth giving a try. There was a similar, not the same, issue on conda repo [here](https://github.com/conda/conda/issues/4957).

rajeeja:
> > Do you think dask should be lower down the order in the yml file?
> 
> I wouldn't expect the order to be an issue as the conda package manager should be solving the entire environment requirements/dependecies when the yml fils is started to execute. However, it might be worth giving a try. There was a similar, not the same, issue on conda repo [here](https://github.com/conda/conda/issues/4957).

You are right, order change did not help. This [ project](https://readthedocs.org/projects/paprika/builds/16076821/) had similar issue, they have sphinx and other  dependencies. We don't need sphinx in environment.yml? I wan't able to get the fix for their issue. They added `icu58` but later removed it: [link](https://github.com/slochower/pAPRika/blame/master/docs/environment.yml)

rajeeja:
> You are right, order change did not help. This [ project](https://readthedocs.org/projects/paprika/builds/16076821/) had similar issue, they have sphinx and other dependencies. We don't need sphinx in environment.yml? I wan't able to get the fix for their issue. They added `icu58` but later removed it: [link](https://github.com/slochower/pAPRika/blame/master/docs/environment.yml)

The issue is described [here](https://github.com/dask/dask/issues/8227)
@michaelavs do you think we can add `docs/requirements-docs.txt `as suggested in the link on the above line?

rajeeja:
> > You are right, order change did not help. This [ project](https://readthedocs.org/projects/paprika/builds/16076821/) had similar issue, they have sphinx and other dependencies. We don't need sphinx in environment.yml? I wan't able to get the fix for their issue. They added `icu58` but later removed it: [link](https://github.com/slochower/pAPRika/blame/master/docs/environment.yml)
> 
> The issue is described [here](https://github.com/dask/dask/issues/8227) @michaelavs do you think we can add `docs/requirements-docs.txt `as suggested in the link on the above line?

What is the use of top level requirements.txt file? - would adding dask to the list solve the RTD issue above?

```

git show d83d9c
commit d83d9c06aad4797fb699c57a511eab9f362084f5
Author: Michaela Sizemore <michaela.sizemore@gmail.com>
Date:   Fri Oct 29 14:12:00 2021 -0600

    update requirements.txt

diff --git a/requirements.txt b/requirements.txt
index eede985..1457852 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1 @@
-cf_xarray>=0.3.1
-cftime
-dask[array]
-eofs
-metpy
-numpy
 xarray

```

erogluorhan:
> > > You are right, order change did not help. This [ project](https://readthedocs.org/projects/paprika/builds/16076821/) had similar issue, they have sphinx and other dependencies. We don't need sphinx in environment.yml? I wan't able to get the fix for their issue. They added `icu58` but later removed it: [link](https://github.com/slochower/pAPRika/blame/master/docs/environment.yml)
> > 
> > 
> > The issue is described [here](https://github.com/dask/dask/issues/8227) @michaelavs do you think we can add `docs/requirements-docs.txt `as suggested in the link on the above line?
> 
> What is the use of top level requirements.txt file? - would adding dask to the list solve the RTD issue above?
> 
> ```
> 
> git show d83d9c
> commit d83d9c06aad4797fb699c57a511eab9f362084f5
> Author: Michaela Sizemore <michaela.sizemore@gmail.com>
> Date:   Fri Oct 29 14:12:00 2021 -0600
> 
>     update requirements.txt
> 
> diff --git a/requirements.txt b/requirements.txt
> index eede985..1457852 100644
> --- a/requirements.txt
> +++ b/requirements.txt
> @@ -1,7 +1 @@
> -cf_xarray>=0.3.1
> -cftime
> -dask[array]
> -eofs
> -metpy
> -numpy
>  xarray
> ```

requiremenents/txt is to be used for PyPI releases and it shouldn't be directly related to RTD builds.

> > > You are right, order change did not help. This [ project](https://readthedocs.org/projects/paprika/builds/16076821/) had similar issue, they have sphinx and other dependencies. We don't need sphinx in environment.yml? I wan't able to get the fix for their issue. They added `icu58` but later removed it: [link](https://github.com/slochower/pAPRika/blame/master/docs/environment.yml)
> > 
> > 
> > The issue is described [here](https://github.com/dask/dask/issues/8227) @michaelavs do you think we can add `docs/requirements-docs.txt `as suggested in the link on the above line?
> 
> What is the use of top level requirements.txt file? - would adding dask to the list solve the RTD issue above?
> 
> ```
> 
> git show d83d9c
> commit d83d9c06aad4797fb699c57a511eab9f362084f5
> Author: Michaela Sizemore <michaela.sizemore@gmail.com>
> Date:   Fri Oct 29 14:12:00 2021 -0600
> 
>     update requirements.txt
> 
> diff --git a/requirements.txt b/requirements.txt
> index eede985..1457852 100644
> --- a/requirements.txt
> +++ b/requirements.txt
> @@ -1,7 +1 @@
> -cf_xarray>=0.3.1
> -cftime
> -dask[array]
> -eofs
> -metpy
> -numpy
>  xarray
> ```

sds

> > > You are right, order change did not help. This [ project](https://readthedocs.org/projects/paprika/builds/16076821/) had similar issue, they have sphinx and other dependencies. We don't need sphinx in environment.yml? I wan't able to get the fix for their issue. They added `icu58` but later removed it: [link](https://github.com/slochower/pAPRika/blame/master/docs/environment.yml)
> > 
> > 
> > The issue is described [here](https://github.com/dask/dask/issues/8227) @michaelavs do you think we can add `docs/requirements-docs.txt `as suggested in the link on the above line?
> 
> What is the use of top level requirements.txt file? - would adding dask to the list solve the RTD issue above?
> 
> ```
> 
> git show d83d9c
> commit d83d9c06aad4797fb699c57a511eab9f362084f5
> Author: Michaela Sizemore <michaela.sizemore@gmail.com>
> Date:   Fri Oct 29 14:12:00 2021 -0600
> 
>     update requirements.txt
> 
> diff --git a/requirements.txt b/requirements.txt
> index eede985..1457852 100644
> --- a/requirements.txt
> +++ b/requirements.txt
> @@ -1,7 +1 @@
> -cf_xarray>=0.3.1
> -cftime
> -dask[array]
> -eofs
> -metpy
> -numpy
>  xarray
> ```

It is used for PyPI releases, and I don't think it is of any direct use in RTD.

erogluorhan:
@rajeeja , in our RTD case, our builds started to fail after adding Dask as a dependency; see [the first failing RTD build](https://readthedocs.org/projects/uxarray/builds/16373141/) and [its corresponding commit](https://github.com/UXARRAY/uxarray/commit/ad785e4c3a67e32d603020ef1bc0c2e706457509). 

I tried the Dask-removed version in a different branch (fix_rtd), and it succeeded, see [here](https://readthedocs.org/projects/uxarray/builds/16396691/). 

Though, it is weird, adding Dask shouldn't be an issue (we have it in different repos and their RTD builds are no problem). There might be a conflict of Dask with another dependency here. Let us keep digging, we could solve the issue soon.

erogluorhan:
Okay, @rajeeja , I tried mamba for this in the fix_rtd branch and it seems to fix the issue. Let me commit that here.

docs/internal_api/index.rst
```diff
@@ -1,20 +1,25 @@
...
-
-For now, please check the
-tentative `Uxarray API
+This page shows already-implemented Uxarray internal API functions. You can also
+check the draft `Uxarray API
 <https://github.com/UXARRAY/uxarray/blob/main/docs/user_api/uxarray_api.md>`_
-documentation and let us know if you have any feedback!
+documentation to see the tentative whole API and let us know if you have any feedback!
 
 Routines
```
erogluorhan, line 11: I'd suggest using a titling here similar to user_api/index, something like: 

```
Grid functions
--------------

grid.Grid.__init__
grid.Grid.__from_file__
...

Grid Helper Modules
--------------------

_exodus._read_exodus
...

```

michaelavs (approved):

rajeeja (commented):
looks good.

rajeeja (commented):
split tests. resolving comments.

uxarray/_exodus.py
rajeeja: Added a variable for the same resolving.

anissa111 (approved):
This looks good! Thank you for all the effort you put in!

PR #1, opened 2021-10-26
CI/CD and doc files

michaelavs:
Added CI/CD and RTD/Sphinx documentation files. Many of the doc pages are populated with placeholder information that generally just routes to the landing page of this repository. 

What immediately comes to mind for what still needs to be done:
- tests
- gitignore
- contributing guide
- logos (I believe we will want all relevant logos from all orgs involved)
- citing info

PR #23, opened 2022-03-04
Adding scrip file reader

michaelavs:
Adds scrip file reader to cf compliant dataset function. 

Items I think will need to be addressed before merging:
- Using OOP with function to better integrate into grid.py 
- Checks for consistency with PR #20 (init file location, data file locations, tests, name conventions, etc)
- Any additional edge cases or general updates for the current tests


uxarray/reader/_scrip.py
```diff
@@ -0,0 +1,78 @@
+# Create list of names that can be used for node coords
+_X_NODES = ("lat", "latitude", "Lat", "Latitude", "X", "x", "grid_corner_lat",
+            "grid_corner_latitude")
+
+_Y_NODES = ("lon", "longitude", "Lon", "Longitude", "Y", "y", "grid_corner_lon",
+            "grid_corner_longitude")
+
+
+def viewkeys(d):
```
erogluorhan, line 9: May this be an internal API function (not intended for user calls)?

uxarray/reader/_scrip.py
```diff
@@ -0,0 +1,78 @@
...
+    """Return either the keys or viewkeys method for a dictionary.
+
+    Args:
```
erogluorhan, line 12: For doc strings let us use the following format for consistency with Pangeo community (This applies to other doc strings throughout):

```
Parameters
----------

d : :obj:`dict`
        Dictionary that holds ... .

... other parameters if any...

Returns
--------

func() : 
    Description of func()'s purposes.
```

uxarray/reader/_scrip.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def _get_nodes(ds):
+    """Outter wrapper function for datasets that are not cf compliant.
+
+    :param ds: xarray.Dataset
+    :return: xarray.Dataset with cf compliant naming of lat/lon variables
+    """
+
+    NODE_X = None
```
erogluorhan, line 31: Is it intentional to use uppercase naming for these two variables?

uxarray/reader/_scrip.py
```diff
@@ -0,0 +1,78 @@
...
+        if name in viewkeys(ds):
+            ds['Mesh2_node_y'] = ds[name]
+            NODE_Y = ds['Mesh2_node_y']
+
+            break
+
+    return NODE_X, NODE_Y
+
+
+def populate_scrip_data(ds, is_cf=True):
```
erogluorhan, line 51: I think even this function is not intended for user calls and it will only be used by Grid object, right? If so, please use internal API function naming conventions.

uxarray/reader/_scrip.py
```diff
@@ -0,0 +1,78 @@
...
+
+    - Error message if variable name is not in current dictionary
+    - Further capability to have a user input for variables not in dictionary
+
+    :param ds: xarray.Dataset used
+    :param cf: Bool tells code if dataset is cf compliant or not, default True
+    :return: Reassigns data variables to be mesh2_node_x and mesh2_node_y
+    """
+
+    NODE_X = None
```
erogluorhan, line 63: Similar comment about uppercase letters.

erogluorhan (commented):
Thanks a lot for putting this together! 

Even though it is hard to make a thorough review on this PR before seeing its integration with `grid.py`, here are my initial comments and inline comments in hopes that they could help you in the meantime:

- Current unit test cases test this scrip functionality against xarray (i.e. its atomic functionality), but we will need test cases that test its integrated functionality with our `Grid` object (I think you also point this out in your description but just wanted to make sure)
- Overall, it looks like a nice, lightweight helper module that could be utilized by `grid.py`. When @rajeeja 's PR #20 gets merged, I think we can move forward with this as well.

michaelavs:
@erogluorhan @anissa111 
I'm looking into the reason this PR is failing after switching to `_populate_scrip_data` and in looking at setup.py for this repo, I wanted to double check a couple of items. 
1. Should we change the layout of the repo to be `src/uxarray` instead of having the two separated as we have now? Looking at comp, it would seem that we should have things set up this way, but I'm not sure if there is a big difference overall between the two.
2. With respect to the last point and updating setup.py, would the package directories look like:
`
package_dir={
        '': 'src',
        'uxarray': 'uxarray',}
`
if we didn't put uxarray into src?

3. Should I put any changes to the setup.py file into a separate PR, or is it ok to have it in this one?

erogluorhan:
> 1. Should we change the layout of the repo to be src/uxarray instead of having the two separated as we have now? Looking at comp, it would seem that we should have things set up this way, but I'm not sure if there is a big difference overall between the two.

I don't think we should make such a change. That's because folder structure shouldn't be the actual reason of a namespacing issue, and also most of the tools within Pangeo ecosystem employs a folder structure of "root/package_name" instead of "root/src/package_name". I was even thinking of changing the geocat-comp (and other GeoCAT tools) to this way.

> 2. With respect to the last point and updating setup.py, would the package directories look like:
package_dir={ '': 'src', 'uxarray': 'uxarray',}
if we didn't put uxarray into src?

I didn't meant we should change right away `setup.py` to have such a `package_dir`. I just it wanted to point out the difference between uxarray and geocat-comp. Sorry for any confusion. However, I am also fine changing it only if needed. I'd recommend looking at multiple ways of namespace handling and figuring out what is really happening in this case here.

> 3. Should I put any changes to the setup.py file into a separate PR, or is it ok to have it in this one?

After the above research, if you believe some changes need to be made in setup.py to help your PR, that's fine to have that in this PR. Otherwise, a separate PR also works.

rajeeja (commented):
There is already a _scrip.py in uxarray/_scrip.py
Can you use that instead of the `reader` folder? 
As discussed during the last meeting, we might organize the repo into subfolders at a later stage.

Readers must have a 
_read_scrip method. See line 153 grid.py

Can you also add a SCRIP writer to this PR?

For tests:
See TODO: line 47 of test_grid.py

uxarray/_scrip.py
```diff
@@ -1,11 +1,211 @@
...
+            pass
+
+        # done processing write the file to disk
+        scrip_ds.to_netcdf(outfile)
+
+    return scrip_ds
+
+
+def _is_scrip(in_ds, out_ds):
+    """If in file is an unstructured SCRIP file, function will reassign SCRIP
```
erogluorhan, line 72: No variable named "in file" here.

erogluorhan (changes requested):
Good effort! In addition to the inline comments, please also add your user-API and internal-API functions into the documentation indices. I can give another review after my current comments are addressed.

uxarray/_scrip.py
```diff
@@ -1,11 +1,121 @@
-from warnings import warn
+import xarray as xr
+import numpy as np
 
 
-def _read_scrip(filepath):
-    """This is just a function placeholder.
+def _to_ugrid(in_ds, outfile):
```
erogluorhan, line 5: Arg name "outfile" does not make sense as `_to_ugrid(ext_ds, ds)` on Line 98 gets in a dataset, `ds`, and populates it.

uxarray/_scrip.py
```diff
@@ -1,11 +1,121 @@
...
+        file to be returned by _populate_scrip_data, used as an empty placeholder file
+        to store reassigned SCRIP variables in UGRID conventions
+    """
+
+    if in_ds['grid_area'].all():
+
+        # Create Mesh2_node_x/y variables from grid_corner_lat/lon
+        # Turn latitude scrip array into 1D instead of 2D
+        corner_lat = in_ds['grid_corner_lat'].values
+        corner_lat = corner_lat.flatten()
```
erogluorhan, line 24: These two can be replaced with `xarray.Dataarray.stack`

uxarray/_scrip.py
```diff
@@ -1,11 +1,121 @@
...
+        lon_idx = np.unique(corner_lon, return_index=True)[1]
+        sort_lon = sorted(lon_idx)
+
+        unique_lon = np.zeros(len(sort_lon))
+        for i in range(len(sort_lon) - 1):
+            unique_lon[i] = (corner_lon[sort_lon[i]])
+
+        # Create Mesh2_node_x/y from unsorted, unique grid_corner_lat/lon
+        outfile['Mesh2_node_x'] = unique_lon
+        outfile['Mesh2_node_y'] = unique_lat
```
erogluorhan, line 47: The dimensions of these two do not match in the test case `test_grid.py:test_read_scrip`, which indicates a problem.

uxarray/_scrip.py
```diff
@@ -1,11 +1,121 @@
...
+        corner_lat = corner_lat.flatten()
+
+        # Return only the index of unique values to preserve order
+        lat_idx = np.unique(corner_lat, return_index=True)[1]
+        sort_lat = sorted(lat_idx)
+
+        # Create placeholder array for the sorted unique values
+        unique_lat = np.zeros(len(sort_lat))
+        for i in range(len(sort_lat) - 1):
+            unique_lat[i] = (corner_lat[sort_lat[i]])
```
erogluorhan, line 33: Is there a need for a for-loop, isn't this just doable with: 

`unique_lat  = corner_lat[sort_lat]`

uxarray/_scrip.py
```diff
@@ -1,11 +1,121 @@
...
+        face_arr = []
+        for i in range(len(in_ds['grid_corner_lat'] - 1)):
+            x = in_ds['grid_corner_lon'][i].values
+            y = in_ds['grid_corner_lat'][i].values
+            face = np.hstack([x[:, np.newaxis], y[:, np.newaxis]])
+            face_arr.append(face)
+
+        face_node = np.asarray(face_arr)
+
+        outfile['Mesh2_face_nodes'] = xr.DataArray(
```
erogluorhan, line 63: The dimensionality of this variable does not seem correct. It should be:

`integer Mesh2_face_nodes(nMesh2_face, nMaxMesh2_face_nodes)`

Please see details [here](https://ugrid-conventions.github.io/ugrid-conventions/#:~:text=the%20current%20convention.-,Example%3A,-dimensions%3A%0AnMesh2_node%20%3D%205).

erogluorhan (changes requested):
I have made several comments about the implementation and code. Please address them, then we can move forward with another cycle.

uxarray/_scrip.py
michaelavs: I tried this chunk without the for loop, and was getting a naming error, so have left it in it's current form. However, this comment did remind me that this chunk (and the lon version below) was implemented when I was looking into the writing function. On further consideration, I don't think there is a strong case to actually keep them overall. Originally, I wanted to preserve the ordering of the data points as I thought there was maybe a pattern I could use for the writing function, but that wasn't the case overall. That said, before removing this code completely, I wanted to double check that there wouldn't be other merit to preserving the data point index order instead of switching to the numpy default of re-ordering by data point value.

uxarray/_scrip.py
michaelavs: Would it be better to just use the same variable names from when this is called in `_read_scrip` in this case?

uxarray/_scrip.py
```diff
@@ -1,11 +1,142 @@
...
+                "cf_role":
+                    "face_node_connectivity",
+                "_FillValue":
+                    -1,
+                "start_index":
+                    np.int32(
+                        0
+                    )  # NOTE: This might cause an error if numbering has holes
+            })
+        # face_arr = []
```
michaelavs, line 79: @erogluorhan I left the original code here mainly for discussion and references purposes. I'm having a little bit of difficulty understanding how to determine the values for the `nMesh2_face_nodes` part of the `Mesh2_face_nodes` variable. From the UGRID website ([for 2D flexible mesh](https://ugrid-conventions.github.io/ugrid-conventions/#:~:text=the%20current%20convention.-,Example%3A,-dimensions%3A%0AnMesh2_node%20%3D%205)) there isn't much to go off of. When looking at `outCSne30.ug`, I also can't seem to place how the data points being selected and placed for this array. Just a quick print out of each so you can see what I am seeing:
Mesh2_node_x:
`array([315.      ,  45.      , 135.      , ..., 141.09969 , 138.033171,
       135.      ])`

Mesh2_node_y:
`array([-35.26439 , -35.26439 , -35.26439 , ...,  40.837604,  39.549078,
        38.143314])`

Mesh2_face_nodes:
`array([[   0.,    8.,  356.,  124.],
       [   8.,    9.,  357.,  356.],
       [   9.,   10.,  358.,  357.],
       ...,
       [5399., 5400.,  299.,  300.],
       [5400., 5401.,  298.,  299.],
       [5401.,  297.,    6.,  298.]])`

It isn't the first two indexes of each row that are giving me much pause, but rather the last two (which should be coming from Mesh2_node_x and Mesh2_node_y) that are, as they don't follow a strict order that would match with their inputs like the first two do.

erogluorhan: I am assuming this is a comment to my comment above that starts with "The dimensionality of this variable does not ...". 

The UGRID website ([for 2D flexible mesh](https://ugrid-conventions.github.io/ugrid-conventions/#:~:text=the%20current%20convention.-,Example%3A,-dimensions%3A%0AnMesh2_node%20%3D%205)) says that "The attribute `face_node_connectivity` points to an index variable identifying for every face the indices of its corner nodes." So, in our case, we are naming that index variable as "Mesh2_face_nodes", which is also the example name given in the same website (Though, if it was a UGRID file reading task, we would first. look at the attribute `face_node_connectivity` and it would give us the exact name of the variable that is stored in that netCDF file). 

From the statement, we understand that `Mesh2_face_nodes` should identify for every face the indices of its corner nodes, i.e. it should have a size of `nMesh2_face` * `nMaxMesh2_face_nodes`. That's to say, we should have as many rows as the number of faces, and in each row, we should have the index of nodes that are on that specific face. More clearly from your example:

Mesh2_face_nodes:
`array([[ 0., 8., 356., 124.], [ 8., 9., 357., 356.], [ 9., 10., 358., 357.], ..., [5399., 5400., 299., 300.], [5400., 5401., 298., 299.], [5401., 297., 6., 298.]])`

That means the face #0 has the nodes with index numbers 0, 8, 356, 124 as its corners; the face #1 has the nodes with index numbers 8, 9, 357, 356 in it, etc. 

A small side-note here, even though your provided `Mesh2_face_nodes` array values look like float (e.g. "356.", "124.", all numbers with a period at the end), if you check its data type, it is int32 instead, because it only holds node indices. 

In your SCRIP case, SCRIP files do not keep track of indices for nodes and faces. So, you should be generating and keeping tracking of indices  for your faces and nodes when you are creating a UGRID data structures from a SCRIP file.

I hope this helps.

erogluorhan (commented):
Provided answers to your comments/questions.

docs/internal_api/index.rst
```diff
@@ -31,3 +31,4 @@ Grid Helper Modules
    _exodus._get_element_type
    _ugrid._write_ugrid
    _ugrid._read_ugrid
+   _scrip._read_scrip
```
erogluorhan, line 34: Add also `_to_ugrid`

uxarray/_scrip.py
```diff
@@ -1,11 +1,123 @@
...
-    warn("Function not implemented yet. FYI, attempted to read SCRIP file: " +
-         str(filepath))
-    # TODO: create ds
+
+    if in_ds['grid_area'].all():
+
+        # Create Mesh2_node_x/y variables from grid_corner_lat/lon
+        # Turn latitude scrip array into 1D instead of 2D
+        corner_lat_xr = in_ds['grid_corner_lat']
+        corner_lat = corner_lat_xr.values.ravel()
```
erogluorhan, line 24: Redundant memory usage through the initialization of `corner_lat_xr` (and similarly `corner_lon_xr` below) looks like avoidable. Those variables are only actually used for their `shape`s, which is for example (384,4) in this dataset, and it could be read through some other variables coming from SCRIP datasets. If you can do that, you can get rid of `corner_lat_xr` and `corner_lon_xr` and initialize `corner_lat ` and `corner_lat` as:

`corner_lat = in_ds['grid_corner_lat'].values.ravel()`

erogluorhan (commented):
This looks good overall; please address a few inline comments though.

uxarray/_scrip.py
michaelavs: @erogluorhan I used the built in PyCharm profiler on the tests for the function and it appears I reduced the memory usage with the most recent push/changes. Please feel free to double check those stats on your end too!

erogluorhan (approved):

rajeeja (approved):

rajeeja:
@michaelavs Good work. Please check and click resolve conversations to be able to merge this.

PR #33, opened 2022-04-12
Integration

rajeeja:
o Use xr.register_dataset_accessor for "integrate function"
o Face area calculation
o Quadrature rules
o Tests
o Some documentation will improve on it.


uxarray/integrate.py
```diff
@@ -0,0 +1,49 @@
+"""xarray accessor functions for integration with uxarray."""
+import xarray as xr
+import uxarray as ux
+from .helpers import calculate_face_area
+
+
+@xr.register_dataset_accessor('integrate')
```
dcherian, line 7: This works but is not recommended. This probably overwrites the existing [Dataset.integrate](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.integrate.html). Instead the recommendation is to create a namespace using the accessor registration e.g. `@xr.register_dataset_accessor('ugrid')` and then do `ds.ugrid.integrate`. 

See the example [here](https://docs.xarray.dev/en/stable/internals/extending-xarray.html) where the namespace is `"geo"` and `plot` is a method in that namespace

uxarray/helpers.py
```diff
@@ -63,10 +66,167 @@ def determine_file_type(filepath):
...
+    nodeCross = cross(dDaG, dDbG)
+    # print("nc", nodeCross)
+    dJacobian = math.sqrt(nodeCross[0] * nodeCross[0] +
+                          nodeCross[1] * nodeCross[1] +
+                          nodeCross[2] * nodeCross[2])
+
+    return dJacobian
+
+
+def cross(a, b):
```
dcherian, line 220: numpy provides cross and dot methods so you could use those.

uxarray/helpers.py
```diff
@@ -1,7 +1,10 @@
-import os
+import math
```
dcherian, line 1: Is there a reason to use the math methods instead of the numpy methods? It seems like this package will pass around a bunch of numpy arrays so it might be better to just use the numpy methods directly

uxarray/helpers.py
```diff
@@ -63,10 +66,167 @@ def determine_file_type(filepath):
...
+
+    dDenomTerm = dInvR * dInvR * dInvR
+
+    dDaG[0] *= dDenomTerm
+    dDaG[1] *= dDenomTerm
+    dDaG[2] *= dDenomTerm
+
+    dDbG[0] *= dDenomTerm
+    dDbG[1] *= dDenomTerm
+    dDbG[2] *= dDenomTerm
```
dcherian, line 208: If these are numpy arrays, could shorten to
```suggestion
    dDaG *= dDenomTerm
    
    dDbG *= dDenomTerm
```

uxarray/integrate.py
erogluorhan: @dcherian , assuming such a namespace, "ugrid", is created, then a call to `integrate` would look like:

```
grid = ux.Grid("filename")
grid.ds.ugrid.integrate("psi")
```

which wouldn't look elegant. Is there a mechanism / best practice that Xarray recommends to the Xarray-extending packages to hide this chain to make the call look like:

'grid.integrate("psi")' 

instead? 

This is because, even though we want to extend Xarray as recommended, we would like the UXarray functionality to show up in the user API as its own, not through `Xarray.dataset`. I'd think of writing wrapper functions for such cases but it would clutter the codebase much.

erogluorhan (commented):
Just a few quick inline comments. More to come. I am also planning to branch off this branch and propose some changes directly in PRs to speed up the reviewing/addressing process (Feel free to accept them partially as you see fit or otherwise)

uxarray/integrate.py
dcherian: The xgcm-like syntax is 
```
grid = xgcm.Grid(...)
grid.integrate(ds.psi, axis=("X", "Y"))
```

But since you're making `ds` a property of `Grid` you can do
```
grid.integrate("psi")
```

This suggests `integrate` should be a method on the `Grid` class, which makes sense.

test/constants.py
```diff
@@ -4,3 +4,7 @@
 NNODES_outCSne30 = 5402
 NNODES_outRLL1deg = 64442
 DATAVARS_outCSne30 = 4
+TRI_AREA = 1.047
+MESH30_AREA = 15.59
```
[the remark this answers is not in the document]
rajeeja: @paullric how do we justify these values?

uxarray/helpers.py
rajeeja: done.

uxarray/helpers.py
```diff
@@ -82,10 +84,149 @@ def determine_file_type(filepath):
...
+    z = r * np.sin(lat)  # z coordinate
+
+    coord = np.array([x, y, z])
+    # make it coord on a sphere with unit radius
+    unit_coord = coord / np.linalg.norm(coord)
+
+    return unit_coord
+
+
+# Calculate the area of all faces.
```
michaelavs, line 135: I think this comment can be removed

uxarray/helpers.py
```diff
@@ -82,10 +84,149 @@ def determine_file_type(filepath):
...
+                dA = dG[p]
+                dB = dG[q]
+                jacobian = calculate_spherical_triangle_jacobian(
+                    node1, node2, node3, dA, dB)
+                area += dW[p] * dW[q] * jacobian
+    return area
+
+
+def calculate_spherical_triangle_jacobian(node1, node2, node3, dA, dB):
+    """Helper function for calculating face area."""
```
michaelavs, line 183: For consistency with the rest of the file, I think this should have a full docstring

michaelavs, line 183: For consistency with the rest of the file, I think this should have a full docstring

uxarray/helpers.py
```diff
@@ -82,10 +84,149 @@ def determine_file_type(filepath):
...
+    ])
+
+    dDenomTerm = dInvR * dInvR * dInvR
+
+    dDaG *= dDenomTerm
+    dDbG *= dDenomTerm
+
+    #  Cross product gives local Jacobian
+    nodeCross = np.cross(dDaG, dDbG)
+    # print("nc", nodeCross)
```
michaelavs, line 227: This can be removed

michaelavs, line 227: This can be removed

uxarray/get_quadratureDG.py
```diff
@@ -0,0 +1,287 @@
+def get_tri_quadratureDG(nOrder):
+    """Triangular Quadrature Points for integration.
+
+    Args:
```
michaelavs, line 4: This should be `Parameters`, same with next function docstring on line 153

michaelavs, line 4: This should be `Parameters`, same with next function docstring on line 153

uxarray/get_quadratureDG.py
```diff
@@ -0,0 +1,287 @@
...
+
+        dW = [
+            0.0666713443086881, 0.1494513491505806, 0.2190863625159820,
+            0.2692667193099963, 0.2955242247147529, 0.2955242247147529,
+            0.2692667193099963, 0.2190863625159820, 0.1494513491505806,
+            0.0666713443086881
+        ]
+
+
+# Scale quadrature points
```
michaelavs, line 270: Indent should match with below line

michaelavs, line 270: Indent should match with below line

uxarray/grid.py
```diff
@@ -113,18 +113,41 @@ def __from_vert__(self):
             })
         self.ds.Mesh2.attrs['topology_dimension'] = self.vertices[0].size
 
+        # if type == "spherical":
```
michaelavs, line 116: Should this and lines 124-129 be removed?

michaelavs, line 116: Should this and lines 124-129 be removed?

uxarray/grid.py
```diff
@@ -225,3 +276,61 @@ def __init_ds_var_names__(self):
...
+                    z.append(0)
+
+            # After getting all the nodes of a face assembled call the  cal. face area routine
+            face_area = calculate_face_area(x, y, z, units)
+            # get the value from the data file
+            face_val = self.ds.get(var_key).to_numpy().data[i]
+
+            integral += face_area * face_val
+
+        # print("Integral of ", var_key, " over the surface is ", integral)
```
michaelavs, line 335:
```suggestion

```

michaelavs, line 335: Can also be removed

michaelavs (commented):
Just some documentation and formatting changes for now. I'm still looking at the profiling of a couple of changes to see if they may help with overall time/memory usage

uxarray/_exodus.py
```diff
@@ -123,6 +123,7 @@ def _read_exodus(filepath, ds_var_names):
 
             # concatenate to the previous blk
             conn = np.concatenate((conn, tmp_conn))
+            conn = conn.astype(int)
```
dcherian, line 126: Instead I suggest using `np.full` instead of `np.empty` and changing dtypes

```
conn = np.full((0, max_face_nodes), 0, dtype=np.int64)
tmp_conn = np.full((value.data.shape[0], max_face_nodes), 0, dtype=conn.dtype)
```

Also why is `conn` needed It seems like `np.concatenate((conn, tmp_conn)) == tmp_conn`?

uxarray/helpers.py
```diff
@@ -82,10 +84,149 @@ def determine_file_type(filepath):
...
+
+    coord = np.array([x, y, z])
+    # make it coord on a sphere with unit radius
+    unit_coord = coord / np.linalg.norm(coord)
+
+    return unit_coord
+
+
+# Calculate the area of all faces.
+def calculate_face_area(x, y, z, type="spherical"):
```
dcherian: Consider using numba for this and the jacobian function for a giant speedup.

Alternatively (and more fun!), could explore vectorizing using `tensordot`, `einsum` and friends from numpy (https://numpy.org/doc/stable/reference/routines.linalg.html#matrix-and-vector-products)

test/test_integrate.py
```diff
@@ -0,0 +1,57 @@
...
+                          [0.57735027, 5.77350269e-01, -0.57735027],
+                          [-0.57735027, 5.77350269e-01, -0.57735027]])
+        vgrid = ux.Grid(verts)
+
+        # get node names for each grid object
+        x_var = vgrid.ds_var_names["Mesh2_node_x"]
+        y_var = vgrid.ds_var_names["Mesh2_node_y"]
+        z_var = vgrid.ds_var_names["Mesh2_node_z"]
+
+        vgrid.ds[x_var].attrs["units"] = "cartesian"
```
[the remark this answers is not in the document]
rajeeja: helpers.py has on line: 168 
'''if (coords_type == "spherical")'''
then convert spherical_to_cartesian_unit, this setting is needed for that.

docs/user_api/index.rst
```diff
@@ -21,10 +21,14 @@ Grid Methods
...
+   grid.Grid.integrate
 
 Helper Functions
 ----------------
 .. autosummary::
    :toctree: _autosummary
 
    determine_file_type
+   calculate_face_area
+   spherical_to_cartesian_unit
```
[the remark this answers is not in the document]
rajeeja: It might be because of multiple defs. I have removed. It is used and declared in helpers.py

uxarray/_exodus.py
rajeeja: Thanks, I've simplified things based on your input.

uxarray/get_quadratureDG.py
```diff
@@ -0,0 +1,287 @@
+def get_tri_quadratureDG(nOrder):
```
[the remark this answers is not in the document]
rajeeja: Removed

uxarray/get_quadratureDG.py
rajeeja: Fixed.

uxarray/get_quadratureDG.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Args:
+        nOrder (integer): Integration order, supports: 12, 10, 8, 4 and 1
+
+    Returns:
+        dG, dW: points and weights, with dimension order x 3
+    """
+    # 12th order quadrature rule (33 points)
+    if (nOrder == 12):
+        dG = [[0.023565220452390, 0.488217389773805, 0.488217389773805],
```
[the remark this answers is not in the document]
rajeeja: Done.

uxarray/get_quadratureDG.py
```diff
@@ -0,0 +1,287 @@
...
+    dXi0 = 0.0
+    dXi1 = 1.0
+    for i in range(nCount):
+        dG[i] = dXi0 + 0.5 * (dXi1 - dXi0) * (dG[i] + 1.0)
+        dW[i] = 0.5 * (dXi1 - dXi0) * dW[i]
+
+    return dG, dW
+
+
+def main():
```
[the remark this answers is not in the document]
rajeeja: yes, removing

test/test_integrate.py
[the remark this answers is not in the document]
dcherian: :+1: My concern is that `units` is a CF standard attribute, so its confusing to use that attribute for something else. Is there a way to determine cartesian vs spherical in the UGRID conventions?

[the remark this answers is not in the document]
dcherian: For reference: http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#units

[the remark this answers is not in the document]
rljacob: We should avoid conflicting with the CF standard.

rajeeja:
@anissa111 we should get your branch here and merge main here. It seems like we'd have some merge conflicts to resolve. Please create a PR from your branch to integrate branch

uxarray/grid.py
```diff
@@ -284,21 +291,24 @@ def integrate(self, var_key):
...
+        all faces in the mesh.
+
         Returns
         -------
 
-        ndarray: area of all the faces in the mesh.
+        area of all the faces in the mesh. : ndarray
 
-        Examples
-        --------
```
dcherian: "Examples" is the standard: https://numpydoc.readthedocs.io/en/latest/format.html#examples

uxarray/grid.py
```diff
@@ -225,3 +255,84 @@ def __init_ds_var_names__(self):
...
+
+        # call function to get area of all the faces as a np array
+        face_areas = self.face_areas
+
+        face_vals = self.ds.get(var_key).to_numpy()
+        integral = np.dot(face_areas, face_vals)
+
+        return integral
+
+    @property
```
erogluorhan, line 292: Regarding this usage, I have created a suggestion PR from a branch off this branch.

ci/environment.yml
```diff
@@ -4,9 +4,13 @@ channels:
 dependencies:
   - dask
   - netcdf4
+  - numba
+  # pinning the numpy version for numba compatibility
+  - numpy>=1.2, <1.22
```
rajeeja, line 9: @anissa111 do we need this? or just requirements is enough.

anissa111: @rajeeja I think we need to have it in the environment file as well to keep the pinned version in our local development environment

erogluorhan:
I will create a series of PRs against this branch to propose my changes. They will be for either bug-fixing or code improvement. The first one is PR #71

uxarray/grid.py
```diff
@@ -228,6 +258,118 @@ def __init_ds_var_names__(self):
...
+            # call function to get area of all the faces as a np array
+            self._face_areas = get_all_face_area_from_coords(
+                x, y, z, face_nodes, dim, quadrature_rule, order, coords_type)
+
+        return self._face_areas
+
+    # use the property keyword for declaration on face_areas property
+    face_areas = property(get_face_areas)
+
     def __init_grid_var_attrs__(self):
```
anissa111, line 373: It looks like there are two definitions for this function

erogluorhan: Good catch! Fixing this in PR #74

anissa111:
See #76 for changes for documentation generation

uxarray/helpers.py
```diff
@@ -39,6 +51,372 @@ def parse_grid_type(filepath, **kw):
...
+
+    # num triangles is two less than the total number of nodes
+    num_triangles = num_nodes - 2
+
+    # Using tempestremap GridElements: https://github.com/ClimateGlobalChange/tempestremap/blob/master/src/GridElements.cpp
+    # loop through all sub-triangles of face
+    for j in range(0, num_triangles):
+        node1 = np.array([x[0], y[0], z[0]], dtype=np.float64)
+        node2 = np.array([x[j + 1], y[j + 1], z[j + 1]], dtype=np.float64)
+        node3 = np.array([x[j + 2], y[j + 2], z[j + 2]], dtype=np.float64)
```
erogluorhan, line 183: This node selection may not always give us the correct triangles, so, is there an assumption about the order of the nodes in the original tempestremap code about these? For example, the below scenario won't give us the correct area:

image

May there be something getting lost from tempestremap to here?

rajeeja: > This node selection scheme may not always give us the correct triangles, so, is there an assumption about the order of the nodes in the original tempestremap code about these? For example, the below scenario won't give us the correct area:
> 
> image
> 
> Am I missing anything here or may there be something getting lost from tempestremap to here?

This is not a valid mesh, there is a function called validate in our grid class. Once that is completed this mesh would be rejected. The assumption is: nodes are in counterclockwise order, and we look through the connectivity of the face breaking it into triangles. Incorrect order of nodes in connectivity will not produce the correct result

erogluorhan: Thanks for the explanation! This reminds me we may also consider putting a CCW ordering utility function into our package in the future?

erogluorhan (approved):

uxarray/__init__.py
```diff
@@ -1,2 +1,3 @@
 from .grid import *
 from .dataset import *
+from .helpers import *
```
dcherian, line 3: Do you really wnat all the helpers to be exposed publicly?

uxarray/grid.py
```diff
@@ -228,6 +259,118 @@ def __init_ds_var_names__(self):
...
+                z = self.ds[self.ds_var_names["Mesh2_node_z"]].data
+
+            # call function to get area of all the faces as a np array
+            self._face_areas = get_all_face_area_from_coords(
+                x, y, z, face_nodes, dim, quadrature_rule, order, coords_type)
+
+        return self._face_areas
+
+    # use the property keyword for declaration on face_areas property
+    face_areas = property(get_face_areas)
```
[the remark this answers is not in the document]
dcherian: AH didn't realize you want this to be configurable.

Would a user ever want to rewrite the face areas on the _same_ Grid object if they changed their mind about the quadrature and order? Presently this is not possible because the cache check is in `get_face_areas`

If you did want that then I would think about
1. Renaming `get_face_areas` to `compute_face_areas`
2. And following the caching pattern in `face_areas` as recommended earlier
``` python
@property
def face_areas(self): 
    if self._face_areas is None:
	self.compute_face_areas()
    return self._face_areas
```
3. Remove the caching check in `compute_face_areas`

Then a user can overwrite the cache by explicitly calling `grid.compute_face_areas` with their favourite options. And then rest of the code which uses `grid.face_areas` will use the updated one.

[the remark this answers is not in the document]
dcherian: ^ Also this seems a somewhat minor interface issue. I agree with merging and resolving later.

uxarray/__init__.py
rajeeja: It seemed possible and I thought it might be useful for some projects (giving our projects added utility), can be removed also (fine either way).

anissa111: The way we have it set up now, some of the functions in helper are available through the user api: https://uxarray.readthedocs.io/en/integration/user_api/index.html

Do you think we should change this?

dcherian: @anissa111 Hm... do those work with just `import uxarray` . Seems like that might require a separate `import uxarray.helpers`. 

You could do `from . import helpers` here and `uxarray.helpers` would always be available

anissa111 (approved):

PR #38, opened 2022-04-21
Updating helpers.py

michaelavs:
To reduce try except blocks and possibly update to remove shapefile support

### DIFF
--- a/uxarray/helpers.py
+++ b/uxarray/helpers.py
@@ -3,70 +3,75 @@ import xarray as xr
 from pathlib import PurePath
 
 
-# helper function to find file type
 def determine_file_type(filepath):
     """Checks file path and contents to determine file type. Supports detection
     of UGrid, SCRIP, Exodus and shape file.
 
-    Parameters: string, required
+    Parameters
+    ----------
+
+    filepath : :class:`string`: required
        Filepath of the file for which the filetype is to be determined.
 
-    Returns: string
-       File type: ug, exo, scrip or shp
+    Returns
+    -------
+    mesh_filetype : :class:`string`:
+        String describing the file type (ug, exo, scrip or shp)
+        to be used by Grid.py
 
     Raises:
        RuntimeError: Invalid file type
     """
     msg = ""
-    # exodus with coord
+    # Extract file name and extension, read into xarray.Dataset to prep
+    path = PurePath(filepath)
+    file_extension = path.suffix
+    # ext_ds = xr.open_dataset(filepath, mask_and_scale=False)
+
+    # Shapefiles (not supported, placeholder)
+    # if file_extension == ".shp":
+    #     mesh_filetype = "shp"
+
+    # NetCDF style files
     try:
-        # extract the file name and extension
-        path = PurePath(filepath)
-        file_extension = path.suffix
-
-        # try to open file with xarray and test for exodus
-        ext_ds = xr.open_dataset(filepath, mask_and_scale=False)["coord"]
-        mesh_filetype = "exo"
-    except KeyError as e:
-        # exodus with coordx
-        try:
-            ext_ds = xr.open_dataset(filepath, mask_and_scale=False)["coordx"]
+        ext_ds = xr.open_dataset(filepath, mask_and_scale=False)
+        if "coord" in ext_ds.data_vars:
             mesh_filetype = "exo"
-        except KeyError as e:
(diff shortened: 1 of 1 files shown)


michaelavs:
Hey @erogluorhan @anissa111 and @rajeeja,
I wanted to get a little bit of feedback on the direction of the updates to this file before I remove the "draft" aspect of the PR. The current version in this PR is a bit messy, but I wanted to make sure the separate options for its direction were shown when asking this question. In the original helpers.py, shapefiles are tested for based off of either one of two possible errors being thrown from trying to read them in with xarray. With that, and that shapefile reading has been delayed for a later date in mind, I have come up with three possible ways to handle shapefiles going forward:
1. Creating a second helper function within helpers.py that will then be integrated into grid.py (seen in the function called "determine_shapefile" starting on line 65)
2. Possibly moving the test for shapefiles to be before the tests for exodus, scrip, and ugrid (roughly what is seen in lines 31-33, but in it's current form, that code chunk doesn't work)
3. Dropping the shapefile aspect from both helpers.py and grid.py until we are ready to implement the shapefile reader later on

rajeeja:
Hey @michaelavs your changes look reasonable. It'd be good if you merge `main` to this, after PR #44 goes to main.

uxarray/helpers.py
```diff
@@ -3,70 +3,75 @@
...
-        # check if this is a shp file
-        # we won't use xarray to load that file
-        if file_extension == ".shp":
-            mesh_filetype = "shp"
-        else:
-            msg = str(e) + ': {}'.format(filepath)
-    except ValueError as e:
-        # check if this is a shp file
-        # we won't use xarray to load that file
+            print("exo")
```
erogluorhan, line 40: Why is the print statement? To me, constructing a "msg", as it was done before, for each case and raising it as an exception/error in except cases or printing it as the report of succesful output sounds like a good way to go.

uxarray/helpers.py
```diff
@@ -3,70 +3,75 @@
...
+
+        elif "coordx" in ext_ds.data_vars:
```
erogluorhan, line 42: Should we combine this with the above `if "coord"...`?

uxarray/helpers.py
```diff
@@ -3,70 +3,75 @@
...
+            mesh_filetype = "exo"
+
+        # SCRIP files
+        elif 'grid_area' in ext_ds.data_vars:
+            mesh_filetype = "scrip"
+            print("scrip")
+
+        # UGRID files
+        elif "Mesh2" in ext_ds.data_vars:
```
erogluorhan, line 51: This part will need to be gotten from PR #44.

uxarray/helpers.py
```diff
@@ -3,70 +3,75 @@
...
+        to be used by Grid.py
 
     Raises:
        RuntimeError: Invalid file type
     """
     msg = ""
-    # exodus with coord
+    # Extract file name and extension, read into xarray.Dataset to prep
+    path = PurePath(filepath)
+    file_extension = path.suffix
```
erogluorhan, line 28: If the filepath is given somehow broken, there will be an exception that we won't catch.

uxarray/helpers.py
```diff
@@ -3,70 +3,75 @@
...
+
+        # Shape files
+        # elif file_extension == ".shp":
+        #     mesh_filetype = "shp"
+
+        return mesh_filetype
+
+    except:
+        # Update with the correct exceptions, not sure which will be the best from below
+        print("Unrecognized file type")
```
erogluorhan, line 62: Agreed! Shouldn't be left as a print statement, isntead should raise an exception/error.

erogluorhan:
> 1. Creating a second helper function within helpers.py that will then be integrated into grid.py (seen in the function called "determine_shapefile" starting on line 65)
> 2. Possibly moving the test for shapefiles to be before the tests for exodus, scrip, and ugrid (roughly what is seen in lines 31-33, but in it's current form, that code chunk doesn't work)
> 3. Dropping the shapefile aspect from both helpers.py and grid.py until we are ready to implement the shapefile reader later on

I'd prefer option (2). i.e. I'd implement the shape file extension case but construct and raise an error message about it being not currently but soon to be implemented. That'd give the reader both information (we will support shapefiles but not now)

erogluorhan (commented):
Please see a few inline comments

Issue #39, opened 2022-04-25
Shouldn't `open_dataset` return Grid object? 


erogluorhan:
Our `open_dataset` function currently returns `Grid.ds` (See [here](https://github.com/UXARRAY/uxarray/blob/328b4448fbf4bb289882ed15d800be4c0cf12cd0/uxarray/dataset.py#L48) ), which is of type `xarray.Dataset`, but I think we should return Grid object itself. Any thoughts?

michaelavs:
Was this addressed with PR #41 yesterday? I see it referenced in here, but want to double check before closing out this issue

rajeeja:
Yes. Merged to main

PR #41, opened 2022-04-25
#39 fix return type 

rajeeja:
#39

### DIFF
--- a/test/test_dataset.py
+++ b/test/test_dataset.py
@@ -24,19 +24,19 @@ class TestDataset(TestCase):
         uds2 = ux.open_dataset(uds2_name)
         uds3 = ux.open_dataset(uds3_name)
 
-        assert (uds1.nMesh2_node.size == constants.NNODES_ov_RLL10deg_CSne4)
-        assert (uds2.nMesh2_node.size == constants.NNODES_outCSne8)
-        assert (uds3.nMesh2_node.size == constants.NNODES_outCSne30)
+        assert (uds1.ds.nMesh2_node.size == constants.NNODES_ov_RLL10deg_CSne4)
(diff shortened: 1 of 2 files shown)


test/test_dataset.py
```diff
@@ -24,25 +24,25 @@ def test_open_grid_only(self):
         uds2 = ux.open_dataset(uds2_name)
         uds3 = ux.open_dataset(uds3_name)
 
-        assert (uds1.nMesh2_node.size == constants.NNODES_ov_RLL10deg_CSne4)
-        assert (uds2.nMesh2_node.size == constants.NNODES_outCSne8)
-        assert (uds3.nMesh2_node.size == constants.NNODES_outCSne30)
+        assert (uds1.ds.nMesh2_node.size == constants.NNODES_ov_RLL10deg_CSne4)
```
erogluorhan, line 27: This use reminds me another discussion: Should we provide a direct `uds1.nMesh2_node` or is it sufficient to access it through `ds` like here, i.e. `uds1.nMesh2_node`?

erogluorhan (commented):
Thanks for quick fix! Just a single inline comment.

test/test_dataset.py
rajeeja: Right, we should not. That would be fixed in another PR, when we fix ugrid reader. It would need wholesale changes to all mentions of "Mesh2", in this case nMesh2_node. It'd be the `variable name` as specified in the original file.

erogluorhan: Sounds good!

erogluorhan (approved):

anissa111 (approved):

Issue #27, opened 2022-03-23
Unable to open UGRID file


clyne:
I believe that the attached NetCDF grid  file is UGRID compliant, but when I try to open it with:

`uds = ux.open_dataset('grid.nc')
`

I get:

_Loading initial grid from file:  /Users/clyne/Data/UGRID/geoflow-ugrid/small/grid.nc
This is not a supported NetCDF file_

Here is an `ncdump -h` of the file:

netcdf grid {
dimensions:
	nMeshNodes = 6000 ;
	nMeshFaces = 3840 ;
	nFaceNodes = 4 ;
	meshLayers = 20 ;
	time = 1 ;
variables:
	int mesh ;
		mesh:cf_role = "mesh_topology" ;
		mesh:long_name = "Topology data of 2D unstructured mesh." ;
		mesh:topology_dimension = 2 ;
		mesh:node_coordinates = "mesh_node_x mesh_node_y" ;
		mesh:face_node_connectivity = "mesh_face_nodes" ;
		mesh:face_dimension = "nMeshFaces" ;
	uint mesh_face_nodes(nMeshFaces, nFaceNodes) ;
		mesh_face_nodes:cf_role = "face_node_connectivity" ;
		mesh_face_nodes:long_name = "Maps every face to its corner nodes." ;
		mesh_face_nodes:_FillValue = 4294967295U ;
		mesh_face_nodes:start_index = 0 ;
	double mesh_node_x(nMeshNodes) ;
		mesh_node_x:standard_name = "longitude" ;
		mesh_node_x:long_name = "Longitude of 2D mesh nodes." ;
		mesh_node_x:units = "degrees_east" ;
	double mesh_node_y(nMeshNodes) ;
		mesh_node_y:standard_name = "latitude" ;
		mesh_node_y:long_name = "Latitude of 2D mesh nodes." ;
		mesh_node_y:units = "degrees_north" ;
	double mesh_depth(meshLayers, nMeshNodes) ;
		mesh_depth:units = "meters" ;
		mesh_depth:positive = "up" ;
		mesh_depth:standard_name = "elevation" ;
		mesh_depth:mesh = "mesh" ;
		mesh_depth:location = "node" ;
		mesh_depth:coordinates = "mesh_node_x mesh_node_y" ;
}

[grid.nc.gz](https://github.com/UXARRAY/uxarray/files/8336834/grid.nc.gz)

rajeeja:
Thanks for the attachment and this issue.

[...]

Issue #28, opened 2022-03-23
Should UGRID coordinate variables appear as xarray.coords?


clyne:
Coordinate variables defined in UGRID files appear as Xarray data variables (xarray.data_vars), not coordinate variables (xarray.coords). The UGRID specification follows the CF conventions and there is sufficient information to programmatically identify the coordinate variables (as XArray does itself). Shouldn't the coordinate variables appear as xarray.coords?

To demonstrate:

load the UXarray test data set "ov_RLL10deg_CSne4.ug" and print its members:

```
uds3 = ux.open_dataset('test/meshfiles/ov_RLL10deg_CSne4.ug')
print(uds3)
```

"Mesh2_node_x" and "Mesh2_node_y" are listed as data variables even though they have 'standard_name' and 'units' attributes that identify them as coordinate variables.

rajeeja:
Yes, they can. 
Would they be better for scaling and working with large meshes with coordinate variables?

clyne:
I'm not sure if there would be any impact on scalability, but this would be consistent with the use of XArray elsewhere, and though I can't think of any examples at the moment, would possibly open the door to using UXarray data with XArray operators that require coordinate information.

rajeeja:
> I'm not sure if there would be any impact on scalability, but this would be consistent with the use of XArray elsewhere, and though I can't think of any examples at the moment, would possibly open the door to using UXarray data with XArray operators that require coordinate information.

Gotcha, I can add them as coordinates.

clyne:
Thanks!

erogluorhan:
PR #44  should have addressed this; please check

PR #44, opened 2022-05-10
fix ugrid reader

rajeeja:


rajeeja:
Fixes #27 #28

michaelavs (approved):
I think this looks good, I was able to run the tests successfully locally! I do want to note that grid.py was having issues at first because I have edits to my version locally (I think from scrip PR/branch) and I think your version and mine were overlapping somewhat, so definitely wait for at least one more review before merging this just in case

uxarray/_ugrid.py
```diff
@@ -10,12 +10,54 @@ def _read_ugrid(filepath):
...
+        # set coordinates
+        xr_ds = xr_ds.set_coords(
+            [var_names_dict["Mesh2_node_x"], var_names_dict["Mesh2_node_y"]])
+    else:
+        xr_ds = xr_ds.set_coords([
+            var_names_dict["Mesh2_node_x"], var_names_dict["Mesh2_node_y"],
+            var_names_dict["Mesh2_node_z"]
+        ])
+
+    return xr_ds
```
erogluorhan, line 56: `var_names_dict` gets updated after leaving this ffunction, but for the sake of code readability, should we also return that in here:

`return xr_ds, var_names_dict`

and change Line 155 in grid.py as follows:

`self.ds, self.var_names_dict = _read_ugrid(self.filepath, self.var_names_dict)`?

I am not  sure which way to go though.

erogluorhan (commented):
Great work overall, thanks! Please add the newly added functions into the docs as well as address inline comments.

uxarray/_ugrid.py
```diff
@@ -10,12 +10,54 @@ def _read_ugrid(filepath):
...
     """
 
+    # TODO: obtain and change to Mesh2 construct, see Issue #27
     # simply return the xarray object loaded
-    return xr.open_dataset(filepath, mask_and_scale=False)
+
+    xr_ds = xr.open_dataset(filepath, mask_and_scale=False)
+
+    #
+    base_dv_mt = list(xr_ds.filter_by_attrs(cf_role="mesh_topology").keys())[0]
```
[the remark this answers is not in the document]
rajeeja: fixed

uxarray/_ugrid.py
rajeeja: makes sense. fixed.

uxarray/grid.py
```diff
@@ -203,3 +205,22 @@ def buildlatlon_bounds(self):
     def validate(self):
         """Not implemented."""
         warn("Function placeholder, implementation coming soon.")
+
+    def __init_var_names__(self):
```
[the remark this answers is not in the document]
rajeeja: done.

docs/internal_api/index.rst
```diff
@@ -16,6 +16,7 @@ Grid Methods
    :toctree: ./generated/
 
    grid.Grid.__init__
+   grid.Grid.__init_var_names__
```
erogluorhan, line 19: Typo: "__init_ds_var_names__"

erogluorhan (commented):
Just a single inline comment before approval

uxarray/helpers.py
```diff
@@ -70,3 +95,14 @@ def determine_file_type(filepath):
             os._exit(0)
 
     return mesh_filetype
+
+
+def check_ugrid_cf_role(cf_role):
```
erogluorhan, line 100: Is this intended to be an internal or user API function? Also, please add it to corresponding docs index.

rajeeja: This isn't used. Removing.

docs/internal_api/index.rst
rajeeja: Fixed

erogluorhan (approved):

uxarray/helpers.py
```diff
@@ -39,13 +40,37 @@ def determine_file_type(filepath):
...
+                # check mesh topology and dimension
                 try:
-                    ext_ds = xr.open_dataset(filepath,
-                                             mask_and_scale=False)["Mesh2"]
-                    mesh_filetype = "ugrid"
+                    standard_name = lambda v: v is not None
+                    # getkeys_filter_by_attribute(filepath, attr_name, attr_val)
+                    # return type KeysView
+                    base_dv_nc = list(
+                        xr.open_dataset(
```
dcherian, line 50: xarray does cache datasets but it might be cleaner to just use `open_dataset` once like `ext_ds` above.

rajeeja: List is for "xr.open_dataset(
                            filepath, mask_and_scale=False).filter_by_attrs(
                                node_coordinates=standard_name).keys()" There could be multiple keys, so taking the first ("[0]") from first element from the list.

dcherian: Thanks. I meant tthat you could
```
ds = xr.open_dataset(filepath, mask_and_scale=False)
```
and then
```
ds.filter_by_attrs(node_coordinates=standard_name).keys()
```
since it gets opened repeatedly (it doesn't actually, because xarray caches the open file)

rajeeja: gotcha, thanks.

Issue #45, opened 2022-05-11
`uxarray.Grid.Mesh2_node_x` vs UGRID's flexible naming 


erogluorhan:
We are currently implementing the flexible variable naming in UGRID conventions through dynamic variable naming in xArray.Dataset, i.e. `Uxarray.Grid.ds`, with the help of a dictionary (See PR #44).

However, we are proposing to have [uxarray.Grid.Mesh2_node_x](https://github.com/UXARRAY/uxarray/blob/main/docs/user_api/uxarray_api.md#:~:text=uxarray.Grid.Mesh2_node_x%3A%20np.float64%20DataArray%20of%20size%20(nMesh2_node)) et al., i.e. standard name variables, in our API, too.

To comply with our API as well as to provide the user with a standardized naming set in addition to keeping track of original dataset names in our `Uxarray.Grid.ds`, should we provide `Uxarray.Grid.mesh2_node_x` et al. functions that would simple access `Uxarray.Grid.ds.<actual variable name>`?

For instance, if we had `my_ugrid_node_x` variable coming from a netCDF file, one would be able to access this through two ways like:

`Uxarray.Grid.ds.my_ugrid_node_x`
`Uxarray.Grid.mesh2_node_x` 

Any concerns such as it'd lead to confusions?

@anissa111 @michaelavs  @paullric @rajeeja @rljacob

rajeeja:
Maybe, we can do this as a part 2 of this fix.

erogluorhan:
I share your concerns @rajeeja

erogluorhan:
> From experience with [cf_xarray](https://cf-xarray.readthedocs.io/en/latest/), I think that having a way to access using "standard names" Grid.XXX and a way to use model-specific names through Grid.ds.YYY is very useful.

> It makes it easier to slowly switch code from model-specific names to using the "standard name" that will generalize to multiple models.

Great point, thanks!

PR #62, opened 2022-07-07
Directly accessing data variables with UGRID conventions

philipc2:
Convenience attributes for accessing variables without the use of a dictionary. 

Assuming the mesh node coordinates for longitude are stored with an input
name of `mesh_node_x`, we store this variable name in the `ds_var_names`
dictionary with the key `Mesh2_node_x`. In order to access it:

>>> x = grid.ds[grid.ds_var_names["Mesh2_node_x"]]

With the help of this function, we can directly access it through the
use of a standardized name (ugrid convention)

 >>> x = grid.Mesh2_node_x


uxarray/grid.py
```diff
@@ -225,3 +227,34 @@ def __init_ds_var_names__(self):
...
+
+        Return Xarray Data Array
+        >>> x = grid.Mesh2_node_x      
+
+        Return Numpy Array
+        >>> x = grid.Mesh2_node_x.values
+        """
+    
+        # For Testing (caused other tests to fail)
+        if self.ds is None:
```
erogluorhan, line 248: I'd expect `ds` not being `None` to be guaranteed until this point, isn't it?

uxarray/grid.py
```diff
@@ -225,3 +227,34 @@ def __init_ds_var_names__(self):
             "nMesh2_face": "nMesh2_face",
             "nMaxMesh2_face_nodes": "nMaxMesh2_face_nodes"
         }
+
+    def __init_ds_var_methods__(self):
```
erogluorhan, line 231: How about `__init_grid_var_attrs__` (or simply `__init_var_attrs__` as we are already in `grid.__init__ method`) for the name of this function? I propose this because we are initializing the grid var attributes here, not the dataset var methods.

uxarray/grid.py
```diff
@@ -225,3 +227,34 @@ def __init_ds_var_names__(self):
...
+        """Initialize variables for directly accessing Coordinate and
+        Data variables through ugrid conventions
+
+        Examples
+        ----------
+        Initialize some Grid Object
+        >>> grid = ux.open_dataset(file_path)
+
+        Return Xarray Data Array
+        >>> x = grid.Mesh2_node_x
```
erogluorhan, line 241: What about providing this "Return Xarray data array " part as follows (it is a bit long though):

Assuming the mesh node coordinates for longitudes come from the input in the name "mesh_node_x". We are storing such longitude coordinate in our variable names dictionary with the key "Mesh2_node_x". So, we'd need to access the data variable as follows:
>>> x = grid.ds[grid.ds._var_names["Mesh2_node_x"]]

With the help of this function, we could also access it directly via Grid through a standardized name:
>>> x = grid.Mesh2_node_x

uxarray/grid.py
```diff
@@ -225,3 +227,34 @@ def __init_ds_var_names__(self):
...
+
+        Examples
+        ----------
+        Initialize some Grid Object
+        >>> grid = ux.open_dataset(file_path)
+
+        Return Xarray Data Array
+        >>> x = grid.Mesh2_node_x      
+
+        Return Numpy Array
```
erogluorhan, line 243: This is (i.e. Xarray data array to Numpy array conversion) beyond the scope of this example and function I believe.

uxarray/grid.py
```diff
@@ -225,3 +227,34 @@ def __init_ds_var_names__(self):
...
+        """
+    
+        # For Testing (caused other tests to fail)
+        if self.ds is None:
+            return
+
+        # Swap Keys & Values (Accessed with UGRID convention) UGRID -> Orig
+        ds_ugrid_var_names = dict((k, v) for k,v in self.ds_var_names.items())
+      
+        for key, value in ds_ugrid_var_names.items():
```
erogluorhan, line 254: I like this implementation much overall (I was personally thinking of writing separate methods instead, but this seems more compact). 

The only concern of mine is that depending on the existence of variables and coords, the user will see some attributes showing up in Grid for some datasets and disappearing for others. I am not sure how this behavior will feel to the user. Any thoughts? @anissa111 @dcherian @rajeeja @michaelavs

test/test_grid.py
```diff
@@ -43,14 +42,31 @@ def test_init_verts(self):
...
         vgrid.write(face_filename)
 
+    def test_init_ds_var_methods(self):
+        """Tests if accessing a variable though self.{NAME} is 
+        equivalent to doing self.ds[{NAME}]
+        """
+        # Variables in UGRID convention
+        path = current_path / "meshfiles" / "outCSne30.ug"
+        grid = ux.open_dataset(path)
+        xr.testing.assert_equal(grid.Mesh2_node_x, grid.ds['Mesh2_node_x'])
```
erogluorhan, line 52: This assertion works correctly for this test case, but FYI: In fact, for every cases, we are accessing the variable through the dictionary as follows:

`grid.ds[grid.ds._var_names["Mesh2_node_x"]]`

erogluorhan (commented):
Thanks a lot for putting this together! Please see my inline comments as well as do not forget to address the items below:

- Link the corresponding issue to this PR
- Add your function name into the docs/internal_api/index.rst for it to be documented
- PR currently fails with the pre-commit tests (i.e. code quality and standardization). You will probably need to install pre-commit in your local then run it on all files; it will fix the failures.

rajeeja:
good utility and use of xr.testing, please run `pre-commit run --all-files`

uxarray/grid.py
philipc2: My thought process was that this approach would only create attributes for variables that are present in the dataset and dictionary. For example, if our dataset doesn't include a `Mesh2_node_z` variable, it wouldn't make sense to add an attribute for it. 

As long as there exists an entry in the `ds_var_names` dictionary that corresponds to some variable in the coordinates or data vars of our dataset, an attribute will be created allowing us to access it with it's UGRID standardized name. For variables that are not covered by UGRID, I didn't add that functionality, but can easily do that if we believe it would be beneficial.

erogluorhan: This makes sense to me, and we can go with your approach if there is no objection from others.

philipc2:
Changes on Latest Commit
o Renamed to `__init_grid_var_attrs__`
o Added attributes for original variable names
o Added checks for NoneTypes and cleaned up code
o Cleaned up and test case

erogluorhan:
> We can also access it through it's original variable name
> 
> > > > x = grid.mesh_node_x

Yes, but that is not something we wanted to have. Our objective (as mentioned in Issue #45) is to provide a "standardized" naming for such variables in directly through our Grid object even though they could come in any name from the datasets. 

Maybe in a future PR, we'd consider to introduce an access to original var names through `ds` like:

`Grid.ds.<original_var_name>` 

instead of 

`Grid.ds[Grid.ds_var_names["Mesh2_node_x"]]`

but not through Grid directly. So, could you please remove this new feature?

philipc2:
> 

Done!

uxarray/grid.py
erogluorhan: How about replacing the warning in _shapefile.py with an exception like this:

`raise RuntimeError("Function not implemented yet. FYI, attempted to read SHAPE file: " + str(filepath))`

Then, calling the assertion lines in `test_read_shpfile` like this:

```
with self.assertRaises(RuntimeError):
            shp_filename = current_path / "meshfiles" / "grid_fire.shp"
            tgrid = ux.Grid(str(shp_filename))
```

I believe with thee, you could remove your `self.ds is None` check.

erogluorhan (approved):
This is great! Thanks for putting this together!

anissa111 (approved):

PR #64, opened 2022-07-07
Maybe we should rename `face_areas` method and avoid property accessor usage

erogluorhan:
I can see the motivation with using the `@property` accessor (i.e. most likely in order to guarantee the `self._face_areas` attribue to be populated before calling `calc_total_face_area` and `integrate`, as well as avoiding to checking  `self._face_areas is None` in those functions?), but I believe the changes in this PR works better because:

- Using the `@property` accessor like the getter/setter approach in Java is not recommended in Python AFAIK
- Avoiding the `@property` accessor and explicitly checking `if self._face_areas is None` whenever needed makes the code much more readable.
- Renaming the function `face_areas` to `calculate_each_face_area` is self-explanatory

### DIFF
--- a/uxarray/grid.py
+++ b/uxarray/grid.py
@@ -212,10 +212,10 @@ class Grid:
         float: Sum of area of all the faces in the mesh
         """
 
-        # call function to get area of all the faces as a np array
-        face_areas = self.face_areas
+        if self._face_areas is None:
+            self.calculate_each_face_area()
 
-        return np.sum(face_areas)
+        return np.sum(self._face_areas)
 
     # Build the node-face connectivity array.
     def build_node_face_connectivity(self):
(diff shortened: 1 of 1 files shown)


rajeeja (approved):
Looks good, thanks!

dcherian:
> Using the @property accessor like the getter/setter approach in Java is not recommended in Python AFAIK

I think it's the [other way](https://stackoverflow.com/questions/6618002/using-property-versus-getters-and-setters). The recommendation is to use `property` instead of `set_face_areas` and `get_face_areas`

Allowing users to access face areas is useful for custom computations, so it's nice to provide a public property.

dcherian:
> My reference was [this](https://stackoverflow.com/questions/15706935/correct-way-to-use-accessors-in-python) though.

Ya previously the property was allowing cached access but hiding the details

"Use property when you have a genuine need to execute code when getting, setting or deleting an attribute. Validation, caching, side effects, etc. all are fair use-cases for properties. "

> we can rename it to the public naming convention (without underscores).

True but then user code has to use the `if` clause too

```
if Grid.face_areas is not None:
    Grid.calculate_face_areas()
```

erogluorhan:
> "Validation, caching, side effects, etc. all are fair use-cases for properties."

> True but then user code has to use the if clause too

These make sense. Even though I still find using the `@property` accessor confusing to human-read the code, I don't have a preference any longer. @rajeeja, we had already merged my suggestion, but please feel free to revert it back if you prefer that way, or if you want to get it reverted, I can do that.

dcherian:
Thanks @erogluorhan

PR #65, opened 2022-07-07
Fix docs by pinning numpy in requirements.txt 

anissa111:
This was a weird one. In summary, RTD installs uxarray from source with:

``` bash
python -m pip install --upgrade --upgrade-strategy eager --no-cache-dir .
```

This uninstalls the pinned `numpy=1.22` and replaces it with the newest version of numpy, regardless of the pinning. See this [raw output](https://readthedocs.org/api/v2/build/17384086.txt):
image

This (as far as we could tell) not be changed. We can, however, edit `.readthedocs.yml` to include a `post_install` step that re-installs `numpy==1.22`.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -9,6 +9,9 @@ build:
   os: "ubuntu-20.04"
   tools:
     python: "mambaforge-4.10"
+  jobs:
+    post_install:
+      - pip install numpy==1.22
 
 python:
   install:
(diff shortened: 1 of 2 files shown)


michaelavs (approved):

anissa111:
> You should update `requirements.txt` to include the numpy pin. Once that gets to pypi this might fix itself.

Yeah I tried that! I'll add it to this PR anyway

anissa111:
> > You should update `requirements.txt` to include the numpy pin. Once that gets to pypi this might fix itself.
> 
> Yeah I tried that! I'll add it to this PR anyway

Huh! It seems like it's working now with `requirements.txt` and not needing my forcing step. Not exactly sure why it wasn't working before, but I'll take it.

PR #68, opened 2022-07-12
Read from URLs or OPeNDAP endpoints instead of just files

ocefpaf:
~~This is not ready for review yet b/c I want to add a few tests.~~ However, I'd love some input on the `determine_file_type` refactor. It was quite complex to follow and I tried to simplify it a bit, hopefully I did not make it too simple :-)


erogluorhan:
This is great contribution, regardless of whether the code is ready to be reviewed! Thanks very much! 

@anissa111 @dcherian @michaelavs @rajeeja , if any of you can give it some feedback, that'd be great!

uxarray/grid.py
```diff
@@ -85,14 +85,12 @@ def __init__(self, *args, **kwargs):
...
         elif isinstance(data_arg, str):
-            # check if file exists
-            if not os.path.isfile(data_arg):
-                raise RuntimeError("File not found: " + data_arg)
-
-            self.filepath = data_arg
-            # call the appropriate reader
-            self.__from_file__()
-
+            if is_url(data_arg) or os.path.isfile(data_arg):
```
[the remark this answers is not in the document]
ocefpaf: Done in https://github.com/UXARRAY/uxarray/pull/68/commits/5615964f505c51bb891d0bcea39acda34c8f8130

At this point I don't want to send more changes b/c there are already some big and breaking ones in this PR. The main is that the interface to load is no longer the class but `ux.open_dataset` instead. The class can take either an xarray.Dataset, or iterable with grid data, and a shapefile corner case that is baked in the `ux.open_dataset`.

The advantages are:
- closer to xugrid
- no need to maintain lots of "load data" boilerplate checking code, just let xarray handle that
- one can pass special kw args to xarray, like time/coord workarounds

ocefpaf:
This one is ready for review! The tests are passing locally and I added a simple OPeNDAP test. See [this notebook](https://nbviewer.org/urls/gist.githubusercontent.com/ocefpaf/5852806460440a2cad16ce11edb377c7/raw/0d7c226849b4ca44e57b59838ce708100f5b2222/uxarray_opendap.ipynb) for an example of it in action.

uxarray/grid.py
```diff
@@ -28,7 +27,7 @@ class Grid:
     >>> mesh.write("outfile.ug")
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, data_arg, **kwargs):
```
erogluorhan, line 30: Th  dev team had thought there would be need for new args in the future, so they did this as `*args`. @rajeeja  @anissa111 any thoughts?

uxarray/grid.py
```diff
@@ -137,24 +127,15 @@ def __from_vert__(self):
...
-        automatically detect if it is a UGrid, SCRIP, Exodus, or shape file.
-
-        Raises:
-
-            RuntimeError: Unknown file format
-        """
-        # call function to set mesh file type: self.mesh_filetype
-        self.mesh_filetype = determine_file_type(self.filepath)
-
+    def __from_ds__(self):
```
erogluorhan, line 130: Could you please reflect this function renaming in documentation, docs/internal_api/index.rst?

uxarray/helpers.py
```diff
@@ -1,91 +1,56 @@
-import os
 import xarray as xr
 from pathlib import PurePath
 
 
-# helper function to find file type
-def determine_file_type(filepath):
-    """Checks file path and contents to determine file type. Supports detection
-    of UGrid, SCRIP, Exodus and shape file.
+def parse_grid_type(filepath, **kw):
```
erogluorhan, line 5: Could you please reflect this function renaming in documentation, docs/user_api/index.rst?

uxarray/helpers.py
```diff
@@ -1,91 +1,56 @@
...
+        mesh_filetype = "scrip"
+    # ugrid topology
+    elif _is_ugrid(xr_ds):
+        mesh_filetype = "ugrid"
+    else:
+        raise RuntimeError(f"Could not recognize {filepath} format.")
+    return mesh_filetype, xr_ds
+
+
+def _is_ugrid(ds):
```
erogluorhan, line 43: Could you please add this new function into the documentation, docs/internal_api/index.rst?

uxarray/grid.py
```diff
@@ -84,18 +81,11 @@ def __init__(self, *args, **kwargs):
...
-            if not os.path.isfile(data_arg):
-                raise RuntimeError("File not found: " + data_arg)
-
-            self.filepath = data_arg
-            # call the appropriate reader
-            self.__from_file__()
-
-        # check if invalid initialization
+        elif isinstance(data_arg, xr.Dataset):
+            self.xr_ds = data_arg
```
erogluorhan, line 85: I am concerned about keeping this as an instance attribute because it introduces redundant memory use taht'd be an issue with large datasets and we'd not actually need this original dataset any longer after we construct the UGRID-aware one.

erogluorhan (commented):
I like this new way that resembles Xarray open dataset better and am fine with most of the code changes but have a few inline comments to be addressed please.

I am looping in the dev team for their reviews.

uxarray/helpers.py
ocefpaf: Done!

uxarray/helpers.py
ocefpaf: Done!

uxarray/grid.py
ocefpaf: I prefer an issue or a roadmap with that instead of "dead code parts or TODOs." It is easy to get lost and never address them. With that said, if I can revert that if you prefer it.

erogluorhan: That sounds good to me. FYI: Since the class constructor was being directly called by the user until this PR, the changing its signature would lead to broken calls on the user end, but now since we are planning the user to call the `open_dataset()` instead, this approach would work. Thanks!

uxarray/grid.py
dcherian: If you assign to `self.ds` then it'll just be overwritten after the "decode" step (e.g. `read_script`)?

uxarray/grid.py
```diff
@@ -38,8 +37,8 @@ def __init__(self, *args, **kwargs):
         Parameters
         ----------
 
-        data_arg : string, ndarray, list, tuple, required
-            - Input file name with extension or
+        data_arg : xarray.Dataset, ndarray, list, tuple, required
+            - Input xarray.Dataset or
```
dcherian, line 41:
```suggestion
        dataset : xarray.Dataset, ndarray, list, tuple, required
            - Input xarray.Dataset or
```

really minor comment...

uxarray/grid.py
erogluorhan: I think that'd make the code confusing and less readable (`self.ds` is the UGRID-aware ds after all the decoding, but it'd seem like initiated with the original ds coming from the data ingest. One would need to track all the way down to see `self.ds` is indeed the ugrid-aware one)

dcherian: OK good point; could just pass `dataset` (or `data_arg`) as arg to `self._from_ds__`

uxarray/dataset.py
```diff
@@ -1,10 +1,10 @@
 """uxarray dataset module."""
 
 from .grid import *
-from pathlib import PurePath
+from .helpers import parse_grid_type
 
 
-def open_dataset(grid_filename, *args):
+def open_dataset(grid_filename, *args, **kw):
```
dcherian, line 7: Not for this PR but IMO we should consider deleting this.

There are ways to create an xarray Dataset of unstructured mesh data without going through `open_dataset` or `open_mfdataset`: for example, through `intake` or `fsspec`. 

And as Filipe had to thread through `**kw` here, simple `open_dataset` and `open_mfdataset` calls can fail (especially the latter!) without special kwargs.

uxarray/grid.py
ocefpaf: I renamed it to dataset everywhere for consistency. We are now passing it as an arg instead of a class attribute, and we close it after the `__from_ds__()` call just in case. Hope that is alright for now.

erogluorhan (approved):
This looks great to me! Thanks very much for offering and making this happen!

rajeeja (approved):
Thanks for the cleanup. Lots of this original code my writing during the inception of the repo.
Looking forward to more from you!

PR #72, opened 2022-07-15
Fix potential bugs due to the use of self.ds.Mesh2_node_x by replacing it with s…

erogluorhan:
Fix by replacing `self.ds.Mesh2_node_x` with `self.Mesh2_node_x` and likewise for y and z.


rajeeja (approved):
Looks good, this was in the era of pre `self.Mesh2..`

PR #75, opened 2022-07-16
Off integration

erogluorhan:
This PR does two things:

1. When Numba JIT is enabled, debugging and profiling those functions decorated with @njit is disabled. For developer's debugging purposes, added a `config.DISABLE_JIT = False` to two scripts where there are njit decorators. It has no effect when it is false, but when the developer wants to debug those functions, they should set this to True IN THEIR LOCAL, but it should always be False in the `main` branch to enable JIT
2. Just alphabetically ordered the `__init...`, private, and public function groups in grid.py and helpers.py.

### DIFF
--- a/uxarray/get_quadratureDG.py
+++ b/uxarray/get_quadratureDG.py
@@ -1,5 +1,7 @@
 import numpy as np
-from numba import njit
+from numba import njit, config
+
+config.DISABLE_JIT = False
 
 
 @njit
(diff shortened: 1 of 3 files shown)


anissa111:
Orhan, I think you need to merge integration in here, it seems like you're pulling along some extra files from previous off_integrate PRs

erogluorhan:
> Orhan, I think you need to merge integration in here, it seems like you're pulling along some extra files from previous off_integrate PRs

Thanks! That was indeed because I alphabetically ordered all the functions in grid.py and helpers.py, which left me with several conflicts against your changes :) I couldn't fix them all and thought starting off a new branch and new PR will be much easier

PR #77, opened 2022-07-16
Add config.DISABLE_JIT that can be set to True by developers for debug

erogluorhan:


erogluorhan:
@rajeeja since this is a small change that can easily be revertible, and since there is one more change after this I'd like to open to a discussion, let me merge this and create the other. Still let me know if you have any concern with this one.

rajeeja:
Thanks, these are timely.

PR #78, opened 2022-07-16
Replace for loop in `get_all_face_area_from_coords` with `xr.apply_ufunc`

erogluorhan:
What do you all (@rajeeja  @dcherian @anissa111 ) think about this?

### DIFF
--- a/uxarray/helpers.py
+++ b/uxarray/helpers.py
@@ -3,7 +3,9 @@ import xarray as xr
 from pathlib import PurePath
 
 from .get_quadratureDG import get_gauss_quadratureDG, get_tri_quadratureDG
-from numba import njit
+from numba import njit, config
+
+config.DISABLE_JIT = False
 
 
 def parse_grid_type(filepath, **kw):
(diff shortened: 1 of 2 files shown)


uxarray/helpers.py
```diff
@@ -248,26 +248,22 @@ def get_all_face_area_from_coords(x,
...
-            face_z = z[face_nodes[i]]
 
-        # After getting all the nodes of a face assembled call the  cal. face area routine
-        face_area = calculate_face_area(face_x, face_y, face_z, quadrature_rule,
-                                        order, coords_type)
-
-        area[i] = face_area
-
-    return area
+    areas = xr.apply_ufunc(
```
dcherian, line 252: numba can't compile functions with xarray objects so this won't work. 

You'll have to follow https://tutorial.xarray.dev/advanced/apply_ufunc/vectorize_1d.html#high-performance-vectorization-gufuncs-numba-guvectorize and stick the compiled `get_all_face_area_from_coords` in the `apply_ufunc` call

erogluorhan: Thanks! I can have a look at it later and incorporate. Hi @rajeeja , we can ignore this for now and follow up with other comments in the integration PR #33 to get it merged.

PR #79, opened 2022-07-19
Updates to use new functionality

michaelavs:


test/test_dataset.py
```diff
@@ -28,14 +28,13 @@ def test_open_grid_only(self):
         ux_ds3 = ux.open_dataset(uds3_name)
 
         # get node names for each grid object
```
erogluorhan, line 30: This comment ("node names" ) does no longer make sense (the actual data, not the names, are being initialized now). It can just be removed, and below three variables could also be removed and `ux_ds1.Mesh2_node_` could be directly used in assertion lines.

test/test_dataset.py
```diff
@@ -44,9 +43,9 @@ def test_open_single_dataset(self):
 
         uds3 = ux.open_dataset(uds3_name, uds3_data_name1)
 
-        n3 = uds3.ds_var_names["Mesh2_node_x"]
+        n3 = uds3.Mesh2_node_x
```
erogluorhan, line 46: I think the name "n3" was also referring to the fact that this was the variable name, but it is the variable itself from now on with this change. So, naming it as "n3" won't make sense. The same comment above could apply to this one; no need for such a variable definition, and `uds3.Mesh2_node_x` could be directly used below I think.

test/test_dataset.py
```diff
@@ -55,9 +54,9 @@ def test_open_multiple_dataset(self):
 
         uds3 = ux.open_dataset(uds3_name, uds3_data_name1, uds3_data_name2)
 
-        n3 = uds3.ds_var_names["Mesh2_node_x"]
+        n3 = uds3.Mesh2_node_x
```
erogluorhan, line 57: Same as above

test/test_ugrid.py
```diff
@@ -24,16 +24,13 @@ def test_read_ugrid(self):
...
-        ux_grid2_node_x_var = ux_grid2.ds_var_names["Mesh2_node_x"]
-        ux_grid3_node_x_var = ux_grid3.ds_var_names["Mesh2_node_x"]
-
-        assert (
-            ux_grid1.ds[ux_grid1_node_x_var].size == constants.NNODES_outCSne30)
-        assert (ux_grid2.ds[ux_grid2_node_x_var].size ==
-                constants.NNODES_outRLL1deg)
-        assert (ux_grid3.ds[ux_grid3_node_x_var].size ==
-                constants.NNODES_ov_RLL10deg_CSne4)
+        ux_grid1_node_x_var = ux_grid1.Mesh2_node_x
```
erogluorhan, line 27: The same with above comments applies here, too.

erogluorhan (commented):
Thanks for doing this! Could you please address a few inline comments as well as a request below?

- This is not actually the scope of this PR, but could you please rename the test case `test_init_ds_var_methods` to `test_init_grid_var_attrs` in test_grid.py? (I think it got forgotten in a prior PR even though we decided to rename the grid method). Its docstring could take a few typo corrections as well.

test/test_grid.py
```diff
@@ -43,8 +43,8 @@ def test_init_verts(self):
         face_filename = current_path / "meshfiles" / "1face.ug"
         vgrid.write(face_filename)
 
-    def test_init_ds_var_methods(self):
-        """Tests to see if accessing variables though set attribtues is equal
+    def test_init_grid_var_attrs(self):
+        """Tests to see if accessing variables though set attributes is equal
```
erogluorhan, line 47:
```suggestion
        """Tests to see if accessing variables through set attributes is equal
```

erogluorhan (approved):

philipc2 (approved):

PR #81, opened 2022-07-22
o Add changes promped by Deepak.

rajeeja:
o Changes allow user to specify quadrature rule and order for property

### DIFF
--- a/uxarray/grid.py
+++ b/uxarray/grid.py
@@ -203,7 +203,7 @@ class Grid:
         """
 
         # call function to get area of all the faces as a np array
-        face_areas = self.get_face_areas(quadrature_rule, order)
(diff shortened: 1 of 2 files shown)


uxarray/grid.py
```diff
@@ -335,7 +335,11 @@ def get_face_areas(self, quadrature_rule="triangular", order=4):
         return self._face_areas
 
     # use the property keyword for declaration on face_areas property
-    face_areas = property(get_face_areas)
+    @property
+    def face_areas(self):
+        if self._face_areas is None:
+            self.compute_face_areas()
+            return self._face_areas
```
dcherian, line 342:
```suggestion
        return self._face_areas
```

PR #84, opened 2022-07-26
Fix the upstream CI and Sphinx 5.1.0 issue

erogluorhan:
- Add numba and pinned numpy to upstream ci. 
- Remove python from ci
- RTD recently switched to Sphinx 5.1.0, which led to docs build issues with functions without docstrings. Fix that for `face_areas`


michaelavs (approved):
Looks good to me!

rajeeja (approved):

PR #85, opened 2022-07-28
Updates to meta.yaml for conda-forge build tests

anissa111:
Summary of changes:
- adds the following to the test requirements
    - numba
    - numpy>=1.2,<1.22
    - scipy>=0.16
    - xarray
- adds `conda info` and `conda list` to the test commands section for debugging help in the future
- adds an additional commented out source option for local building and testing

For reference, this was tested only locally after recreating the issues on the conda-forge pipeline on my machine. The errors on the pipeline can be recreated locally by running `conda build .`

We found that conda didn't actually need numpy, numba, or scipy to build the package, but instead the pipeline was only failing on the testing step. Looking at the differences in the development environment where the tests pass and the created testing environment during the build process, these seem to be the minimal changes in order to successfully run the tests during the build process.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -14,6 +14,7 @@ source:
   git_tag: v{{ version }}   # For building from a specific tag
   git_url: https://github.com/UXARRAY/uxarray
   #url: https://github.com/UXARRAY/uxarray/archive/v{{ version }}.tar.gz   # For building from a specific tag's tarball
+  #path: . # for building from a local directory
 
 requirements:
   host:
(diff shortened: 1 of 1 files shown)


michaelavs (approved):

marodrig (approved):

erogluorhan:
Just to understand: Our only intent with having a conda recipe file here in this repo (in addition to one at -feedstock) is for debug/test purposes, right?

anissa111:
Yes, though ideally the one in the feedstock repo should be identical to the one here. Theoretically we could have just updated the feedstock meta.yaml I think, but that would mean we would not be running the same process in local testing and on the deployment azure pipeline.

There should be (or already is?) a corresponding change in a PR for the feedstock repo

Issue #40, opened 2022-04-25
Regarding Grid.filename


erogluorhan:
Our Uxarray API has a `Grid.filename` attribute for Grid that is initiated through files. There are two things to review about it as follows:

1. It is currently implemented to keep track of the full file path (not only file name), see [here](https://github.com/UXARRAY/uxarray/blob/328b4448fbf4bb289882ed15d800be4c0cf12cd0/uxarray/grid.py#L90). Should it be file name or file path; pros/cons with either? 
2. Shouldn't we also keep track of file names/paths when opening multiple dataset files through `open_dataset`. We currently implement this to initiate the `Grid` object using the Grid definition file (so, keeping track of its file path), then other dataset files are just merged to `Grid.ds` [here](https://github.com/UXARRAY/uxarray/blob/328b4448fbf4bb289882ed15d800be4c0cf12cd0/uxarray/dataset.py#L45). My proposal for this line of code could be  something like:

`ux_grid.add_dataset(all_data, args)`

Then, we could define a `add_dataset` function as a helper function for grid.py as something like:

```
def add_dataset(ds_to_add, filename):
     self.ds = xr.merge([self.ds, ds_to_add]) 
     self.filepath = self.filepath + ', ' + filename
```

Or, a different way to keep multiple strings for filename/filepath.

rajeeja:
What is the use case of keeping track of filename or paths?
Once loaded, we're done. No need to keep track or am I missing something?

We might need the "path" for only when saving a file; that also we can assume that we'll save in the original path supplied by the user.

erogluorhan:
Good question, @rajeeja ! I don't have an immediate answer. Any thoughts? @brianpm @zarzycki @clyne @paullric @rljacob

clyne:
If there are multiple files involved in constructing a single data set, maintaining the path names would be helpful for error reporting to the user.

brianpm:
No strong opinion about this. I think the full file name or list of files would be my slight preference. My only use case would be along the lines that John said. If something looks wrong, check that the expected files were used... if still suspicious, go investigate the files directly.

zarzycki:
I'll just note I agree with the above. Not a fan of the potential of leveraging the file path to save any new file in said path (I could envision unexpected behavior there)
... however for diagnostics (either straight up error trap by code itself or valid, but odd, behavior the user must investigate) it seems prudent to be able to pass information about the full path/filename back.

paullric:
[...]

PR #87, opened 2022-08-04
Add source_grid and source_datasets into Grid attributes

erogluorhan:
This PR addresses that tasks mentioned in Issue #40. Namely,

- Adds 'source_grid' and 'source_datasets' attributes into Grid class, for keeping track of source files for diagnostics and reporting purposes only.
- Originally this attribute was anticipated to be named `Grid.filename`, but since there might be multiple files (i.e. grid file and datasets file(s)) and it can be either a file path or URL, I ended up with the names 'source_grid' and 'source_datasets' instead
   - Modified the draft API document to reflect this accordingly
- Adds assertions into the test cases to ensure this behavior works properly
- Updates the docstrings of the `dataset.py` module to fix some forgotten older names and to better reflect the changes of this PR

### DIFF
--- a/docs/user_api/uxarray_api.md
+++ b/docs/user_api/uxarray_api.md
@@ -16,10 +16,6 @@ Describes an unstructured grid.
   MaxNumFacesPerNode (max number of faces per node),Two, Three, Four}`
 
 
-- uxarray.Grid.filename: string \
-  Original filename for this uxarray.Grid.
-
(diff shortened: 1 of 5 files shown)


rajeeja (approved):
Apt changes that provide good reporting and diagnostic variables.

anissa111 (approved):

PR #104, opened 2022-08-12
Write scrip

michaelavs:
Adds SCRIP writer function

Updates User API for helpers.py section:
- Ensures alphabetical ordering
- Updates names for consistency


uxarray/_scrip.py
```diff
@@ -112,3 +113,87 @@ def _read_scrip(ext_ds):
...
+
+    More information on structured vs unstructured SCRIP files can be found here:
+    https://earthsystemmodeling.org/docs/release/ESMF_6_2_0/ESMF_refdoc/node3.html
+
+    Parameters
+    ----------
+    ext_ds : :class:`xarray.Dataset`
+        Original UGRID dataset of interest being used
+
+    outfile : :class:`string`
```
dcherian, line 136: You might considering architecting this as an "encoder" function.

That way you can

```
ds = grid.encode_scrip()  # or various other formats
# Now ds.to_netcdf or ds.to_zarr or various other formats
```

michaelavs:
Did we have a solution for the grid.write() function not being able to detect scrip based on a suffix (as it does for ugrid and exodus)?

uxarray/_scrip.py
erogluorhan: Good suggestion! I've created Issue #109 to address this. This PR could just be merged (if it is ready), and that issue could be addressed afterwards as a refactoring effort. 

@rajeeja @anissa111 @dcherian  thoughts?

uxarray/helpers.py
```diff
@@ -434,3 +434,33 @@ def _is_ugrid(ds):
         return True
     else:
         return False
+
+
+def grid_center_lat_lon(ds):
```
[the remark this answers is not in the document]
michaelavs: When Anissa and I were working on this part of the SCRIP writer function, we felt it would be a good function that users could access outside of just being for creating the grid_center_lat/lon variables in SCRIP files. There is also an implication from the ESMF gist I was looking at that implies having this as a stand alone function would be helpful in both creating ESMF and SCRIP style datasets. That said, I am more than happy to put it into the scrip file and have it as an internal API function instead. Whatever is the preference from everyone else is completely fine by me as well!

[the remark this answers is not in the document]
erogluorhan: Ok, that sounds reasonable! Then, please document its existence in user API accordingly

erogluorhan:
> Did we have a solution for the grid.write() function not being able to detect scrip based on a suffix (as it does for ugrid and exodus)?

I don't know who this question is for, but currently there does not seem to be any condition that is covering the scrip case in  `grid.write()`. @rajeeja @anissa111 any thoughts?

rljacob:
I don't think there is enough reliable metadata in mapping files to tell if one is SCRIP.  We could make that the default:  if its not UGRID or exodus, must be SCRIP.   Else we need a way for the user to optionally specify.

erogluorhan:
> I don't think there is enough reliable metadata in mapping files to tell if one is SCRIP. We could make that the default: if its not UGRID or exodus, must be SCRIP. Else we need a way for the user to optionally specify.

Thanks, @rljacob , this helps! So, I think we can proceed with @michaelavs making that default decision (the former in your suggestion), and with us creating an issue to point this out and explore if we should implement something for the latter.

docs/user_api/index.rst
```diff
@@ -35,5 +35,6 @@ Helper Functions
    helpers.calculate_spherical_triangle_jacobian
    helpers.calculate_spherical_triangle_jacobian_barycentric
    get_gauss_quadratureDG
+   helpers.grid_center_lat_lon
```
michaelavs, line 38: @erogluorhan I just added this function for the user API. I was wondering if I should re-organize this list to be in alphabetical order and add `helpers.` in front of `get_gauss_quadrature` and `get_tri_quadrature`?

erogluorhan: Yes, please do so!

erogluorhan: You can also mention this change in the PR description for future reference.

test/test_scrip.py
```diff
@@ -36,6 +36,32 @@ def test_scrip_is_not_ugrid(self):
         with self.assertRaises(KeyError):
             new_ds['grid_corner_lat']
 
+    def test_scrip_writer(self):
```
erogluorhan, line 39: Neither of these two test cases below seems to be testing against actual values, which definitely needs to be the case. I think the current implementation would give us errors at least for `lat_nodes` and `lon_nodes` calculations (i.e. they should be `double` but our calculation converts them to `int`, which is described in one of the comments below)

If we don't have a SCRIP file for that we have a corresponding UGRID file, I think the way for testing should be:

- Read in a SCRIP file and save it as a UGRID file
- Read in that saved UGRID file, and save it as a SCRIP file
- Compare all of the values in that output SCRIP file, which you calculated in `write_scrip`, to the ones in the original input SCRIP file

erogluorhan (changes requested):
Thanks for this! Please address my inline comments. In addition:

This PR is currently not touching `grid.write` at all. That means, there will not be write to SCRIP functionality from the grid object even if we get this merged. At least, please implement the assumption that was:

_Originally posted by @rljacob in https://github.com/UXARRAY/uxarray/issues/104#issuecomment-1234854790_

and 

_Originally posted by @erogluorhan in https://github.com/UXARRAY/uxarray/issues/104#issuecomment-1235583964_

After then, Issue #114 can be implemented by you or someone else.

uxarray/_scrip.py
```diff
@@ -112,3 +114,91 @@ def _read_scrip(ext_ds):
...
+    y_val = grid.Mesh2_node_y
+    x_val = grid.Mesh2_node_x
+
+    # Create empty arrays to hold lat/lon data
+    lat_nodes = np.zeros_like(f_nodes)
+    lon_nodes = np.zeros_like(f_nodes)
+
+    for i in range(len(f_nodes)):
+        lat_nodes[i] = y_val[int(f_nodes[i])]
+        lon_nodes[i] = x_val[int(f_nodes[i])]
```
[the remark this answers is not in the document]
michaelavs: I updated grid_corner and grid_center lat/lon to be type `float` (which I believe is the Python version of a double, correct me if I'm wrong). As for the suggested change for lat/lon nodes, when trying to implement that, there is an issue with the dimension of the arrays which prevents it from working. So I have just left that part as is for the time being.

test/test_scrip.py
michaelavs: I updated this test, let me know if there should be any further changes to it. I'll mark this as resolved if it looks good to you

michaelavs:
> Thanks for this! Please address my inline comments. In addition:
> 
> This PR is currently not touching `grid.write` at all. That means, there will not be write to SCRIP functionality from the grid object even if we get this merged. At least, please implement the assumption that was:
> 
> _Originally posted by @rljacob in [#104 (comment)](https://github.com/UXARRAY/uxarray/pull/104#issuecomment-1234854790)_
> 
> and
> 
> _Originally posted by @erogluorhan in [#104 (comment)](https://github.com/UXARRAY/uxarray/pull/104#issuecomment-1235583964)_
> 
> After then, Issue #114 can be implemented by you or someone else.

I have updated grid.write to now take a user input for what the desired output format will be and updated relevant tests for this change

uxarray/grid.py
```diff
@@ -160,32 +160,35 @@ def __from_ds__(self, dataset):
...
                 raise RuntimeError("File directory not found: " + outfile)
 
-        if extension == ".ugrid" or extension == ".ug":
+        if extension == "ugrid":
             _write_ugrid(self.ds, outfile, self.ds_var_names)
-        elif extension == ".g" or extension == ".exo":
+        elif extension == "exodus":
             _write_exodus(self.ds, outfile, self.ds_var_names)
+        elif extension == "scrip":
+            _write_scrip(self, outfile)
```
erogluorhan, line 191: This is not how it is supposed to be called.

test/test_helpers.py
```diff
@@ -60,3 +62,19 @@ def test_quadrature(self):
...
+        scrip_center_lon = ds_ne8['grid_center_lon']
+        scrip_center_lat = ds_ne8['grid_center_lat']
+
+        # Calculate the center_lat/lon using same dataset's corner_lat/lon
+        calc_center = ux.grid_center_lat_lon(ds_ne8)
+        calc_lat = calc_center[0]
+        calc_lon = calc_center[1]
+
+        # Test that calculated center_lat/lon is the same as actual center_lat/lon
+        self.assertEqual(scrip_center_lat.any(), calc_lat.any())
```
erogluorhan, line 79: Should test for the actual array values, not for `arrays.any()`

erogluorhan (commented):
Please address the changes needed

erogluorhan (approved):

anissa111 (approved):

PR #156, opened 2022-10-05
Unify coordinates conversion

hongyuchen1030:
Unifying the coordination fuctions:

Get rid of the `_spherical_to_cartesian_unit_`
And use `_convert_node_xyz_to_lonlat_rad` and `_convert_node_lonlat_rad_to_xyz` all the time


uxarray/helpers.py
```diff
@@ -100,33 +100,6 @@ def parse_grid_type(filepath, **kw):
     return mesh_filetype
 
 
-@njit
-def _spherical_to_cartesian_unit_(node, r=6371):
-    """Converts spherical (lat/lon) coordinates to cartesian (x,y,z).
-
-    Final output is cartesian coordinates on a sphere of unit radius
```
rljacob, line 107: Are these cartesian coordinates on a cube that the sphere is enclosed in?

rajeeja: we removed the above, these are not on a cube; our intent is just to convert from degree or radians to cartesian for area calculation.

rajeeja (approved):

PR #160, opened 2022-10-18
Fesom Support

pgierz:
Dear Colleagues,

thanks for developing an xarray extension with ugrid support! We would be interested in contributing :-)

At AWI we are developing and using a finite volume ocean model which runs on an unstructured grid, [FESOM](https://github.com/FESOM). We would be happy if our data could also be included for brainstorming and further developing uxarray. I've included a very small set of test data, as well as fixing one mistake in the `compute_face_areas` function. We would be happy to contribute more (either test data or even manpower) since we currently rely mostly on an internally developed toolbox, and if we can instead contribute to the community I would be very happy!

All the best,
Paul

Info also for @koldunovn


rajeeja:
Hi Paul, @pgierz 

Thanks for this. We are in the middle of a code reorganization and will review this soon. [ fix ](https://github.com/UXARRAY/uxarray/pull/160/commits/232cd4e569ffdf1d989c3a9b986cd2fcba8b442c#diff-93f6ca66ccba77109d776037f2925e663079c29d5c68233ec90ab1b58258cec0R303) is a good one.

philipc2:
Hi Paul,

Adding to what @rajeeja said, the fix in `compute_face_areas` highlighted an inconsistency in our data access attributes. The current functionality only handles UGRID variables (Mesh2_node_x, Mesh2_face_nodes, etc) and doesn't support UGRID dimensions (nMesh2_node, nMesh2_face, etc). I'll look into `__init_grid_var_attrs__` and add support for dimensions. 

Great work and we're looking forward to hearing more from you all!

erogluorhan:
Hi @pgierz, it's great to see such a contribution! This is very helpful for pushing UXarray further! Please keep contributing to it and discussing your use-cases, etc!

uxarray/grid.py
```diff
@@ -308,7 +308,8 @@ def compute_face_areas(self, quadrature_rule="triangular", order=4):
             dim = self.Mesh2.attrs['topology_dimension']
 
             # initialize z
-            z = np.zeros((self.ds.nMesh2_node.size))
+            da_nMesh2_node = getattr(self.ds, self.ds_var_names['nMesh2_node'])
+            z = np.zeros((da_nMesh2_node.size))
```
erogluorhan, line 312: You can use the following code instead of these two lines, using the new functionality @philipc2 implemented in PR #161:

`z = np.zeros((self.nMesh2_node)) `

test/test_face_areas.py
```diff
@@ -0,0 +1,24 @@
...
+import xarray as xr
+
+from unittest import TestCase
+from pathlib import Path
+
+import uxarray as ux
+
+current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+
+class TestFaceAreas(TestCase):
```
erogluorhan, line 12: We write one test module (with possible several test classes with multiple test cases in each) per one code module (I just noticed though we oversaw this with `test_integrate.py` as well; we can fix that, too). 

So, could you please remove this test module, instead put this `TestFaceAreas ` class into the `test_grid.py` module?

uxarray/grid.py
erogluorhan, line 312: Could you please use the following code instead of these two lines, using the new functionality @philipc2 just implemented in PR #161 for this PR:

`z = np.zeros((self.nMesh2_node.size)) `

erogluorhan (changes requested):
Thanks very much for this PR! Could you please address the inline comments and just a few requests as follows:

- pre-commit tests for code quality currently fails. Please look at the [UXarray Contributor's Guide](https://uxarray.readthedocs.io/en/latest/contributing.html), and then it should be an easy fix.
- Please document the Fesom test files under `/test/meshfiles/README.md`

test/test_face_areas.py
pgierz: OK! I can do that for `test_integrate` as well in this PR. I guess it is the entire `class TestIntegrate` that needs to go in the other file?

uxarray/grid.py
pgierz: OK, done.

test/test_face_areas.py
pgierz: Ah, that's already done, I forgot to remerge main.

pgierz:
Hi @erogluorhan, I think everything should be good now.

pgierz:
Hi @erogluorhan, all done 😄. All tests also locally pass on my system, so if you are happy this could go in.

test/test_face_areas.py
erogluorhan: That's correct; the entire class!

ci/environment.yml
```diff
@@ -13,3 +13,6 @@ dependencies:
   - pytest-cov
   - scipy
   - xarray
+  # Tools run by pre-commit so they can be in the same conda environment
```
erogluorhan, line 16: Conda environment is supposed to represent the minimum required dependencies for development purposes. These tools are configured and run by pre-commit through the [.pre-commit-config.yaml](https://github.com/UXARRAY/uxarray/blob/main/.pre-commit-config.yaml) file and do not need to be in the conda environment.

erogluorhan (commented):
This is great overall and almost there to be merged. Please address just one inline comment

erogluorhan (approved):

PR #162, opened 2022-10-20
Load grids with xarray and then create uxarray grid object

rajeeja:
#151 and Discussion: https://github.com/UXARRAY/uxarray/discussions/154


uxarray/grid.py
```diff
@@ -304,7 +304,7 @@ def compute_face_areas(self, quadrature_rule="triangular", order=4):
             if not "degree" in self.Mesh2_node_x.units:
                 coords_type = "cartesian"
 
-            face_nodes = self.Mesh2_face_nodes.data
+            face_nodes = self.Mesh2_face_nodes.data.astype(np.int64)
```
philipc2, line 307: Explicitly setting face nodes as integers is great, however there are some small inconsistencies throughout the code (some sections call for `np.int64` which others only `int`)

test/test_helpers.py
```diff
@@ -30,7 +30,7 @@ def test_face_area_coords(self):
         x = np.array([0.57735027, 0.57735027, -0.57735027])
         y = np.array([-5.77350269e-01, 5.77350269e-01, 5.77350269e-01])
         z = np.array([-0.57735027, -0.57735027, -0.57735027])
-        face_nodes = np.array([[0, 1, 2]])
+        face_nodes = np.array([[0, 1, 2]]).astype(int)
```
philipc2, line 33: See previous comment

uxarray/helpers.py
```diff
@@ -250,16 +241,15 @@ def get_all_face_area_from_coords(x,
     """
     num_faces = face_nodes.shape[0]
     area = np.zeros(num_faces)  # set area of each face to 0
-
     for i in range(num_faces):
-
+        # print(i, face_nodes[i])
```
philipc2, line 245: Remove

uxarray/helpers.py
```diff
@@ -250,16 +241,15 @@ def get_all_face_area_from_coords(x,
...
         face_z = np.zeros(len(face_nodes[i]))
 
-        face_x = x[face_nodes[i]]
-        face_y = y[face_nodes[i]]
+        face_x = x[face_nodes[i].astype(np.int64)]
+        face_y = y[face_nodes[i].astype(np.int64)]
         # check if z dimension
         if dim > 2:
-            face_z = z[face_nodes[i]]
+            face_z = z[face_nodes[i].astype(np.int64)]
```
philipc2, line 252: Might be better to set the type to `int` outside of the loop, instead of doing it while indexing

philipc2 (commented):
Great work @rajeeja! I've left a couple comments about some slightly inconsistencies with casting face nodes to integers. Other than that, the implementation is looking great!

philipc2:
Small suggestion for the `Grid` constructor, but also applies to most of our other functions: Do we want to implement type hints?

Our current constructor for grid is:
```__init__(self, dataset, **kwargs):```

A simple change to emphasize that we are using `xr.Dataset` would be as follows
```__init__(self, dataset : xr.Dataset, **kwargs)```:

File: uxarray/grid.py [hunk 5]
```diff
@@ -70,32 +68,88 @@ def __init__(self, dataset, **kwargs):
...
             self.vertices = dataset
             self.__from_vert__()
             self.source_grid = "From vertices"
             self.source_datasets = None
-
         # check if initializing from string
         # TODO: re-add gridspec initialization when implemented
         elif isinstance(dataset, xr.Dataset):
+            self.ds = dataset
+            self.mesh_filetype = parse_grid_type(dataset, **kwargs)
```
erogluorhan, line 84: Would `mesh_type` be a better name as we dropped file reading support?

uxarray/helpers.py
```diff
@@ -8,14 +8,14 @@
...
+def parse_grid_type(dataset, **kw):
     """Checks input and contents to determine grid type. Supports detection of
     UGrid, SCRIP, Exodus and shape file.
 
     Parameters
     ----------
-    filepath : str
-       Filepath of the file for which the filetype is to be determined.
+    dataset : Xarray dataset
+       Xarray dataset of the file input
```
erogluorhan, line 18: The term "file input" might be confusing. As we are not getting a file inut anymore, let's say "... of the grid."

uxarray/grid.py
erogluorhan, line 84: The `**kwargs` in the functon call was being input thorugh `open_dataset` before. I believe feeding the Grid constructor's `**kwargs` into `parse_grid_type ` wouldn't fit or be needed.

uxarray/grid.py
```diff
@@ -155,9 +209,10 @@ def __from_ds__(self, dataset):
         elif self.mesh_filetype == "ugrid":
             self.ds, self.ds_var_names = _read_ugrid(dataset, self.ds_var_names)
         elif self.mesh_filetype == "shp":
-            self.ds = _read_shpfile(self.filepath)
+            self.ds = _read_shpfile(dataset)
```
erogluorhan, line 212: This might need a revisit; maybe we should open an issue for this because:

- There won't be a shape file reading anymore (We will instead read in a dataset(?) that is already constructed from a shape file, how?)
- Our parse will ned to dete a shape file from such a dataset, not from a file extension, etc.

uxarray/grid.py
```diff
@@ -155,9 +209,10 @@ def __from_ds__(self, dataset):
...
         else:
-            raise RuntimeError("unknown file format: " + self.mesh_filetype)
+            raise RuntimeError("unknown file format: ")
```
erogluorhan, line 214: "unknown mesh type"

erogluorhan (changes requested):
Thank very much for doing this!  Please address a few things as follows:

- Reflect the removal of the `dataset` functionality in the documentation (e.g. user API, draft UXarray API, etc.) 
- Inline comments

uxarray/_exodus.py
```diff
@@ -221,7 +221,7 @@ def _encode_exodus(ds, ds_var_names, outfile=None):
     # process face nodes, this array holds num faces at corresponding location
     # eg num_el_all_blks = [0, 0, 6, 12] signifies 6 TRI and 12 SHELL elements
     num_el_all_blks = np.zeros(ds[ds_var_names["nMaxMesh2_face_nodes"]].size,
-                               "i4")
+                               "i8")
```
philipc2, line 224: Since face nodes are always natural numbers, would it make more sense to use unsigned integers `u8` instead of signed integers?

philipc2: Additionally, do we need 64 bit precision for our face nodes? Using a 32 bit unsigned integer would cap our max number of face nodes at slightly over 4 billion. Is there any scenario where we would have a grid with over 4 billion face nodes?

erogluorhan (approved):
Thanks very much for doing this!

uxarray/helpers.py
```diff
@@ -251,8 +242,9 @@ def get_all_face_area_from_coords(x,
     num_faces = face_nodes.shape[0]
     area = np.zeros(num_faces)  # set area of each face to 0
 
-    for i in range(num_faces):
+    face_nodes = face_nodes.astype(np.int64)
```
philipc2, line 245: We should check if our data is already stored as an integer to avoid unnecessary conversions if our data is another integer (uint32, int32, uint64, etc). 

Can do something like this:
```Python
if not np.issubdtype(face_nodes.dtype, np.integer):
    face_nodes = face_nodes.astype(np.int64)    

```

uxarray/_scrip.py
```diff
@@ -182,7 +182,7 @@ def _encode_scrip(mesh2_face_nodes, mesh2_node_x, mesh2_node_y, face_areas):
     ds = xr.Dataset()
 
     # Make grid corner lat/lon
-    f_nodes = mesh2_face_nodes.values.ravel()
+    f_nodes = mesh2_face_nodes.values.astype(np.int64).ravel()
```
philipc2, line 185: Same as previous comment. 

Fix with something like this:
```Python
if not np.issubdtype(mesh2_face_nodes.dtype, np.integer):
    mesh2_face_nodes = mesh2_face_nodes.astype(np.int64)    

f_nodes = mesh2_face_nodes.values.ravel()

```

uxarray/grid.py
```diff
@@ -304,7 +298,7 @@ def compute_face_areas(self, quadrature_rule="triangular", order=4):
             if not "degree" in self.Mesh2_node_x.units:
                 coords_type = "cartesian"
 
-            face_nodes = self.Mesh2_face_nodes.data
+            face_nodes = self.Mesh2_face_nodes.data.astype(np.int64)
```
philipc2, line 301: Same as previous comments

uxarray/_exodus.py
```diff
@@ -221,12 +221,12 @@ def _encode_exodus(ds, ds_var_names, outfile=None):
...
     # this list stores connectivity without filling
     conn_nofill = []
 
     # store the number of faces in an array
-    for row in ds[ds_var_names["Mesh2_face_nodes"]].data:
+    for row in ds[ds_var_names["Mesh2_face_nodes"]].astype(np.int64).data:
```
philipc2, line 229: Same as previous comments.

philipc2 (changes requested):
Thank you once again for putting this together @rajeeja. A few more comments on casting to integers, but nothing directly related to functionality. Great work!

philipc2 (approved):
Changes look good! Great work

Issue #166, opened 2022-10-28
Grid.nMesh2_node and Grid.nMesh2_face returns range, not total count


philipc2:
Calling `Grid.nMesh2_node` and `Grid.nMesh2_face` returns a Numpy array with range `[0 to nMesh2_node)` and `[0 to nMesh2_face)` respectively. 

Current Behavior:
```Python
>>> n_nodes = ugrid.nMesh2_node.values
array([   0,    1,    2, ..., 5997, 5998, 5999], dtype=int64)
```

Expected Behavior:
```Python
>>> n_nodes = ugrid.nMesh2_node.values
6000
```

Edit:
Also applies to `ugrid.nMaxMesh2_face_nodes`

erogluorhan:
This is a great catch! I had also though about this while working PRs #160 and #161 but didn't attempt further actions. The expected behavior you mention would save memory usage.

philipc2:
> This is a great catch! I had also though about this while working PRs #160 and #161 but didn't attempt further actions. The expected behavior you mention would save memory usage.

I'll work on getting a hotfix in by the end of today!

rajeeja:
> > This is a great catch! I had also though about this while working PRs #160 and #161 but didn't attempt further actions. The expected behavior you mention would save memory usage.
> 
> I'll work on getting a hotfix in by the end of today!

Well done. thanks!

PR #168, opened 2022-10-28
Fix dimension attributes returning a range instead of counts

philipc2:
Fix to issue #166


erogluorhan (approved):
Thanks very much for doing this!

rajeeja (approved):

Issue #141, opened 2022-09-27
grid validate function


rajeeja:

rajeeja:
Things to handle in this PR:
=======
```
        xr_dataset = xr.open_dataset(ug_30)
        grid = ux.Grid(xr_dataset)
        assert (grid.validate())
```
Check validity BASIC (Eg. uxds.ugrid.validate)
 - check if the mesh conforms to UGRID conventions (minimum set of keys required by UGRID)
 - check for duplicate nodes
 - check the connectivity of the mesh 

Check validity COMPREHENSIVE ((Eg. uxds.ugrid.validate_all)
 - check if the mesh conforms to UGRID conventions (minimum set of keys required by UGRID)
 - check for duplicate nodes
 - check the connectivity of the mesh 
 - area non zero (this and following steps can be expensive and not included in the basic)
 - Jacobian non zero
 - angle between faces not greater than 180 degrees
  
Report mesh quality (Eg. uxds.ugrid.quality)
- area
- aspect ratio
- Jacobian
- angle between faces

=======
TACKLE THIS LATER:

Fix bad mesh
 - Remove duplicate nodes
 - Remove duplicate elements
 - Fix concave elements
 - Mesh refinement, derefinement, smoothing etc.

Please review and suggest if you want to add or remove things from this list.

PR #173, opened 2022-11-08
validate mesh

rajeeja:
#141

### DIFF
--- a/test/test_grid.py
+++ b/test/test_grid.py
@@ -30,6 +30,13 @@ class TestGrid(TestCase):
     tgrid2 = ux.Grid(xr_ds2)
     tgrid3 = ux.Grid(xr_ds3)
 
+    def test_validate(self):
+        """Test to check the validate function."""
+        path = current_path / "meshfiles" / "outCSne30.ug"
+        xr_dataset = xr.open_dataset(path)
+        grid = ux.Grid(xr_dataset)
+        assert (True == grid.validate())
+
     def test_encode_as(self):
         """Reads a ugrid file and encodes it as `xarray.Dataset` in various
         types."""
--- a/uxarray/grid.py
+++ b/uxarray/grid.py
@@ -8,7 +8,7 @@ from ._exodus import _read_exodus, _encode_exodus
 from ._ugrid import _read_ugrid, _encode_ugrid
 from ._shapefile import _read_shpfile
 from ._scrip import _read_scrip, _encode_scrip
-from .helpers import get_all_face_area_from_coords, parse_grid_type
+from .helpers import get_all_face_area_from_coords, parse_grid_type, _is_ugrid
 
 int_dtype = np.uint32
 
@@ -210,6 +210,21 @@ class Grid:
             raise RuntimeError("unknown mesh type")
 
         dataset.close()
+
+    def validate(self):
+        """Validate a grid object as per UGRID conventions.
+        Raises
+        ------
+        RuntimeError
+            If unsupported grid type provided
+        """
+        # If the mesh file is loaded correctly, we have the underlying file format as UGRID
+        # Test if the file is a valid ugrid file format or not
+        valid = _is_ugrid(self.ds)
+
+        # TODO: Add more checks
+
+        return valid
 
     def encode_as(self, grid_type):
         """Encodes the grid as a new `xarray.Dataset` per grid format supplied


rajeeja:
> The flow suggested in this PR (i.e. calling `validate()` through `Grid` object) does not make sense to me because we are already creating any `Grid`, representing in the UGRID conventions, regardless of the original grid type coming from the file. In other words, any dataset that does not conform to UGRID standards would lead our code to throw exceptions in the `Grid` creation stage, wouldn't it?
> 
> I'd assume a flow instead like: `ux.validate(xr.open_dataset(file))` (This would require us to modify the draft UXarray API, too)
> 
> Am I missing anything here?

good suggestion, I will try to tweak and implement the same.

uxarray/grid/grid.py
```diff
@@ -193,6 +196,62 @@ def from_face_vertices(cls,
...
+        unique_nodes, indices = np.unique(coords1, axis=0, return_index=True)
+        duplicate_indices = np.setdiff1d(np.arange(len(coords1)), indices)
+        if duplicate_indices.size > 0:
+            print("WARNING: Duplicate nodes found in the mesh. # ",
+                  duplicate_indices.size, " nodes are duplicates.")
+            return True
+        else:
+            print("No duplicate nodes found in the mesh.")
+            return False
+
```
philipc2, line 254: Can we move these validation helper functions into a separate module, maybe `grid/valdiation.py` 

We should keep the `Grid.validate` function where it is currently.

rajeeja: will do, good idea.

uxarray/grid/validation.py
```diff
@@ -0,0 +1,44 @@
...
+
+# validation helper functions
+def check_connectivity(self):
+    """Check if all nodes are referenced by at least one element.
+
+    If not, the mesh may have hanging nodes and may not a valid UGRID
+    mesh
+    """
+
+    # Check if all nodes are referenced by at least one element
```
aaronzedwick, line 14: Should this check if all nodes are referenced by at least two elements? Wouldn't it still be an invalid UGRID mesh if there was just a single line connecting to it?

rajeeja: > Should this check if all nodes are referenced by at least two elements? Wouldn't it still be an invalid UGRID mesh if there was just a single line connecting to it?

Sometimes there is ocean or land mesh, basically anything not fully covering.
There is a metric used in quality `degree,` the number of faces that a node is a part of, the min/max of this number can be reported in a future PR.

aaronzedwick (approved):
This looks good to me, I don't see anything that would cause problems. I am sure there are a lot more features that could probably go into this but definitely a good initial implementation!

uxarray/grid/validation.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
+from uxarray.io._ugrid import _is_ugrid
+from uxarray.constants import ERROR_TOLERANCE
+
+
+# validation helper functions
+def check_connectivity(self):
```
philipc2, line 7: Please include any new functions in our internal and user api

uxarray/grid/grid.py
```diff
@@ -209,6 +214,41 @@ def from_face_vertices(cls,
...
+        check_duplicate_nodes(self)
+        check_connectivity(self)
+
+        # check face area
+        areas = self.face_areas
+        # Check if area of any face is close to zero
+        if np.any(np.isclose(areas, 0, atol=ERROR_TOLERANCE)):
+            print(
+                "-WARNING: At least one face area is close to zero. Mesh may contain inverted elements"
+            )
```
philipc2, line 246: I suggest using the `warnings` module instead of prints. https://docs.python.org/3/library/warnings.html

uxarray/grid/grid.py
```diff
@@ -209,6 +214,41 @@ def from_face_vertices(cls,
...
+        RuntimeError
+            If unsupported grid type provided
+        """
+        # If the mesh file is loaded correctly, we have the underlying file format as UGRID
+        # Test if the file is a valid ugrid file format or not
+        print("Validating the mesh...")
+
+        # check if the grid is a valid ugrid format
+        print("-Valid ugrid format"
+              if _is_ugrid(self._ds) else "-WARNING: Not a valid ugrid format")
```
philipc2, line 234: Same regarding `warnings`

uxarray/grid/grid.py
```diff
@@ -209,6 +214,41 @@ def from_face_vertices(cls,
...
+        # call the check_connectivity and check_duplicate_nodes functions from validation.py
+        check_duplicate_nodes(self)
+        check_connectivity(self)
+
+        # check face area
+        areas = self.face_areas
+        # Check if area of any face is close to zero
+        if np.any(np.isclose(areas, 0, atol=ERROR_TOLERANCE)):
+            print(
+                "-WARNING: At least one face area is close to zero. Mesh may contain inverted elements"
```
philipc2, line 245: Same regarding `warnings`

uxarray/grid/validation.py
```diff
@@ -0,0 +1,44 @@
...
+    if (nodes_in_conn.size == self.nMesh2_node):
+        print("-All nodes are referenced by at least one element.")
+        return True
+    else:
+        print("-WARNING: Some nodes may not referenced by any element.",
+              nodes_in_conn.size, self.nMesh2_node)
+        return False
+
+
+def check_duplicate_nodes(self):
```
philipc2, line 30: This should be an internal method
```suggestion
def _check_duplicate_nodes(self):
```

uxarray/grid/validation.py
philipc2, line 7: This should be an internal method also
```suggestion
def _check_connectivity(self):
```

uxarray/grid/validation.py
philipc2: `self` is also typically reserved for use in classes, use `grid` instead.

uxarray/grid/validation.py
philipc2: same comment about the `self`

uxarray/grid/validation.py
```diff
@@ -0,0 +1,44 @@
...
+    """Check if there are duplicate nodes in the mesh."""
+
+    coords1 = np.column_stack(
+        (np.vstack(self.Mesh2_node_x), np.vstack(self.Mesh2_node_y)))
+    unique_nodes, indices = np.unique(coords1, axis=0, return_index=True)
+    duplicate_indices = np.setdiff1d(np.arange(len(coords1)), indices)
+
+    if duplicate_indices.size > 0:
+        print("-WARNING: Duplicate nodes found in the mesh. # ",
```
philipc2, line 39: Same regarding `warnings`

philipc2 (changes requested):
This is some great functionality, see my comments about a few suggestions.

uxarray/grid/grid.py
```diff
@@ -198,6 +202,37 @@ def from_face_vertices(cls,
...
+
+        # check face area
+        areas = self.face_areas
+        # Check if area of any face is close to zero
+        if np.any(np.isclose(areas, 0, atol=ERROR_TOLERANCE)):
+            warn(
+                "At least one face area is close to zero. Mesh may contain inverted elements",
+                RuntimeWarning)
+        else:
+            print("-No face area is close to zero.")
```
philipc2, line 232: Can we turn this into a `_check_area()` helper, like above?

philipc2 (changes requested):

uxarray/grid/grid.py
rajeeja: good suggestions, thanks for the review, pushed the fixes.

uxarray/grid/grid.py
```diff
@@ -198,6 +200,28 @@ def from_face_vertices(cls,
...
+        # If the mesh file is loaded correctly, we have the underlying file format as UGRID
+        # Test if the file is a valid ugrid file format or not
+        print("Validating the mesh...")
+
+        # call the check_connectivity and check_duplicate_nodes functions from validation.py
+        _check_duplicate_nodes(self)
+        _check_connectivity(self)
+        _check_area(self)
+
+        return True
```
philipc2, line 223: Does this return statement make sense in the context of the above functions? It should only return `True` if all three of the above helpers return true, right?

uxarray/grid/validation.py
```diff
@@ -0,0 +1,60 @@
...
+        warn(
+            "Duplicate nodes found in the mesh. {0} nodes are duplicates.".
+            format(duplicate_indices.size), RuntimeWarning)
+        return False
+    else:
+        print("-No duplicate nodes found in the mesh.")
+        return True
+
+
+def _check_area(self):
```
philipc2, line 49: Add this to internal API

uxarray/grid/validation.py
```diff
@@ -0,0 +1,60 @@
...
+        warn(
+            "Some nodes may not be referenced by any element. {0} and {1}".
+            format(nodes_in_conn.size, self.n_node), RuntimeWarning)
+        return False
+
+
+def _check_duplicate_nodes(self):
+    """Check if there are duplicate nodes in the mesh."""
+
+    coords1 = np.column_stack((np.vstack(self.node_x), np.vstack(self.node_y)))
```
philipc2, line 35: This should be `self.node_lon` and `self.node_lat` if you are intending to use Spherical coordinates (ignore if you meant Cartesian, but if so you should also check `node_z`

philipc2 (changes requested):
Looks great! I'll approve after these two final comments are addressed.

rajeeja:
Test failures seem to come from reliance on this ftp mesh file unavailability:
`
https://web.lcrc.anl.gov/public/e3sm/inputdata/share/meshes/mpas/ocean/oQU480.230422.nc
`
am I missing something @philipc2 ?

philipc2:
> Test failures seem to come from reliance on this ftp mesh file unavailability: `https://web.lcrc.anl.gov/public/e3sm/inputdata/share/meshes/mpas/ocean/oQU480.230422.nc` am I missing something @philipc2 ?

This is correct. See #582

philipc2 (approved):
Great work!

docs/user_api/index.rst
```diff
@@ -154,6 +154,7 @@ IO
    Grid.to_polycollection
    Grid.to_linecollection
    Grid.to_shapely_polygons
+   Grid.validate_mesh
```
philipc2, line 157: One last catch, this should be
```suggestion
   Grid.validate
```

PR #188, opened 2022-12-12
add Grid.__populate_lonlat_coord and Grid.__populate_cartesian_xyz_coord

hongyuchen1030:
This PR request will add two functions in the uxarray.Grid class: `Grid._populate_lonlat_coord`, `Grid._populate_cartesian_xyz_coord`

1.`Grid._populate_lonlat_coord`   populates the longitude and latitude and store
        it into the Mesh2_node_x and Mesh2_node_y. This is called when the
        input data has "Mesh2_node_x", "Mesh2_node_y", "Mesh2_node_z" in
        meters. Since we want "Mesh2_node_x" and "Mesh2_node_y" always have the
        "degree" units.

2. `Grid._populate_cartesian_xyz_coord`  populates the xyz attribute in UXarray.ds.
        This function is called when we need to use the cartesian coordinates
        for each node to do the calculation but the input data only has the
        "Mesh2_node_x" and "Mesh2_node_y" in degree.

The reason why we want this is that In the UXarray, we abide the UGRID convention and make sure the following attributes will always have their corresponding units as stated below:

        `Mesh2_node_x
         unit:  "degree_east"
        Mesh2_node_y
         unit:  "degree_north"
        Mesh2_node_z
         unit:  "m"
        Mesh2_node_cart_x
         unit:  "m"
        Mesh2_node_cart_y
         unit:  "m"
        Mesh2_node_cart_z
         unit:  "m"`

`Grid._populate_cartesian_xyz_coord` will be used in the `Grid.generate_latlon_bounds`  and `Grid_zonal_average`functions. 
`Grid._populate_cartesian_xyz_coord`  will have a pre-check every time it's being called: to check if the ` Mesh2_node_cart_x` already exists so that it won't re-populate.


rajeeja:
Looks good. Thanks for adding this @hongyuchen1030 

Please add tests, add new or extend the current test case(s) that create a mesh with cart-coords to use these functions and get lat/lon and vice-versa?

rajeeja:
Looks good. Thanks for adding this @hongyuchen1030 

How about extending the current test case(s) that create a mesh with cart coords to use these functions and get lat/lon and vice-versa? or create new tests to cover these functions.

test/test_grid.py
```diff
@@ -214,3 +211,42 @@ def test_compute_face_areas_fesom(self):
         grid_2_ds = xr.open_dataset(fesom_grid_small)
         grid_2 = ux.Grid(grid_2_ds)
         grid_2.compute_face_areas()
+
+
+class TestPopulateCoordinates(TestCase):
```
hongyuchen1030, line 216: Hi @rajeeja,

Here're the corresponding test cases for these functions

hongyuchen1030:
> Looks good. Thanks for adding this @hongyuchen1030
> 
> How about extending the current test case(s) that create a mesh with cart coords to use these functions and get lat/lon and vice-versa? or create new tests to cover these functions.

Thanks for your information.

There should already be two test cases in testGrid.py for these two new functions `test_populate_cartesian_xyz_coord` and `test_populate_lonlat_coord`. I have also pinned you at the codes Please let me know if any further changes are needed. Thank you very mcuh.

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        Mesh2_node_cart_z
+         unit:  "m"
+        """
+
+        # Check if the cartesian coordinates are already populated
+        if "Mesh2_node_cart_x" in self.ds.keys():
+            return
+
+        # check for units and create Mesh2_node_cart_x/y/z set to self.ds
+        num_nodes = self.ds.Mesh2_node_x.size
```
philipc2, line 406: Number of nodes is already stored as an attribute in the Grid class: `self.nMesh2_node`

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        for i in range(num_nodes):
+            if "degree" in self.ds.Mesh2_node_x.units:
+                node = [
+                    np.deg2rad(self.ds["Mesh2_node_x"].values[i]),
+                    np.deg2rad(self.ds["Mesh2_node_y"].values[i])
+                ]  # [lon, lat]
+                node_cart = _convert_node_lonlat_rad_to_xyz(node)  # [x, y, z]
+                node_cart_list_x[i] = node_cart[0]
+                node_cart_list_y[i] = node_cart[1]
+                node_cart_list_z[i] = node_cart[2]
```
philipc2, line 419: The majority of this loop should be vectorizable, avoiding the need to iterate over every single node individually

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        num_nodes = self.ds.Mesh2_node_x.size
+        node_cart_list_x = [0.0] * num_nodes
+        node_cart_list_y = [0.0] * num_nodes
+        node_cart_list_z = [0.0] * num_nodes
+        for i in range(num_nodes):
+            if "degree" in self.ds.Mesh2_node_x.units:
+                node = [
+                    np.deg2rad(self.ds["Mesh2_node_x"].values[i]),
+                    np.deg2rad(self.ds["Mesh2_node_y"].values[i])
+                ]  # [lon, lat]
```
philipc2, line 415: `np.deg2rad()` can be applied to every node by passing through the entire 1D array

```Python
lon = np.deg2rad(self.Mesh2_node_x.values)
lat = np.deg2rad(self.Mesh2_node_y.values)
```

`lon` and `lat` will have shape $(N_{nodes},1)$

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+                node_cart = _convert_node_lonlat_rad_to_xyz(node)  # [x, y, z]
```
philipc2, line 416: Same idea as last comment, passing through the entire `lon` and `lat` arrays would return our nodes in carteisian coordinates with shape $(3, N_{nodes})$

uxarray/grid.py
philipc2, line 419: Not needed with vectorized approach

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+
+        # Check if the cartesian coordinates are already populated
+        if "Mesh2_node_cart_x" in self.ds.keys():
+            return
+
+        # check for units and create Mesh2_node_cart_x/y/z set to self.ds
+        num_nodes = self.ds.Mesh2_node_x.size
+        node_cart_list_x = [0.0] * num_nodes
+        node_cart_list_y = [0.0] * num_nodes
+        node_cart_list_z = [0.0] * num_nodes
```
philipc2, line 409: Not needed with vectorized approach

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+                    np.deg2rad(self.ds["Mesh2_node_y"].values[i])
+                ]  # [lon, lat]
+                node_cart = _convert_node_lonlat_rad_to_xyz(node)  # [x, y, z]
+                node_cart_list_x[i] = node_cart[0]
+                node_cart_list_y[i] = node_cart[1]
+                node_cart_list_z[i] = node_cart[2]
+
+        self.ds["Mesh2_node_cart_x"] = xr.DataArray(data=node_cart_list_x)
+        self.ds["Mesh2_node_cart_y"] = xr.DataArray(data=node_cart_list_y)
+        self.ds["Mesh2_node_cart_z"] = xr.DataArray(data=node_cart_list_z)
```
philipc2, line 423: Use `node_cart` instead of `node_cart_list_xyz`, i.e.
```Python
self.ds["Mesh2_node_cart_x"] = xr.DataArray(data=node_cart[0])

File: test/test_grid.py [hunk 8]
```diff
@@ -214,3 +211,42 @@ def test_compute_face_areas_fesom(self):
...
+        vgrid._populate_lonlat_coord()
+        num_nodes = vgrid.ds.Mesh2_node_x.size
+        for i in range(0, num_nodes):
+            lon = vgrid.ds["Mesh2_node_x"].values[i]
+            lat = vgrid.ds["Mesh2_node_y"].values[i]
+            nodes = list(verts[i])
+            [lon_cal,
+             lat_cal] = ux.helpers._convert_node_xyz_to_lonlat_rad(nodes)
+            nt.assert_almost_equal(np.deg2rad(lon), lon_cal, decimal=12)
+            nt.assert_almost_equal(np.deg2rad(lat), lat_cal, decimal=12)
```
philipc2, line 252: I would re-think both of these test cases. When you call `ux.Grid._populate_....`, you are then running the same algorithm in the for loop, essentially testing the function against itself. 

To test the correctness of your functions, I suggest doing the follwing:
1. Create two toy datasets. One should be the pre-converted values $D_{test}$ and the other should be the expected values after running the algorithm/function $D_{gold}$. Note that $D_{gold}$ should be manually computed from $D_{test}$ using something other than the algorithm you are testing
2. Run `ux.Grid._populate_....` on $D_{test}$ to obtain $D_{out}$. 
3. Compare $D_{out}$ to $D_{gold}$

uxarray/grid.py
```diff
@@ -371,3 +372,111 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+
+    def _populate_lonlat_coord(self):
```
philipc2, line 425: Comments from the previous function apply to this one too, but won't duplicate them since most would be essentially the same.

philipc2 (changes requested):
Hi Hongyu!

Thank you for putting this together! After looking through it, there doesn't appear to be any computational issues, however I've left some comments regarding the test cases and general style.

The functions you wrote can be vectorized and I've left some comments on how I would approach it. Let me know if anything needs some clarification and I would be more than willing to sit down for a pair programming session at any point.

test/test_grid.py
hongyuchen1030: Thank you for your comments here.

I just wondering what other algorithms I should use to generate the expected results? I have also been bugged by this problem for a while. So far I can only find one conversion method as implemented in the `ux.helpers._convert_node_xyz_to_lonlat_rad` or vice versa. `ux.helpers._convert_node_xyz_to_lonlat_rad` have passed their own test cases and being merged into the main already. 

I wonder if you have an any better idea than that? Thanks

uxarray/grid.py
hongyuchen1030: Thank you for your information! I wonder if you have any suggestions for the detailed vectorizations here? I know the `map` function and haven't tried it much.  Thank you for your help!

philipc2: You're welcome! The comments below this one should be a good starting point for an efficient vectorization. Essentially, instead of iterating over each node and indexing into arrays (which can be extremely slow in Python), we pass through our entire arrays into the Numpy functions. The following article is a good read! 

https://www.geeksforgeeks.org/vectorized-operations-in-numpy/

test/test_grid.py
philipc2: For the expected results, I suggest using some known algorithm that you can compute the old-fashioned way using pen and paper and apply it to a small dataset for testing. It doesn't have to be anything large, just enough to test the function that you wrote.

hongyuchen1030:
I have finished the vectorization and updated the test cases for these two functions. @philipc2 Thank you so much for your help! Can you take a look at the latest version to see if it works?

Also, for some reasons, I suddenly couldn't pass the the `Read the Docs` tests. 
The error message is `We encountered a problem with a command while building your project. To resolve this error, double check your project configuration and installed dependencies are correct and have not changed recently.`

When I ran `make html` locally, it shows that everything is populated correctly. I wonder if anyone knows how to fix it? Thank you very much. @rajeeja @anissa111

anissa111:
@hongyuchen1030 I believe this is coming from [this known bug](https://github.com/numba/numba/issues/8615) in numba. Interestingly, I was able to recreate the issue locally, but only the first time I generated the docs.

I tried various things on my own branch, including removing/pinning numba and numpy from the docs dependencies and adding it to sphinx's "mocked modules" in various ways, but nothing seemed to solve the issue, but I don't think that it was necessarily anything that you added in this PR that's causing the breakage

uxarray/grid.py
```diff
@@ -371,3 +371,134 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+
+        self.ds["Mesh2_node_cart_x"] = xr.DataArray(
+            data=[val[0] for val in nodes_cart],
+            dims=["nMesh2_node"],
+            attrs={
+                "standard_name": "cartesian x",
+                "units": "m",
+            })
+        self.ds["Mesh2_node_cart_y"] = xr.DataArray(
+            data=[val[1] for val in nodes_cart],
```
philipc2, line 418:
```Python
data=nodes_cart[:, 1]

File: uxarray/grid.py [hunk 11]
```diff
@@ -371,3 +371,134 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+            dims=["nMesh2_node"],
+            attrs={
+                "standard_name": "cartesian y",
+                "units": "m",
+            })
+        self.ds["Mesh2_node_cart_z"] = xr.DataArray(
+            data=[val[2] for val in nodes_cart],
```
philipc2, line 425:
```Python
data=nodes_cart[:, 2]

uxarray/grid.py
philipc2: This implementation looks great! Same comments from above apply to this one (converting to numpy to avoid the iterations)

philipc2 (changes requested):
This looks great! Left a few comments regarding the way you were constructing the `xr.DatArray`, but other than that, the vectorized implementation looks good.

hongyuchen1030:
> @hongyuchen1030 I believe this is coming from [this known bug](https://github.com/numba/numba/issues/8615) in numba. Interestingly, I was able to recreate the issue locally, but only the first time I generated the docs.
> 
> I tried various things on my own branch, including removing/pinning numba and numpy from the docs dependencies and adding it to sphinx's "mocked modules" in various ways, but nothing seemed to solve the issue, but I don't think that it was necessarily anything that you added in this PR that's causing the breakage

Thank you so much for your help! At this moment, is there anything I can do on this branch to avoid this bug?

Or in other words, should we just merge this branch even though the `readthedocs` CI run fails? Or we should still wait until it's fixed? thanks!

hongyuchen1030:
@philipc2 Thanks for your suggestion. I have updated the codes accordingly.

At this moment I think the only concern is the `readthedocs` bugs in the CI run

hongyuchen1030 (commented):
All requested changes are made now

philipc2 (approved):
These changes look great. Great work on the vectorization and test cases.

hongyuchen1030:
Hi @rajeeja, @erogluorhan 
Can you please take a look at this PR and see if it's ready to be merged? Thank you very much!

rajeeja:
> Hi @rajeeja, @erogluorhan Can you please take a look at this PR and see if it's ready to be merged? Thank you very much!

Are all the `2.xxx` files needed? why were they introduced?

hongyuchen1030:
> > Hi @rajeeja, @erogluorhan Can you please take a look at this PR and see if it's ready to be merged? Thank you very much!
> 
> Are all the `2.xxx` files needed? why were they introduced?

I think that might be introduced when trying to fix the `doc-read` bugs we encountered before, @anissa111 I wonder if you have any hint about them? Thank you very much

rajeeja (approved):
thanks for the changes.

Issue #130, opened 2022-09-13
Implement `Grid.build_node_face_connectivity`


anissa111:
Implement `Grid.build_node_face_connectivity`

anissa111:
@paullric I'm working on this implementation now, could you provide a little clarification about what this function is supposed to do?

Does the function build `node_face_connectivity` from scratch? Or is this function just building a reverse relationship from a pre-existing `face_node_connectivity`?

paullric:
It should have to build it from scratch, if face_node_connectivity is unavailable.  If it is available then you can leverage it and transpose.

~ Paul

anissa111:
> It should have to build it from scratch, if face_node_connectivity is unavailable. If it is available then you can leverage it and transpose.

Oh, okay, I see thank you! Is there a prefered method for doing this calculation or is something similar implemented in `tempestremap` I should look at?

Huite:
Hi @anissa111,

I butt in here occasionally, I've implemented this in `xugrid` if you're looking for inspiration:

https://github.com/Deltares/xugrid/blob/main/xugrid/ugrid/ugrid2d.py#L406

It mostly uses some utils defined here:
https://github.com/Deltares/xugrid/blob/main/xugrid/connectivity.py

Which I find make it relatively easy to these transformations on connectivity arrays.

In general, I found it convenient to use scipy.sparse functionality to build/represent these arrays here and there. In particular, constructing a CSR (https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html) matrix does a lot of sorting (which is C++ behind the scenes). Alternatively, you require a lot of numpy sorting calls, which aren't that easy to do efficiently if the arrays are really not that rectangular; or you'd have to write a lot of numba code, I suppose.

anissa111:
@Huite, thank you! I will definitely have a look at the links you've provided.

philipc2:
Closed with the merge of #205

PR #191, opened 2022-12-13
DRAFT: Add `Grid.build_face_edges_connectivity`

hongyuchen1030:
add the function `Grid.build_face_edges_connectivity` and the corresp…onding test case in test_grid.py. docs also updated


philipc2:
I've created an issue similar to #130 to discuss

hongyuchen1030:
Hi @philipc2 @erogluorhan @rajeeja ,

I have already finished the vectorization of the Grid.build_face_edges_connectivity functions. And inside this PR, I expose the `normalized_in_place`, `convert_node_lonlat_rad_to_xyz` and `convert_node_xyz_to_lonlat_rad` to the users (changing them from internal APIs to User APIs)

Can you take a look into it and see if can be merged? Thank you very much!

uxarray/grid.py
```diff
@@ -372,6 +373,79 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        elif _FillValue_ is not -1:
+            mesh2_face_nodes[mesh2_face_nodes == _FillValue_] = -1
+
+        # Then do the padding for each face to close the polygon
+        closed = np.full((n, m + 1), -1, dtype=np.intp)
+        closed[:, :-1] = np.array(mesh2_face_nodes, dtype=np.intp)
+        # We only want the index of first occurrence of -1
+        first_fill_value_index = np.array([
+            np.array(list(*np.where(row == -1)))[0] for row in closed
+        ])  # np.array(list(zip(*np.where(closed == -1))))
```
philipc2, line 406: Can replace this with an elegant one-liner (see [here](https://stackoverflow.com/questions/16243955/numpy-first-occurrence-of-value-greater-than-existing-value) for explanation) 
```Python
first_fill_value_index = np.argmax(closed == -1, axis=1)

**Review by philipc2 — CHANGES_REQUESTED** [review]
Some initial comments

File: test/test_grid.py [hunk 2]
```diff
@@ -116,6 +116,31 @@ def test_init_grid_var_attrs(self):
...
+            tgrid1.build_face_edges_connectivity()
+            mesh2_face_edges = tgrid1.ds.Mesh2_face_edges
+            mesh2_edge_nodes = tgrid1.ds.Mesh2_edge_nodes
+
+            # Assert if the mesh2_face_edges sizes are correct.
+            self.assertEqual(mesh2_face_edges.sizes["nMesh2_face"],
+                             mesh2_face_nodes.sizes["nMesh2_face"])
+            self.assertEqual(mesh2_face_edges.sizes["nMaxMesh2_face_edges"],
+                             mesh2_face_nodes.sizes["nMaxMesh2_face_nodes"])
+            self.assertEqual(mesh2_face_edges.sizes["Two"], 2)
```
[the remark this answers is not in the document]
hongyuchen1030: When I tried writing them as
`            # Assert if the mesh2_face_edges sizes are correct.
>           self.assertEqual(mesh2_face_edges.nMesh2_face,
                             mesh2_face_nodes.nMesh2_face)
`

It throws the following errors:
`
self = <xarray.DataArray 'nMesh2_face' (nMesh2_face: 5400)>
array([ True,  True,  True, ...,  True,  True,  True])
Dimensions without coordinates: nMesh2_face
`

Here I just need to compare the size of these two arrays, I wonder how should I solve these errors.

I also tried `mesh2_face_edges.nMesh2_face.sizes()`, and it throws `'Frozen' object is not callable`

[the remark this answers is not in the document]
philipc2: The attributes I mentioned above are through the `Grid` object, not through the face edges or node variables. They are set according to the UGRID conventions.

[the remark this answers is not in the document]
hongyuchen1030: > ```python
> nMaxMesh2_face_edges
> ```

I am still a bit confused here. So when I tried using
`
            self.assertEqual(tgrid1.nMaxMesh2_face_edges,
                             tgrid1.nMaxMesh2_face_nodes)
`
it will throw the following exception
`
            # Assert if the mesh2_face_edges sizes are correct.
>           self.assertEqual(tgrid1.nMaxMesh2_face_edges,
                             tgrid1.nMaxMesh2_face_nodes)
E           AttributeError: 'Grid' object has no attribute 'nMaxMesh2_face_edges'
`
Even though the `tgrid1.build_face_edges_connectivity()` is successfully called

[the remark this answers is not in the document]
philipc2: Let me take a look into it, that dimension might have not been explicitly set as an attribute (only ones from face nodes are iirc)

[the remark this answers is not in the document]
hongyuchen1030: > Let me take a look into it, that dimension might have not been explicitly set as an attribute (only ones from face nodes are iirc)

I tried adding the dimension at `__init_ds_var_names__` but it still doesn't work. I have push my changes in `grid.py` and pushed it. Thank you very much!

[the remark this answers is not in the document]
philipc2: Let's leave what you currently have until we have the UXarray redesign is merged. Currently `__init_grid_var_attrs__()` supports the following attributes. This explains why `nMaxMesh2_face_edges` doesn't work, but the following do. I'd like to extend this to include other attributes, such as the ones you are using. 
```Python
self.ds_var_names = {
            "Mesh2": "Mesh2",
            "Mesh2_node_x": "Mesh2_node_x",
            "Mesh2_node_y": "Mesh2_node_y",
            "Mesh2_node_z": "Mesh2_node_z",
            "Mesh2_face_nodes": "Mesh2_face_nodes",
            # initialize dims
            "nMesh2_node": "nMesh2_node",
            "nMesh2_face": "nMesh2_face",
            "nMaxMesh2_face_nodes": "nMaxMesh2_face_nodes"
        }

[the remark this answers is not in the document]
Comment (hongyuchen1030) [reply test/test_grid.py:139]: > Let's leave what you currently have until we have the UXarray redesign is merged. Currently `__init_grid_var_attrs__()` supports the following attributes. This explains why `nMaxMesh2_face_edges` doesn't work, but the following do.
> 
> ```python
> self.ds_var_names = {
>             "Mesh2": "Mesh2",
>             "Mesh2_node_x": "Mesh2_node_x",
>             "Mesh2_node_y": "Mesh2_node_y",
>             "Mesh2_node_z": "Mesh2_node_z",
>             "Mesh2_face_nodes": "Mesh2_face_nodes",
>             # initialize dims
>             "nMesh2_node": "nMesh2_node",
>             "nMesh2_face": "nMesh2_face",
>             "nMaxMesh2_face_nodes": "nMaxMesh2_face_nodes"
>         }
> ```

Thank you so much for your help!
Just curious, I Also try adding the ` "nMaxMesh2_face_edges"` inside the `__init_grid_var_attrs__()` but it doesn't solve the problem. I wonder if you know why? Thanks!
> 
> ```python
> self.ds_var_names = {
>             "Mesh2": "Mesh2",
>             "Mesh2_node_x": "Mesh2_node_x",
>             "Mesh2_node_y": "Mesh2_node_y",
>             "Mesh2_node_z": "Mesh2_node_z",
>             "Mesh2_face_nodes": "Mesh2_face_nodes",
>             # initialize dims
>             "nMesh2_node": "nMesh2_node",
>             "nMesh2_face": "nMesh2_face",
>             "nMaxMesh2_face_nodes": "nMaxMesh2_face_nodes"
>            "nMaxMesh2_face_edges": "nMaxMesh2_face_edges",
>            "Two": "Two"
>         }
> ```

**Comment by hongyuchen1030** [comment]
Hi @philipc2 ,
Can you look at the codes and see if they can be merged now? Thank you very much!

File: uxarray/grid.py [hunk 3]
```diff
@@ -372,6 +373,82 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
 
         return integral
 
+    # A one-time-used helper function that will only be used in the build_face_edges_connectivity map() function usage
+    def __replace_fill_val(self, row, index, rep_val):
+        row[index] = rep_val
+        return row
```
philipc2, line 379: Function not necessary with the comments below. 

Also, I suggest not using double `_` because it has a special purpose in Python: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

uxarray/grid.py
```diff
@@ -372,6 +373,82 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+
+        # In mesh2_edge_nodes, we want to remove all dummy edges (edge that has "-1" node index)
+        # But we want to preserve that in our mesh2_face_edges so make the datarray has the same dimensions
+        cumprod = (mesh2_edge_node_copy[:, 0] + 1) * \
+                  (mesh2_edge_node_copy[:, 1] + 1) # In this case, former 0 will be positive while former -1 will be 0
+        has_fill_value = np.ma.masked_where(cumprod != 0, cumprod).mask
+        mesh2_edge_nodes = mesh2_edge_node_copy[has_fill_value] if isinstance(
+            has_fill_value, np.ndarray) else mesh2_edge_node_copy
+        inverse_indices = inverse_indices.reshape(n, m)
+        mesh2_face_edges = mesh2_edge_node_copy[inverse_indices]
```
philipc2, line 441: Would you mind giving me an overview of what is happening in this section?

uxarray/grid.py
```diff
@@ -372,6 +373,82 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        # We only want the index of first occurrence of -1
+        first_fill_value_index = np.argmax(closed == -1, axis=1)
+        first_node = mesh2_face_nodes[:, 0]
+
+        # Now replace the first -1 at each row to be the first node
+        pad_results = np.array(
+            list(
+                map(self.__replace_fill_val, closed, first_fill_value_index,
+                    first_node)))
+
```
philipc2, line 413: Can eliminate the need for the `__replace_fill_val` helper function and mapping by indexing `closed` as a 1D array to replace values (doing this operating on 2D arrays in numpy is messy and I haven't been able to get it to work well). I've tested the following approach and it is equivalent to what you have.

```Python
# Create 1D index into a 2D array
first_fill_idx = (m+1)*np.arange(0, n) + first_fill_value_index

# replace all values at 1D index with the first_node 
np.put(closed.ravel(), first_fill_idx, first_node)

pad_results = closed

File: uxarray/grid.py [hunk 6]
```diff
@@ -8,9 +8,10 @@
 from ._ugrid import _read_ugrid, _encode_ugrid
 from ._shapefile import _read_shpfile
 from ._scrip import _read_scrip, _encode_scrip
-from .helpers import get_all_face_area_from_coords, parse_grid_type, _convert_node_xyz_to_lonlat_rad, _convert_node_lonlat_rad_to_xyz
+from .helpers import get_all_face_area_from_coords, parse_grid_type, node_xyz_to_lonlat_rad, node_lonlat_rad_to_xyz
 
 int_dtype = np.uint32
+_FillValue = float("nan")
```
philipc2, line 14: This should be defined within the `Grid` class and it shouldn't be internal since being able to access `Grid.FillValue` makes sense outside of the scope of the class (unless we want to keep it as `_FillValue` to align closer with the UGRID conventions)

Include it in the `Grid` class and reference it as `self.FillValue`

philipc2 (changes requested):

uxarray/grid.py
hongyuchen1030: Understood, but this function is meant to be a super private class function that will only be used once in the `Grid.build_face_edge_connectivity`

uxarray/grid.py
hongyuchen1030: I am not very sure about the design of this implementation as well because of the following reasons:
1. We still haven't decided what's our universal `_Fill_Value`
2. What's the final dataset read-in and `Grid` initialization function
3. What does the `Grid` class look like finally

But I am happy to make any change to make the design look more reasonable

uxarray/grid.py
hongyuchen1030: The basic idea behind this is our `mesh2_face_nodes` will have Filled_values

> uxarray.Grid.Mesh2_face_nodes: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace)
> A DataArray of indices for each face node, corresponding to coordinates in uxarray.Grid.node_*. Faces can have arbitrary length, with _FillValue=-1 used when faces have fewer nodes than MaxNumNodesPerFace. Nodes are in counter-clockwise order.

And also the fact that our `mesh2_face_edges` has fixed dimensions

> uxarray.Grid.Mesh2_face_edges: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace) (optional)
> A DataArray of indices indicating edges that are neighboring each face.

So in order to incorporate Huite's implementation from XUGRID, we need to do the following
1. For the `mesh2_edge_nodes`, we don't want it to contain any "dummy edges" (the edge that has `-1` as its node)
2. However, for the  `Grid.Mesh2_face_edges`, we need to preserve the dummy edges to make sure it has the same dimensions, For example,
A face with node [0 ,3 5 , -1] means it's a triangle [(0, 3), (3, 5), (5, 0)], however, this will make it only have a size of 3, so we need it to look like [(0, 3), (3, 5), (5, 0), (0, -1)]. For the last `dummy edge`, we don't care what's the valid node number, as long as this edge has `-1` as its node, it will be considered as the  `dummy edge`

Therefore, after line 429, we have the unique edges set `mesh2_edge_node_copy` and it may look like this:

> [[0, 1],
> [1, 2],
> ...
> [1, -1]
> ...
> [0, -1]
> ]

And from line 435 to line line 438 we need to filter out all "valid edge" by removing the "invalid edge". AKA, we want to know the index of all edge that contains -1. Naturally, we will try `mesh2_edge_node_copy[0] * mesh2_edge_node_copy[1]` and get whatever index that's negative. But `0 * -1` will be equal to `0 * 1`(or any other thing). So we're going to increase every index by 1, which means 0 is now 1 and -1 will be 0. 

So we just need to find out
`mesh2_edge_node_copy[:, 0] * mesh2_edge_node_copy[:,1] != 0`, and that will be the valid edge index

uxarray/grid.py
hongyuchen1030: Changed, thank you very much

uxarray/grid.py
```diff
@@ -372,6 +373,82 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        """
+        mesh2_face_nodes = self.Mesh2_face_nodes.values
+        n = self.nMesh2_face
+        m = self.nMaxMesh2_face_nodes
+
+        # First identify the _FillValue in the mesh2_face_nodes
+        # We will unify the _FillValue used in this function as -1:
+        if np.isnan(_FillValue):
+            mesh2_face_nodes = np.nan_to_num(mesh2_face_nodes, nan=-1)
+        elif _FillValue is not -1:
```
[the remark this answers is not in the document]
hongyuchen1030: changed, thanks

uxarray/grid.py
hongyuchen1030: Currently I just leave it as the `self.FillValue`

uxarray/grid.py
hongyuchen1030: Another thing I am thinking about right now is how to make every face edge in the counter-clockwise order, @philipc2 I wonder if you have any idea about it? thanks

uxarray/grid.py
philipc2: I would of still stuck with a single `_` , but I understand where you are coming from

uxarray/grid.py
philipc2: Pinging @erogluorhan to see what his thoughts are about this.

uxarray/grid.py
```diff
@@ -405,11 +406,13 @@ def build_face_edges_connectivity(self):
...
-        # Now replace the first -1 at each row to be the first node
-        pad_results = np.array(
-            list(
-                map(self.__replace_fill_val, closed, first_fill_value_index,
-                    first_node)))
+        # Create 1D index into a 2D array
+        first_fill_idx = (m + 1) * np.arange(0, n) + first_fill_value_index
+
+        # replace all values at 1D index with the first_node
+        np.put(closed.ravel(), first_fill_idx, first_node)
```
philipc2, line 412: small variable name changes & comments. might be worth adding some comments about the 1D indexing, but I think it's self explanatory 

```Python
first_node = mesh2_face_nodes[:, 0]

# first index along columns where a fill value is present
first_fill_idx = np.argmax(closed == -1, axis=1)

# above index into a flattened (1D) array 
first_fill_idx_1d = (m + 1) * np.arange(0, n) + first_fill_idx

# replace first fill value with the first node
np.put(closed.ravel(), first_fill_idx, first_node)

File: uxarray/grid.py (hunk 3 above)
Comment (hongyuchen1030) [reply uxarray/grid.py:379]: Can you explain in more detail about why a single `_` will be more appropriate in this situation? Thanks

Comment (philipc2) [reply uxarray/grid.py:379]: I see what you mean, ignore what I said!

File: uxarray/grid.py (hunk 4 above)
Comment (philipc2) [reply uxarray/grid.py:443]: Thank you for the explanation!  

You mentioned that

>And also the fact that our mesh2_face_edges has fixed dimensions

 > > uxarray.Grid.Mesh2_face_edges: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace) (optional)
A DataArray of indices indicating edges that are neighboring each face.

In your implementation, `Mesh2_face_edges` has dimensions `[nMesh2_face, nMaxMesh2_face_nodes, 2]`. Can you explain why you have the extra dimension in `Mesh2_face_edges`?

Comment (hongyuchen1030) [reply uxarray/grid.py:443]: > Thank you for the explanation!
> 
> You mentioned that
> 
> > And also the fact that our mesh2_face_edges has fixed dimensions
> 
> > > uxarray.Grid.Mesh2_face_edges: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace) (optional)
> > > A DataArray of indices indicating edges that are neighboring each face.
> 
> In your implementation, `Mesh2_face_edges` has dimensions `[nMesh2_face, nMaxMesh2_face_nodes, 2]`. Can you explain why you have the extra dimension in `Mesh2_face_edges`?

I guess it's because I am still not very familiar with the Xarray "dimension" attributes. I thought it refers to the size of each matrix dimensions

So the Mesh2_face_edges look like this

> [[0,1],[1,2],[2,3],[3,0],
>  [0,2],[2,4],[4,5],[5,0],
> [0,2],[2,4],[4,0],[-1,-1],
> ....
> ]

So it will have the `nMesh2_face` of rows, in each row, it will have the `nMaxMesh2_face_nodes` number of edges, and then inside each edge, the size of the edge array is 2.

Please let me know if you have any better idea about the data structure. Thanks!

File: uxarray/grid.py [hunk 9]
```diff
@@ -372,6 +373,79 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
 
         return integral
 
+    # A one-time-used helper function that will only be used in the build_face_edges_connectivity map() function usage
+    def __replace_fill_val(self, row, index, rep_val):
```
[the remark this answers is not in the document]
hongyuchen1030: I thought about that, But it needs to access the grid.mesh2_face_nodes. I feel like passing the entire copy of grid.mesh2_face_nodes is not a very good idea here

I personally don't like creating a one-time-used function as well. However, since we need to avoid looping, we need to use the python. `map()` function in this case. (I couldn't figure out any better python vectorization at this moment), and to use a `map()` function in this specific scenario, I have to create a function for it. (I also tried lambda function, but lambda doesn't support this value replacement)

Please let me know if you have any better suggestion for this.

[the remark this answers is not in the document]
hongyuchen1030: > Instead of being a grid.py function, it looks like a good candidate as an internal function for helprs.py?

I thought about that, But it needs to access the grid.mesh2_face_nodes. I feel like passing the entire copy of grid.mesh2_face_nodes is not a very good idea here

I personally don't like creating a one-time-used function as well. However, since we need to avoid looping, we need to use the python. `map()` function in this case. (I couldn't figure out any better python vectorization at this moment), and to use a `map()` function in this specific scenario, I have to create a function for it. (I also tried lambda function, but lambda doesn't support this value replacement)

Please let me know if you have any better suggestion for this.

uxarray/grid.py
```diff
@@ -372,6 +373,79 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+        row[index] = rep_val
+        return row
+
+    def build_face_edges_connectivity(self):
+        """A DataArray of indices indicating edges that are neighboring each
```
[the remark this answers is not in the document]
hongyuchen1030: > May be start with "Builds" or "Constructs" and continue "a DataArray ..."

The name is based on the previous API function when I was given the `grid.py`

> 
> ``` python
>     # Build the node-face connectivity array.
>     def build_node_face_connectivity(self):
>         """Not implemented."""
>         warn("Function placeholder, implementation coming soon.")

>     # Build the edge-face connectivity array.
>     def build_edge_face_connectivity(self):
>         """Not implemented."""
>         warn("Function placeholder, implementation coming soon.")
> ```

I will be happy to rename it to a better function name. Just need to make sure they has the same function name syntax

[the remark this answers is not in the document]
hongyuchen1030: Ohh you mean the comments? Got it. Thanks!

uxarray/grid.py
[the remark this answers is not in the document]
erogluorhan: Ok, that makes sense, thanks!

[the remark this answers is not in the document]
philipc2: One approach could be to store these one-off utilities in a separate helper file i.e. `_build_face_edges_helpers.py` or `_grid_helpers.py` and simply exclude it from our API. This should hide these functions enough to prevent any accidents described in your second bullet. As for your first bullet, I think having them stored in a separate helper file as @erogluorhan mentioned would be best.

For now though, it seems like this function isn't needed anymore after some of the changes you've made, but I think as our codebase grows, we'll need to keep organizational issues like this in mind.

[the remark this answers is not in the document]
philipc2: You could also define this helper function as an [inner function](https://realpython.com/inner-functions-what-are-they-good-for/)

philipc2 (commented):
Changes in #241 will address the inconsistencies with the data types and fill values. Once it's merged and the changes are made here, I'll give this PR another review

hongyuchen1030:
> > Changes in #241 will address the inconsistencies with the data types and fill values. Once it's merged and the changes are made here, I'll give this PR another review
> 
> @hongyuchen1030 @philipc2 as PR #241 has been merged now, can you work together and move this PR forward to utilize the FillValue solution from that PR?
Thank you so much for your help, I will wrap up the remaining fixes and bugs soon.

philipc2:
@hongyuchen1030 

I've pushed the corresponding changes to use the standardized dtype and fill value. The CI is still failing, but it doesn't appear to be because of the changes I've made

uxarray/grid.py
```diff
@@ -379,6 +384,100 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+                    continue
+                elif match_index_cur_edge == 1:
+                    # Now switch two indexes
+                    temp = mesh2_face_nodes[i][j][0]
+                    mesh2_face_nodes[i][j][0] = last_node
+                    mesh2_face_nodes[i][j][1] = temp
+
+        self.ds["Mesh2_face_edges"] = xr.DataArray(
+            data=mesh2_face_edges,
+            dims=["nMesh2_face", "nMaxMesh2_face_edges", "Two"],
```
philipc2, line 473:
```suggestion
            dims=["nMesh2_face", "nMaxMesh2_face_edges"],
```

From the UGRID conventions:
```
integer Mesh2_face_edges(nMesh2_face, nMaxMesh2_face_nodes) ;

File: uxarray/grid.py [hunk 12]
```diff
@@ -379,6 +384,100 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+            attrs={
+                "cf_role": "face_edges_connectivity",
+                "start_index": 0
```
philipc2, line 476:
```suggestion
                "start_index": 0,
                "_FillValue": INT_FILL_VALUE
```

uxarray/grid.py
```diff
@@ -379,6 +384,100 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+    def build_face_edges_connectivity(self):
+        """A DataArray of indices indicating edges that are neighboring each
+        face.
+
+        Notes
+        -----
+        This function will add `Grid.ds.Mesh2_face_edges` to the `Grid` class, which is an integer
+        DataArray of size (nMesh2_face, MaxNumNodesPerFace)
+        """
+        mesh2_face_nodes = self.Mesh2_face_nodes.values
```
philipc2, line 401: No changes needed, but for future reference with #241 merged, we can guarantee that `self.Mesh2_face_nodes` will always have `dtype=INT_DTYPE` and `_FillValue=INT_FILL_VALUE`

philipc2 (changes requested):
A couple things I caught when setting up the standardized data types and fill values, will provide a more comprehensive review once this one is passing

hongyuchen1030:
> @hongyuchen1030
> 
> I've pushed the corresponding changes to use the standardized dtype and fill value. The CI is still failing, but it doesn't appear to be because of the changes I've made

I am implementing the counterclockwise functionality now, probably that's the problem. I will convert this PR into Draft

philipc2:
> > @hongyuchen1030
> > I've pushed the corresponding changes to use the standardized dtype and fill value. The CI is still failing, but it doesn't appear to be because of the changes I've made
> 
> I am implementing the counterclockwise functionality now, since I need to figure out a way to make Huite's method compatible with our requirement here. probably that's the problem. I am trying to convert this branch into Draft again, however, I can't find the option here. Can you take a look into that? Thanks!

Sounds good. I've marked it as a draft for now, but for the future it's under the "reviewers" tab on the right hand side. 
image

uxarray/grid.py
hongyuchen1030: > The basic idea behind this is our `mesh2_face_nodes` will have Filled_values
> 
> > uxarray.Grid.Mesh2_face_nodes: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace)
> > A DataArray of indices for each face node, corresponding to coordinates in uxarray.Grid.node_*. Faces can have arbitrary length, with _FillValue=-1 used when faces have fewer nodes than MaxNumNodesPerFace. Nodes are in counter-clockwise order.
> 
> And also the fact that our `mesh2_face_edges` has fixed dimensions
> 
> > uxarray.Grid.Mesh2_face_edges: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace) (optional)
> > A DataArray of indices indicating edges that are neighboring each face.
> 
> So in order to incorporate Huite's implementation from XUGRID, we need to do the following
> 
> 1. For the `mesh2_edge_nodes`, we don't want it to contain any "dummy edges" (the edge that has `-1` as its node)
> 2. However, for the  `Grid.Mesh2_face_edges`, we need to preserve the dummy edges to make sure it has the same dimensions, For example,
>    A face with node [0 ,3 5 , -1] means it's a triangle [(0, 3), (3, 5), (5, 0)], however, this will make it only have a size of 3, so we need it to look like [(0, 3), (3, 5), (5, 0), (0, -1)]. For the last `dummy edge`, we don't care what's the valid node number, as long as this edge has `-1` as its node, it will be considered as the  `dummy edge`
> 
> Therefore, after line 429, we have the unique edges set `mesh2_edge_node_copy` and it may look like this:
> 
> > [[0, 1],
> > [1, 2],
> > ...
> > [1, -1]
> > ...
> > [0, -1]
> > ]
> 
> And from line 435 to line line 438 we need to filter out all "valid edge" by removing the "invalid edge". AKA, we want to know the index of all edge that contains -1. Naturally, we will try `mesh2_edge_node_copy[0] * mesh2_edge_node_copy[1]` and get whatever index that's negative. But `0 * -1` will be equal to `0 * 1`(or any other thing). So we're going to increase every index by 1, which means 0 is now 1 and -1 will be 0.
> 
> So we just need to find out `mesh2_edge_node_copy[:, 0] * mesh2_edge_node_copy[:,1] != 0`, and that will be the valid edge index

@philipc2 This is the description for why the FillValue = -1 will be important in my implementation here. Please let me know if you any question! Thanks!

philipc2: @hongyuchen1030 

Thank you! I'll give this another read through this week 😀

hongyuchen1030:
> @hongyuchen1030
> 
> I've pushed the corresponding changes to use the standardized dtype and fill value. The CI is still failing, but it doesn't appear to be because of the changes I've made

The CI is failing because the codes that convert the default Fill_Value to -1 are removed. As mentioned, replacing the default Fill_Value with -1 is critical for the following implementation. I am trying to recover this section now. So before we came out with a better solution, please don't modify this section of codes.

philipc2:
@hongyuchen1030 

There was a bug when handling `nan` fill values in the recently merged #241. `self.ug_filename2` had `Mesh2_face_nodes` as floating point values with a `_FillValue` of `nan`

hongyuchen1030:
> ### From the UGRID Conventions:
> Optionally the topology may have the following attributes:
> 
> 1. `face_edge_connectivity` pointing to an index variable identifying for every face the indices of its edges
> 2. The edges should be specified in anticlockwise direction as viewed from above
> 3. This connectivity array will be a matrix of size `nFaces x MaxNumNodesPerFace`
> 4. If a face has fewer corners/edges than `MaxNumNodesPerFace` then the last edge indices shall be equal to _FillValue`
> 
> ### Implementation
> Addressing some of the above bullet points:
> 
> 1. No comment needed
> 2. Since `face_nodes` are presumed to already be in a counter clockwise orientation, when we create our  `edge_nodes` derived from `face_nodes`, they will already be in a counter clockwise orientation since we are simply linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1] for example
> 3. You have the shape of `face_edges` set as `nFaces x MaxNumNodesPerFace x 2`, which is different than what is expected from the UGRID Conventions. Are you storing the edges instead of of the edge indices?
> 4. No comment needed, other than it should use the same standardized fill value as `face_nodes`

1. No comment

2. This is exactly my previous implementation using the `frozenset()`. However, in Huite's implementation, it unravel the `face_nodes` and collage them together, then use the `np.sort()` combined with `np.unique()`, which will destroy the "counter-clockwise" in the `face_nodes`. In other words, in Huite's method, we are **not** simply linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1]. So we need another nested loop the restore edges. That was also my concern when I was asked to use his way. And also why I am thinking about reverting back to my previous implementation. 

3. Yes, I am storing the edges instead of the edge indices according to my understanding of our api [uxarray_api.md](https://github.com/UXARRAY/uxarray/blob/main/docs/user_api/uxarray_api.md) `uxarray.Grid.Mesh2_face_edges` and `uxarray.Grid.Mesh2_edge_faces: integer DataArray of size (nMesh2_edge, Two) (optional)` 

So if we're only storing the edge indices, then we probably need to re-evaluate problem 2, since at this moment, we don't need to  `linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1]` in the `face_edges` anymore. And preserving the edge direction in the `edge_nodes`  is also not needed. We can cast each edge as a `frozen_set ()` object to highlight it's unidirectional feature or not. I am fine with either implementation

4. No comment

Please let me know what's your opinion about the pb2 and p3, I am fine with any proper modification.

hongyuchen1030:
> @hongyuchen1030
> 
> Correct me if I'm wrong, but I don't believe that having the fill value as `-1` is critical. It seems like it's only used for the dummy edge detection, using the formula `mesh2_edge_node_copy[:, 0] * mesh2_edge_node_copy[:,1] != 0` to filter out all the dummy edges.
> 
> Using `INT_FILL_VALUE` would require us to change this approach, but instead we could search down each column to find any occurrences of `INT_FILL_VALUE`.
> 
> If any index in `mesh2_edge_node_copy[:, 0]` or `mesh2_edge_node_copy[:, 1`] contains at least one fill value, it can be marked as a dummy edge. Below is an example
> 
> ```python
> import numpy as np
> 
> INT_FILL_VALUE = 999
> 
> # example data 
> face_edges = np.array([[0, 3], 
>                        [3, 5], 
>                        [5, 0], 
>                        [0, 999], 
>                        [999, 999],
>                        [999, 0]])
> 
> # true for any edge that contains a fill value
> np.logical_or(face_edges[:, 0] == INT_FILL_VALUE,
>               face_edges[:, 1] == INT_FILL_VALUE)
> 
> >>> array([False, False, False,  True,  True,  True])
> ```

This method works for me.

philipc2:
> 
> 1. No comment
> 2. This is exactly my previous implementation using the `frozenset()`. However, in Huite's implementation, it unravel the `face_nodes` and collage them together, then use the `np.sort()` combined with `np.unique()`, which will destroy the "counter-clockwise" in the `face_nodes`. In other words, in Huite's method, we are **not** simply linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1]. So we need another nested loop the restore edges. That was also my concern when I was asked to use his way. And also why I am thinking about reverting back to my previous implementation.
> 3. Yes, I am storing the edges instead of the edge indices according to my understanding of our api [uxarray_api.md](https://github.com/UXARRAY/uxarray/blob/main/docs/user_api/uxarray_api.md) `uxarray.Grid.Mesh2_face_edges` and `uxarray.Grid.Mesh2_edge_faces: integer DataArray of size (nMesh2_edge, Two) (optional)`
> 
> So if we're only storing the edge indices, then we probably need to re-evaluate problem 2, since at this moment, we don't need to `linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1]` in the `face_edges` anymore. And preserving the edge direction in the `edge_nodes` is also not needed. We can cast each edge as a `frozen_set ()` object to highlight it's unidirectional feature or not. I am fine with either implementation
> 
> 4. No comment
> 
> Please let me know what's your opinion about the pb2 and p3, I am fine with any proper modification.

## Point 2
I'll need to do some more local testing on how these algorithms unfold, but my initial thoughts are that `edge_nodes` does NOT have to store all edges in a counter-clockwise manner, only `face_edges` does, since `face_edges` is an indexing variable (similar to `face_nodes`). Our `edge_nodes` will contain the set of edges, excluding any dummy and duplicates. We simply need to keep a mapping of which index we need to use into `edge_nodes`

### Summary
* `edge_nodes` contains the set of unique edges that compose our unstructured grid, with no spatial relationship
* `face_edges` is used to construct a face from its edges, using indicies into `edge_nodes` to select the deseired `edge_nodes`

### Simple Example
```Python
# edge_nodeswith dummy and duplicate edges
raw_edge_nodes = np.array([[0, 1], 
                           [1, 2], 
                           [2, 0], 
                           [0, 999], 
                           [0, 1],
                           [1, 3],
                           [3, 0],
                           [0, 999]])

# filterededge_nodes (final result)
edge_nodes = np.array([[0, 1],  # index 0
                       [1, 2],  # index 1
                       [2, 0],  # index 2
                       [1, 3],  # index 3
                       [3, 0]]) # index 4

# indicies into edge_nodes with fill values
face_edges = np.array([[0, 1, 2, 999],
                       [0, 3, 4, 999]])

```

Above is what the expected results should be, I'll need to keep exploring algorithms to construct these.

## Point 3
Directly from the API, which aligns with the UGRID conventions.

> uxarray.Grid.Mesh2_face_edges: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace) (optional)
> A DataArray of indices indicating edges that are neighboring each face.

> uxarray.Grid.Mesh2_edge_nodes: integer DataArray of size (nMesh2_edge, Two) (optional)
> A DataArray of indices for each edge. Nodes are in arbitrary order.

uxarray/grid.py
philipc2: Resolved by #241

hongyuchen1030:
> > 1. No comment
> > 2. This is exactly my previous implementation using the `frozenset()`. However, in Huite's implementation, it unravel the `face_nodes` and collage them together, then use the `np.sort()` combined with `np.unique()`, which will destroy the "counter-clockwise" in the `face_nodes`. In other words, in Huite's method, we are **not** simply linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1]. So we need another nested loop the restore edges. That was also my concern when I was asked to use his way. And also why I am thinking about reverting back to my previous implementation.
> > 3. Yes, I am storing the edges instead of the edge indices according to my understanding of our api [uxarray_api.md](https://github.com/UXARRAY/uxarray/blob/main/docs/user_api/uxarray_api.md) `uxarray.Grid.Mesh2_face_edges` and `uxarray.Grid.Mesh2_edge_faces: integer DataArray of size (nMesh2_edge, Two) (optional)`
> > 
> > So if we're only storing the edge indices, then we probably need to re-evaluate problem 2, since at this moment, we don't need to `linking together indices [0, 1] to [1, 2] to [2, 3] to [3, 1]` in the `face_edges` anymore. And preserving the edge direction in the `edge_nodes` is also not needed. We can cast each edge as a `frozen_set ()` object to highlight it's unidirectional feature or not. I am fine with either implementation
> > 
> > 4. No comment
> > 
> > Please let me know what's your opinion about the pb2 and p3, I am fine with any proper modification.
> 
> ## Point 2
> I'll need to do some more local testing on how these algorithms unfold, but my initial thoughts are that `edge_nodes` does NOT have to store all edges in a counter-clockwise manner, only `face_edges` does, since `face_edges` is an indexing variable (similar to `face_nodes`). Our `edge_nodes` will contain the set of edges, excluding any dummy and duplicates. We simply need to keep a mapping of which index we need to use into `edge_nodes`
> 
> ### Summary
> * `edge_nodes` contains the set of unique edges that compose our unstructured grid, with no spatial relationship
> * `face_edges` is used to construct a face from its edges, using indicies into `edge_nodes` to select the deseired `edge_nodes`
> 
> ### Simple Example
> ```python
> # edge_nodes with dummy and duplicate edges
> raw_edge_nodes = np.array([[0, 1], 
>                            [1, 2], 
>                            [2, 0], 
>                            [0, 999], 
>                            [0, 1],
>                            [1, 3],
>                            [3, 0],
>                            [0, 999]])
> 
> # filtered edge_nodes (final result)
> edge_nodes = np.array([[0, 1],  # index 0
>                        [1, 2],  # index 1
>                        [2, 0],  # index 2
>                        [1, 3],  # index 3
>                        [3, 0]]) # index 4
> 
> # indicies into edge_nodes with fill values
> face_edges = np.array([[0, 1, 2, 999],
>                        [0, 3, 4, 999]])
> ```
> 
> Above is what the expected results should be, I'll need to keep exploring algorithms to construct these.
> 
> ## Point 3
> Directly from the API, which aligns with the UGRID conventions.
> 
> > uxarray.Grid.Mesh2_face_edges: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace) (optional)
> > A DataArray of indices indicating edges that are neighboring each face.
> 
> > uxarray.Grid.Mesh2_edge_nodes: integer DataArray of size (nMesh2_edge, Two) (optional)
> > A DataArray of indices for each edge. Nodes are in arbitrary order.

My latest commit should satisfy the ## Point 3

But for ## Point 2,  the reason why I said we can use the `frozen_set()` for each edge in `edge_nodes` is because

> ```python
># Original face_nodes input:
> raw_face_nodes = np.array(
>                                                 [[0,7,2,3],
>                                                 [7,4,5,2],
>                                                  ])
> # edge_nodes with dummy and duplicate edges
> raw_edge_nodes = np.array([[0, 7], 
>                            [7, 2], 
>                            [2, 3], 
>                            [3, 0], 
>                            [7, 4], 
>                            [4, 5], 
>                            [5, 2],
>                            [2, 7])
> # After we use `np.sort()` combined with `np.unique()` to get the filered edge_nodes
> # filtered edge_nodes (final result)
> edge_nodes = np.array([[0, 7],  # index 0
>                        [2,7],  # index 1
>                        [2, 3],  # index 2
>                        [0, 3],  # index 3
>                        [4, 7]]) # index 4
>                        [4, 5]]) # index 5
>                        [2, 5]]) # index 6
> 
> # indicies into edge_nodes with fill values
> face_edges = np.array([[0, 1, 2, 3],#face 0
>                        [4, 5, 6, 1]])#face 1
> ```

And if someone want to reconstruct the face using these edge indices, it will get:

> ```python
># Original face_nodes input:
>face_0 = [[0, 7], [2,7],[2, 3],[0, 3]] # The nodes are not in the counter-clockwise order, but since our edge is undirectional, >                                                           # though [2,7] = [7, 2], but [7, 2] is the "correct order".
>face_1 = [[4, 7], [4,5],[2, 5],[2, 7]] # The nodes are not in the counter-clockwise order, but since our edge is undirectional, >                                                           # though [2,7] = [7, 2], but [2,7] is the "correct order".
> ```

I just wonder if this is the result we want users to have?  Also though the nodes in each face is counter clock-wise, it can start at any index, suggesting that the reconstructed face_edges will be more confusing.

philipc2:
> Another thing for ## Point 2 is: You're also trying to use INT_FILL_VALUES, in here 999 to represet a dummy edge in face_edges array. But face_edges stores the index to edge_nodes array, which apparently unlikely to have the size of INT_FILL_VALUES + 1. So it might cause some index out-of-bound error if user trying to get the edge from edge_nodes array without any checking, for example:

This is expected when adhering to the UGRID conventions. There isn't a much better approach unless we explore alternative data structures, such as numpy masked arrays or scipy sparese arrays

> Again, if a face has fewer corners/edges than MaxNumNodesPerFace then the last edge indices shall be equal to _FillValue

> I just wonder if this is the result we want users to have? Also though the nodes in each face is counter clock-wise, it can start at any index, suggesting that the reconstructed face_edges will be more confusing.

The result is expected at least from my understanding. Can you explain by what you mean with starting at any index?

hongyuchen1030:
> I'm unsure on why we would need to add one to it though, can you elaborate?
This actually is just the index and sizing things in general. if an array has the maximum index as `999`  then the size should be `999+1 = 1000`. So don't even worry about it.

>Can you explain by what you mean by starting at any index?

``` python
# Original face_nodes input:
raw_face_nodes = np.array(
                                                [[0,7,2,3],
                                                [7,4,5,2],
                                                 ])

another_raw_face_nodes = np.array(
                                                [[7,2,3,0],
                                                [2,7,4,5],
                                                 ])# This also represents the exact same faces as the previous `raw_face_nodes`
```
So for the user point of view, if they use the filtered and ordered `edge_nodes`, they will get
```python
face_0 = [[2,7],[2, 3],[0, 3], [0.7]] 
face_1 = [[2,7], [4,7],[4, 5],[2, 5]] 
```
  Basically, they can't have the correct ordering unless they dig into the original `face_nodes` again. But if this is what they should expect to have, then we don't need to worry about it here.

uxarray/grid.py
```diff
@@ -420,6 +420,8 @@ def build_face_edges_connectivity(self):
         mesh2_edge_node_copy, inverse_indices = np.unique(ar=mesh2_edge_nodes,
                                                           return_inverse=True,
                                                           axis=0)
+        # TODO: Make the dummy edge index in the mesh2_face_edges as INT_FILL_VALUE
```
hongyuchen1030, line 423: @philipc2 Can you help me implement this part, so that the dummy edge index in the `mesh2_face_edges` is `INT_FILL_VALUE`? Thanks

philipc2:
> ``` python
> # Original face_nodes input:
> raw_face_nodes = np.array(
>                                                 [[0,7,2,3],
>                                                 [7,4,5,2],
>                                                  ])
> 
> another_raw_face_nodes = np.array(
>                                                 [[7,2,3,0],
>                                                 [2,7,4,5],
>                                                  ])# This also represents the exact same faces as the previous `raw_face_nodes`
> ```
> So for the user point of view, if they use the filtered and ordered `edge_nodes`, they will get
> ```python
> face_0 = [[2,7],[2, 3],[0, 3], [0.7]] 
> face_1 = [[2,7], [4,7],[4, 5],[2, 5]] 
> ```
>   Basically, they can't have the correct ordering unless they dig into the original `face_nodes` again. But if this is what they should expect to have, then we don't need to worry about it here.            

Correct, but indeed it is expected. Below is how I internally think of it:

At the lowest level, our grid is essentially just an unstructured set of points, with no spatial relationship to each other when stored in memory as 1D array (`node_x` and `node_y`).

The connectivity variables `face_nodes` and `edges_nodes` are then used to construct the geometries of our grid/mesh. `face_nodes` indexes `node_x` and `node_y` to obtain the nodes/vertices that outline each face. Similarly, `edge_nodes` indexes `node_x` and `node_y` to obtain the nodes/vertices that make up each edge.

I think of `face_edges` as adding an extra level of indirection to the process. Instead of outlining the face with nodes/verticies, we want to do so with edges. Therefore, we need to index `edge_nodes`, which is already built from `node_x` and `node_y`

hongyuchen1030:
> ```python
> ```python
> # Original face_nodes input:
> raw_face_nodes = np.array(
>                                                 [[0,7,2,3],
>                                                 [7,4,5,2],
>                                                  ])
> 
> another_raw_face_nodes = np.array(
>                                                 [[7,2,3,0],
>                                                 [2,7,4,5],
>                                                  ])# This also represents the exact same faces as the previous `raw_face_nodes`
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> So for the user point of view, if they use the filtered and ordered edge_nodes, they will get
> ```python
> face_0 = [[2,7],[2, 3],[0, 3], [0.7]] 
> face_1 = [[2,7], [4,7],[4, 5],[2, 5]] 
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> Basically, they can't have the correct ordering unless they dig into the original face_nodes again. But if this is what they should expect to have, then we don't need to worry about it here.
> 
> Correct, but indeed it is expected. Below is how I internally think of it:
> At the lowest level, our grid is essentially just an unstructured set of points, with no spatial relationship to each other when stored in memory as 1D array (node_x and node_y).
> The connectivity variables face_nodes and edges_nodes are then used to construct the geometries of our grid/mesh. face_nodes indexes node_x and node_y to obtain the nodes/vertices that outline each face. Similarly, edge_nodes indexes node_x and node_y to obtain the nodes/vertices that make up each edge.
> I think of face_edges as adding an extra level of indirection to the process. Instead of outlining the face with nodes/verticies, we want to do so with edges. Therefore, we need to index edge_nodes, which is already built from node_x and node_y
> ```

This is exactly what we’re doing now. 

My question is just: user will get the “edge” whose node index is not intuitively matching the order as in `face_nodes ` if they are using the `edge _nodes ` for indexing. And you said this is what they should expect. So it should be fine

philipc2:
@hongyuchen1030 

I think we are on the same page now then! I think this discussion really brings to light that we should have proper documentation and usage examples about what each of the variables is used for, other than what's already in the API.

A section on how the UGRID conventions are used throughout the package would be a great addition to our documentation.

philipc2:
@hongyuchen1030 

I've created a separate branch and PR to focus on adding the `edge_node_connecitvity` (#263)

Once we have that merged, we can work on having a final implementation of `Grid.build_face_edges_connectivity`

PR #200, opened 2023-01-04
Pin numpy in requirements.txt for documentation generation

### LINKED ISSUE #199: Documentation generation failing
Our documentation is failing on PRs due to some incompatibility with the versions of numpy and numba that RTD's environment is trying to use.

## An explanation:
We have numpy pinned in the relevant environment files, but RTD is a little tricky sometimes and will eager-update packages pinned in a yml file that are not also pinned in the package `requirements.txt`. See the following part of the [build logs](https://readthedocs.org/projects/uxarray/builds/19068875/) here:
```
Installing collected packages: numpy, uxarray
  Attempting uninstall: numpy
    Found existing installation: numpy 1.23.5
    Uninstalling numpy-1.23.5:
      Successfully uninstalled numpy-1.23.5
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
numba 0.56.4 requires numpy<1.24,>=1.18, but you have numpy 1.24.1 which is incompatible.
Successfully installed numpy-1.24.1 uxarray-2022.12.0
```

Even though we've pinned numpy for causing the exact errors we're getting in the builds, this step in the RTD build overrides the pin without the additional pin in `requirements.txt`. 

This also means that the errors weren't easily reproducible locally, as local generation would use the environment created from the relevant yml file with the correct pinning.

anissa111:
Closes #199 

Summary of changes:
- pins numpy<1.24 in `requirements.txt`

See passing RTD build for this branch [here](https://readthedocs.org/projects/uxarray/builds/19068929/)

@hongyuchen1030 This should fix the issues you're experiencing in #188

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 xarray
-numpy
+numpy<1.24


philipc2 (approved):
Thank you!

rajeeja (approved):
I should have fixed this, when fixing the code dependency. Thanks 👍

PR #203, opened 2023-01-11
Latlonbound Test Cases

hongyuchen1030:
This PR contains the test case of the `build_latlonbound`

The main test cases will be implemented in the `uxarray/test/test_grid.py` using quantitative methods. The associate utility functions will be at `uxarray/test/latlonbound_test_utilities.py`


test/latlonbound_test_utilities.py
```diff
@@ -0,0 +1,386 @@
...
+        The second endpoint of the great circle arc [lon, lat] in degree east.
+
+    Returns
+    -------
+    float
+        maximum latitude in radian
+    """
+
+    # Find the parametrized equation for the great circle passing through v1 and v2
+    err_tolerance = 1.0e-15
```
anissa111, line 28: What do you think about adding tolerance as an optional input to the function?

test/latlonbound_test_utilities.py
```diff
@@ -0,0 +1,386 @@
...
+                    max_section = [w1_new, v_c]
+
+        b_lonlat = uxarray.helpers._convert_node_xyz_to_lonlat_rad(copy.deepcopy(max_section[0]))
+        c_lonlat = uxarray.helpers._convert_node_xyz_to_lonlat_rad(copy.deepcopy(max_section[1]))
+
+    return np.average([b_lonlat[1], c_lonlat[1]])
+
+
+def min_latitude_rad(v1, v2):
+    """Quantitative method to find the minimum latitude between in a great circle arc recursively
```
anissa111, line 106: The function as written here isn't recursive, is it meant to be?

test/latlonbound_test_utilities.py
```diff
@@ -0,0 +1,386 @@
...
+
+    # Only stop the iteration when two endpoints are extremely closed
+    while np.absolute(b_lonlat[1] - c_lonlat[1]) >= err_tolerance or np.absolute(
+            b_lonlat[0] - c_lonlat[0]) >= err_tolerance:
+        max_lat = -np.pi  # reset the max_latitude for each while loop
+        v_b = max_section[0]
+        v_c = max_section[1]
+
+        # Divide the angle of v1/v2 into 10 subsections, the leftover will be put in the last one
+        # Update v0 based on max_section[0], since the angle is always from max_section[0] to v0
```
anissa111, line 45: Why 10 subsections here?

test/latlonbound_test_utilities.py
```diff
@@ -0,0 +1,386 @@
...
+        angle_v1_v2_rad = uxarray.helpers.angle_of_2_vectors(v_b, v_c)
+        v0 = np.array(uxarray.helpers._normalize_in_place(np.cross(np.cross(v1_cart, v2_cart), v_b)))
+        avg_angle_rad = angle_v1_v2_rad / 10
+
+        for i in range(0, 10):
+            angle_rad_prev = avg_angle_rad * i
+            if i >= 9:
+                angle_rad_next = angle_v1_v2_rad
+            else:
+                angle_rad_next = angle_rad_prev + avg_angle_rad
```
anissa111, line 55: This section could be vectorized, but I'm not entirely clear of the purpose. How many times does this while loop run? Does it ever get stuck? I don't necessarily think it's necessary to vectorize this small of a loop if it's only run a few times.

anissa111 (commented):
I've left a few comments, but I think I need a little clarity in to the algorithm here to make a better review.

test/latlonbound_test_utilities.py
hongyuchen1030: The error tolerance here is not for the user input but for the python floating point precision and machine epsilon issues.

In other words, when we're doing the floating point calculation in python, we cannot get the exact same results every time (even though the data and the function are the same), we can at most achieve a certain level of precision genrerally. Therefore, after discussing with @paullric , we decided to make 1.0e-12 the precision level of all our calculations (When users use our function, they can expect the results to be at most 12 decimal precise).

The reason why I use 1.0e-15 here is just for precaution.

test/latlonbound_test_utilities.py
hongyuchen1030: Yes, it's meant to be non-recursive. 

Because python actually doesn't support recursion that well and might cause stack over flow easily under some marginal cases

test/latlonbound_test_utilities.py
hongyuchen1030: It's just a conventional range division numbers for this iterative method

test/latlonbound_test_utilities.py
hongyuchen1030: Thank you for your information. This while loop should keep iterating until it converges. With all current input files, the while loop will converge within 10 iterations. 

As I said, this is only a test-case function. And that's why I am unsure about the need to vectorize these functions.

hongyuchen1030:
The Algo behind each testcase of the build_latlon_bounds:
`test_generate_Latlon_bounds_latitude_max` sudo codes:

```python 
    for face in all_faces:
        for edge in face:
            # n1, n2 are the two endpoints of the edge
            max_lat_edge_i = _latlonbound_utilities.max_latitude_rad(n1, n2)
       
      max_lat_face = max(max_lat_edge_i) # i range from 1...n
```
`max_latitude_rad(v1, v2)` sudo codes:
- v1 and v2 are the two endpoints of the edge
- divide the GCR (v1, v2) into 10 sub-intervals $(v_i, v_{i+1}), i \in (0,...,n)$
- compare the latitude at $v_i,  i \in (0,...,n)$ to find the largest one (aka.$ i = argmax(v_i), i \in (0,...,n)$)
- Divide this newly found interval i into 10 sub-intervals and search within this interval recursively
- Stop this recursively search until the current interval (v1, v2) has the "same" attitude  (difference less than the epsilon)

Similar for the `test_generate_Latlon_bounds_latitude_min` and `min_latitude_rad`

hongyuchen1030:
Algos for the longitude boundaries detection:

`def minmax_Longitude_rad(v1, v2)` sudo codes: 
- This function is based on the rule that edge's longitude span cannot be larger than 180 degree. And it will quantitative method to find the minimum("left") and maximum("right") Longitude between in a great circle
- First sort v1 v2 based on the longnitude and get `lon_min` and `lon_max`
- If `lon_max - lon_min <= 180` , `return [lon_min  lon_max]`
- Else, return [lon_max  lon_min]

`def test_generate_Latlon_bounds_longitude_minmax` sudo codes:
    for face in all_faces:
         minmax_lon_rad_face = [left_lon, right_lon]
        for edge in face:
            # n1, n2 are the two endpoints of the edge
           # if one of the point is the pole point, then insert the another point only:
           pre-treated_edge = _latlonbound_utilities.minmax_Longitude_rad(n1, n2)
           new_minmax_lon_rad_face = _latlonbound_utilities.expand_longitude_rad(old_minmax_lon_rad_face, pre-treated_edge)

Issue #207, opened 2023-01-26
CI and Upstream CI failures


anissa111:
Summary of issues:
- Upstream CI
    - pip is causing some dependency clashes since we use pip to install upstream packages after solving a conda environment
- Nightly CI
    - It seems like the UMass tredds server is... down? And has been for at least two weeks. The `test_read_ugrid_opendap` test fails because it cannot connect, not because anything on our side is broken. 
    - We're getting a [warning](https://github.com/UXARRAY/uxarray/actions/runs/4001972562/jobs/6868752083#step:6:92). Its not breaking anything currently, but we should probably address it
```
Encountered the use of a type that is scheduled for deprecation: type 'reflected list' found for argument 'node_coord' of function '_convert_node_lonlat_rad_to_xyz'.
```

rajeeja:
How do you plan to address it? Disable it or let it fail gracefully or something else?

anissa111:
> How do you plan to address it? Disable it or let it fail gracefully or something else?

I was going to ask about that in my PR. I currently have a branch that is just try/catching the url before trying tests and printing a warning if it can't find the file, but ideally I think it would be good to do something like
1. find an alternative OPeNDAP url
2. include a backup OPeNDAP url in case the first one goes down for maintenance or something
3. print a warning that the test didn't run if, by some chance, both OPeNDAP sources fail

rajeeja:
3 sounds good.

anissa111:
Yes, the only problem is that I don't know any other two ugrid file on separate OPeNDAP sources. Any ideas there?

rljacob:
What are we trying to get from UMass?

dcherian:
Can you vendor or generate some "nice" small files for testing? It seems like a bad idea to depend on remote endpoints that you don't control.

anissa111:
@rljacob, here is the test. It was contributed by @ocefpaf in https://github.com/UXARRAY/uxarray/pull/68. 

``` python
def test_read_ugrid_opendap(self):
    """Read an ugrid model from an OPeNDAP URL."""

    url = "http://www.smast.umassd.edu:8080/thredds/dodsC/FVCOM/NECOFS/Forecasts/NECOFS_GOM3_FORECAST.nc"
    xr_grid = xr.open_dataset(url, drop_variables="siglay")
    ugrid = ux.Grid(xr_grid)
    assert isinstance(getattr(ugrid, "Mesh2_node_x"), xr.DataArray)
[...]

PR #208, opened 2023-01-26
CI failures

anissa111:
Addresses #207 

### Summary of changes directly addressing #207:
- To address the broken OPeNDAP url: adds a try/catch before attempting remainder of test (in addition to @ocefpaf's contribution in https://github.com/UXARRAY/uxarray/pull/209)
- Does not touch reflected list deprecation in numba, that will be a separate PR since it's not currently causing tests to fail
- To address failing upstream ci caused by pip installs overriding conda pins:
    - removes upstream pip installs from `upstream-dev-environment.yml`
    - take dependency install out of `conda_setup` step
    - add `Install select upstream dependencies` step
    - re-add conda dependencies install _after_ pip dependencies

### Bonus changes:
- adds a `conda list` step to `upstream-dev-ci.yml` and `ci.yml`
- updated upstream CI to run on python 3.10
- added python 3.10 to version matrix for regular CI, too
- renames environment in `upstream-dev-ci.yml` to be `uxarray_upstream_build` instead of `uxarray_build` to be distinct from the environment name in `environment.yml` and `ci.yml`
- changes pip install from git to zip for speed
- uses conda instead of mamba for more consistent environment solving only for upstream ci (errors persist if mamba is used)

See passing upstream ci for this branch here: https://github.com/UXARRAY/uxarray/actions/runs/4021274777/jobs/6910001710

### DIFF
--- a/ci/upstream-dev-environment.yml
+++ b/ci/upstream-dev-environment.yml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - netcdf4
   - numba
-  - numpy
+  - numpy>=1.2, <1.25
   - pathlib
   - python
   - pytest
--- a/ci/upstream-dev-environment.yml
+++ b/ci/upstream-dev-environment.yml
@@ -1,17 +1,12 @@
-name: uxarray_build
+name: uxarray_upstream_build
 channels:
   - conda-forge
 dependencies:
   - netcdf4
   - numba
-  # pinning the numpy version for numba compatibility
   - numpy
   - pathlib
   - python
   - pytest
   - pytest-cov
-  - pip
   - scipy
-  - pip:
-    - git+https://github.com/pydata/xarray.git
-    - git+https://github.com/dask/dask.git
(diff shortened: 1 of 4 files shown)


ci/upstream-dev-environment.yml
```diff
@@ -1,17 +1,12 @@
-name: uxarray_build
+name: uxarray_upstream_build
 channels:
   - conda-forge
 dependencies:
   - netcdf4
   - numba
-  # pinning the numpy version for numba compatibility
-  - numpy
+  - numpy>=1.2, <1.25
```
erogluorhan, line 7: Is this version pinning needed? I think it was resolved a few releases ago for the regular CI.

erogluorhan (commented):
It looks good overall but a single inline comment

ci/upstream-dev-environment.yml
anissa111: Oh good point! It was something I was trying along the way. Let me revert this and see if everything still holds

erogluorhan (approved):

philipc2 (approved):
Looks good!

PR #209, opened 2023-01-26
use a more stable URL

### LINKED ISSUE #207: CI and Upstream CI failures

ocefpaf:
I will send a PR to use pytest-vcr later but it will be more involved b/c I don't know if it works with the unittest setup.

Partially closes #207

### DIFF
--- a/test/test_ugrid.py
+++ b/test/test_ugrid.py
@@ -39,7 +39,7 @@ class TestUgrid(TestCase):
     def test_read_ugrid_opendap(self):
         """Read an ugrid model from an OPeNDAP URL."""
 
-        url = "http://www.smast.umassd.edu:8080/thredds/dodsC/FVCOM/NECOFS/Forecasts/NECOFS_GOM3_FORECAST.nc"
+        url = "http://test.opendap.org:8080/opendap/ugrid/NECOFS_GOM3_FORECAST.nc"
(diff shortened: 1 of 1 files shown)


anissa111 (approved):

anissa111:
This will work. We can merge this change in to main and then I'll merge it in to my other PR with the other updates

rajeeja (approved):
This is a better dataset. I hope they don't move or rename it.

ocefpaf:
> This is a better dataset. I hope they don't move or rename it.

Wait, remote servers renaming, moving, disappearing with data? That never happens! ;-p

PR #211, opened 2023-02-01
Organize test files into directories

### LINKED ISSUE #210: Organize test files
This issue points two things to be fixed with our current test structure:

1. Test files are not organized, and it is hard to follow them through README file. A grid format-based organization would work better (e.g. /ugrid, /ugrid/fesom, /ugrid/geoflow-small, /exoduus, /scrip, etc.).

2. This part has now been moved into a new issue (#212):
~~I believe for any test format/case, the minimal set of files we need to have is a grid topology file along with at least two variable files (just like we currently do with the "outCSne30" or "fesom" cases). However, for many existing test files, we only have a grid topology file, which is not sufficient for testing dataset openning operations (only a grid topology can be opened with that file).~~
   - ~~We need to find and add at least two different variable files for the following cases:~~
      - ~~outCSne8 (does not have any var files)~~
      - ~~outRLL1deg (has only vortex var file)~~
      - ~~ov_RLL10deg_CSne4 (has only vortex var file)~~
      - ~~mesh.nc (does not have any var files)~~

@rajeeja @anissa111

erogluorhan:
Closes #210 

This PR handles the following:

- Organizes /test/meshfiles into grid format-based directories
   - With such organization, no need for an overarching README file. Get rid of it (actually shrink it to only fesom-specific information and move it to ugrid/fesom folder)
   - New structure looks like:
      - /meshfiles
         - /exodus
            - /mixed
            - /outCSne8  
         - scrip
            - /outCSne8  
         - shp
         - ugrid 
           - fesom
           - geoflow-small
           - outCSne30
           -  outRLL1deg
           - ov_RLL10deg_CSne4
- Get rid of mesh.nc as it is the same as ugrid/geoflow-small/grid.nc (point tests to this file instead wherever applicable)
- Update test cases accordingly with these changes


philipc2 (approved):
These changes look great and this approach is much more organized and flexible compared to using a single README file for bookkeeping. Thank you for putting this together!

anissa111 (approved):

Issue #181, opened 2022-12-07
Provide an example that shows how to apply Xarray operators to UXarray grids


clyne:
It would be useful to have an example (or documentation) showing how Xarray operators that do not require mesh information (such as min or max) can be applied to a UXarray grid.

zarzycki:
To clarify, the idea is to provide an example of using uxarray to...

1. Open a file and read a field on an unstructured grid (say, surface temperature contained in a NetCDF output from MPAS)
2. Print a global maximum or minimum (say 321.1 K and 260.2 K).

Or something more?

philipc2:
Adressed in #216 with the refactor of the following [usage example](https://uxarray.readthedocs.io/en/latest/examples/001-working-with-unstructured-grids.html)

PR #216, opened 2023-02-02
Redesign

erogluorhan:
This PR introduces a prototype redesign of UXarray based on the conversation in Discussion #185 . This prototype's backbone is the `uxarray.core.Grid` object and a brand new `uxarray.core.UxDataset` that inherits from `xarray.Dataset`.

There were a lot of hesitation (including mine) and warnings against inheritance from xarray in Discussion #185., so one major goal of the prototype is to trigger the team and collaborators to point out potential drawbacks of the inheritance scheme. 

Disclosure: I ended up finding the inheritance most suitable way for UXarray's case after figuring immediate issues with the dataset accessors and not being able to seeing robust implementations of xarray's custom indexes other than current.experiments/trials.

This PR does:

* Create a new namespace `uxarray.core` that houses the existing `grid` as well as new `api` and `dataset` modules.

* NOT change the `grid.py` functionality much but adds a new `equals` function to check if any two grid topologies are the same. Other functions are kept the same for now; however, functions like `interpolate` that can better be applied on `UxDataset` instead of a `grid` can be and should  most likely be removed in the future if tis prototype gets hired.

* Create xarray-like `api.py` module to handle user API functions such as `open_dataset`, `open_grid`, etc by wrapping xarray's functions.

* Create a new `uxarray.core.UxDataset` that inherits from `xarray.Dataset` (Likewise a DataArray version can be considered in the future).
   * Constructor adds a `uxarray.core.Grid` object as property to itself at the initialization (similar to xarray's dataset accessor), so the connection between each uxarray dataset and grid topology
   * Has only one function `integrate()` defined for now. Algorithm is the same, but function call is focused on the dataset instead of on the grid, e.g.:
      ```
      uxds = uxarray.open_dataset(grid_file, dataset_file)
      integral = uxds.integrate()
      ```

* Add `test_api.py` to tests and update existing test cases to reflect the design changes

@UXARRAY/uxarray-dev 

EDIT (2/16/2023): 

@philipc2 @dcherian Now added `UxDataArray` to this prototype and got all `xr.DataArray` fields of any `UxDataset` to be represented as `UxDataArray`. `UxDataArray`also has a `uxgrid` property. So, whenever a function that generates a new dataset or a data array out of `UxDataset` is executed, the result is expected to be either a new `UxDataset` and `UxDataArray` instead of `xr.Dataset` or `xr.DataArray`, respectively, so we can keep uxarray functionality valid on these new things. 

EDIT (3/30/2023): 

Uxarray's new design can be tried out following the instructions in the [How to test new design](https://docs.google.com/document/d/1Ql2AClMazZchSvgGDdxtK5BUARV3KrpTvut0vDOgJEs/edit?usp=sharing) document.


uxarray/core/grid.py
```diff
@@ -123,30 +123,30 @@ def __init_grid_var_attrs__(self) -> None:
...
 
     def __from_vert__(self):
         """Create a grid with one face with vertices specified by the given
         argument.
 
         Called by :func:`__init__`.
         """
-        self.ds = xr.Dataset()
-        self.ds["Mesh2"] = xr.DataArray(
+        self._ds = xr.Dataset()
```
philipc2, line 148: This creates a separate  `xr.Dataset` that is stored under the `xugrid` accessor, acting like `Grid.ds` in our current implementation. It appears that currently only data variables are being store in the parent `xr.Dataset` class and our grid connectivity and other variables are stored under the `uxgrid._ds` dataset. 


Is this intentional? Do we want to separate our data variables and grid topology this way? At a high level, can we think of an `uxarray.Dataset` containing **_only_** data variables inside the inherited `xarray.Dataset` with additional functionalities and data structures  handled through accessors.

What about using the inherited `xarray.Dataset` to store **_both_** our data and grid information?

hongyuchen1030:
> Disclosure: I ended up finding the inheritance most suitable way for UXarray's case after figuring immediate issues with the dataset accessors and not being able to seeing robust implementations of xarray's custom indexes other than current.experiments/trials.
> 
> This PR does:
> 
> * Create a new namespace `uxarray.core` that houses the existing `grid` as well as new `api` and `dataset` modules.
> * NOT change the `grid.py` functionality much but adds a new `equals` function to check if any two grid topologies are the same. Other functions are kept the same for now; however, functions like `interpolate` that can better be applied on `UxDataset` instead of a `grid` can be and should  most likely be removed in the future if tis prototype gets hired.
> * Create xarray-like `api.py` module to handle user API functions such as `open_dataset`, `open_grid`, etc by wrapping xarray's functions.
> * Create a new `uxarray.core.UxDataset` that inherits from `xarray.Dataset` (Likewise a DataArray version can be considered in the future).
>   
>   * Constructor adds a `uxarray.core.Grid` object as property to itself at the initialization (similar to xarray's dataset accessor), so the connection between each uxarray dataset and grid topology
>   * Has only one function `integrate()` defined for now. Algorithm is the same, but function call is focused on the dataset instead of on the grid, e.g.:
>     ```
>     uxds = uxarray.open_dataset(grid_file, dataset_file)
>     integral = uxds.integrate()
>     ```
> * Add `test_api.py` to tests and update existing test cases to reflect the design changes
> 
> @UXARRAY/uxarray-dev

My current understanding is `grid` class hold the original information of the mesh grid (basically the read-in file information), and all the connectivity information are stored in a separate class called "dataset". And to link them together we do 
```python
this_grid = uxarray.open_grid("gridfile.sh")
this_grid_dataset = uxarray.open_dataset(this_grid, "datasetfile.ds")
```
If that so, my concerns will be:
1. Will any operation(now or in the future)  in `the xarray.datase` change the "original information" of the grid? (Basically the coherence in updating the information and avoid dirty read & write)

2. If users are going to export the data, what are we going to give them?

Please let me know if I misunderstand or being unclear of anything

hongyuchen1030:
> 

Resolved in the meeting:
1. ux_ds = uxarray.open_dataset(grid_file, "datasetfile.ds")
2. ux_ds.uxgrid
3. ux_ds.dataset_wise_funcs()
4. ux_ds.uxgrid.grid_wise_funcs()
5. ux_ds.var1.uxgrid.dataarray_wise_funcs()

erogluorhan:
EDIT (2/16/2023): 

@philipc2 @dcherian Now added `UxDataArray` to this prototype and got all `xr.DataArray` fields of any `UxDataset` to be represented as `UxDataArray`. `UxDataArray`also has a `uxgrid` property. So, whenever a function that generates a new dataset or a data array out of `UxDataset` is executed, the result is expected to be either a new `UxDataset` and `UxDataArray` instead of `xr.Dataset` or `xr.DataArray`, respectively, so we can keep uxarray functionality valid on these new things.

philipc2:
Attempting to merge two `UxDataset` objects, each lying on the same grid with different data variables that agree with the grid topology fails because of the `if not isinstance(obj, (DataArray, Dataset, dict)):` in `xr.merge()` complains about our `Grid` object. We will need to implement our own version of the functions mentioned [here](https://docs.xarray.dev/en/stable/user-guide/combining.html) to be grid-aware. 

```Python
uxds_v1 = ux.open_grid(gridfile_geoflow, dsfile_v1_geoflow)
uxds_v2 = ux.open_grid(gridfile_geoflow, dsfile_v2_geoflow)
# fails because of the grid accessor 
uxds_pair = xr.merge([uxds_v1, uxds_v2])

File: uxarray/core/grid.py [hunk 2]
```diff
@@ -319,6 +320,31 @@ def compute_face_areas(self, quadrature_rule="triangular", order=4):
 
         return self._face_areas
 
+    def equals(self, other):
```
philipc2, line 323: Implementing this functionality using ``__eq__(self, other):`` and ``__ne__(self, other):`` methods would be more practical 

```Python
# instead of 
uxds_01.uxgrid.equals(uxds_02.uxgrid)
# can check
uxds_01.uxgrid == uxds_02.uxgrid
# or alternatively 
uxds_01.uxgrid != uxds_02.uxgrid

**Comment by philipc2** [comment]
Creating a shallow or deep copy using the `Xarray` copy method on a `UxDataset` does not preserve the `uxgrid` accessor

```Python
uxds = ux.open_dataset(gridfile_geoflow, dsfile_v1_geoflow)
# uxgrid accessor cleared to None
uxds_copy = uxds.copy(deep=False)

**Comment by philipc2** [comment]
Similarly, calling a method such as `.astype()` also does not preserve the `uxgrid `accessor. Taking the route of inherence with custom accessors, does this mean that we must still deal with the statefullness issue, however we can "patch" it up with modifications to methods that return a Dataset or DataArray?

**Comment by dcherian** [comment]
FWIW I brought up this issue at the Xarray dev meeting. We're discussing [here](https://github.com/pydata/xarray/discussions/7538) and it seems like we're converging to a solution. 

For this discussion the relevant piece is that if you created an xarray.Index object to store grid state, then all xarray operations will propagate these indexes; and any accessors can access state information in the Index. That particular discussion is about propagating the Index when a DataArray is extracted, this does not work currently.

> however we can "patch" it up with modifications to methods that return a Dataset or DataArray

You could hack this by using `__getattr__` but it'll be painful.

**Review by hongyuchen1030 — COMMENTED** [review]
For all my Algorithms implementing experience, I don't see any trouble using the current design.

The following algorithms' implementation is not heavily relied on the Xarray functionality. It uses more numpy features and are highly portable and self-robust

File: uxarray/core/dataset.py [hunk 3]
```diff
@@ -0,0 +1,138 @@
...
+
+from uxarray.core.dataarray import UxDataArray
+from uxarray.core.grid import Grid
+
+
+class UxDataset(xr.Dataset):
+
+    _uxgrid = None
+
+    def __init__(self,
```
dcherian, line 16: @headtr1ck if you have the time, your comments on subclassing here  (also `dataarray.py`) would be valuable.

headtr1ck:
A major method that you are missing is `DataArray/Dataset._replace`.
This will always be called when a copy is returned (so basically 90% of the methods use it).

Notice, that xarrays version calls `type(self)(...)` so it will call the init of your custom class but without the `uxgrid` arg, so you have to call `super()._replace` and then afterwards set the uxgrid.

uxarray/core/dataarray.py
```diff
@@ -0,0 +1,35 @@
...
+
+class UxDataArray(xr.DataArray):
+
+    _uxgrid = None
+
+    def __init__(self, *args, uxgrid: Grid = None, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.uxgrid = uxgrid
+        # TODO: Weird that below if-check leads miscreation of UxDataArray object
```
headtr1ck, line 17: Can you provide a MVCE that lets it fail?
For me this seems to work fine.

rajeeja: > Can you provide a MVCE that lets it fail? For me this seems to work fine.

@erogluorhan After uncommenting the section, I tried running this function to see the error:

https://github.com/UXARRAY/uxarray/blob/5082ca28d961923cf57935dc8192f1e5a5b910f1/test/test_dataset.py#L40

The problem is https://github.com/UXARRAY/uxarray/blob/prototype_redesign/uxarray/core/dataset.py#L135 

`self[var_key].to_numpy()
`This following above calls the dataarray constructor with the uxgrid object set to None and so the assertion is triggered.

rajeeja:
I would like to reorganize a bit: add two folders:
1. utils (helpers , get quadrature etc.) 
2. io (_exodus etc.)

uxarray/core/dataarray.py
headtr1ck: This should be fixed if you use your own implementation of `_replace`

rajeeja: Thanks for the pointer, took a little while to fix the root of this. `_replace` wasn't needed, my last commit fixes it.  It was a bug in `__setitem__` function.

headtr1ck: You are still going to need it. Many instance methods are using it, like `isel`, `rename`, `transpose` etc.

uxarray/core/dataset.py
```diff
@@ -0,0 +1,138 @@
...
+        else:
+            return xarr
+
+    def __setitem__(self, key, value):
+        """Override to check if the value being set is an instance of
+        xarray.DataArray.
+
+        If so, convert to UxDataArray.
+        """
+        if isinstance(value, xr.DataArray):
```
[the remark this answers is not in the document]
rajeeja: The idea is to return a `uxdataarray` object and attach `uxgrid` and other properties with 

`value = UxDataArray(value)`

uxarray/core/dataset.py
```diff
@@ -0,0 +1,138 @@
...
+    def __getitem__(self, key):
+        """Override to check if the result is an instance of xarray.DataArray.
+
+        If so, convert to UxDataArray.
+        """
+        xarr = super().__getitem__(key)
+
+        if isinstance(xarr, xr.DataArray):
+            return UxDataArray(xarr)
+        else:
```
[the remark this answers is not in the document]
rajeeja: you mean `xarr = super().__getitem__(key)` could return a list of datasets and the check should be xarr as list with each element being `xr.DataArray` I don't think we can handle that case, so skip it, should we skip it.

`elif isinstance(xarr, list): skip`

uxarray/core/dataarray.py
rajeeja: Right. Thanks for the details

uxarray/core/dataset.py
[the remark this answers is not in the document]
dcherian: I mean it should be 
``` python
else:
    assert isintance(xarr, xr.Dataset):
    return UxDataset(xarr, uxgrid=self.uxgrid)
```

erogluorhan:
EDIT (3/30/2023): 

Uxarray's new design can be tried out following the instructions in the [How to test new design](https://docs.google.com/document/d/1Ql2AClMazZchSvgGDdxtK5BUARV3KrpTvut0vDOgJEs/edit?usp=sharing) document.

philipc2:
> EDIT (3/30/2023):
> 
> Uxarray's new design can be tried out following the instructions in the [How to test new design](https://docs.google.com/document/d/1Ql2AClMazZchSvgGDdxtK5BUARV3KrpTvut0vDOgJEs/edit?usp=sharing) document.

This could be a good candidate for a separate section under the `README` for the prototype branch (see Shapely 2.0 modified readme for their redesign branch [here](https://github.com/shapely/shapely/tree/shapely-2.0))

uxarray/core/dataarray.py
erogluorhan: `_replace()` and many other required overridings have been made.

uxarray/core/dataset.py
erogluorhan: we received some invaluable comments from @headtr1ck at the time; resolving this now

rajeeja (approved):

uxarray/core/grid.py
erogluorhan: Discussed well in #324

ci/docs.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
     - ipython
     - jupyter_client
     - matplotlib-base
-    - sphinx-book-theme
+    - sphinx-book-theme<=0.13.1 # pydata-sphinx-theme 0.13.2 broke sphinx_book_theme import
```
philipc2, line 21: I don't believe this is necessary anymore.

uxarray/core/api.py
```diff
@@ -0,0 +1,263 @@
...
+        Additional arguments passed on to ``xarray.open_dataset``. Refer to the
+        [xarray
+        docs](https://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html)
+        for accepted keyword arguments.
+
+    Returns
+    -------
+
+    object : uxarray.Grid
+        Dataset with the unstructured grid.
```
philipc2, line 64:
```suggestion
    uxgrid : uxarray.Grid
        Initialized Grid Object from Input Grid File 
```

uxarray/core/api.py
```diff
@@ -0,0 +1,263 @@
...
+    Examples
+    --------
+
+    Open dataset with a grid topology file
+
+    >>> import uxarray as ux
+    >>> uxgrid = ux.open_grid("grid_filename.g")
+    """
+
+    ## Grid definition
```
philipc2, line 75:
```suggestion
    # Grid definition
```

uxarray/core/api.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def open_dataset(grid_filename_or_obj: str,
+                 filename_or_obj: str,
+                 gridspec: Optional[str] = None,
+                 vertices: Optional[list] = None,
+                 islatlon: Optional[bool] = False,
+                 isconcave: Optional[bool] = False,
+                 use_dual: Optional[bool] = False,
+                 **kwargs: Dict[str, Any]) -> xr.Dataset:
```
philipc2, line 108: Should this be `xr.Dataset` or `ux.UxDataset` ?

uxarray/core/api.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def open_mfdataset(grid_filename_or_obj: str,
+                   paths: Union[str, os.PathLike],
+                   gridspec: Optional[str] = None,
+                   vertices: Optional[list] = None,
+                   islatlon: Optional[bool] = False,
+                   isconcave: Optional[bool] = False,
+                   use_dual: Optional[bool] = False,
+                   **kwargs: Dict[str, Any]) -> xr.Dataset:
```
philipc2, line 188: Similar comment to above.

uxarray/core/api.py
```diff
@@ -0,0 +1,263 @@
...
+        Additional arguments passed on to ``xarray.open_dataset``. Refer to the
+        [xarray
+        docs](https://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html)
+        for accepted keyword arguments.
+
+    Returns
+    -------
+
+    object : uxarray.UxDataset
+        Dataset with the unstructured grid.
```
philipc2, line 153:
```suggestion
    uxds: uxarray.UxDataset
        Dataset with linked unstructured grid accessor.
```

uxarray/core/api.py
```diff
@@ -0,0 +1,263 @@
...
+        Additional arguments passed on to ``xarray.open_mfdataset``. Refer to the
+        [xarray
+        docs](https://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html)
+        for accepted keyword arguments.
+
+    Returns
+    -------
+
+    object : uxarray.UxDataset
+        Dataset with the unstructured grid.
```
philipc2, line 232:
```suggestion
    uxds: uxarray.UxDataset
        Dataset with linked unstructured grid accessor.
```

uxarray/core/dataarray.py
```diff
@@ -0,0 +1,99 @@
...
+import xarray as xr
+
+from typing import Optional
+
+from uxarray.core.grid import Grid
+
+
+class UxDataArray(xr.DataArray):
+    """N-dimensional ``xarray.DataArray``-like array. Inherits from
+    `xarray.DataArray` and has its own unstructured grid-aware array operators
```
philipc2, line 11:
```suggestion
    ``xarray.DataArray`` and has its own unstructured grid-aware array operators
```

uxarray/core/dataarray.py
```diff
@@ -0,0 +1,99 @@
...
+        grid topology it belongs to.
+
+        If `None`, it needs to be an instance of `uxarray.Grid`.
+
+    Other Parameters
+    ----------------
+    *args:
+        Arguments for the `xarray.DataArray` class
+    **kwargs:
+        Keyword arguments for the `xarray.DataArray` class
```
philipc2, line 27:
```suggestion
        Keyword arguments for the ``xarray.DataArray`` class
```

uxarray/core/dataarray.py
```diff
@@ -0,0 +1,99 @@
...
+    uxgrid : uxarray.Grid, optional
+        The `Grid` object that makes this array aware of the unstructured
+        grid topology it belongs to.
+
+        If `None`, it needs to be an instance of `uxarray.Grid`.
+
+    Other Parameters
+    ----------------
+    *args:
+        Arguments for the `xarray.DataArray` class
```
philipc2, line 25:
```suggestion
        Arguments for the ``xarray.DataArray`` class
```

uxarray/core/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+
+class UxDataset(xr.Dataset):
+    """A ``xarray.Dataset``-like, multi-dimensional, in memory, array database.
+    Inherits from ``xarray.Dataset`` and has its own unstructured grid-aware
+    dataset operators and attributes through the ``uxgrid`` accessor.
+
+    Parameters
+    ----------
+    uxgrid : uxarray.Grid, optional
+        The `Grid` object that makes this array aware of the unstructured
```
philipc2, line 22:
```suggestion
        The ``Grid`` object that makes this array aware of the unstructured
```

uxarray/core/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+        grid topology it belongs to.
+
+        If `None`, it needs to be an instance of `uxarray.Grid`.
```
philipc2, line 25:
```suggestion
        If ``None``, it needs to be an instance of ``uxarray.Grid``.
```

uxarray/core/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+
+    Other Parameters
+    ----------------
+    *args:
+        Arguments for the `xarray.Dataset` class
```
philipc2, line 30:
```suggestion
        Arguments for the ``xarray.Dataset`` class
```

uxarray/core/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+    **kwargs:
+        Keyword arguments for the `xarray.Dataset` class
```
philipc2, line 32:
```suggestion
        Keyword arguments for the ``xarray.Dataset`` class
```

uxarray/core/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+    def from_dict(cls, data, **kwargs):
+        """Override to make the result a ``uxarray.UxDataset`` class."""
+
+        return cls({key: ('index', val) for key, val in data.items()},
+                   coords={'index': range(len(next(iter(data.values()))))},
+                   **kwargs)
+
+    def info(self, buf: IO = None, show_attrs=False) -> None:
+        """Concise summary of a Dataset variables and attributes including grid
+        topology information stored in the ``uxgrid`` accessor.
```
philipc2, line 199:
```suggestion
        """Concise summary of Dataset variables and attributes including grid
        topology information stored in the ``uxgrid`` accessor.
```

uxarray/core/grid.py
```diff
@@ -1,61 +1,65 @@
...
+from uxarray.utils.constants import INT_DTYPE, INT_FILL_VALUE
 
 
 class Grid:
-    """
+    """Unstructured grid topology definition.
+
+    Can be used standalone to explore an unstructured grid topology, or
+    can be seen as the property of `uxarray.UxDataset` and `uxarray.DataArray`
+    and makes them ustrudtured grid-aware data sets and arrays.
```
philipc2, line 22:
```suggestion
    Can be used standalone to explore an unstructured grid topology, or
    can be seen as the property of ``uxarray.UxDataset`` and ``uxarray.DataArray``
    and makes them unstructured grid-aware data sets and arrays.
```

uxarray/core/grid.py
```diff
@@ -1,61 +1,65 @@
...
+
+    Parameters
+    ----------
+    input_obj : xarray.Dataset, ndarray, list, tuple, required
+        Input `xarray.Dataset` or vertex coordinates that form faces.
```
philipc2, line 27:
```suggestion
    input_obj : xarray.Dataset, ndarray, list, tuple, required
        Input ``xarray.Dataset`` or vertex coordinates that form faces.
```

philipc2 (changes requested):

uxarray/core/api.py
erogluorhan: Good catch; done!

uxarray/core/dataset.py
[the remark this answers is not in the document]
erogluorhan: Done; thanks for catching this!

uxarray/core/api.py
erogluorhan: Done

uxarray/core/dataset.py
erogluorhan: Done

uxarray/core/api.py
erogluorhan: Done, slightly differently

uxarray/core/dataset.py
erogluorhan: Done

uxarray/core/dataset.py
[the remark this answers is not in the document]
erogluorhan: I see! Yes, or even we may not be needing to override this function at all. I think UxDataset could handle addition of UxDataArray(s) into it via this func by itself. Regardless, I have created Issue #344 for this, and I am commenting it out in this PR to see.

ci/docs.yml
erogluorhan: Looks so!

uxarray/core/grid.py
```diff
@@ -261,24 +232,283 @@ def __from_ds__(self, dataset):
...
+    @property
+    def Mesh2_node_cart_x(self):
+        """Coordinate Variable ``Mesh2_node_cart_x``, which contains the x
+        location in meters.
+
+        Dimensions (``nMesh2_node``)
+        """
+        if "Mesh2_node_cart_x" not in self._ds:
+            self._populate_cartesian_xyz_coord()
+        return self._ds['Mesh2_node_cart_x']
```
philipc2, line 388: @hongyuchen1030 @erogluorhan 

I've added these properties to API, it should adequality expose the functionality of `Grid._populate_cartesian_xyz_coord` to the user now, without needing to explicitly call the function.

uxarray/core/api.py
```diff
@@ -0,0 +1,271 @@
...
+                       use_dual=use_dual)
+
+    # Data variables
+    if filename_or_obj is None:
+        # no data variables provided, defaults to empty
+        # TODO: parse grid file for data variables
+        ds = xr.Dataset()
+    else:
+        ds = xr.open_dataset(filename_or_obj, decode_times=False,
+                             **kwargs)  # type: ignore
```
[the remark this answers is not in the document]
erogluorhan: I'd suggest we avoid this quick addition witout testing it thoroughly, so I am reverting this commit. Instead, after merging this PR firs,  please a new PR (and issue) for that case, and work on i, including test cases as well.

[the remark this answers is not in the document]
erogluorhan: > For the case of MPAS grids or simply creating a blank dataset with a grid, the user may not want to immediately pass through a dataset file.

`open_grid` should also be considered with this case when working on it.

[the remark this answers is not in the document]
philipc2: Sounds good, I'll investigate a better way of doing this and fill out the issue!

philipc2 (approved):
Functionality wise everything is looking fantastic!

Before merging, I strongly suggest that we get a few more eyes into the documentation, especially the Getting Started & Usage Example sections. I've looked over them, however since I wrote them there's bound to be some things I haven't caught.

Great work on this everyone!

erogluorhan:
Thanks very much for your review, @philipc2 ! 

Just to note, I am planning to do two more things tonight:
1. Update the [draft API](https://github.com/UXARRAY/uxarray/blob/prototype_redesign/docs/user_api/uxarray_api.md) to reflect the changes (I am sorry I know this is not ideal to do that after the approvals)
2. Add an "ANNOUNCEMENT" under "For Users" (and maybe "For Developers) section(s) of the documentation to talk about this design change, which will stay there for a few months

docs/examples/001-working-with-unstructured-grids.ipynb
hongyuchen1030, line 1: Here it said: "For both the single and complete instances of UxDataset, the uxgrid property contains the same grid information, however they are each instantiated separately."

Can you provide more details about what's the difference between the `simple` and the `complete` instances?

docs/examples/002-grid-topology.ipynb
hongyuchen1030, line 1: 1. Can you give more details explaining the difference between the `Grid` object and the `UxDataset` here?

docs/user_api/index.rst
hongyuchen1030, line 1: Can you provide more explanation between the relationship of `UxDataset`, `UxArray` and `UxGrid`?

hongyuchen1030 (commented):
Thank you so much for putting these together. It looks awesome overall. Just left some comments about the documentation

uxarray/core/grid.py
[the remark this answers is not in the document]
erogluorhan: For now in this PR's scope, I'd suggest we stick to the Xarray's way (what we currently implemented here), but we can think about it in the future

docs/getting-started/overview.rst
```diff
@@ -0,0 +1,67 @@
+.. currentmodule:: uxarray
+
+======================
+Overview: Why UXarray?
+======================
+UXarray aims to address the geoscience community need for tools that enable standard
+data analysis techniques to operate directly on unstructured grids. It extends upon
+and inherits from the commonly used Xarray Python package to provide a powerful and
+familiar interface for working with unstructured grids in Python.UXarray provides
```
ifranda, line 9:
```suggestion
familiar interface for working with unstructured grids in Python. UXarray provides
```

docs/examples/002-grid-topology.ipynb
erogluorhan: Great catch, thanks! Added new parts and modified the content of the beginning of the notebook to make the distinction

docs/examples/001-working-with-unstructured-grids.ipynb
erogluorhan: Renamed "complete" to "multiple" and also modified the text to make the difference more clear

docs/user_api/index.rst
erogluorhan: Added more explanation into the entries for each of those classes. There is not a separate section that explains the relation between these as this is an API document, but I hope the new text under each of these classes is more clear now (e.g. Grid's text says it can be used standalone to explore grid topology or as part of ``uxarray.UxDataset`` and ``uxarray.DataArray`` as well)

philipc2:
@erogluorhan 

I have updated our "Working with Unstructured Grid Data" to reflect the comments in #181 and the work that we've done with the inheritance notebook.

Let me know if there is anything else I should add!

docs/examples/002-grid-topology.ipynb
```diff
@@ -0,0 +1,546 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "ux.INT_FILL_VALUE"
+   ],
+   "metadata": {
+    "collapsed": false
+   }
+  },
```
ifranda, line 348: Should we add a bit of text here to separate the constants (that are from the UGRID conventions) from the data (that are specific to the file we loaded)? We could add something simple like this, unless it seems redundant:

```suggestion
  },
  {
   "cell_type": "markdown",
   "source": [
    "The grid's connectivity information can be accessed as follows:"
   ],
   "metadata": {
    "collapsed": false
   }
  },
```

ifranda (approved):

docs/getting-started/overview.rst
```diff
@@ -0,0 +1,67 @@
...
+
+
+Unstructured Grids
+==================
+The "U" in UXarray stands for "Unstructured Grids". These types of grids differ from
+typical Structured Grids in terms of complexity, requiring additional overhead to
+store and represent their geometry and topology. However, these types of
+grids are extremely flexible and scalable.
+
+UXarray uses the UGRID conventions as a
```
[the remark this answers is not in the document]
philipc2: Great idea!

docs/user_api/index.rst
hongyuchen1030: Thanks! That looks great now.

uxarray/core/grid.py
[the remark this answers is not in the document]
hongyuchen1030: Got it.

hongyuchen1030 (approved):
Thank you for your hard work! The documentation looks good to me now

PR #219, opened 2023-02-04
DRAFT: zonal average chen

hongyuchen1030:


hongyuchen1030:
TODO: Need to fix the double counting edges when the constant latitude overlaps with an edge that's shared by two faces

PR #222, opened 2023-02-13
Fix Reading in Data usage example

philipc2:
Fixes Reading in Data usage example by modifying the path to correctly take into account the changes made in #211


erogluorhan (approved):

hCraker (approved):
Looks good! It's best practice to restart your kernel and run all cells before pushing changes so the meta data (e.g. execution_count) doesn't change so much. It makes it easier for us to see what you actually changed about the content. Whether or not you do that for this example, I think it's good to go.

PR #224, opened 2023-03-04
o reorganize files into folders, fix tests and docs

rajeeja:


erogluorhan:
Thanks very much for this!

PR #228, opened 2023-03-08
Read MPAS 

philipc2:
### Overview
Provides support for reading in MPAS grid datasets and encoding them in the UGRID conventions, written according to the MPAS specifications outlined in [this](https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf) document. 
* Similar approach to what is already implemented for other grid conventions (`_scrip.py`, `_exodus.py`)
* Supports converting both the Primal and Dual Meshes to the UGRID conventions.
* Assigns all required UGRID variables for a 2D topology, with the addition of `Mesh2_face_x/y` and `Mesh2_edge_nodes` when available 

### Usage
```Python
import xarray as xr 
import uxarray as ux

# path to grid file in the mpas format
mpas_grid_path = "/path/to/mpas/grid.nc"

# load with xarray 
mpas_xr_ds = xr.open_dataset(mpas_grid_path)

# load the primal mesh
ugrid_primal = ux.Grid(mpas_xr_ds)
# alternatively 
# ugrid_primal = ux.Grid(mpas_xr_ds, use_dual=False)

# load the dual mesh 
ugrid_dual = ux.Grid(mpas_xr_ds, use_dual=True)

### CODE REVIEW

**Comment by rljacob** [comment]
Are the contents of an MPAS grid file specified somewhere?

**Comment by philipc2** [comment]
> Are the contents of an MPAS grid file specified somewhere?

https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf

**Comment by philipc2** [comment]
Hi @cdeciampa 

Considering your recent work with visualizing MPAS grids, I'd like to get your input on converting from MPAS to UGRID. Let me know if you have any questions!

**Review by cdeciampa — COMMENTED** [review]
Looks good, the only suggestion I have is to convert lonCell, latCell, lonVertex, and latVertex from radians to degrees. I'm curious, though, why use the dual-mesh instead of the primary mesh?

**Comment by philipc2** [comment]
> Looks good, the only suggestion I have is to convert lonCell, latCell, lonVertex, and latVertex from radians to degrees. I'm curious, though, why use the dual-mesh instead of the primary mesh?

Whoops! I think I mixed up the Dual and Primary meshes when writing the comments. This implementation should be from the _Primary_ mesh, not the Dual one. Good catch!

**Comment by rljacob** [comment]
@xylar check this out.

File: uxarray/_mpas.py [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+    """
+    # convert to unsigned integers
+    verticesOnCell = verticesOnCell.astype(int_dtype)
+
+    # iterate over each primal mesh cell
+    for D, pad_start_idx in enumerate(nEdgesOnCell):
+        # convert non-padded values to zero-index
+        verticesOnCell[D, 0:pad_start_idx] -= 1
+        # replace remaining padding or zeros with fill_value
+        verticesOnCell[D, pad_start_idx:] = fill_val
```
xylar, line 123: Iterating over cells in python is quite slow.  I have always iterated over the `maxEdges` dimension instead in loops like this one:
```suggestion
    maxEdges = verticesOnCell.shape[1]
    for vert_idx in range(maxEdges):
        mask = vert_idx < nEdgesOnCell
        # convert non-padded values to zero-index
        verticesOnCell[mask, vert_idx] -= 1
        # replace remaining padding or zeros with fill_value
        verticesOnCell[np.logical_not(mask), vert_idx] = fill_val
```

uxarray/helpers.py
```diff
@@ -41,6 +41,8 @@ def parse_grid_type(dataset):
     # ugrid topology
     elif _is_ugrid(dataset):
         mesh_type = "ugrid"
+    elif "verticesOnCell" in dataset:
+        mesh_type = "mpas"
```
xylar, line 45: It looks like the docstring needs to be updated to include MPAS (and maybe remove shape files, which don't in fact seems to be supported yet).

Also, is there any concern that the approach of looking for a specific variable name to detect a given format won't be robust?  `verticesOnCell` seems pretty safe for detecting MPAS but `coord` and `coordx` seem like they could be pretty generic names.  I know from ParaView that it can be frustrating to have files misidentified because the reader's identification of a format is too generic (every NetCDF file gets identified as a CAM atmosphere file for some reason).

test/test_mpas.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+from pathlib import Path
+
+current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+
+
+class TestMPAS(TestCase):
+
+    mpas_small_grid_path = current_path / 'meshfiles' / "mpas" / "x1" / 'x1.40962.static.nc'
+    mpas_large_grid_path = current_path / 'meshfiles' / "mpas" / "x1" / 'x1.655362.grid_subset.nc'
```
xylar, line 16: These examples are pretty big files.  In my experience, having multi MB files in a repo gets cumbersome.  I would suggest something much coarser.  As examples, we have this very coarse mesh in MPS-Tools:
https://github.com/MPAS-Dev/MPAS-Tools/blob/master/mesh_tools/mesh_conversion_tools/test/mesh.QU.1920km.151026.nc

xylar (commented):
This looks really exciting!

I do think having support for the dual mesh will be quite important, as we often want to manipulate and visualize quantities on both meshes.  But the primary mesh is certainly the right starting point.

uxarray/_mpas.py
xylar: I ran `pytest test` locally with these changes and it passes (not sure how rigorous a test that is).

I'm happy to do some other testing but not sure how much can be done with uxarray yet.  The docs don't seem to go much beyond reading in files and getting connectivity info in various ways. (I fully understand that this is a work in progress.)

cdeciampa: Now that I'm taking a closer look at this (it's been a long week), I'm noticing that within `verticesOnCell`, the shapes that have sides n < 7 are all padded out with the same value, which doesn't actually exist as a vertex index in the mesh file. While I was trying to test this on my own with Polymesh, I ran into downstream issues with numpy indexing for coordinate values (ie; `lonCell[verticesOnCell]`, `latCell[verticesOnCell]`).

I've done this differently within my own code by having each individual cell connect back to each individual "origin" vertex. It's a little clunky, but is marginally faster this way as well (~20-30 ms):

```
%%time
# MPAS indexing starts at 1, not 0 (any original 0 value ~ NaN)
face_nodes = c_mesh_ds.verticesOnCell.values - 1

# Replaces all -1 values with first value in array to force 7-sided shapes
face_nodes[:, 4] = np.where(face_nodes[:, 4] == -1, face_nodes[:, 0], face_nodes[:, 4])
face_nodes[:, 5] = np.where(face_nodes[:, 5] == -1, face_nodes[:, 0], face_nodes[:, 5])
face_nodes[:, 6] = np.where(face_nodes[:, 6] == -1, face_nodes[:, 0], face_nodes[:, 6])

# Ensures face nodes are returned as integers
face_nodes = face_nodes.astype(np.uint32)
```

```
CPU times: user 27.9 ms, sys: 32.2 ms, total: 60.1 ms
Wall time: 58.2 ms
```

xylar: @cdeciampa, yes, that's another approach that's good for visualization (and I believe also the approach expected by the SCRIP format).  I don't know if that's an appropriate solution for uxarray more broadly.  That presumably needs to be a decision on what the standard uxarray format should be.

test/test_mpas.py
rljacob: I agree we should avoid large binary files in a git repo.

uxarray/helpers.py
philipc2: > It looks like the docstring needs to be updated to include MPAS (and maybe remove shape files, which don't in fact seems to be supported yet).
> 
> Also, is there any concern that the approach of looking for a specific variable name to detect a given format won't be robust? `verticesOnCell` seems pretty safe for detecting MPAS but `coord` and `coordx` seem like they could be pretty generic names. I know from ParaView that it can be frustrating to have files misidentified because the reader's identification of a format is too generic (every NetCDF file gets identified as a CAM atmosphere file for some reason).

Good catch on the docstring. 

I agree that the way that we check what format the grid is in is not the most robust. I'll raise an issue about it today, since having more complete functions to check what format a dataset is in would be better (i.e. `_is_mpas()`, `_is_scrip()`, etc)

philipc2:
> This looks really exciting!
> 
> I do think having support for the dual mesh will be quite important, as we often want to manipulate and visualize quantities on both meshes. But the primary mesh is certainly the right starting point.

@erogluorhan @rajeeja 

How would you suggest handling both meshes within `UXarray`? Since we currently have a singular dataset under our `Grid` object, storing both the Primal and Dual mesh would not be ideal. Should we add a parameter to let the user chose which grid to load in?

uxarray/_mpas.py
xylar: @philipc2, my version accounts for all of that.  The mask array will be true for cells that have a vertex with the given index and will be false for cells that don't have it.  It is vastly more efficient to iterate over 7 to 9 vertices than it is to iterate over potentially tens of millions of cells.  So I stand by my version being much preferable, especially for very large MPAS meshes.  The result should be the same as your loop, as you can check.

rljacob:
If a file has multiple meshes in it, then yes there should be a parameter to let the user specify which one to load.  And if they go back to that mesh file for a second mesh, thats a second grid object. 

MPAS files I assume have easy ways to find one or the other mesh.  Not sure how to do that in general

uxarray/helpers.py
philipc2: #238

philipc2:
@xylar 

All changes should be addressed now, with the added support for the Dual mesh.

uxarray/_mpas.py
```diff
@@ -0,0 +1,266 @@
...
+    """
+
+    # convert to unsigned integers
+    verticesOnCell = verticesOnCell.astype(int_dtype)
+
+    # max vertices/edges per cell
+    maxEdges = verticesOnCell.shape[1]
+
+    # is verticesOnCell one-indexed
+    one_idx = [True if verticesOnCell.min() == 1 else False]
```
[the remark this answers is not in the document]
xylar:
```suggestion
    one_idx = (verticesOnCell.min() == 1)
```
This is currently working for completely the wrong reasons.  It is a one-item list, whereas it is intended to be a boolean.  Then, `if one_idx` evaluates to True later on because the list is not empty.

xylar (changes requested):
I don't think the conversion to zero-based indexing is working for the right reasons, see inline comments below.

uxarray/_mpas.py
```diff
@@ -0,0 +1,266 @@
...
+            "_FillValue": fill_val,
+            "start_index": int_dtype(0)
+        })
+
+    # vertex indices that saddle a given edge
+    verticesOnEdge = in_ds['verticesOnEdge'].values
+
+    # convert to zero-indexed
+    if verticesOnEdge.min() == 1:
+        verticesOnEdge -= 1
```
xylar, line 89: I just noticed these kinds of checks.  These also aren't going to work for the same reason I pointed out -- zeros as invalid values.

uxarray/_mpas.py
```diff
@@ -0,0 +1,266 @@
...
+        latVertex,
+        dims=["nMesh2_face"],
+        attrs={
+            "standard_name": "latitude",
+            "long_name": "latitude of center nodes",
+            "units": "degrees_north",
+        })
+
+    # vertex indices that surround each dual-mesh cell
+    cellsOnVertex = in_ds['cellsOnVertex'].values
```
xylar, line 158: For the dual mesh, we need to talk about how to handle vertices with fewer than 3 neighboring cells.  These are indicated by zeros in `cellsOnVertex`.  The occur in planar or spherical meshes where some cells have been removed (e.g. land cells removed from the ocean).  For visualization purposes, a vertex with one or more invalid neighboring cells is degenerate.  But for mathematical purposes, they remain useful.  I don't know how uxarray plans to handle such situations.

In any case, some code needs to be added to convert zero values to mask values here.

philipc2:
> how about point this to prototype branch rather than main? I think it will break a few things.
> 
> Good work with the changes for dual and prime mesh. We should document this in a notebook example.

We discussed the idea of linking this PR to the prototype branch during our Raijin meeting. We agreed that it would be best to keep this linked to the main branch, since the implementation and algorithms can be done disjoint from any of the redesign work. 

I think having a notebook example would be a great idea.

xylar (approved):
Excellent!  I think this looks right to me.  Nice testing!

uxarray/_mpas.py
philipc2: This is a great point. I think this would be a good candidate for a helper function.

xylar: I think your call to `_replace_zeros()` has taken care of this nicely.

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    Parameters
+    ----------
+    verticesOnCell : np.ndarray
+        Vertex indices that surround a given cell
+
+    nEdgesOnCell : np.ndarray
+        Number of edges on a given cell
+
+    Outputs
+    ----------
```
anissa111, line 229:
```suggestion
    Returns
    -------
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    """Replaces all instances of a zero (invalid/missing MPAS value) with a
+    fill value.
+
+    Parameters
+    ----------
+    grid_var : np.ndarray
+        Grid variable that may contain zeros that need to be replaced
+
+     Outputs
+    ----------
```
anissa111, line 258:
```suggestion
    Returns
    -------
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    return verticesOnCell
+
+
+def _replace_zeros(grid_var):
+    """Replaces all instances of a zero (invalid/missing MPAS value) with a
+    fill value.
+
+    Parameters
+    ----------
+    grid_var : np.ndarray
```
anissa111, line 254:
```suggestion
    grid_var : numpy.ndarray
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+        Grid variable that may contain zeros that need to be replaced
+
+     Outputs
+    ----------
+    grid_var : np.ndarray
```
anissa111, line 259:
```suggestion
    grid_var : numpy.ndarray
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    return grid_var
+
+
+def _to_zero_index(grid_var):
+    """Given an input using that is one-indexed, subtracts one from all non-
+    fill value entries to convert to zero-indexed.
+
+    Parameters
+    ----------
+    grid_var : np.ndarray
```
anissa111, line 278:
```suggestion
    grid_var : numpy.ndarray
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+        Grid variable that is one-indexed
+
+     Outputs
+    ----------
```
anissa111, line 282:
```suggestion
    Returns
    -------
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    grid_var : np.ndarray
```
anissa111, line 283:
```suggestion
    grid_var : numpy.ndarray
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+def _read_mpas(ext_ds, use_dual=False):
+    """Function to read in a MPAS Grid dataset and encode either the Primal or
+    Dual mesh in the UGRID conventions.
+
+    Parameters
+    ----------
+    ext_ds : xarray.Dataset, required
+        MPAS datafile of interest
+    use_dual : bool
+        Flag to select whether to encode the Dual-Mesh
```
anissa111, line 304:
```suggestion
    use_dual : bool, optional
        Flag to select whether to encode the Dual-Mesh. Defaults to False
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+                "topology_dimension": 2,
+                "node_coordinates": "Mesh2_node_x Mesh2_node_y",
+                "node_dimension": "nMesh2_node",
+                "face_node_connectivity": "Mesh2_face_nodes",
+                "face_dimension": "nMesh2_face"
+            })
+    except:
+        print("Variables not in recognized MPAS form. Please refer to",
+              "https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf",
+              "for more information on MPAS Grid file formatting")
```
anissa111, line 336: Could we convert this print statement to an explicit error or warning?

uxarray/grid.py
```diff
@@ -206,6 +209,12 @@ def __from_ds__(self, dataset):
             self.ds, self.ds_var_names = _read_ugrid(dataset, self.ds_var_names)
         elif self.mesh_type == "shp":
             self.ds = _read_shpfile(dataset)
+        elif self.mesh_type == "mpas":
+            # select whether to use the dual mesh
+            if self.use_dual is not None:
```
anissa111, line 214: Why are you using `is not None` here? Wouldn't this theoretically produce unexpected results if `self.use_dual` is set, but set to False?

test/test_mpas.py
```diff
@@ -0,0 +1,107 @@
...
+    mpas_xr_ds = xr.open_dataset(mpas_grid_path)
+
+    # fill value (remove once there is a unified approach in uxarray)
+    int_dtype = np.uint32
+    fv = np.iinfo(int_dtype).max
+
+    def test_primal_to_ugrid_execution(self):
+        """Tests execution of _primal_to_ugrid()"""
+        ugrid_ds = xr.Dataset()
+        _primal_to_ugrid(self.mpas_xr_ds, ugrid_ds)
```
anissa111, line 27: Aren't these tests up here duplicating some work done by tests below? Do we need separate tests for execution and correctness since correctness will also fail if its execution fails?

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    """Replaces the padded values in verticesOnCell defined by nEdgesOnCell
+    with a fill-value.
+
+    Parameters
+    ----------
+    verticesOnCell : np.ndarray
+        Vertex indices that surround a given cell
+
+    nEdgesOnCell : np.ndarray
+        Number of edges on a given cell
```
anissa111, line 226:
```suggestion
    verticesOnCell : numpy.ndarray
        Vertex indices that surround a given cell

    nEdgesOnCell : numpy.ndarray
        Number of edges on a given cell
```

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Outputs
+    ----------
+    verticesOnCell : np.ndarray
```
anissa111, line 230:
```suggestion
    verticesOnCell : numpy.ndarray
```

anissa111 (changes requested):
Here's my first pass, thanks for working on this!

I still want to take a look at whether we can replace some of the for loops and numpy functions with more dask friendly alternatives, but I don't want to leave you waiting for a review for that

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+        out_ds['mesh_spec'] = in_ds.mesh_spec
+    if "on_a_sphere" in in_ds:
+        out_ds['on_a_sphere'] = in_ds.on_a_sphere
+        # required attributes if mesh lies on a sphere
+        if in_ds.on_a_sphere == "YES":
+            out_ds['is_periodic'] = in_ds.is_periodic
+            # required attributes if mesh is periodic
+            if in_ds.is_periodic == "YES":
+                out_ds['x_period'] = in_ds.x_period
+                out_ds['y_period'] = in_ds.y_period
```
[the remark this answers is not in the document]
philipc2: Good idea, a warning would be appropriate.

uxarray/grid.py
philipc2: Since `use-dual` is part of the `**kwargs` of our `Grid` class, it's set to `None` if the user doesn't pass it through. See the usage example in the description for how it's supposed to be used.

uxarray/_mpas.py
```diff
@@ -0,0 +1,338 @@
...
+    if 'sphere_radius' in in_ds:
+        out_ds['sphere_radius'] = in_ds.sphere_radius
+    if 'mesh_id' in in_ds:
+        out_ds['mesh_id'] = in_ds.mesh_id
+    if 'mesh_spec' in in_ds:
+        out_ds['mesh_spec'] = in_ds.mesh_spec
+    if "on_a_sphere" in in_ds:
+        out_ds['on_a_sphere'] = in_ds.on_a_sphere
+        # required attributes if mesh lies on a sphere
+        if in_ds.on_a_sphere == "YES":
```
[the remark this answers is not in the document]
philipc2: Good catch, thanks!

uxarray/grid.py
anissa111: Ah, yeah I see that now!

uxarray/_mpas.py
```diff
@@ -0,0 +1,356 @@
...
+                "node_coordinates": "Mesh2_node_x Mesh2_node_y",
+                "node_dimension": "nMesh2_node",
+                "face_node_connectivity": "Mesh2_face_nodes",
+                "face_dimension": "nMesh2_face"
+            })
+    except:
+        raise Exception(
+            "Variables not in recognized MPAS form. Please refer to",
+            "https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf",
+            "for more information on MPAS Grid file formatting")
```
anissa111, line 354: I'm a bit concerned that this exception is broad enough, and there's enough that could go wrong in the try block, that this error message might not appropriately reflect what the root cause of an issue may be. 

I ran into this situation when I was running the tests while messing with some of the suggestions later in this review. I had made a typo in a variable name and the tests all triggered this exception message.

Would it be possible to wrap only the parts of `_dual_to_ugrid` and `_primal_to_ugrid` that are extracting the MPAS attributes in a try/catch instead of the entire read-in process?

anissa111 (commented):
I'll approve after these are addressed!

uxarray/_mpas.py
xylar: Oh, I missed this but I agree with @anissa111 that the `try`/`except` is covering too much code.  In addition, it is best to avoid raising a generic `Exception` because as @anissa111 says, it catches *everything* including typos in the code or even someone trying to abort the code with `^c` (a `KeyboardInterrupt`).  If existing exceptions like `ValueError` or `OSError` aren't sufficient, you should strongly consider implementing your own exception.

philipc2: Good points, I didn't give this part enough attention (based it off what was already implemented in `_scrip.py`).

I'll work on setting up some more robust catches.

xylar: It might be better not to have this `try`/`except` at all and to focus on individual errors within each function that are more specific to the problems actually triggering the error.  As a user of `uxarray`, I wouldn't find this blanket exception very useful because it would replace a different exception that could point me toward the problem in more detail.

philipc2: I've removed the `try` / `except` entirely. 

Once we have more robust grid format checks, we should be able to handle errors and exceptions outside of these I/O modules. (see #238)

Example:
* User inputs a subset of an MPAS grid that excludes necessary variables to convert to UGRID
* We should be able to catch this before jumping into `_read_mpas()` and raise an appropriate exception

xylar: That sounds like a much cleaner plan  Thanks!

anissa111 (approved):

philipc2:
@xylar 

Apologies for the re-request, you can ignore it (the grey checkmark confused me 😄) 

We need one more approval from anyone part of the Uxarray Group (@erogluorhan, @rajeeja , @rljacob) to get this merged.

xylar (approved):

xylar:
No worries.  Happy to give you an approval that doesn't quite count :-)

Nice work on this!  No doubt there will be a few more hiccups to iron out as we try this in our workflows but it looks great to me.

rajeeja (approved):

PR #231, opened 2023-03-10
`==` and `!=` Operators for Grid Object 

philipc2:
### Overview 
Modifies the previous implementation of `uxarray.Grid.equals(self, other)` to use the `==` operator. Additionally implements the `!=` operator. Both implementations are done using operator overloading. 

### Example
```Python
# Check if two grids are equal
uxds_01.uxgrid == uxds_02.uxgrid
# Check if two grids are not equal
uxds_01.uxgrid != uxds_02.uxgrid


rajeeja:
Does this fix the python 3.7 failure on the prototype branch?

philipc2:
> Does this fix the python 3.7 failure on the prototype branch?

Had to merge the most recent branch to double check, but it appears to be passing.

rajeeja (approved):

rajeeja:
> @rajeeja 
> 
> While I'd love to claim that my simple changes here fixed the CI issues, it appears that it must of been an issue somewhere with one of the upstream packages, since re-running the failed checks seemed to fix the issue. Glad it works now though! 

sounds good. Let's see how it does after this merge!

Issue #189, opened 2022-12-13
The current data read-in will make the FillValue=nan instead of -1


hongyuchen1030:
Our current implementation to construct Grid from 
```
xr_tgrid1 = xr.open_dataset(str(ug_filename1))
tgrid1 = ux.Grid(xr_tgrid1)

```
According to our API documentation, the `uxarray.Grid.Mesh2_face_nodes` should have the following properties
> uxarray.Grid.Mesh2_face_nodes: integer DataArray of size (nMesh2_face, MaxNumNodesPerFace)
> A DataArray of indices for each face node, corresponding to coordinates in uxarray.Grid.node_*. Faces can have arbitrary length, with _FillValue=-1 used when faces have fewer nodes than MaxNumNodesPerFace. Nodes are in counter-clockwise order.

However, the current Grid construction will make the _FillValue=nan instead of -1. This already caused a malfunction in my implemented `build_edge_face_connectivity` which I will take care of later. But such inconsistency might bring more problems in the following code developments

erogluorhan:
Thanks! I think defaulting to NaN is better than to -1, so we need to change the draft API doc. Is it also what you meant to fix the inconsistency?

paullric:
Why do you think defaulting to NaN is better than to -1?

_FillValue = nan is notorious for breaking things, though perhaps that is a legacy issue.

erogluorhan:
I'd say, 
- NaN is consistently used for missing data representation in Python, 
- `np.nan` could enable vectorized execution and hence I'd expect improved performance
- I also personally like the fact that `-1 == -1` is true but `NaN == NaN` is false, which would be useful for cases such as comparing indices, etc. 
- Despite not applicable to the case here (i.e. -1 can never be an index), using particular values for missing data might lead to not-breaking but wrong-executing codes.

Though, NaN might be converting the data arrays it resides in to floating point even if they are integer (which is this case since we are working on index numbers). I don't know if this'd be an issue.

hongyuchen1030:
> Thanks! I think defaulting to NaN is better than to -1, so we need to change the draft API doc. Is it also what you meant to fix the inconsistency?

Yes. I was just saying no matter what kind of FillValue we choose finally, it's better to have a clear description in the documentation so that we can base the following development on that.

Currently, I have considered FillValue = -1 and nan in my following implementation to eliminate possible marginal cases.

rljacob:
[...]

PR #241, opened 2023-03-16
Standardized Fill Values and Data Types for Connectivity Variables

### LINKED ISSUE #239: Add Standardized Data Types and Fill Values for UGRID connectivity variables
Use a standardized data type and fill values for connectivity variables. Continutation of #189

philipc2:
Closes #239, #189

### Overview 
* Introduces the `constants.py` module, which currently only stores the standardized integer data type `INT_DTYPE` and fill value `INT_FILL_VALUE` which is derived from the integer data type
* As discussed in #189, the standard data type for integer indexing variables will be set to `uintp`, with a corresponding fill value of `max(uintp`)
* `Mesh2_face_nodes` is set to always be of type `uintp` (either `uint32` or `uint64` depending on system) and use the fill value as described above
* Changes to I/O modules `_scrip.py`, `_ugrid.py` and `_exodus.py` were made to correctly represent `Mesh2_face_nodes`
* Introduction of the `replace_fill_values()` helper function, which can be used on any grid connectivity variable to replace all instances of a face value with a new one, with the addition of converting data types 

### Expected Use
* With the fill value now being standardized, they can be either accessed through the `_FillValue` attribute under `Mesh2_face_nodes` or through `uxarray.constants.INT_FILL_VALUE`
* Additional connectivity variables, such as `Mesh2_edge_nodes`, should adhere to the standards set in `uxarray.constants`


uxarray/constants.py
```diff
@@ -0,0 +1,5 @@
+import numpy as np
+
+# is 32-bit safe or should we use 64-bit?
+INT_DTYPE = np.uint32
+FILL_VALUE = np.iinfo(INT_DTYPE).max
```
rljacob, line 5: Should this be INTFILL_VALUE ?   What are doing for data on the grid that is in float and comes with a _FillValue?  Do we still want those to be Nan?

rljacob: In https://github.com/UXARRAY/uxarray/issues/189#issuecomment-1428509976 you thought we'd need FILL_VALUE32 and FILL_VALUE64.

philipc2: Good point, we should specify that this fill value is for integer arrays, I like the name `INT_FILL_VALUE`

The `replace_fill_values()` method handles converting data types and pre-defined fill values to the standardized versions, with correct edge case handling.

I think `FLOAT_FILL_VALUE` should be set to `NaN`, but this PR is only standardizing the connectivity variables, which should never be floats. Once we get these changes merged we can standardize the float variables to use a consistent fill value.

philipc2: > In [#189 (comment)](https://github.com/UXARRAY/uxarray/issues/189#issuecomment-1428509976) you thought we'd need FILL_VALUE32 and FILL_VALUE64.

After giving it some more thought, for connectivity variables I don't see a situation where we'd need more than 32-bit precision. However, we could use [`np.uintp`](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.intp:~:text=%27p%27-,numpy.uintp,%23,-Alias%20for%20the) which would dynamically change to either 32 or 64 bit depending on the system.

erogluorhan (commented):
Although I like this overall, I have a few in-line comments

uxarray/helpers.py
```diff
@@ -556,3 +557,45 @@ def _normalize_in_place(node):
         raise RuntimeError("Input array should have a length of 3: [x, y, z]")
 
     return list(np.array(node) / np.linalg.norm(np.array(node), ord=2))
+
+
+def replace_fill_values(grid_var, original_fill, new_fill, new_dtype=None):
```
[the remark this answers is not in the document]
philipc2: 1. Changing it to an internal helper is safer, change made is most recent commit 
2. Will work on this in the redesign branch, current location is fine for now

erogluorhan (approved):
Approving this per its scope, but please be sure to create a corresponding issue to keep track of the broader requests made in this convo (e.g. floating point, moving from helpers.py to a better place, etc). Thanks for this!

Issue #248, opened 2023-03-24
Add Standardized Data Types and Fill Values for UGRID floating point variables 


philipc2:
https://github.com/UXARRAY/uxarray/pull/241#pullrequestreview-1351271149

Use a standardized data type and fill values for floating point variables.

PR #259, opened 2023-03-29
Fix to `nan` Values Breaking UGRID Data Type and Fill Values

philipc2:
# Overview
* Implementation from #241 broke when `nan` was used for `Mesh2_face_nodes`, since `np.isnan` needs to be used to locate any occurrences of `nan`
* This hotfix modifies `test_standardized_dtype_and_fill` to separate datasets that contain a fill value and those that don't
* Changes were made to `_read_ugrid` to correctly parse the original fill value 
* `_replace_fill_values` now uses `np.isnan` to find all occurrences of `nan`

### DIFF
--- a/test/test_ugrid.py
+++ b/test/test_ugrid.py
@@ -82,8 +82,14 @@ class TestUgrid(TestCase):
         ux_grid3 = ux.Grid(xr_grid3)
 
         # check for correct dtype and fill value
-        grids = [ux_grid1, ux_grid2, ux_grid3]
-        for grid in grids:
+        grids_with_fill = [ux_grid2]
+        for grid in grids_with_fill:
+            assert grid.Mesh2_face_nodes.dtype == INT_DTYPE
+            assert grid.Mesh2_face_nodes._FillValue == INT_FILL_VALUE
+            assert INT_FILL_VALUE in grid.Mesh2_face_nodes.values
+
+        grids_without_fill = [ux_grid1, ux_grid2]
+        for grid in grids_without_fill:
             assert grid.Mesh2_face_nodes.dtype == INT_DTYPE
             assert grid.Mesh2_face_nodes._FillValue == INT_FILL_VALUE
 
@@ -93,9 +99,9 @@ class TestUgrid(TestCase):
 
         with dask chunking
         """
-        ug_filename = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30.ug"
+        ug_filename = current_path / "meshfiles" / "ugrid" / "outRLL1deg" / "outRLL1deg.ug"
         xr_grid = xr.open_dataset(str(ug_filename), chunks={'nMesh2_node': 100})
         ux_grid = ux.Grid(xr_grid)
         assert ux_grid.Mesh2_face_nodes.dtype == INT_DTYPE
         assert ux_grid.Mesh2_face_nodes._FillValue == INT_FILL_VALUE
-        pass
+        assert INT_FILL_VALUE in ux_grid.Mesh2_face_nodes.values
--- a/test/test_ugrid.py
+++ b/test/test_ugrid.py
@@ -88,7 +88,7 @@ class TestUgrid(TestCase):
             assert grid.Mesh2_face_nodes._FillValue == INT_FILL_VALUE
             assert INT_FILL_VALUE in grid.Mesh2_face_nodes.values
 
-        grids_without_fill = [ux_grid1, ux_grid2]
+        grids_without_fill = [ux_grid1, ux_grid3]
         for grid in grids_without_fill:
             assert grid.Mesh2_face_nodes.dtype == INT_DTYPE
             assert grid.Mesh2_face_nodes._FillValue == INT_FILL_VALUE
--- a/uxarray/helpers.py
+++ b/uxarray/helpers.py
@@ -584,7 +584,10 @@ def _replace_fill_values(grid_var, original_fill, new_fill, new_dtype=None):
     """
 
     # locations of fill values
-    fill_val_idx = grid_var.ravel() == original_fill
+    if original_fill is not np.nan:
+        fill_val_idx = grid_var.ravel() == original_fill
+    else:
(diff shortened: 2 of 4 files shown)


philipc2:
@erogluorhan @anissa111 @rajeeja 

This is a hotfix PR to the issue @hongyuchen1030 discovered in #191. Hoping we can get this merged ASAP so that it doesn't inhibit her work.

uxarray/helpers.py
```diff
@@ -584,7 +584,10 @@ def _replace_fill_values(grid_var, original_fill, new_fill, new_dtype=None):
     """
 
     # locations of fill values
-    fill_val_idx = grid_var.ravel() == original_fill
+    if original_fill is not np.nan:
+        fill_val_idx = grid_var.ravel() == original_fill
+    else:
+        fill_val_idx = np.isnan(grid_var.ravel())
```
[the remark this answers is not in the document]
philipc2: Forgot about this :)

uxarray/helpers.py
```diff
@@ -584,7 +584,10 @@ def _replace_fill_values(grid_var, original_fill, new_fill, new_dtype=None):
...
 
     # locations of fill values
-    fill_val_idx = grid_var.ravel() == original_fill
+    if original_fill is not np.nan:
+        fill_val_idx = grid_var == original_fill
+    else:
+        fill_val_idx = np.isnan(grid_var)
 
     # convert to new data type
     if new_dtype != grid_var.dtype and new_dtype is not None:
```
hongyuchen1030, line 593: what will happen if `new_dtype != grid_var.dtype and new_dtype is None:` ?

uxarray/helpers.py
```diff
@@ -598,6 +601,6 @@ def _replace_fill_values(grid_var, original_fill, new_fill, new_dtype=None):
                          f' dtype: {grid_var.dtype}')
```
hongyuchen1030, line 601: If it's possible that the `new_fill` is also a `nan`? I am just thinking about some extremely marginal case in the future. But since we have already settled down the `fill_value` to be `INT_MAX` it should be fine now.

test/test_ugrid.py
```diff
@@ -82,8 +82,14 @@ def test_standardized_dtype_and_fill(self):
...
-        grids = [ux_grid1, ux_grid2, ux_grid3]
-        for grid in grids:
+        grids_with_fill = [ux_grid2]
+        for grid in grids_with_fill:
+            assert grid.Mesh2_face_nodes.dtype == INT_DTYPE
+            assert grid.Mesh2_face_nodes._FillValue == INT_FILL_VALUE
+            assert INT_FILL_VALUE in grid.Mesh2_face_nodes.values
+
+        grids_without_fill = [ux_grid1, ux_grid2]
+        for grid in grids_without_fill:
```
hongyuchen1030, line 92: Why `ux_grid2` is showing in both `grids_without_fill` and `grids_with_fill` and where is `ux_grid3`?

test/test_ugrid.py
```diff
@@ -93,9 +99,9 @@ def test_standardized_dtype_and_fill_dask(self):
 
         with dask chunking
         """
-        ug_filename = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30.ug"
+        ug_filename = current_path / "meshfiles" / "ugrid" / "outRLL1deg" / "outRLL1deg.ug"
```
hongyuchen1030, line 102: Can directly call `ux_grid3` here

test/test_ugrid.py
```diff
@@ -93,9 +99,9 @@ def test_standardized_dtype_and_fill_dask(self):
...
         xr_grid = xr.open_dataset(str(ug_filename), chunks={'nMesh2_node': 100})
```
hongyuchen1030, line 103: What will be generated if the input `ug_filename` doesn't have any value (an empty file)?

hongyuchen1030 (commented):
I think this fix looks good in general! Thank you for your work! Just left some little questions/comments that I am not clear with.

uxarray/helpers.py
philipc2: It actually is possible and likely (see #248), since we will use `nan` as a fill value for floating point arrays.

test/test_ugrid.py
philipc2: Good catch

test/test_ugrid.py
philipc2: It should raise an exception according to the Xarray API

hongyuchen1030: Then probably we can use a `try catch` here

philipc2: I'm not fully understanding what the concern here is though, at the load dataset line?

philipc2:
Documentation failing due to an issue upstream, can ignore for now (see https://github.com/pydata/pydata-sphinx-theme/issues/1274)

anissa111:
> Documentation failing due to an issue upstream, can ignore for now (see [pydata/pydata-sphinx-theme#1274](https://github.com/pydata/pydata-sphinx-theme/issues/1274))

It looks like the sphinx-book-theme made a release last night ([v1.0.1](https://github.com/executablebooks/sphinx-book-theme/releases/tag/v1.0.1)). I'd try re-running the docs without the pin now

anissa111 (approved):

test/test_ugrid.py
hongyuchen1030: What I meant is inside our `standardized_dtype_and_fill_dask():`, did we handle the case when the input dataset is empty?

philipc2: I think this is beyond the scope of this PR. If we load in an empty, or incomplete dataset, it should be caught by `parse_grid` within `grid.py`

hongyuchen1030 (approved):

test/test_ugrid.py
hongyuchen1030: So if we can make sure `parse_grid` will prevent empty or incomplete dataset, then we're fine here.

Issue #193, opened 2022-12-23
Face Geometry Variable for Mixed-Topology Grids 


philipc2:
#189 discusses multiple approaches to handling the`_FillValue`, which is used to to indicate faces with fewer nodes than MaxNumNodesPerFace for a mixed-topology grid [(UGRID Conventions)](https://ugrid-conventions.github.io/ugrid-conventions/#2d-flexible-mesh-mixed-triangles-quadrilaterals-etc-topology:~:text=The%20use%20of%20_FillValue%20to%20indicate%20faces%20with%20fewer%20nodes%20than%20MaxNumNodesPerFace%20extends%20to%20the%20coordinate%20bounds%20variables%3B%20this%20is%20an%20extension%20of%20the%20current%20convention.)

From what I've read, there doesn't appear to be a variable in the UGRID conventions that specifies the geometry of each face. However, knowing what geometry is constructed from each face is important for proper indexing and for creating geometries for visualization. See the figure below: 

![gitt](https://user-images.githubusercontent.com/67855069/209246516-e6c023bb-bc37-4df8-8e7d-e8fd787ac925.png)

In this example, we have a mixed topology with MaxNumNodesPerFace = 6. Proposed in the introduction of a new variable i.e. `Mesh2_face_geometry` which would contain the geometry of each face based on the location of `_FillValue`. Alternately, the value representing each face would act as a maximum index for slicing, used to only return the values representing our geometry (no `_FillValue` is returned)

kjnam:
Hi, Philip. We use a hybrid grid, and this would be a nice (and a necessary) feature for us. Adding a new uxarray variable seems OK if it makes the code efficient.

PR #263, opened 2023-04-03
Add `Grid._build_edge_node_connectivity` & `helpers.close_face_nodes`

### LINKED ISSUE #266: Add `Grid._build_edge_node_connectivity`
Add a function for constructing the connectivity variable `Mesh2_edge_nodes` as described in the UGRID conventions for a Flexible 2D Mesh Topology

philipc2:
Closes #266 and splits #191

## Overview
* Function for constructing `Mesh2_edge_nodes` as described in the UGRID conventions for a Flexible 2D Mesh Topology 
* Introduces the `close_face_nodes` helper function
* Infrastructure in Grid constructor for calling build connectivity functions
* New test class for connectivity testing


uxarray/grid.py
```diff
@@ -380,6 +380,48 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+
+        # all edge nodes that contain a fill value
+        fill_value_mask = np.logical_or(edge_nodes[:, 0] == INT_FILL_VALUE,
+                                        edge_nodes[:, 1] == INT_FILL_VALUE)
+
+        # all edge nodes that do not contain a fill value
+        non_fill_value_mask = np.logical_not(fill_value_mask)
+
+        # filter out all invalid edges
+        edge_nodes = edge_nodes[non_fill_value_mask]
```
philipc2, line 408: @hongyuchen1030

This is the only section that we still need to finalize. The question that remains is whether we want to filter out unique edges from here. Since entries in `Mesh2_edge_nodes` don't have any spatial relationship to each other, as long as we properly index into it using `Mesh2_face_edges`, it should not be an issue. What do you think?

uxarray/grid.py
```diff
@@ -380,6 +380,48 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+            dims=["nMesh2_edge", "Two"],
+            attrs={
+                "cf_role":
+                    "edge_node_connectivity",
+                "long_name":
+                    "Maps every edge to the two nodes that it connects",
+                "start_index":
+                    INT_DTYPE(0),
+                "original_fill_value_mask":
+                    fill_value_mask,
```
philipc2, line 422: To my above point, we can use this mask to derive `Mesh2_face_edges`, similar to what is discussed in #193

uxarray/grid.py
hongyuchen1030: I am not very clear why we don't need to use unique edges here, Can you elaborate it a bit with some usage examples? Thanks

philipc2: Actually scrap my idea, I did a bit more digging into how `np.unique()` works, and we should be able to use the inverse indices to construct our `face_edges`. Will be pushing a few changes here shortly

hongyuchen1030: I think that's exactly what I did in the `build_face_edges_connectivity`, and also why I think building `build_face_edges_connectivity` and `build_edge_nodes_connectivity` together will be more efficient. But probably from the users' standpoints, having another `build_edge_nodes_connectivity` will be more intuitive.

philipc2: I prefer having them separate, since it's easier to debug and test. 

I also made the slight change to make these functions protected, since they will be called within the Grid object, not by the user.

hongyuchen1030: Works for me

uxarray/grid.py
```diff
@@ -380,6 +384,65 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
+                    "edge_node_connectivity",
+                "long_name":
+                    "Maps every edge to the two nodes that it connects",
+                "start_index":
+                    INT_DTYPE(0),
+                "inverse_indices":
+                    inverse_indices,
+                "fill_value_mask":
+                    fill_value_mask
+            })
```
philipc2, line 444: The addition of this DataArray in our `Grid.ds` is breaking the `TestIntegrate.test_calculate_total_face_area_triangle` cases. It causes the the `face_area` to be improperly calculated.

@rajeeja Do you have any idea why this is breaking? I attempted to do some digging, but wasn't able to find anything.  All other test cases pass locally.

rajeeja: strange, how different are the values?

philipc2: The same occurs when calculating `area_triangular`

test/test_grid.py
```diff
@@ -299,3 +293,27 @@ def test_populate_lonlat_coord(self):
...
+
+        # arrange edge nodes in the same manner as Grid._build_edge_node_connectivity
+        edge_nodes_expected.sort(axis=1)
+        edge_nodes_expected = np.unique(edge_nodes_expected, axis=0)
+
+        # construct edge nodes
+        mpas_grid_ux._build_edge_node_connectivity()
+        edge_nodes_output = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
+
+        assert np.array_equal(edge_nodes_expected, edge_nodes_output)
```
hongyuchen1030, line 319: Try using the Euler formula (n_face = n_edges - n_nodes + 2) to make sure we get the correct edge numbers, especially for grid file that has fill_values (different face geometry)

hongyuchen1030 (dismissed):
Thank you for your work! This looks good to me

uxarray/grid.py
rajeeja: Has your [commit](https://github.com/UXARRAY/uxarray/pull/263/commits/96d66d4c65604880b437d4a370b969ad502ec0ac) fixed the issue? no failures reported now.

uxarray/grid.py
```diff
@@ -87,6 +87,11 @@ def __init__(self, dataset, **kwargs):
         # initialize convenience attributes
         self.__init_grid_var_attrs__()
 
+        # construct connectivity
+        if self.source_grid != "From vertices":
+            if "Mesh2_edge_nodes" not in self.ds:
+                self._build_edge_node_connectivity()
```
philipc2, line 93: @rajeeja 

This here acts acts as a workaround for now.

test/test_grid.py
```diff
@@ -299,3 +293,34 @@ def test_populate_lonlat_coord(self):
...
                                    lat_deg[i],
                                    decimal=12)
+
+
+class TestConnectivity(TestCase):
+    ugrid_filepath = current_path / "meshfiles" / "ugrid" / "outRLL1deg" / "outRLL1deg.ug"
+    mpas_filepath = current_path / "meshfiles" / "mpas" / "QU" / "mesh.QU.1920km.151026.nc"
+    exodus_filepath = current_path / "meshfiles" / "exodus" / "outCSne8" / "outCSne8.g"
+
+    def test_build_edge_nodes(self):
```
hongyuchen1030, line 303: @philipc2 
Can you add more test cases here to make sure it works?
Especially the test cases that include different face geometry (The input Grid that contains the fill_values)

You can refer to my initial commit for this PR, which contains 3 input files as test cases for `test_build_face_edges_connectivity`

hongyuchen1030 (commented):
After more thinking, I think we need to include the test cases for the faces with filled values here

test/test_grid.py
philipc2: Added a separate test case for Euler's formula using the datasets you mentioned.

All seems well, except one of the datasets that I tested `ugrid/geoflow-small/grid.nc` is not adhering to euler's formula. Will investigate this, but hoping it's not an issue with our implementation.

philipc2: Additionally, `exodus/mixed/mixed.nc` is also failing.

erogluorhan: All tests seem to be passing now; so did you fix them?

erogluorhan (commented):
Code looks good overall! There are a few inline questions; will approve when they are answered.

test/test_grid.py
hongyuchen1030: > 
Since this implementation is the same as my old ones, the implementation itself should be all right. I have tested on mixed geometry faces and it still adheres the Euler's formula. 

There might be little bugs inside the codes that cause the inconsistency. Have you checked that all mixed faces are correctly patched and identified?

hongyuchen1030: > 
Since this implementation is the same as my old ones, the implementation itself should be all right. I have tested on mixed geometry faces and it still adheres the Euler's formula. 

There might be little bugs inside the codes that cause the inconsistency. Have you checked that all mixed faces are correctly patched and identified?

philipc2: @erogluorhan 

I've been locally testing the two failing datasets. I'm trying to figure out why those two specific datasets don't adhere to Euler's formula, especially since the geoflow dataset does not contain any fill values.

hongyuchen1030: @philipc2 

I digged into the codes a little bit for `geoflow-small/grid.nc`, 

I noticed that some of the edges are used only once in the face, indicating that there might be holes in the grid.

The Euler Formular in the form of `V-E+F=2.` is only valid for convex polyhedron [Euler](https://en.wikipedia.org/wiki/Euler_characteristic)

In our cases, it means that the Grids should cover the entire sphere (A.K.A. every edge will be used twice, A.K.A. there's no hole on the sphere)

Can you check if this `geoflow-small/grid.nc` has a hole or not? If there is, we don't need to worry about here, if there's not, we will dig more into what happened. Thanks!

philipc2: > @philipc2
> 
> I digged into the codes a little bit for `geoflow-small/grid.nc`,
> 
> I noticed that some of the edges are used only once in the face, indicating that there might be holes in the grid.
> 
> The Euler Formular in the form of `V-E+F=2.` is only valid for convex polyhedron ([Euler Formula](https://en.wikipedia.org/wiki/Euler_characteristic))
> 
> In our cases, it means that the Grids should cover the entire sphere (A.K.A. every edge will be used twice, A.K.A. there's no hole on the sphere)
> 
> Can you check if this `geoflow-small/grid.nc` has a hole or not? If there is, we don't need to worry about here, if there's not, we will dig more into what happened. Thanks!

This is a great catch, let me look into it.

philipc2: It doesn't seem like there is any holes in `geoflow-small/grid.n`, however this is the dataset I used last summer, and it does have multiple faces that cross the antemeridian. It uses positive 180 lon regardless of whether the face is on the left hand side (< 0 lon) or right hand side (>0 lon). Could this be part of the issue?

hongyuchen1030: > It doesn't seem like there is any holes in `geoflow-small/grid.n`, however this is the dataset I used last summer, and it does have multiple faces that cross the antemeridian. It uses positive 180 lon regardless of whether the face is on the left hand side (< 0 lon) or right hand side (>0 lon). Could this be part of the issue?

What I found for this dataset is that some unique edges are only used once, which should not happen if this dataset doesn't have any holes inside.

According to your explanation, it might be caused by the inconsistent longitude representation. Can you elaborate what does " It uses positive 180 lon regardless of whether the face is on the left hand side (< 0 lon) or right hand side (>0 lon)." mean here?

All my algorithms will base on the fact that all vertexes are represented as `[lon_degree, lat_degree]`, where `0<= lon_degree<=360`(I should take care of the 360 degree cases in my latlonbound function), and `-90<= lat_degree<=90`

philipc2: > What I found for this dataset is that some unique edges are only used once, which should not happen if this dataset doesn't have any holes inside. 

From my testing, it appears that each edge is used at least twice. Below is what I used to test it:
```Python
        grid_xr = xr.open_dataset(self.ugrid_filepath_04)
        grid_ux = ux.Grid(grid_xr)
        edge_nodes = grid_ux.Mesh2_edge_nodes.values.ravel()
        _, counts = np.unique(edge_nodes, return_counts=True)
```

> Can you elaborate what does " It uses positive 180 lon regardless of whether the face is on the left hand side (< 0 lon) or right hand side (>0 lon)." mean here?

Assuming our longitude values are [-180, 180] and the antimeridian is defined at +- 180, I'm saying that if a face is on the left hand plane (longitude < 0) and contains an edge that touches the antemeridian, it uses **positive** 180 instead of **negative** 180.

hongyuchen1030: I used a python dict to count the occurrence of each unique edge inside the `Grid._build_edge_node_connectivity` after line 427

```python
        # unique edge nodes
        edge_nodes_unique, inverse_indices = np.unique(edge_nodes,
                                                       return_inverse=True,
                                                       axis=0)

        # Use python dict to count the occurance of each face
        counts = {}
        for num in inverse_indices:
            if num in counts:
                counts[num] += 1
            else:
                counts[num] = 1
```
And here's what inside the `count`

```python
{0: 1, 3: 2, 9: 2, 1: 1, 2: 1, 5: 2, 11: 2, 4: 1, 7: 2, 13: 2, 6: 1, 8: 1, 15: 2, 12: 2, 18: 2, 10: 1, 14: 2, 20: 2, 16: 2, 22: 2, 17: 1, 24: 2, 21: 2, 27: 2, 19: 1, 23: 2, 29: 2, 25: 2, 31: 2, 26: 1, 33: 2, 30: 2, 36: 1, 28: 1, 32: 2, 37: 1, 34: 2, 38: 1, 35: 1, 39: 1, 40: 1, 43: 2, 49: 2, 41: 1, 42: 1, 45: 2, 51: 2, 44: 1, 47: 2, 53: 2, 46: 1, 48: 1, 55: 2, 52: 2, 58: 2, 50: 1, 54: 2, 60: 2, 56: 2, 62: 2, 57: 1, 64: 2, 61: 2, 67: 2, 59: 1, 63: 2, 69: 2, 65: 2, 71: 2, 66: 1, 73: 2, 70: 2, 76: 1, 68: 1, 72: 2, 77: 1, 74: 2, 78: 1, 75: 1, 79: 1, 80: 1, 83: 2, 89: 2, 81: 1, 82: 1, 85: 2, 91: 2, 84: 1, 87: 2, 93: 2, 86: 1, 88: 1, 95: 2, 92: 2, 98: 2, 90: 1, 94: 2, 100: 2, 96: 2, 102: 2...
```

hongyuchen1030: > > What I found for this dataset is that some unique edges are only used once, which should not happen if this dataset doesn't have any holes inside.
> 
> From my testing, it appears that each edge is used at least twice. Below is what I used to test it:
> 
> ```python
>         grid_xr = xr.open_dataset(self.ugrid_filepath_04)
>         grid_ux = ux.Grid(grid_xr)
>         edge_nodes = grid_ux.Mesh2_edge_nodes.values.ravel()
>         _, counts = np.unique(edge_nodes, return_counts=True)
> ```
> 
> > Can you elaborate what does " It uses positive 180 lon regardless of whether the face is on the left hand side (< 0 lon) or right hand side (>0 lon)." mean here?
> 
> Assuming our longitude values are [-180, 180] and the antimeridian is defined at +- 180, I'm saying that if a face is on the left hand plane (longitude < 0) and contains an edge that touches the antemeridian, it uses **positive** 180 instead of **negative** 180.

I tried your testing method. 

First, an edge should **NOT** shared by more than 2 faces.

Second, why do you want to use `ravel()` here? We want to count the occurrence of each edge, aka the occurrence of  each unique vertice pair

philipc2: @hongyuchen1030 

I will revisit this PR this weekend / early next week since it is a busy week for me.

philipc2: Thank you!

I haven't been able to find anything wrong with this dataset, but on an unrelated note it has also broken something with the redesign. I think it's fair to leave it out of testing until we can figure out more information about it.

Considering the other test cases, especially the one that compares it to the MPAS edge nodes, I think this should be ready to get merged

uxarray/grid.py
philipc2: It appears to be working now with the latest merges

hongyuchen1030 (approved):
Thanks, I think it's ready to merge

test/test_grid.py
hongyuchen1030: Works for me. Thanks

rajeeja (approved):

PR #268, opened 2023-04-06
Enable initializing `Grid` object from multiple face vertices

hongyuchen1030:
Add another private function `Grid.__from_multi_faces_vert__(self, dataset)` inside `Grid` that allow user to initialize the Grid object with multiple faces vertices. The skeleton codes are oriented from the `Grid.__from_vert__()`

And since our Grid initialization only supported the input vertices in the ndarray form, users also have to pass the faces vertices in a nested `ndarray` form, please see the corresponding test case for detailed usage examples.

As presented in the test case, users are allowed to form a one-face `Grid` using a two dimensions ndarray or 3 dimensions ndarray.


erogluorhan:
@rajeeja what was the motivation behind limiting `Grid.__from_vert__` to a single-face grid at first?

rajeeja:
To create a minimalistic test example from scratch. Specifying each point by hand and testing out the minimum requirements to create an unstructured grid.

uxarray/grid.py
```diff
@@ -141,6 +151,60 @@ def __init_grid_var_attrs__(self) -> None:
...
 
+    def __from_multi_faces_vert__(self, dataset):
+        """Create a grid with multiple faces with vertices specified by the
+        given argument. Called by :func:`__init__`.
+
+        Parameters
+        ----------
+        dataset : ndarray
+            ndarry of vertex coordinates that form multiple faces.
+        """
```
philipc2, line 162: If this implementation works works for an arbitrary number of faces (including just a single one), we should patch up `__from_vert__` instead of creating a separate function.

rajeeja:
> @rajeeja what was the motivation behind limiting `Grid.__from_vert__` to a single-face grid at first? 

It helps understanding the data structure and testing primitive design.

uxarray/grid.py
hongyuchen1030: This one will work for a single one (as illustrated in the `test__from_multi_faces_vert__`)

The reason why I created a separate function for multiple test cases is that I noticed the input for `Grid.__from_vers__()` is 

```python
np.array([[0, 0], [2, 0], [0, 2], [2, 2]])
```

Adding another function to handle the `multiple faces` will allow users to initialize a single face through

```python
np.array([[0, 0], [2, 0], [0, 2], [2, 2]])
```

or 

```python
np.array([
            np.array([[150, 10], [160, 20], [150, 30], [135, 30], [125, 20],
                      [135, 10]])])
```

But either patching and separating works for me, as long as we can initialized the`Grid` object through multiple face objects

uxarray/grid.py
```diff
@@ -141,6 +151,60 @@ def __init_grid_var_attrs__(self) -> None:
...
+            attrs={
+                "cf_role": "mesh_topology",
+                "long_name": "Topology data of unstructured mesh",
+                "topology_dimension": -1,
+                "node_coordinates": "Mesh2_node_x Mesh2_node_y Mesh2_node_z",
+                "node_dimension": "nMesh2_node",
+                "face_node_connectivity": "Mesh2_face_nodes",
+                "face_dimension": "nMesh2_face"
+            })
+        self.ds.Mesh2.attrs['topology_dimension'] = dataset.ndim
```
hongyuchen1030, line 174: @philipc2 
`dataset.ndim` will only work for `ndarray` here, I wonder if you can help me expand the type of `dataset` from `ndarray` to [xarray.Dataset](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html#xarray.Dataset), ndarray, [list](https://docs.python.org/3/library/stdtypes.html#list), [tuple](https://docs.python.org/3/library/stdtypes.html#tuple), as specified in our API here? Thanks!

hongyuchen1030:
Another thing to be noticed here is that:

Although in the API we stated that the input dataset can be ([xarray.Dataset](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html#xarray.Dataset), ndarray, [list](https://docs.python.org/3/library/stdtypes.html#list), [tuple](https://docs.python.org/3/library/stdtypes.html#tuple))

But when I tried to pass the face vertices into the `Grid.__init()__` it seems that only `np.ndarray` will work due to the difference naming for size in different data structures.

uxarray/grid.py
erogluorhan: I am favoring to expand the `__from_vert__ ` function's capability to handle both single-face and multi-face grid creation instead of separate functions; is that fine @rajeeja ?

rajeeja: Absolutely.

erogluorhan: @hongyuchen1030 Could you please make changes to get it done as an expansion to the existing `__from_vert__` as suggested? Thanks very much for the added functionality!

uxarray/grid.py
philipc2: See my fix below, should also address this

rajeeja (approved):
Looks good, we should later on add the edge nodes test to this also.

philipc2:
It should be fine for this PR, but we should considering move ` __from_vert__`, ` __from_ds__` and `encode_as` into their own modules, similar to all of our other I/O modules (`_scrip.py`, `_mpas.py`, etc).

uxarray/grid.py
hongyuchen1030: > @hongyuchen1030 Could you please make changes to get it done as an expansion to the existing `__from_vert__` as suggested? Thanks very much for the added functionality!

Working on it

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
...
+        """
+             7---------6
+            /|        /|
+           / |       / |
+          3---------2  |
+          |  |      |  |
+          |  4------|--5
+          | /       | /
+          |/        |/
+          0---------1
```
philipc2, line 71: love the diagram :)

hongyuchen1030:
@erogluorhan @philipc2 I have expand the `Grid.__from_vert__(dataset)` to incorporate multiple face vertices.

Things to be noticed are listed as following:

1. Since inside the new `Grid.__from_vert__(dataset)`, we called the `np.unique()` function to identify unique vertexes coordinates, the users shouldn't expect the `face_nodes` index will be in the same order when passing in the vertexes coordinates, and should always refers to the `mesh2_face_nodes` for the correct index for each nodes in `Mesh2_node_x/y/z` 

2. To maintain consistency, we support users to create a single-face Grid from either
`np.array([[0, 0], [2, 0], [0, 2], [2, 2]])` or `np.array([np.array([[150, 10], [160, 20], [150, 30], [135, 30], [125, 20],
                      [135, 10]])])`

uxarray/grid.py
```diff
@@ -158,35 +172,43 @@ def __from_vert__(self):
...
+        x_coord = dataset[:, :, 0].flatten()
+        y_coord = dataset[:, :, 1].flatten()
+        if dataset[0][0].size > 2:
+            z_coord = dataset[:, :, 2].flatten()
+
+        # Identify unique vertices and their indices
+        unique_verts, indices = np.unique(dataset.reshape(
+            -1, dataset.shape[-1]),
+                                          axis=0,
+                                          return_inverse=True)
```
philipc2, line 192: Fix the formatting/indexing here, maybe consider the reshape right above the function call?

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
         grid.encode_as("exodus")
 
     def test_init_verts(self):
-        """Create a uxarray grid from vertices and saves a ugrid file.
+        """Create a uxarray grid from multiple face' vertices and saves a ugrid
+        file.
```
philipc2, line 56:
```suggestion
        """Create a uxarray grid from multiple face vertices and saves a ugrid
        file.
```

uxarray/grid.py
```diff
@@ -158,35 +172,43 @@ def __from_vert__(self):
...
+        num_nodes = unique_verts.shape[0]
+
+        # Create coordinate DataArrays
+        self.ds["Mesh2_node_x"] = xr.DataArray(data=xr.DataArray(
+            unique_verts[:, 0]),
```
philipc2, line 197: This comment applies to multiple lines below also.

We can keep the data as a numpy array, I don't believe there is a need to explicitly set it to a `xr.DataArray`

uxarray/grid.py
```diff
@@ -158,35 +172,43 @@ def __from_vert__(self):
...
-        if self.vertices[0].size > 2:
-            self.ds["Mesh2_node_z"] = xr.DataArray(data=xr.DataArray(z_coord),
+        if dataset.shape[-1] > 2:
+            self.ds["Mesh2_node_z"] = xr.DataArray(data=xr.DataArray(
+                unique_verts[:, 2]),
                                                    dims=["nMesh2_node"],
                                                    attrs={"units": z_units})
 
+        # Create connectivity array using indices of unique vertices
+        connectivity = indices.reshape(dataset.shape[:-1]).tolist()
```
philipc2, line 211: Is it necessary to convert to a list if we are converting straight back to an array in the next line?

uxarray/grid.py
```diff
@@ -141,11 +152,14 @@ def __init_grid_var_attrs__(self) -> None:
...
-        argument.
+    def __from_vert__(self, dataset):
+        """Create a grid with multiple faces with vertices specified by the
+        given argument. Called by :func:`__init__`.
 
-        Called by :func:`__init__`.
+        Parameters
+        ----------
+        dataset : ndarray, list, tuple, required
+            Input xarray.Dataset or vertex coordinates that form one face.
```
philipc2, line 162: I don't think it can ever be an `xarray.Dataset` because of the following control flow in the `Grid`

```Python
if isinstance(dataset, (list, tuple, np.ndarray)):

File: uxarray/grid.py [hunk 9]
```diff
@@ -73,9 +73,20 @@ def __init__(self, dataset, **kwargs):
...
+                self.source_grid = "From vertices"
+            elif dataset.ndim == 2:
+                dataset = np.array([dataset])
+                self.__from_vert__(dataset)
+                self.source_grid = "From vertices"
+            else:
+                raise RuntimeError(
+                    "Input vertices don't have correct dimensions (Expected dimension will be 3: [nMesh2_face, "
+                    "nMesh2_node, Two/Three]). Or dimesion of 2 when only one face is passed"
+                )
```
philipc2, line 89: Any way that you could split up this exception to catch the "Or dimesion of 2 when only one face is passed" more directly?

uxarray/grid.py
```diff
@@ -73,9 +73,20 @@ def __init__(self, dataset, **kwargs):
 
         # check if initializing from verts:
         if isinstance(dataset, (list, tuple, np.ndarray)):
-            self.vertices = dataset
-            self.__from_vert__()
-            self.source_grid = "From vertices"
+            dataset = np.asarray(dataset)
+            if dataset.ndim == 3:
+                # This Grid contains multiple faces
```
philipc2, line 78:
```suggestion
            # grid with multiple faces
            if dataset.ndim == 3:
```

uxarray/grid.py
```diff
@@ -73,9 +73,20 @@ def __init__(self, dataset, **kwargs):
...
+                self.__from_vert__(dataset)
+                self.source_grid = "From vertices"
+            elif dataset.ndim == 2:
```
philipc2, line 81:
```suggestion
            # grid with a single face
            elif dataset.ndim == 2:
```

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
...
+            /|        /|
+           / |       / |
+          3---------2  |
+          |  |      |  |
+          |  4------|--5
+          | /       | /
+          |/        |/
+          0---------1
+
+        """
```
philipc2, line 73: Include this in the docstring under an "Example" heading

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
...
+        # Test for ndarray
```
philipc2, line 74: Can you be a little more specific in the comment?

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
...
+        vgrid.encode_as("ugrid")
+
+        vgrid = ux.Grid(faces_verts_ndarray,
+                        vertices=True,
+                        islatlon=True,
+                        concave=False)
+        assert (vgrid.source_grid == "From vertices")
+        assert (vgrid.nMesh2_face == 3)
+        assert (vgrid.nMesh2_node == 14)
+        vgrid.encode_as("ugrid")
```
philipc2, line 138: Add some comments to this block

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
         grid.encode_as("exodus")
 
     def test_init_verts(self):
```
philipc2, line 54: This is a very comprehensive test case, great work!

Can you create another case to see what happens if the user passes through an array/list/tuple with invalid dimensions, especially if only a small part of the input is invalid, such as the following

```Python
faces_verts_tuples = [
            ((150, 10, 12345678), (-1,-1), (150, 30), (135, 30), (125, 20), (135, 10)),
            ((125, 20), (135, 30), (125, 60), (110, 60), (100, 30), (105, 20)),
            ((95, 10), (105, 20), (100, 30), (85, 30), (75, 20, 1234, 4321), (85, 10))
        ]

File: test/test_exodus.py [hunk 16]
```diff
@@ -23,7 +23,7 @@ def test_read_exodus(self):
     def test_init_verts(self):
         """Create a uxarray grid from vertices and saves a 1 face exodus
         file."""
-        verts = np.array([[0, 0], [2, 0], [0, 2], [2, 2]])
+        verts = [[[0, 0], [2, 0], [0, 2], [2, 2]]]
```
philipc2, line 26: Any particular reason for changing this?

test/test_grid.py
```diff
@@ -52,16 +52,147 @@ def test_open_non_mesh2_write_exodus(self):
...
+        ]
+
+        faces_coords = []
+        for face in face_vertices:
+            face_coords = []
+            for vertex_index in face:
+                x, y, z = cart_x[vertex_index], cart_y[vertex_index], cart_z[
+                    vertex_index]
+                face_coords.append([x, y, z])
+            faces_coords.append(face_coords)
```
philipc2, line 116: Add a comment to this block

philipc2 (changes requested):
Great work putting this together so quickly! A few comments:

test/test_exodus.py
hongyuchen1030: Ideally, we still want the user to pass in the one face vertexes in this way `[[[0, 0], [2, 0], [0, 2], [2, 2]]]` instead of just the `meshface_nodes * coord` since that's should be the correct format.

I supported the `meshface_nodes * coord` input just because the origial `__from_vert__()` did that, and forbiding this input type will require more underlying changes in other functions if such functions calls `__from_vert__()`

philipc2: Makes sense!

uxarray/grid.py
hongyuchen1030: These implementations were from the original `Grid.__from_vert__()`

uxarray/grid.py
hongyuchen1030: fixed

uxarray/grid.py
hongyuchen1030: "Or dimesion of 2 when only one face is passed" is not an exception, it's what's allowed

Basically this exception error is just telling users we can only accept the dimension as 2 or 3. And the `ndim==2` case only applies to when a single face is passed

But feel free to rephrase this sentence to make the message more clear

test/test_grid.py
hongyuchen1030: The initial `Grid.__from_vert()__` doesn't do any values filtering, so as this expanded `Grid.__from_vert()__`. The `(-1,-1)` will cause the `        unique_verts, indices = np.unique(dataset.reshape(-1, dataset.shape[-1]),axis=0,return_inverse=True)` to throw an error.

I also noticed that the original `Grid.__from_vert__()` couldn't handle the input vertices with `Fill_values` as well.

I think for this PR, we can only focus on the multiple-face cases, and create a new PR for the data screening

uxarray/grid.py
philipc2: I would still consider removing them, unless @rajeeja had a specific reason for doing this.

test/test_grid.py
philipc2: That's fine with me

uxarray/grid.py
```diff
@@ -211,7 +214,8 @@ def __from_vert__(self, dataset):
             # Update indices accordingly
             for i, idx in enumerate(false_indices):
                 # TODO: setting the INT_FILL_VALUE as np.iinfo(INT_DTYPE).max will cause "OverflowError: Python int too large to convert to C long" here
-                indices[indices == idx] = -1 # TODO: Replace -1 with  INT_FILL_VALUE later
+                indices[indices ==
```
hongyuchen1030, line 217: @philipc2 
If I tried 
```python
indices[indices == idx] = INT_FILL_VALUE
```

It will throw
```python
Victoria Chan
            # Update indices accordingly
            for i, idx in enumerate(false_indices):
>               indices[indices == idx] = indices.dtype.type(INT_FILL_VALUE)
E               OverflowError: Python int too large to convert to C long
```

So I just use `-1` as the fill_value for now. Can you help me fix this so that we can use `INT_FILL_VALUE` here? Thanks

hongyuchen1030: Currently, I temperately solved it by converting `-1` back to `INT_FILL_VALUE` later
in the connectivity array

```python
        connectivity = indices.reshape(dataset.shape[:-1])
        connectivity = np.where(connectivity == -1, INT_FILL_VALUE,
                                connectivity)
```

But I wonder if there's a better solution for it

uxarray/grid.py
hongyuchen1030: I should already remove it in my following commits

uxarray/grid.py
philipc2: You need to ensure that your arrays are all of type `INT_DTYPE`, in this case `indicies`.

uxarray/grid.py
philipc2: looks good, thanks!

uxarray/grid.py
```diff
@@ -158,41 +173,74 @@ def __from_vert__(self):
...
+            fill_value_mask = np.logical_or(
+                unique_verts[:, 0] == INT_FILL_VALUE,
+                unique_verts[:, 1] == INT_FILL_VALUE,
+                unique_verts[:, 2] == INT_FILL_VALUE)
+
+        # Get the indices of all the False values in fill_value_mask
+        false_indices = np.where(fill_value_mask == True)[0]
+
+        # Check if any False values were found
+        indices = indices.astype(np.int64)
```
philipc2, line 208:
```suggestion
        indices = indices.astype(INT_DTYPE)
```

hongyuchen1030:
@erogluorhan @philipc2 



> @erogluorhan @philipc2 I have expand the `Grid.__from_vert__(dataset)` to incorporate multiple face vertices.
> 
> Things to be noticed are listed as following:
> 
> 1. Since inside the new `Grid.__from_vert__(dataset)`, we called the `np.unique()` function to identify unique vertexes coordinates, the users shouldn't expect the `face_nodes` index will be in the same order when passing in the vertexes coordinates, and should always refers to the `mesh2_face_nodes` for the correct index for each nodes in `Mesh2_node_x/y/z`
> 2. To maintain consistency, we support users to create a single-face Grid from either
>    `np.array([[0, 0], [2, 0], [0, 2], [2, 2]])` (`face_nodes * nodes_coor` dimensions) or `np.array([np.array([[150, 10], [160, 20], [150, 30], [135, 30], [125, 20], [135, 10]])])` (`mesh_face * face_nodes * nodes_coor` dimensions)

By now this PR should be ready to merge now.

The latest commits will allow users to use `FIll_VALUE` to pass in different face geometries like below

```python
    def test_init_verts_fill_values(self):
        faces_verts_filled_values = [[[150, 10], [160, 20], [150, 30],
                                      [135, 30], [125, 20], [135, 10]],
                                     [[125, 20], [135, 30], [125, 60],
                                      [110, 60], [100, 30],
                                      [ux.INT_FILL_VALUE, ux.INT_FILL_VALUE]],
                                     [[95, 10], [105, 20], [100, 30], [85, 30],
                                      [ux.INT_FILL_VALUE, ux.INT_FILL_VALUE],
                                      [ux.INT_FILL_VALUE, ux.INT_FILL_VALUE]]]
        vgrid = ux.Grid(faces_verts_filled_values,
                        vertices=True,
                        islatlon=False,
                        concave=False)
        assert (vgrid.source_grid == "From vertices")
        assert (vgrid.nMesh2_face == 3)
        assert (vgrid.nMesh2_node == 12)
```
The current `Grid.__from_vert__()` should handle most of the comon usage for now
For any further input data screening like [@philipc2 mentioned before](https://github.com/UXARRAY/uxarray/pull/268#discussion_r1160315213) we can open a new PR for it.

Please take a look into this PR and see if it can be merged now. Thanks!

uxarray/grid.py
```diff
@@ -141,11 +153,14 @@ def __init_grid_var_attrs__(self) -> None:
...
-        argument.
+    def __from_vert__(self, dataset):
+        """Create a grid with multiple faces with vertices specified by the
+        given argument. Called by :func:`__init__`.
 
-        Called by :func:`__init__`.
+        Parameters
+        ----------
+        dataset : ndarray, list, tuple, required
+            To be Input vertex coordinates that form one face.
```
philipc2, line 163:
```suggestion
        dataset : ndarray, list, tuple, required
             Input vertex coordinates that form our face(s)
```

uxarray/grid.py
```diff
@@ -158,41 +173,70 @@ def __from_vert__(self):
...
+        # Create connectivity array using indices of unique vertices
+        connectivity = indices.reshape(dataset.shape[:-1])
         self.ds["Mesh2_face_nodes"] = xr.DataArray(
             data=xr.DataArray(connectivity).astype(INT_DTYPE),
             dims=["nMesh2_face", "nMaxMesh2_face_nodes"],
             attrs={
                 "cf_role": "face_node_connectivity",
-                "_FillValue": INT_FILL_VALUE,
+                "_FillValue":
+                    INT_FILL_VALUE,  # TODO: Replace -1 with  INT_FILL_VALUE later
```
philipc2, line 239:
```suggestion
                "_FillValue":
                    INT_FILL_VALUE, 
```

uxarray/grid.py
```diff
@@ -141,11 +153,14 @@ def __init_grid_var_attrs__(self) -> None:
                 if value in self.ds.dims:
                     setattr(self, key, len(self.ds[value]))
 
-    def __from_vert__(self):
-        """Create a grid with one face with vertices specified by the given
-        argument.
+    def __from_vert__(self, dataset):
+        """Create a grid with multiple faces with vertices specified by the
+        given argument. Called by :func:`__init__`.
```
philipc2, line 158:
```suggestion
        """Create a grid with faces constructed from vertices specified by the
        given argument. 
```

uxarray/grid.py
```diff
@@ -73,9 +73,21 @@ def __init__(self, dataset, **kwargs):
...
+                self.source_grid = "From vertices"
+            # grid with a single face
+            elif dataset.ndim == 2:
+                dataset = np.array([dataset])
+                self.__from_vert__(dataset)
+                self.source_grid = "From vertices"
+            else:
+                raise RuntimeError(
+                    "Input vertices don't have correct dimensions (Expected dimension will be 3: [nMesh2_face, "
+                    "nMesh2_node, Two/Three]). Or dimesion of 2 when only one face is passed"
```
philipc2, line 89: Make sure to run pre-commit on this, since inline code changes through Github can sometimes break with formatting.

```suggestion
                raise RuntimeError(
                    f"Invalid Input Dimension: {dataset.ndim}. Expected dimension should be "
                    f"3: [nMesh2_face, nMesh2_node, Two/Three] or 2 when only one "
                    f"one face is passed in."
```

philipc2 (changes requested):
Looks great! Only a couple comment formatting changes left

uxarray/grid.py
```diff
@@ -73,9 +73,21 @@ def __init__(self, dataset, **kwargs):
...
+            # grid with a single face
+            elif dataset.ndim == 2:
+                dataset = np.array([dataset])
+                self.__from_vert__(dataset)
+                self.source_grid = "From vertices"
+            else:
+                raise RuntimeError(
+                    f"Invalid Input Dimension: {dataset.ndim}. Expected dimension should be "
+                    f"3: [nMesh2_face, nMesh2_node, Two/Three] or 2 when only one "
+                    f"one face is passed in.")
```
philipc2, line 90:
```suggestion
                    f"3: [nMesh2_face, nMesh2_node, Two/Three] or 2 when only "
                    f"one face is passed in.")
```

philipc2 (approved):
Nice work!

rajeeja (approved):

PR #270, opened 2023-04-11
Bug fixed for `Grid.__from_vert__()`

hongyuchen1030:
A small but important bug fix for `Grid.__from_vert__()`
 The previous version doesn't have `& (indices != INT_FILL_VALUE)` which will decrease the `INT_FILL_VALUE` by 1, deprecating all following `INT_FILL_VALUE` detection
``` python
            # Update indices accordingly
            for i, idx in enumerate(false_indices):
                indices[indices == idx] = INT_FILL_VALUE
                indices[(indices > idx) ] -= 1
```

### DIFF
--- a/uxarray/grid.py
+++ b/uxarray/grid.py
@@ -214,7 +214,7 @@ class Grid:
             # Update indices accordingly
             for i, idx in enumerate(false_indices):
                 indices[indices == idx] = INT_FILL_VALUE
-                indices[indices > idx] -= 1
+                indices[(indices > idx) & (indices != INT_FILL_VALUE)] -= 1
 
         # Create coordinate DataArrays
         self.ds["Mesh2_node_x"] = xr.DataArray(data=unique_verts[:, 0],


rajeeja (commented):
Good find @hongyuchen1030 

Do you think we can add a test that would fail without this change?

philipc2 (commented):
I agree with @rajeeja. An additional test case would be beneficial

hongyuchen1030:
> Good find @hongyuchen1030
> 
> Do you think we can add a test that would fail without this change?

I am unable to think out of a single testcase without invoking other functions for this now.

This bug will show when you called other functions which involve the `INT_FILL_VALUE` check. 

I will greatly appreciate if you can help me implement one. Thanks

philipc2 (approved):
After giving it some more thought, adding a test for this isn't necessary. 

If you did however want to add a check for it, you could add something similar to what I have here:

https://github.com/UXARRAY/uxarray/blob/cb63de0c57609ec7b7b885d5361ffb3cbd04ebb0/test/test_ugrid.py#L86-L89

rajeeja (approved):

PR #275, opened 2023-04-15
HOTFIX: MPAS Attributes not stored correctly 

philipc2:


rajeeja (approved):
How did you catch this? Did a particular mesh file load failed?

_primal_to_ugrid -> calls set attrs -> looks for _is_periodic, this wasn't checked properly?

philipc2:
> How did you catch this? Did a particular mesh file load failed?
> 
> _primal_to_ugrid -> calls set attrs -> looks for _is_periodic, this wasn't checked properly?

While working on the MPAS usage example notebook, it became clear the attributes were not being stored correctly. I'll add in a test case to check for this.

Issue #282, opened 2023-04-18
FESOM grid connectivity stored as tranpose


philipc2:
The dimension of `Mesh2_face_nodes` is `(3 x nFace)` instead of `(nFace x 3)`

From the UGRID Conventions:

> The face_dimension attribute specifies which netcdf dimension is used to indicate the index of the face in the connectivity arrays. This is needed because some applications store the data with the fastest varying index first, and some with that index last. The default is to use the num_faces as fastest dimension; e.g. a (num_faces, 3) array for triangles, but some applications might use a (3, num_faces) order, in which case the face_dimension attribute is required to help the client code disambiguate. The edge_dimension attribute is similar for the edge connectivity arrays.

PR #283, opened 2023-04-18
Face Area Fix, Face Dimension Variable, & Data Type Fix

philipc2:
## Overview
* Switches to `np.intp` instead of `np.uintp` due to `numpy` indexing being written for `intp`
* Added test case to test for invalid indexing
* Introduces the `Mesh2_face_dimension` variable, used for storing the maximum number of non-fill-value vertices/nodes used for constructing a face
* Modified Face Area functions to use `Mesh2_face_dimension`
* Added test case for Unit Sphere Area from the ESDS meeting


test/test_mpas.py
```diff
@@ -122,3 +122,16 @@ def test_set_attrs(self):
         # check if all expected attributes are set
         for mpas_attr in expected_attrs:
             assert mpas_attr in uxgrid.ds.attrs
+
+    def test_sphere_face_area(self):
```
rajeeja, line 126: pre commit might not have liked this.

test/test_grid.py
```diff
@@ -359,14 +393,15 @@ def test_compute_face_areas_geoflow_small(self):
...
-        grid_2.compute_face_areas()
+    # removed test until fix to tranposed face nodes
+    # def test_compute_face_areas_fesom(self):
+    #     """Checks if the FESOM PI-Grid Output can generate a face areas
+    #     output."""
+    #
+    #     fesom_grid_small = current_path / "meshfiles" / "ugrid" / "fesom" / "fesom.mesh.diag.nc"
+    #     grid_2_ds = xr.open_dataset(fesom_grid_small)
+    #     grid_2 = ux.Grid(grid_2_ds)
+    #     grid_2.compute_face_areas()
```
rajeeja, line 404: #282 should be fixed later.

uxarray/grid.py
```diff
@@ -99,6 +99,9 @@ def __init__(self, dataset, **kwargs):
         # initialize convenience attributes
         self.__init_grid_var_attrs__()
 
+        # build face dimension
+        self._build_face_dimension()
```
rajeeja, line 103: An expensive process for large grids.
If the target problem solved does not need face_dimension this will cause unnecessary overhead. 

We can think about guarding it later. Please put a TODO comment about it.

philipc2:
One concern is that we are bending the interpretation of `face_dimension`. Below is how it is defined in the UGRID conventions:

> The face_dimension attribute specifies which netcdf dimension is used to indicate the index of the face in the connectivity arrays. This is needed because some applications store the data with the fastest varying index first, and some with that index last. The default is to use the num_faces as fastest dimension; e.g. a (num_faces, 3) array for triangles, but some applications might use a (3, num_faces) order, in which case the face_dimension attribute is required to help the client code disambiguate. The edge_dimension attribute is similar for the edge connectivity arrays.

However, @rajeeja and I believe that this name is appropriate for this use case. Any other suggestions for a different name, other than `face_geometry` or `face_dimension`?

I was considering `nMesh2_face_nodes`, since it would give us the number of face nodes per face, but the name is confusing, since it's not a singular "dimension"

rajeeja (approved):

test/test_grid.py
```diff
@@ -336,6 +350,26 @@ def test_calculate_total_face_area_file(self):
...
+        dual_face_area = dual_grid.calculate_total_face_area()
+
+        nt.assert_almost_equal(primal_face_area,
+                               constants.UNIT_SPHERE_AREA,
+                               decimal=3)
+
+        nt.assert_almost_equal(dual_face_area,
+                               constants.UNIT_SPHERE_AREA,
+                               decimal=3)
+
```
philipc2, line 372: @areanddee

The behavior is now expected for MPAS grids that lie on a unit sphere.

rajeeja:
@erogluorhan please review, this should fix the primal and dual mesh area calc.

erogluorhan (approved):
Thanks for addressing this really fast!

PR #292, opened 2023-04-22
Separate Files for Connectivity and Coordinate Grid Functions

philipc2:
## Overview
* The `ux.Grid` class has become cluttered with functions, many of which can be separate into individual files
* Introduced the `connectivity.py` and `coordinates.py` files to store corresponding functions
* Did not move anything out of `helpers.py`, but these could act as a spot for specific helpers


rajeeja:
It'd be good to say

> ## Overview
> * The `ux.Grid` class has become cluttered with functions, many of which can be separate into individual files
> * Introduced the `connectivity.py` and `coordinates.py` files to store corresponding functions
> * Did not move anything out of `helpers.py`, but these could act as a spot for specific helpers

This is a good start in breaking down grid.py, the functions in the newly introduced files are essentially helper functions for coords and connectivity.  With regards to uxarray as a whole coords and connectivity could mean more generally.  There could be more thought into splitting things into core and functionality. It'd be better if such changes are made after merging the redesign and for now we focus on getting the redesign into main.

rajeeja:
It'd be good to say

> ## Overview
> * The `ux.Grid` class has become cluttered with functions, many of which can be separate into individual files
> * Introduced the `connectivity.py` and `coordinates.py` files to store corresponding functions
> * Did not move anything out of `helpers.py`, but these could act as a spot for specific helpers

The functions in the newly introduced files are essentially helper functions for coords and connectivity.  With regards to uxarray as a whole coords and connectivity could mean representing

philipc2:
@rajeeja 

Fair enough, I'll migrate some of these changes and discussion into a PR going into the redesign branch.

Issue #290, opened 2023-04-22
MPAS Data Read-in has duplicate `Mesh2_edge_nodes` connectivity building


hongyuchen1030:
**Describe the bug**

So according to our logics, the `Grid._build_edge_node_connectivity` will build the `Mesh2_edge_nodes` and generate an `inverse_indices` for `Grid._build_face_edges_connectivity`.

And to avoid the duplicate call of  `Grid._build_edge_node_connectivity`, we have the following condition for `Grid.__init__()`:

```python
        # construct connectivity
        if "Mesh2_edge_nodes" not in self.ds:
            self._build_edge_node_connectivity()
```
However, since the mpas datafile will build a `Mesh2_edge_nodes` when reading in (`_mpas._dual_to_ugrid` and `_mpas._primal_to_ugrid` themselves will try to build the not  `Mesh2_edge_nodes` without generating the inverse_indices), it will bypass our `self._build_edge_node_connectivity()` and hence don't have the generated inverse_indices` for `Grid._build_face_edges_connectivity`.

**Solutions**
1. Force every grid object to call `self._build_edge_node_connectivity()` in the  `Grid.__init__()`
2. Or, we remove the `Mesh2_edge_nodes` attribute setup in `_mpas._dual_to_ugrid` and `_mpas._primal_to_ugrid`'
3. Change the anti-duplication condition to `        if "Mesh2_edge_nodes" not in self.ds or self.mesh_type == "mpas":
            self._build_edge_node_connectivity()`

@rajeeja @philipc2 @erogluorhan I wonder if you have any preferred solution for this? Thanks

philipc2:
If the grid already comes with edge nodes, we shouldn't need to re-compute them. Instead, we can add a check inside of `Grid._build_edge_node_connectivity` to see whether we already have edge nodes. If we already have them, we can skip this block of code:

https://github.com/UXARRAY/uxarray/blob/e6508fd63db0dfd0c288d3f844f1f2196a5dba0b/uxarray/grid.py#L465-L477

I believe everything below that block should be able to remain the same. We would remove the following check and add it inside of `Grid._build_edge_node_connectivity()`

https://github.com/UXARRAY/uxarray/blob/e6508fd63db0dfd0c288d3f844f1f2196a5dba0b/uxarray/grid.py#L104-L105

### Proposed Change 

```Python
    def _build_edge_node_connectivity(self):
        """Constructs the UGRID connectivity variable (``Mesh2_edge_nodes``)
[...]

PR #297, opened 2023-04-25
Add `Grid._build_face_edges_connectivity`

hongyuchen1030:
Implement the `grid.build face edges connectivity`, resolve the MPAS existed `grid.edge_nodes_connectivity` confilcts


uxarray/grid.py
```diff
@@ -528,6 +531,33 @@ def _build_edge_node_connectivity(self):
...
+        inverse_indices = inverse_indices.reshape(self.nMesh2_face,
+                                                  self.nMaxMesh2_face_nodes)
+        mesh2_face_edges = inverse_indices  # We only need to store the edge index
+
+        self.ds["Mesh2_face_edges"] = xr.DataArray(
+            data=mesh2_face_edges,
+            dims=["nMesh2_face", "nMaxMesh2_face_edges"],
+            attrs={
+                "cf_role": "face_edges_connectivity",
+                "start_index": 0
```
philipc2, line 555:
```suggestion
                "start_index": INT_DTYPE(0)
```

uxarray/grid.py
```diff
@@ -528,6 +531,33 @@ def _build_edge_node_connectivity(self):
...
+            self._build_edge_node_connectivity(repopulate=True)
+        inverse_indices = self.ds['Mesh2_edge_nodes'].inverse_indices
+        inverse_indices = inverse_indices.reshape(self.nMesh2_face,
+                                                  self.nMaxMesh2_face_nodes)
+        mesh2_face_edges = inverse_indices  # We only need to store the edge index
+
+        self.ds["Mesh2_face_edges"] = xr.DataArray(
+            data=mesh2_face_edges,
+            dims=["nMesh2_face", "nMaxMesh2_face_edges"],
+            attrs={
```
philipc2, line 553: Can you add the long name? 

> Maps every face to its edges

philipc2: Can you add the fill value as an attribute as well?

uxarray/grid.py
```diff
@@ -528,6 +531,33 @@ def _build_edge_node_connectivity(self):
...
+                "cf_role": "face_edges_connectivity",
+                "start_index": 0
+            })
+
+        # set standardized attributes
+        setattr(self, "Mesh2_face_edges", self.ds["Mesh2_face_edges"])
+
```
philipc2, line 560:
```suggestion
setattr(self, "nMaxMesh2_face_edges", mesh2_face_edges.shape[1])
```

uxarray/grid.py
```diff
@@ -507,7 +511,6 @@ def _build_edge_node_connectivity(self):
             if inverse_indices[i] != INT_FILL_VALUE:
                 inverse_indices[i] -= indexes[i]
 
-        # add Mesh2_edge_nodes to internal dataset
         self.ds['Mesh2_edge_nodes'] = xr.DataArray(
```
philipc2, line 514: If `Mesh2_edge_nodes` is already in our dataset, we probably have to rewrite this initialization right here, since a DataArray already exists for it

philipc2 (changes requested):
Nice work so far!

uxarray/grid.py
```diff
@@ -528,6 +531,33 @@ def _build_edge_node_connectivity(self):
         setattr(self, "Mesh2_edge_nodes", self.ds['Mesh2_edge_nodes'])
         setattr(self, "nMesh2_edge", edge_nodes_unique.shape[0])
 
+    def build_face_edges_connectivity(self):
```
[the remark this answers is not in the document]
hongyuchen1030: My thought is: it's good to make it a user API so that user can build the connectivity under their need. Especially if our MPAS will only provide the `edge_nodes_connectivity` but not the `face_edge_connectivity`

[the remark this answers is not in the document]
philipc2: See my comments below, we should be calling the `build_face_edge_connectivity` by default. Nothing is preventing the user from using this function on their own accord from the internal api, however it's not suggested / practical for them to do so.

By default, our `Grid` constructor should build all connectivity variables that we support. We can add an argument (something like `build_connectivity`) which can indicate whether the user wants to skip the construction of edge nodes, face edges, etc.

uxarray/grid.py
```diff
@@ -528,6 +531,33 @@ def _build_edge_node_connectivity(self):
...
+        """Build the `Grid.face_edges_connectivity`, a DataArray of indices
+        indicating edges that are neighboring each face.
+
+        The `Grid.face_edges_connectivity` will be an integer DataArray
+        of size (nMesh2_face, MaxNumNodesPerFace) (optional)
+        """
+        if "Mesh2_edge_nodes" not in self.ds.keys() or (
+                self.source_grid != "From vertices" and
+                self.mesh_type == "mpas"):
+            self._build_edge_node_connectivity(repopulate=True)
```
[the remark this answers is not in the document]
hongyuchen1030: > Before entering `self._build_edge_node_connectivity()`, we need to determine whether we need to repopulate the edge nodes, which is when `Mesh2_edge_nodes` is not in our dataset. The only time we need to repopulate is when we already have the edge nodes, but they don't contain the inverse indices. We could do something like this inside of the grid constructor:
> 
> ```python
> # determine if we need to repopulate our edge nodes
> if "Mesh2_edge_nodes" not in self.ds:
>     repopulate = True    
> else:
>     repopulate = False
>       
> # construct edge node connectivity 
> self._build_edge_node_connectivity(repopulate) 
>  
> # construct face edge connectivity
> self._build_face_edge_connectivity()
>   
> # build face dimension, possibly safeguard for large datasets 
> self._build_face_dimension() 
> ```

1. The `repopulate` flag, as indicated in its name, it's for the case where the `Grid.edge_nodes_connectivity` exists but the `inverse_indices` doesn't. So this change will be unnecessary. Imagine when we read-in the MPAS data:
It has the  "Mesh2_edge_nodes", so `repopulate = False`, making this entire block equivalent to the initial implementation
```python
if "Mesh2_edge_nodes" not in self.ds:
    self._build_edge_node_connectivity()
```

2. I put the check in the `Grid.build_face_nodes_connectivity` by checking if the source is MPAS. So that if the user doesn't need the `Grid.Build_face_edges_connectivity`, it won't be repopulated again

3. `Grid.Build_face_edges_connectivity` is just a optional API, indicating that not everyone needs the "Mesh2_face_edges", so we don't wanna waste time calling it in the initialization. We will only call it when needed

uxarray/grid.py
hongyuchen1030: If `Mesh2_edge_nodes` is already in our dataset but the `inverse_indices` doesn't exist, what we do is we will force-repopulate the `Grid.edge_nodes_connectivity`  meaning that we're going to rewrite things in our own way

uxarray/grid.py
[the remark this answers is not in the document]
hongyuchen1030: We had that discussion last summer, the conclusion is only called the `Grid.build_face_edge_connectivity` if it's needed. However, we can make it an internal API if that's more appropriate.

And I wonder why we want to ban users from calling this connectivity. I thought it was the functionality they want. And we already has the check inside the function to prevent duplicate call,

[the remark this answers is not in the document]
philipc2: I don't recall that discussion, but that was a while ago. Do you remember if it was discussed in an issue or in our meeting notes? 

I'm open to finding a good solution. As I described below, I think having an argument in the grid object to determine whether we want to build the connectivity could be useful. However I think we should give this design some more thought, since I also want to restructure and move these functions into a separate file that can then be called by the `Grid` class. Then, we can consider both an internal (modifies `Grid` dataset) and user (returns the desired connectivity) versions of the function.

If you're not opposed, let's still to keeping this as an internal for now until we set up the `connectivity.py` file

uxarray/grid.py
philipc2: It appears that when `repopulate` is true, it computes derives the `edge_nodes` using our approach, and when it is false, it only does the inverse indices. Is this the behavior that we want?

uxarray/grid.py
```diff
@@ -528,6 +540,34 @@ def _build_edge_node_connectivity(self):
...
+                    "face_edges_connectivity",
+                "start_index":
+                    INT_DTYPE(0),
+                "long_name":
+                    "Maps every edge to the two nodes that it connects",
+            })
+
+        # set standardized attributes
+        setattr(self, "nMaxMesh2_face_edges", mesh2_face_edges.shape[1])
+
```
philipc2, line 570: We also want to set an attribute for `Mesh2_face_edges`
```suggestion
setattr(self, "Mesh2_face_edges", self.ds["Mesh2_face_edges"])
```

uxarray/grid.py
```diff
@@ -528,6 +540,34 @@ def _build_edge_node_connectivity(self):
         setattr(self, "Mesh2_edge_nodes", self.ds['Mesh2_edge_nodes'])
         setattr(self, "nMesh2_edge", edge_nodes_unique.shape[0])
 
+    def _build_face_edges_connectivity(self):
+        """Constructs the UGRID connectivity variable (``Mesh2_face_edges``)
+        and stores it within the internal (``Grid.ds``) and through the
+        attribute (``Grid.Mesh2_face_edges``)."""
+        if "Mesh2_edge_nodes" not in self.ds.keys(
+        ) or (self.ds['Mesh2_edge_nodes'].attrs.get("inverse_indices") is None):
```
philipc2, line 548: We can tidy up this check by doing the following:
```suggestion
        if ("Mesh2_edge_nodes" not in self.ds or  
            "inverse_indices" not in self.ds['Mesh2_edge_nodes'].attrs):
```

test/test_grid.py
```diff
@@ -534,9 +509,225 @@ def test_edge_nodes_euler(self):
...
+                face_nodes_dict[face_idx] = after_swapped_remove
+
+            # Convert the dictionary to a list
+            res_face_nodes_connectivity = []
+            for face_idx in range(len(tgrid.ds["Mesh2_face_edges"].values)):
+                res_face_nodes_connectivity.append(face_nodes_dict[face_idx])
+                while len(res_face_nodes_connectivity[face_idx]
+                         ) < tgrid.ds["Mesh2_face_nodes"].values.shape[1]:
+                    res_face_nodes_connectivity[face_idx].append(
+                        ux.INT_FILL_VALUE)
```
philipc2, line 609: Can you give me brief rundown of the motivation for writing this? Also, there is a lot of duplicated code in the next case case. We should consider abstracting these code away into a test helper function that can be called.

uxarray/grid.py
```diff
@@ -528,6 +540,35 @@ def _build_edge_node_connectivity(self):
...
+        and stores it within the internal (``Grid.ds``) and through the
+        attribute (``Grid.Mesh2_face_edges``)."""
+        if ("Mesh2_edge_nodes" not in self.ds or
+                "inverse_indices" not in self.ds['Mesh2_edge_nodes'].attrs):
+            self._build_edge_node_connectivity(repopulate=True)
+
+        inverse_indices = self.ds['Mesh2_edge_nodes'].inverse_indices
+        inverse_indices = inverse_indices.reshape(self.nMesh2_face,
+                                                  self.nMaxMesh2_face_nodes)
+        mesh2_face_edges = inverse_indices  # We only need to store the edge index
```
philipc2, line 554: This line isn't necessary, we can simply pass through `inverse_indices` into the `data=` in our Data Array

test/test_grid.py
```diff
@@ -1,6 +1,7 @@
 import os
 import numpy as np
 import xarray as xr
+from scipy.sparse import csr_matrix
```
philipc2, line 4:
```suggestion
```

test/test_grid.py
```diff
@@ -534,9 +509,225 @@ def test_edge_nodes_euler(self):
...
-            n_node = grid_ux.nMesh2_face
+            n_face = grid_ux.nMesh2_face
+            n_node = grid_ux.nMesh2_node
             n_edge = grid_ux.nMesh2_edge
 
             # euler's formula (n_face = n_edges - n_nodes + 2)
             assert (n_face == n_edge - n_node + 2)
+
+    def test_build_face_edges_connectivity(self):
+        """Generates Grid.Mesh2_edge_nodes from Grid.Mesh2_face_nodes."""
```
philipc2, line 520: A more detailed docstring would be good considering the length of this test case

test/test_grid.py
```diff
@@ -534,9 +509,225 @@ def test_edge_nodes_euler(self):
...
+                         mesh2_face_nodes.sizes["nMaxMesh2_face_nodes"])
+
+        # Assert if the mesh2_edge_nodes sizes are correct.
+        # Euler formular for determining the edge numbers: n_face = n_edges - n_nodes + 2
+        num_edges = mesh2_face_edges.sizes["nMesh2_face"] + tgrid.ds[
+            "Mesh2_node_x"].sizes["nMesh2_node"] - 2
+        size = mesh2_edge_nodes.sizes["nMesh2_edge"]
+        self.assertEqual(mesh2_edge_nodes.sizes["nMesh2_edge"], num_edges)
+
+    def test_build_face_edges_connectivity_fillvalues(self):
```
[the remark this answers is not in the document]
philipc2: Same concerns from above apply:

> A more detailed docstring would be good considering the length of this test case

philipc2 (changes requested):

uxarray/grid.py
```diff
@@ -453,28 +451,41 @@ def integrate(self, var_ds, quadrature_rule="triangular", order=4):
...
 
-        # second index includes second node up to padded value
-        edge_nodes[:, 1] = padded_face_nodes[:, 1:].ravel()
+            # first index includes starting node up to non-padded value
+            edge_nodes[:, 0] = padded_face_nodes[:, :-1].ravel()
+
+            # second index includes second node up to padded value
+            edge_nodes[:, 1] = padded_face_nodes[:, 1:].ravel()
+        else:
+            # If "Mesh2_edge_nodes" already exists, directly return the function call
```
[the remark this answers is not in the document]
hongyuchen1030: The reason why I did this is because it actually did the exactly same thing as the previous implementation. See my comments below about why I deleted the redundant "mpas" test case

test/test_grid.py
```diff
@@ -497,32 +498,6 @@ class TestConnectivity(TestCase):
...
-
-        assert np.array_equal(edge_nodes_expected, edge_nodes_output)
-
-        # euler's formula (n_face = n_edges - n_nodes + 2)
-        n_face = mpas_grid_ux.nMesh2_node
-        n_node = mpas_grid_ux.nMesh2_face
-        n_edge = edge_nodes_output.shape[0]
-
-        assert (n_face == n_edge - n_node + 2)
-
```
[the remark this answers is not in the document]
hongyuchen1030: By examining the MPAS read-in process, it can be discovered that the `np.unique` function in `Grid._build_edge_nodes_connectivity` does not actually perform any unique operations. This is because MPAS has its own "edge_nodes_connectivity," which contains only unique edges. Therefore, the returned `edge_nodes_unique` and `inverse_indices` cannot be reshaped into the `face_edges_connectivity` 2D matrix. As a result, the `face_edges_connectivity` cannot be built unless the `edge_nodes_connectivity` and `inverse_indices` are rebuilt from the MPAS `face_nodes_connectivity`. The deleted testcase is essentially a redundant one, as it mirrors the exact same process as `Grid._build_edge_nodes_connectivity` and therefore does not provide any additional value.

Let's say the MPAS data has the `mesh2_face_nodes` as 162 face * 6 nodePer_face
1. MPAS has its own "edge_nodes_connectivity", which are all unique edges (480 unique edges)
2. So in the `Grid._build_edge_nodes_connectivity`, the `np.unique` does nothing
```python
        # unique edge nodes
        edge_nodes_unique, inverse_indices = np.unique(edge_nodes,
                                                       return_inverse=True,
                                                       axis=0)
```
The returned ` edge_nodes_unique` is still `480` and the `inverse_indices` is also 480 (The correct size should be 162*6=972), which means this `inverse_indices` cannot be reshaped into the `face_edges_connectivity` (A 162*6 2D matrix). 

As I mentioned in issue #290, there's no way we can build the `face_edges_connectivity` unless we force rebuilding the `edge_nodes_connectivity` and `inverse_indices` from the MPAS `face_nodes_connectivity`

3. So since the unique didn't do any actual work. So if we summarize the pathway when the MPAS read in in the original `Grid._build_edge_nodes_connectivity` using the deleted testcase, it will be:

- Since `if "Mesh2_edge_nodes" not in self.ds or repopulate:` is false, it will run `edge_nodes = self.ds["Mesh2_edge_nodes"]`
- `edge_nodes.sort(axis=1)`
- `edge_nodes_unique = edge_nodes` This is what actually happened when we called `np.unique()`
- All the following masking and fill_value or duplicate edges treatment is invalidate

4. Now we look back to your deleted test case, again th `np.unique()` will not do any actual work instead of reordering the data:
```python
        # grid with known edge node connectivity
        mpas_grid_xr = xr.open_dataset(self.mpas_filepath)
        mpas_grid_ux = ux.Grid(mpas_grid_xr)
        edge_nodes_expected = mpas_grid_ux.ds['Mesh2_edge_nodes'].values

        # arrange edge nodes in the same manner as Grid._build_edge_node_connectivity
        edge_nodes_expected.sort(axis=1)
        edge_nodes_expected = np.unique(edge_nodes_expected, axis=0)
```
It's identical to what's inside the `Grid._build_edge_nodes_connectivity`.

test/test_grid.py
```diff
@@ -534,9 +509,225 @@ def test_edge_nodes_euler(self):
...
+        face_edges_connectivity = uds.ds["Mesh2_face_edges"].values
+        face_nodes_connectivity = uds.ds["Mesh2_face_nodes"].values
+
+        # Create a dictionary to store the face indices for each edge
+        face_nodes_dict = {}
+
+        # Loop through each face and edge to build the dictionary
+        for face_idx, face_edges in enumerate(face_edges_connectivity):
+            for edge_idx in face_edges:
+                if edge_idx != ux.INT_FILL_VALUE or np.isnan(edge_idx):
```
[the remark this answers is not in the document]
hongyuchen1030: Including support for `nan` values is a precautionary measure in case users prefer to use `nan` instead of the default values. This feature does not cause any harm and may potentially be beneficial in the future.

test/test_grid.py
hongyuchen1030: As explained in the comments, this testcase will revert the `edge_nodes_connectivity` and `face_edges_connectivity` back to the `face_nodes_connectivity`, and will compare the reverted one and the original one and see if they're the same

I can move it to elsewhere. Although I didn't see its future usage in other functions

test/test_grid.py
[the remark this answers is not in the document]
philipc2: Internally we are always using integers for our connectivity arrays. It should never contain any nans

test/test_grid.py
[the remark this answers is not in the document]
philipc2: I can see that with the addition of `repopulate` this test case wasn't performing as expected. However, the purpose of this test was to test our algorithm for obtaining the edge nodes with a dataset that _already_ contained edge nodes.

I would like to keep this test, but we would need to add `repopulate=True` to the call. Then, we will still be comparing the output of our algorithm versus the provided edge nodes.

test/test_grid.py
philipc2: Makes sense, this is a great test.

We wouldn't move it into our API, we would still keep it contained within our tests. It would make understanding the test much easier and would reduce the need to duplicate this section twice for both of your tests.

test/test_grid.py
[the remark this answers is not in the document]
hongyuchen1030: `repolulate` actually will not change the original case at all. In fact, the addition of `repopulate` helps.

Previously, we only has this line in `Grid.build_edge_nodes_connectivity`
```python
if "Mesh2_edge_nodes" not in self.ds:
    # Build the connectivity

```
As explained above, since MPAS already has its own `Mesh2_edge_nodes`, it will never go into the actual 
building process and won't build the `inverse_indices`. I also described this phenomenon in issue #290 

By adding the flag `repopulate`, we can manually call the process bypassing the existed `Mesh2_edge_nodes`

But I can call it with`repopulate=True` and compare it with the original ones

philipc2:
In the future, try to avoid naming branches with the " ` " character both at the front and back, it makes checking out the branch quite confusing.

hongyuchen1030:
Comment from @philipc2 
I believe the issue lies here within your most recent PR [#297](https://github.com/UXARRAY/uxarray/pull/297)

https://github.com/UXARRAY/uxarray/blob/5e8aebe4f68d8186a8019412c43c0181f09f6409/uxarray/grid.py#L486-L488

If we have already parsed our input dataset and stored Mesh2_edge_nodes within our Grid, we still want to ensure that they are sorted and unique (lines below). This also makes it so that we will have our inverse_indices and inverse_mask computed.

The following test (same one that was remove in [#297](https://github.com/UXARRAY/uxarray/pull/297)) operates as expected with the following change.

```python
if "Mesh2_edge_nodes" not in self.ds or repopulate:
     # same body as before, unchanged
else:
    edge_nodes = self.ds["Mesh2_edge_nodes"].values
This is the slightly modified test case to use the repopulate parameter

    def test_build_edge_nodes_connectivity(self):
        """ Give a set of known (``Mesh2_edge_nodes``) from an MPAS dataset,
        compare result computed in (``Grid._build_edge_node_connectivity``) in
        order to ensure both sorted unique, sorted versions are equivalent
        """

        mpas_grid_xr = xr.open_dataset(self.mpas_filepath)
        mpas_grid_ux = ux.Grid(mpas_grid_xr)

        # unique, sorted version of MPAS edge nodes
        edge_nodes_expected = mpas_grid_ux.ds['Mesh2_edge_nodes'].values

        # computed edges nodes derived from padded face nodes
        mpas_grid_ux._build_edge_node_connectivity(repopulate=True)
        edge_nodes_output = mpas_grid_ux.ds['Mesh2_edge_nodes'].values

        nt.assert_array_equal(edge_nodes_expected, edge_nodes_output)
```
Shouldn't this also eliminate the need to call Grid._build_edge_node_connectivity(repopulate=True) from within Grid._build_face_edge_connectivity because our edge nodes will always have the expected attributes.

hongyuchen1030:
> Comment from @philipc2 I believe the issue lies here within your most recent PR [#297](https://github.com/UXARRAY/uxarray/pull/297)
> 
> https://github.com/UXARRAY/uxarray/blob/5e8aebe4f68d8186a8019412c43c0181f09f6409/uxarray/grid.py#L486-L488
> 
> If we have already parsed our input dataset and stored Mesh2_edge_nodes within our Grid, we still want to ensure that they are sorted and unique (lines below). This also makes it so that we will have our inverse_indices and inverse_mask computed.
> 
> The following test (same one that was remove in [#297](https://github.com/UXARRAY/uxarray/pull/297)) operates as expected with the following change.
> 
> ```python
> if "Mesh2_edge_nodes" not in self.ds or repopulate:
>      # same body as before, unchanged
> else:
>     edge_nodes = self.ds["Mesh2_edge_nodes"].values
> This is the slightly modified test case to use the repopulate parameter
> 
>     def test_build_edge_nodes_connectivity(self):
>         """ Give a set of known (``Mesh2_edge_nodes``) from an MPAS dataset,
>         compare result computed in (``Grid._build_edge_node_connectivity``) in
>         order to ensure both sorted unique, sorted versions are equivalent
>         """
> 
>         mpas_grid_xr = xr.open_dataset(self.mpas_filepath)
>         mpas_grid_ux = ux.Grid(mpas_grid_xr)
> 
>         # unique, sorted version of MPAS edge nodes
>         edge_nodes_expected = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> 
>         # computed edges nodes derived from padded face nodes
>         mpas_grid_ux._build_edge_node_connectivity(repopulate=True)
>         edge_nodes_output = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> 
>         nt.assert_array_equal(edge_nodes_expected, edge_nodes_output)
> ```
> 
> Shouldn't this also eliminate the need to call Grid._build_edge_node_connectivity(repopulate=True) from within Grid._build_face_edge_connectivity because our edge nodes will always have the expected attributes.

Have you tested your test case with the `Grid.build_face_edges_connectivity`?

Indeed, everything seems fine when you only call the `Grid.build_edge_nodes_connectivity`. However, we need to think about the compatibility with the following functions (like `Grid.build_face_edges_connectivity`). With your implementation, when calling the `Grid.build_face_edges_connectivity`, since it sees the `reverse_indices`, will still think it doesn't need to repopulate again from the beginning. But as I explained before, the `reverse_indices` now is incorrect and will cause issues

The reason why I made that decision is: in this way, we will not create a incorrect (and meaningless) `inverse_indices` that will cause trouble later in the `Grid.face_edges_connectivity`

I have placed back your initial test case with some modifications in my latest commit if that's your concern

philipc2:
I have just run into the issue and agree with the meaningless inverse indices. So if we chose to compute the connecitvity, do we plan to always override any existing `edge_nodes`? 

> > Comment from @philipc2 I believe the issue lies here within your most recent PR [#297](https://github.com/UXARRAY/uxarray/pull/297)
> > https://github.com/UXARRAY/uxarray/blob/5e8aebe4f68d8186a8019412c43c0181f09f6409/uxarray/grid.py#L486-L488
> > 
> > If we have already parsed our input dataset and stored Mesh2_edge_nodes within our Grid, we still want to ensure that they are sorted and unique (lines below). This also makes it so that we will have our inverse_indices and inverse_mask computed.
> > The following test (same one that was remove in [#297](https://github.com/UXARRAY/uxarray/pull/297)) operates as expected with the following change.
> > ```python
> > if "Mesh2_edge_nodes" not in self.ds or repopulate:
> >      # same body as before, unchanged
> > else:
> >     edge_nodes = self.ds["Mesh2_edge_nodes"].values
> > This is the slightly modified test case to use the repopulate parameter
> > 
> >     def test_build_edge_nodes_connectivity(self):
> >         """ Give a set of known (``Mesh2_edge_nodes``) from an MPAS dataset,
> >         compare result computed in (``Grid._build_edge_node_connectivity``) in
> >         order to ensure both sorted unique, sorted versions are equivalent
> >         """
> > 
> >         mpas_grid_xr = xr.open_dataset(self.mpas_filepath)
> >         mpas_grid_ux = ux.Grid(mpas_grid_xr)
> > 
> >         # unique, sorted version of MPAS edge nodes
> >         edge_nodes_expected = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> > 
> >         # computed edges nodes derived from padded face nodes
> >         mpas_grid_ux._build_edge_node_connectivity(repopulate=True)
> >         edge_nodes_output = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> > 
> >         nt.assert_array_equal(edge_nodes_expected, edge_nodes_output)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Shouldn't this also eliminate the need to call Grid._build_edge_node_connectivity(repopulate=True) from within Grid._build_face_edge_connectivity because our edge nodes will always have the expected attributes.
> 
> Have you tested your test case with the `Grid.build_face_edges_connectivity`?
> 
> Indeed, everything seems fine when you only call the `Grid.build_edge_nodes_connectivity`. However, we need to think about the compatibility with the following functions (like `Grid.build_face_edges_connectivity`). With your implementation, when calling the `Grid.build_face_edges_connectivity`, since it sees the `reverse_indices`, will still think it doesn't need to repopulate again from the beginning. But as I explained before, the `reverse_indices` now is incorrect and will cause issues
> 
> The reason why I made that decision is: in this way, we will not create a incorrect (and meaningless) `inverse_indices` that will cause trouble later in the `Grid.face_edges_connectivity`
> 
> I have placed back your initial test case with some modifications in my latest commit if that's your concern



I have just run into the issue and agree with the meaningless inverse indices. So if we chose to compute the connectivity, do we plan to always override any existing `edge_nodes`, since not doing so creates meaningless inverse indices? 

Alternately, is there a way we can run `Grid.face_edges_connectivity` on existing `edge_nodes`, assuming that we have them in a unique, sorted order?

hongyuchen1030:
> I have just run into the issue and agree with the meaningless inverse indices. So if we chose to compute the connecitvity, do we plan to always override any existing `edge_nodes`?
> 
> > > Comment from @philipc2 I believe the issue lies here within your most recent PR [#297](https://github.com/UXARRAY/uxarray/pull/297)
> > > https://github.com/UXARRAY/uxarray/blob/5e8aebe4f68d8186a8019412c43c0181f09f6409/uxarray/grid.py#L486-L488
> > > 
> > > If we have already parsed our input dataset and stored Mesh2_edge_nodes within our Grid, we still want to ensure that they are sorted and unique (lines below). This also makes it so that we will have our inverse_indices and inverse_mask computed.
> > > The following test (same one that was remove in [#297](https://github.com/UXARRAY/uxarray/pull/297)) operates as expected with the following change.
> > > ```python
> > > if "Mesh2_edge_nodes" not in self.ds or repopulate:
> > >      # same body as before, unchanged
> > > else:
> > >     edge_nodes = self.ds["Mesh2_edge_nodes"].values
> > > This is the slightly modified test case to use the repopulate parameter
> > > 
> > >     def test_build_edge_nodes_connectivity(self):
> > >         """ Give a set of known (``Mesh2_edge_nodes``) from an MPAS dataset,
> > >         compare result computed in (``Grid._build_edge_node_connectivity``) in
> > >         order to ensure both sorted unique, sorted versions are equivalent
> > >         """
> > > 
> > >         mpas_grid_xr = xr.open_dataset(self.mpas_filepath)
> > >         mpas_grid_ux = ux.Grid(mpas_grid_xr)
> > > 
> > >         # unique, sorted version of MPAS edge nodes
> > >         edge_nodes_expected = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> > > 
> > >         # computed edges nodes derived from padded face nodes
> > >         mpas_grid_ux._build_edge_node_connectivity(repopulate=True)
> > >         edge_nodes_output = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> > > 
> > >         nt.assert_array_equal(edge_nodes_expected, edge_nodes_output)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Shouldn't this also eliminate the need to call Grid._build_edge_node_connectivity(repopulate=True) from within Grid._build_face_edge_connectivity because our edge nodes will always have the expected attributes.
> > 
> > 
> > Have you tested your test case with the `Grid.build_face_edges_connectivity`?
> > Indeed, everything seems fine when you only call the `Grid.build_edge_nodes_connectivity`. However, we need to think about the compatibility with the following functions (like `Grid.build_face_edges_connectivity`). With your implementation, when calling the `Grid.build_face_edges_connectivity`, since it sees the `reverse_indices`, will still think it doesn't need to repopulate again from the beginning. But as I explained before, the `reverse_indices` now is incorrect and will cause issues
> > The reason why I made that decision is: in this way, we will not create a incorrect (and meaningless) `inverse_indices` that will cause trouble later in the `Grid.face_edges_connectivity`
> > I have placed back your initial test case with some modifications in my latest commit if that's your concern
> 
> I have just run into the issue and agree with the meaningless inverse indices. So if we chose to compute the connectivity, do we plan to always override any existing `edge_nodes`, since not doing so creates meaningless inverse indices?
> 
> Alternately, is there a way we can run `Grid.face_edges_connectivity` on existing `edge_nodes`, assuming that we have them in a unique, sorted order?

Technically, we do, but it still requires going over the `Grid.face_nodes_connectivity` again, making it no different than overriding.

philipc2:
> > I have just run into the issue and agree with the meaningless inverse indices. So if we chose to compute the connecitvity, do we plan to always override any existing `edge_nodes`?
> > > > Comment from @philipc2 I believe the issue lies here within your most recent PR [#297](https://github.com/UXARRAY/uxarray/pull/297)
> > > > https://github.com/UXARRAY/uxarray/blob/5e8aebe4f68d8186a8019412c43c0181f09f6409/uxarray/grid.py#L486-L488
> > > > 
> > > > If we have already parsed our input dataset and stored Mesh2_edge_nodes within our Grid, we still want to ensure that they are sorted and unique (lines below). This also makes it so that we will have our inverse_indices and inverse_mask computed.
> > > > The following test (same one that was remove in [#297](https://github.com/UXARRAY/uxarray/pull/297)) operates as expected with the following change.
> > > > ```python
> > > > if "Mesh2_edge_nodes" not in self.ds or repopulate:
> > > >      # same body as before, unchanged
> > > > else:
> > > >     edge_nodes = self.ds["Mesh2_edge_nodes"].values
> > > > This is the slightly modified test case to use the repopulate parameter
> > > > 
> > > >     def test_build_edge_nodes_connectivity(self):
> > > >         """ Give a set of known (``Mesh2_edge_nodes``) from an MPAS dataset,
> > > >         compare result computed in (``Grid._build_edge_node_connectivity``) in
> > > >         order to ensure both sorted unique, sorted versions are equivalent
> > > >         """
> > > > 
> > > >         mpas_grid_xr = xr.open_dataset(self.mpas_filepath)
> > > >         mpas_grid_ux = ux.Grid(mpas_grid_xr)
> > > > 
> > > >         # unique, sorted version of MPAS edge nodes
> > > >         edge_nodes_expected = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> > > > 
> > > >         # computed edges nodes derived from padded face nodes
> > > >         mpas_grid_ux._build_edge_node_connectivity(repopulate=True)
> > > >         edge_nodes_output = mpas_grid_ux.ds['Mesh2_edge_nodes'].values
> > > > 
> > > >         nt.assert_array_equal(edge_nodes_expected, edge_nodes_output)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Shouldn't this also eliminate the need to call Grid._build_edge_node_connectivity(repopulate=True) from within Grid._build_face_edge_connectivity because our edge nodes will always have the expected attributes.
> > > 
> > > 
> > > Have you tested your test case with the `Grid.build_face_edges_connectivity`?
> > > Indeed, everything seems fine when you only call the `Grid.build_edge_nodes_connectivity`. However, we need to think about the compatibility with the following functions (like `Grid.build_face_edges_connectivity`). With your implementation, when calling the `Grid.build_face_edges_connectivity`, since it sees the `reverse_indices`, will still think it doesn't need to repopulate again from the beginning. But as I explained before, the `reverse_indices` now is incorrect and will cause issues
> > > The reason why I made that decision is: in this way, we will not create a incorrect (and meaningless) `inverse_indices` that will cause trouble later in the `Grid.face_edges_connectivity`
> > > I have placed back your initial test case with some modifications in my latest commit if that's your concern
> > 
> > 
> > I have just run into the issue and agree with the meaningless inverse indices. So if we chose to compute the connectivity, do we plan to always override any existing `edge_nodes`, since not doing so creates meaningless inverse indices?
> > Alternately, is there a way we can run `Grid.face_edges_connectivity` on existing `edge_nodes`, assuming that we have them in a unique, sorted order?
> 
> Technically, we do, but it still requires going over the `Grid.face_nodes_connectivity` again, making it no different than overriding.

That's fair then, I'll revisit this PR tomorrow morning with this in mind. I've been considering the case where we want to re-use edge nodes, but will give this a re-review assuming that we will always construct edge nodes again if we are deriving other connectivity's.

uxarray/grid.py
```diff
@@ -100,9 +100,7 @@ def __init__(self, dataset, **kwargs):
         self.__init_grid_var_attrs__()
 
         # construct connectivity
-        if self.source_grid != "From vertices":
-            if "Mesh2_edge_nodes" not in self.ds:
-                self._build_edge_node_connectivity()
+        self._build_edge_node_connectivity()
```
philipc2, line 103: Apologies for the late review, finals season has been extremely busy.

Let's include a new `kwarg` in our `Grid` constructor to indicate whether we want to build connectivities, perhaps call it `build_connectivity`. If it is `True`, we should construct all supported connectivtities.  Additionally, let's create a helper function called `Grid.build_grid_connectivity` that constructs all of our supported connectivities, which will both be called by the constructor (if the previous case is true) or if the user desires to construct the connectivtiies after initializing a `Grid`. This implementation should remove the need for the `repopulate` flag, however we can keep it for testing purposes given it's an internal function. If we chose to expose these individual construction functions to the user, we can create a wrapper with additional safeguards. Below is a rough idea of what I'm thinking 

```Python
kwargs_list = [
            'gridspec', 'vertices', 'islatlon', 'concave', 'source_grid',
            'use_dual', `build_connectivity`
        ]
```

```Python
# construct all supported connectivities, overriding existing parsed connectivities
if self.build_connectivity is not None and self.build_connectivity is True:
    self.build_grid_connectivity()
```
```Python
def build_grid_connectivity(sef):
    self._build_edge_node_connectivity()
    self._build_face_edges_connectivity()
    # any other connectivities will be added here
```

philipc2 (changes requested):

erogluorhan (commented):
This is good overall! Please just address a few inline comments and discussion points.

uxarray/grid.py
philipc2: @erogluorhan 

I was considering this type of approach, except instead of having separate arguments for each connectivity that we would like to build, we could have `build_connectivity` either be boolean for building all of them, or a list of strings if we want to build a subset of them. 

I'm still favoring keeping this PR simple enough to get merged ASAP and handling this issue elsewhere, and I am leaning in the direction of a boolean `build_connectivity` with public helper functions for building each individual connectivity with safeguards when calling the internal build functions.

hongyuchen1030 (commented):
If we both agrees to move the build connectivity APIs to a new issue, this branch should be ready for merge now

philipc2 (approved):
Looks good! I'll tackle the API related issues next week. Thank you for putting this together and for bearing with the slower than usual reviews.

erogluorhan (commented):
I will approve this but need one clarification: Currently `_build_face_edges_connectivity` is not called anywhere in the Grid object construction (i.e. this PR only introduces its implementation and tests but don't start to actually use it), and it will be handled by @philipc2 in a new API PR, right? 

If so, please create an issue for that task, and let us get this one merged.

erogluorhan (approved):
Great job!

PR #299, opened 2023-04-27
Add `Grid.build_node_face_connectivity`

hongyuchen1030:
Add `Grid._build_node_face_connectivity` internal API using the sparse matrix and its corresponding testcases


philipc2 (changes requested):
Please look into the two mesh files & logo that was included in this PR; maybe they got accidently committed?

hongyuchen1030:
> Please look into the two mesh files & logo that was included in this PR; maybe they got accidently committed?
I just saw your comments. The weird thing is I didn't do anything with the image at all. But I will see if I can just directly remove them

uxarray/helpers.py
```diff
@@ -673,3 +673,53 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
...
+    ...     face_nodes_conn, INT_FILL_VALUE)
+
+    >>> face_indices = np.array([0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3])
+    >>> nodes_indices = np.array([3, 4, 5, 3, 0, 2, 5, 3, 4, 1, 0, 0, 1, 2])
+    >>> non_zero_flag = np.array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
+    """
+    n_rows, n_cols = dense_matrix.shape
+    flattened_matrix = dense_matrix.ravel()
+    valid_node_mask = ~np.isnan(flattened_matrix) & (flattened_matrix !=
+                                                     fill_value)
```
philipc2, line 721: Not necessary to check for `nan` in face nodes

uxarray/helpers.py
```diff
@@ -673,3 +673,53 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
     np.put(closed.ravel(), first_fv_idx_1d, first_node_value)
 
     return closed
+
+
+def _convert_face_node_conn_to_sparse_matrix(
+        dense_matrix: np.ndarray, fill_value=INT_FILL_VALUE) -> tuple:
```
philipc2, line 679: Rename to `dense_face_nodes` or similar to indicate this function is only used for face nodes

uxarray/helpers.py
```diff
@@ -673,3 +673,53 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
...
+    """Converts a given dense matrix connectivity to a sparse matrix format
+    where the locations of non-zero entries are stored using COO (coordinate
```
philipc2, line 681: Throughout this function, the sparse matrix format contains the location of **non fill-value entries**, not non zero locations.

philipc2 (changes requested):
Before moving onto the core functionality, a few comments about the helper function

uxarray/helpers.py
hongyuchen1030: I think that's what we want here. Why do we want the non-zero locations (what does the non-zero location mean?)?

uxarray/helpers.py
hongyuchen1030: the current function name already has the `face_nodes` keywords inside. I think it already refect the purpose

uxarray/helpers.py
philipc2: Im referring to the name of the variable and the doc string.

hongyuchen1030: ohh Got it. Thanks for pointing it out

uxarray/helpers.py
```diff
@@ -673,3 +673,49 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
     np.put(closed.ravel(), first_fv_idx_1d, first_node_value)
 
     return closed
+
+
+def _convert_face_node_conn_to_sparse_matrix(dense_matrix: np.ndarray) -> tuple:
```
philipc2, line 678: This function looks good now! I'd consider renaming it to something shorter since the name is quite long.

Maybe something like: `face_nodes_to_sparse_matrix`?

uxarray/grid.py
```diff
@@ -540,6 +542,73 @@ def _build_edge_node_connectivity(self, repopulate=False):
         setattr(self, "Mesh2_edge_nodes", self.ds['Mesh2_edge_nodes'])
         setattr(self, "nMesh2_edge", edge_nodes_unique.shape[0])
 
+    def _build_node_faces_connectivity(self):
```
philipc2, line 545: Should be `node_face` instead of `node_faces` throughout

```suggestion
    def _build_node_face_connectivity(self):
```

uxarray/grid.py
```diff
@@ -540,6 +542,73 @@ def _build_edge_node_connectivity(self, repopulate=False):
...
+        """Builds the `Grid.Mesh2_node_faces`: integer DataArray of size
```
philipc2, line 546:
```suggestion
        """Builds the `Grid.Mesh2_node_face`: integer DataArray of size
```

uxarray/grid.py
```diff
@@ -540,6 +542,73 @@ def _build_edge_node_connectivity(self, repopulate=False):
...
+
+        Raises
+        ------
+        RuntimeError
+            If the Mesh object does not contain a 'Mesh2_face_nodes' variable.
+        """
+        if "Mesh2_face_nodes" not in self.ds:
+            raise RuntimeError(
+                "Currently, node_face_connectivity can only be built with Mesh2_face_nodes"
+            )
```
philipc2, line 561: I believe that we will always have `Mesh2_face_nodes`, we should keep these types of checks outside these build functions. As we discussed in a previous PR, we'll have separate functions exposed to the user that will handle these checks.

uxarray/helpers.py
```diff
@@ -673,3 +673,49 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
...
+    tuple
+        A tuple containing three arrays:
+
+        - face_indices : np.ndarray
+            Array containing the face indices for each non-zero element.
+        - node_indices : np.ndarray
+            Array containing the node indices for each non-zero element.
+        - non_zero_elements_flag : np.ndarray
+            Array containing flags indicating if a non-zero element is present in the corresponding row and column
+            index.
```
philipc2, line 700: Change the description and variable name to mention non-fill-value entries instead of non-zero

uxarray/helpers.py
```diff
@@ -673,3 +673,49 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
...
+    >>> face_nodes_conn = np.array([[3, 4, 5, INT_FILL_VALUE],
+    ...                             [3, 0, 2, 5],
+    ...                             [3, 4, 1, 0],
+    ...                             [0, 1, 2, -999]])
+    >>> face_indices, nodes_indices, non_zero_flag = _convert_face_node_conn_to_sparse_matrix(
+    ...     face_nodes_conn, INT_FILL_VALUE)
+
+    >>> face_indices = np.array([0, 0, 0, 1, 1, 1, 1, 2, 2, 2, 2, 3, 3, 3])
+    >>> nodes_indices = np.array([3, 4, 5, 3, 0, 2, 5, 3, 4, 1, 0, 0, 1, 2])
+    >>> non_zero_flag = np.array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
```
philipc2, line 713: same as above regarding the non-zero

uxarray/helpers.py
```diff
@@ -673,3 +673,49 @@ def close_face_nodes(Mesh2_face_nodes, nMesh2_face, nMaxMesh2_face_nodes):
...
+    """
+    n_rows, n_cols = dense_matrix.shape
+    flattened_matrix = dense_matrix.ravel()
+    valid_node_mask = flattened_matrix != INT_FILL_VALUE
+    face_indices = np.repeat(np.arange(n_rows), n_cols)[valid_node_mask]
+    node_indices = flattened_matrix[valid_node_mask]
+    non_zero_element_flags = np.ones(len(node_indices))
+    return face_indices, node_indices, non_zero_element_flags
```
philipc2, line 721: same as above regarding non-zero

uxarray/grid.py
```diff
@@ -540,6 +542,73 @@ def _build_edge_node_connectivity(self, repopulate=False):
...
+
+        # First we need to build a matrix such that: the row indices are face indexes and the column indices are node
+        # indexes (similar to an adjacency matrix)
+        face_indices, node_indices, non_zero_element_flags = _convert_face_node_conn_to_sparse_matrix(
```
philipc2, line 565: same comment regarding the non_zero name

docs/internal_api/index.rst
```diff
@@ -20,6 +20,7 @@ Grid Methods
    grid.Grid.__from_vert__
    grid.Grid.__init_grid_var_attrs__
    grid.Grid._build_edge_node_connectivity
+   grid.Grid._build_node_faces_connectivity
```
philipc2, line 23:
```suggestion
   grid.Grid._build_node_face_connectivity
```

uxarray/grid.py
```diff
@@ -540,6 +542,73 @@ def _build_edge_node_connectivity(self, repopulate=False):
...
+        for node_index in range(self.nMesh2_node):
+            node_face_connectivity[node_index] = cols[
+                start_indices[node_index]:end_indices[node_index] + 1]
+            if len(node_face_connectivity[node_index]) < nMaxNumFacesPerNode:
+                node_face_connectivity[node_index] = np.append(
+                    node_face_connectivity[node_index],
+                    np.full(
+                        nMaxNumFacesPerNode -
+                        len(node_face_connectivity[node_index]),
+                        self.ds["Mesh2_face_nodes"].attrs["_FillValue"]))
```
philipc2, line 598: When referencing the fill value, use the important constant `INT_FILL_VALUE` instead.

uxarray/grid.py
```diff
@@ -540,6 +542,73 @@ def _build_edge_node_connectivity(self, repopulate=False):
...
+        self.ds["Mesh2_node_faces"] = xr.DataArray(
+            node_face_connectivity,
+            dims=["nMesh2_node", "nMaxNumFacesPerNode"],
+            attrs={
+                "long_name": "Maps every node to the faces that "
+                             "it connects",
+                "nMaxNumFacesPerNode": nMaxNumFacesPerNode,
+                "_FillValue": self.ds["Mesh2_face_nodes"].attrs["_FillValue"]
```
philipc2, line 606: Same as above regarding `INT_FILL_VALUE`

uxarray/grid.py
philipc2, line 598: Make sure this is the correct datatype as defined in the constant `INT_DTYPE`

philipc2 (changes requested):
A few more comments! Once these are addressed, I'll tackle the test case & functionality, but generally everything is looking good!

uxarray/grid.py
hongyuchen1030: According to our naming conventions, it should be `node_faces` since it records the neighboring faces of a node. Same as our `edge_nodes_connectivity`, `face_nodes_connectivity`

uxarray/grid.py
hongyuchen1030: we can delete it later after we have that checking function working. Currently, I don't feel comfortable exposing the API without any pre-check.

uxarray/grid.py
hongyuchen1030: This is the point that has bothering me for a while:

1. If we allow the user to customize the `fill_value`, why do we keep removing the customer fill_value support from our functions? 
2. If we don't allow the user to customize the `fill_value`, why do we store the "_FillValue" `attrs` in the Grid.ds`?

philipc2: 1. The user currently can't customize the fill value. If we add support for this, it'll be done globally through the `constants.py` module, so using the important constant would take those changes into account
2. I've thought about this and it does seem redundant, but for now we keep it since it's an expected attribute according to the UGRID conventions

uxarray/grid.py
philipc2: This works with me

uxarray/grid.py
philipc2: Got it, let's keep it the way it is then. Even after looking at the UGRID document 1000x times I still mess up these little things :)

uxarray/grid.py
hongyuchen1030: I can change back to the imported constant.

However, this Fill_Value design here might lead to a big loophole: Basically,
 we are exposing an API access that can modify the `_FillValue` attrs to users, while this `_FillValue` modification is not linked to the `constant.py` module.
 
 So users might do the following
 ```python
 
 # Use the default constant fill_value
 verts = [1.001, 1.001, INT_FILLVALUE]
 vgrid = Grid(verts, fillvalue = INT_FILLVALUE)
 
 # User thinks they are changing the fill value globally:
 vgrid.ds["Mesh2_face_nodes"].attrs["_FillValue"] = -1
 
 # And if they call our connectivity population function, the codes break down
 vgrid.populate_connectivity()
 ```
 
 This is also the reason why I am so conservative when incorporating the "FIll_VALUE" check in my functions. Never too much for a more adaptive/strict function call/check

test/test_grid.py
```diff
@@ -771,3 +748,72 @@ def test_build_face_edges_connectivity_fillvalues(self):
         self.assertTrue(
             np.array_equal(res_face_nodes_connectivity,
                            uds.ds["Mesh2_face_nodes"].values))
+
+    def test_node_face_connectivity_from_verts(self):
```
philipc2, line 752: Include a short docstring.

test/test_helpers.py
```diff
@@ -178,3 +178,23 @@ def test_replace_fill_values_invalid(self):
                                                    original_fill=-1,
                                                    new_fill=INT_FILL_VALUE,
                                                    new_dtype=np.int16)
+
+
+class TestSparseMatrix(TestCase):
+
+    def test_convert_face_node_conn_to_sparse_matrix(self):
```
philipc2, line 185: Include a short docstring

philipc2 (changes requested):
Everything looks great functionality wise, gave it a walkthrough locally and all appears to check out. A few minor comments, but other than that this is looking ready!

test/test_grid.py
```diff
@@ -771,3 +748,72 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+
+            # use the dictionary method to build the node_face_connectivity
+            node_face_connectivity = {}
+            for i in range(grid_ux.nMesh2_face):
+                face_nodes = grid_ux.ds["Mesh2_face_nodes"].values[i]
+                for j in face_nodes:
+                    if j != ux.INT_FILL_VALUE:
+                        if j not in node_face_connectivity:
+                            node_face_connectivity[j] = []
+                        node_face_connectivity[j].append(i)
```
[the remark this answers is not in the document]
hongyuchen1030: Thank you for your suggestion

erogluorhan:
@philipc2 resolve merge conflicts please

docs/internal_api/index.rst
```diff
@@ -88,10 +88,18 @@ Methods
 .. autosummary::
    :toctree: _autosummary
 
-   Grid.__init_grid_var_names__
-   Grid.__from_ds__
-   Grid.__from_vert__
-   Grid.__init_grid_var_attrs__
+<<<<<<< HEAD
```
erogluorhan, line 91:
```suggestion
```

uxarray/grid/connectivity.py
```diff
@@ -230,3 +232,109 @@ def _build_face_edges_connectivity(grid):
...
+
+    Raises
+    ------
+    RuntimeError
+        If the Mesh object does not contain a 'Mesh2_face_nodes' variable.
+    """
+    if "Mesh2_face_nodes" not in grid._ds:
+        raise RuntimeError(
+            "Currently, node_face_connectivity can only be built with Mesh2_face_nodes"
+        )
```
philipc2, line 253: The minimum set of variables always available are `Mesh2_node_x`, `Mesh2_node_y`, `Mesh2_face_nodes`, so this check shouldn't need to be necessary. 

Additionally, we should move these checks inside of the `Grid` property attribute in the future if they are needed.

uxarray/utils/helpers.py
```diff
@@ -0,0 +1,721 @@
...
+from numba import njit, config
+from uxarray.utils.constants import ENABLE_JIT_CACHE, ENABLE_JIT
+import math
+
+from uxarray.utils.constants import INT_DTYPE, INT_FILL_VALUE
+
+config.DISABLE_JIT = not ENABLE_JIT
+
+
+def parse_grid_type(dataset):
```
philipc2, line 14: This whole module appears to have been left over from a merge, it is not needed anymore and can be deleted. 

Please move any new functions in here into their appropriate module, which I would say is `grid/connectivity.py`

docs/internal_api/index.rst
```diff
@@ -88,10 +88,17 @@ Methods
...
+   __init_grid_var_names__
+   __from_ds__
+   __from_vert__
+   __init_grid_var_attrs__
+   _build_edge_node_connectivity
+   _build_face_edges_connectivity
+   _build_node_faces_connectivity
+   _build_nNodes_per_face
+   _populate_cartesian_xyz_coord
+   _populate_lonlat_coord
```
philipc2, line 100: These appear to be duplicates with the following sections, can remove. 

https://github.com/UXARRAY/uxarray/blob/9314962d46993a41289b7fea019814ff25a47d41/docs/internal_api/index.rst?plain=1#L128-L132

https://github.com/UXARRAY/uxarray/blob/9314962d46993a41289b7fea019814ff25a47d41/docs/internal_api/index.rst?plain=1#L149-L150

philipc2 (changes requested):
Looks good! A few small comments

philipc2 (approved):
Great work!

uxarray/grid/connectivity.py
```diff
@@ -230,3 +232,104 @@ def _build_face_edges_connectivity(grid):
             "start_index": INT_DTYPE(0),
             "long_name": "Maps every edge to the two nodes that it connects",
         })
+
+
+def _build_node_faces_connectivity(grid):
```
aaronzedwick, line 237: Should this be added to the internal_api?

hongyuchen1030: I added it to the internal_api.rst under the `helper` section before the redesign. However, I am open to @philipc2 opinion about where this function should belong to.

philipc2: Good catch @aaronzedwick, that function should be listed under here:

https://github.com/UXARRAY/uxarray/blob/a08e11c2dc7aefb8371c49cb381689740792983e/docs/internal_api/index.rst?plain=1#L114-L126C26

test/test_grid.py
```diff
@@ -778,3 +778,72 @@ def test_build_face_edges_connectivity_fillvalues(self):
         self.assertTrue(
             np.array_equal(res_face_nodes_connectivity,
                            uds._ds["Mesh2_face_nodes"].values))
+
+    def test_node_face_connectivity_from_verts(self):
+        """Test generating Grid.Mesh2_node_faces from array input."""
+        node_latlon_degree = [[162., 30], [216., 30], [70., 30], [162., -30],
```
erogluorhan, line 784: actually "lonlat", not "latlon"

test/test_grid.py
```diff
@@ -778,3 +778,72 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+                        islatlon=True,
+                        concave=False)
+        expected = np.array([
+            np.array([0, 1, ux.INT_FILL_VALUE]),
+            np.array([1, 3, ux.INT_FILL_VALUE]),
+            np.array([0, 1, 2]),
+            np.array([1, 2, 3]),
+            np.array([0, 2, ux.INT_FILL_VALUE]),
+            np.array([2, 3, ux.INT_FILL_VALUE])
+        ])
```
erogluorhan, line 811: Two things this `expected`: 

1. These magic numbers cannot be understood by the reader of the code. Maybe a geometric justification or something else as a proof that this is the expected node-face connectivity is needed
2. I may be wrong, but my calculation on paper gives me the following node-face connectivity instead (this also shows the need for the ask (1) ):
```
[[1,2,3],
  [2,3],
  [1,3],
  [0,1,2],
  [0,2],
  [0,1]]
```

which is the same with your expected result but has different node order

erogluorhan (commented):
This is going great; please address a few comments.

test/test_grid.py
hongyuchen1030: The input will be re-ordered inside the `Grid.build_face_nodes_connectivity` while the topology is the same. We when checking with the expectation, we need to make sure the `Grid._node_face_connectivity` coheres with the `Grid._face_nodes_connectivity`. But I will try to add more explanation in the codes

| Input |         |   | Actually Store in `Grid._face_nodes_connectivity` |         |
|-------|---------|---|---------------------------------------------------|---------|
| Face  | Node    |   | Face                                              | Node    |
| 0     | 3,4,5   |   | 0                                                 | 2,4,0   |
| 1     | 3,0,2,5 |   | 1                                                 | 2,3,1,0 |
| 2     | 3,4,1,0 |   | 2                                                 | 2,4,5,3 |
| 3     | 0,1,2   |   | 3                                                 | 3,5,1   |


| Expected `Grid._node_face_connectivity` |       |
|-----------------------------------------|-------|
| Node                                    | Face  |
| 0                                       | 0, 1  |
| 1                                       | 1,3   |
| 2                                       | 0,1,2 |
| 3                                       | 1,2,3 |
| 4                                       | 0,2   |
| 5                                       | 2,3   |
|                                         |       |

hongyuchen1030:
@philipc2 The CI suddenly fails for all non-windows test. I wonder if you have any clue about that? Thank you very much

philipc2:
> @philipc2 The CI suddenly fails for all non-windows test. I wonder if you have any clue about that? Thank you very much

Thanks for pointing this out! It shouldn't be related to anything you've done and is failing in our Nightly tests too. I'll take a look.

aaronzedwick (approved):
Looks good, nice work!

rajeeja (approved):

test/test_grid.py
erogluorhan: Thanks! There is no `Grid.build_face_nodes_connectivity` but I got you meant `Grid.Mesh2_face_nodes`.

test/test_grid.py
```diff
@@ -778,3 +778,85 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+            ux.INT_FILL_VALUE)
+
+        for i, face_nodes_conn_index_row in enumerate(face_nodes_conn_index):
+            for j, node_index in enumerate(face_nodes_conn_index_row):
+                if node_index != ux.INT_FILL_VALUE:
+                    face_nodes_conn_lonlat[i,
+                                           j] = node_lonlat_degree[node_index]
+
+        # Now we don't need the face_nodes_conn_index anymore. But calling del face_nodes_conn_index will result in
+        # Failing CI for macos and ubuntu (windows is fine). So we just do nothing about it here
```
erogluorhan, line 809: No need for the above two lines; please remove

test/test_grid.py
```diff
@@ -778,3 +778,85 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+
+    def test_node_face_connectivity_from_verts(self):
+        """Test generating Grid.Mesh2_node_faces from array input."""
+
+        # We used the following codes to generate the testing face_nodes_connectivity in lonlat,
+        # The index of the nodes here is just for generation purpose and ensure the topology.
+        # These information will not be used in the Grid.
+        # The input will be re-ordered inside the `Grid.build_face_nodes_connectivity` while the topology is the same.
+        # When checking with the expectation, we need to make sure the `Grid._node_face_connectivity` coheres
+        # with the `Grid._face_nodes_connectivity`.
```
erogluorhan, line 790: How about changing the whole above comment with the following:

"This nodes list is only for vertices creation purposes and the nodes' order will not be used the 
same in the `Grid` object; i.e. the `Grid` object instantiation will instead use the below 
`face_nodes_conn_lonlat ` connectivity variable and determine the actual node orders by itself."

erogluorhan (commented):
This looks great, and I am happy to approve it once you address the last few small comments.

PR #303, opened 2023-05-02
Customized IntervalTree for both `float` and arbitrary precision usage

hongyuchen1030:
1. Implement a subclass for https://github.com/chaimleib/intervaltree, which supports the inclusive interval on both ends (interval begin and end)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,10 @@
-xarray
+xarray~=2022.12.0
 numpy<1.24
+
+pathlib~=1.0.1
+yaml~=0.2.5
+pyyaml~=6.0
+numba~=0.56.4
+intervaltree~=3.1.0
+sortedcontainers~=2.4.0
+setuptools~=66.0.0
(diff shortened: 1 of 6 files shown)


dcherian:
FYI Pandas provides [IntervalIndex](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.IntervalIndex.html) which builds on a private IntervalTree, and support `closed="both"`. 

Would it satisfy the usecases you have in mind? Pandas is already in your dependency list as a hard dependency for xarray.

hongyuchen1030:
> FYI Pandas provides [IntervalIndex](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.IntervalIndex.html) which builds on a private IntervalTree, and support `closed="both"`.
> 
> Would it satisfy the usecases you have in mind? Pandas is already in your dependency list as a hard dependency for xarray.

Thank you very much for your information. I will look into that, thanks

PR #308, opened 2023-05-09
Rename `face_dimension` connectivity to `nNodes_per_face`

### LINKED ISSUE #307: Rename `face_dimension` connectivity to `nNodes_per_face`
Current implementation of `face_dimension` does not align with the UGRID conventions. `nNodes_per_face` is a more appropriate name for the current implementation of `face_dimension`

philipc2:
Closes #307

## Overview

* #283 introduced the `Mesh2_face_dimension` connectivity variable to our API, which was used to store the number of non-fill-value nodes per face. However, this name conflicted with definition of `Mesh2_face_dimension` in the UGRID conventions
* All instances of this `face_dimension` were renamed to `nNodes_per_face`, which is a more appropriate name

## Expected Usage

```Python
import uxarray as ux

uxgrid = ux.Grid(grid_path)

# access nNodes_per_face through internal dataset
uxgrid.ds['nNodes_per_face']

# access nNodes_per_face through attribute
uxgrid.nNodes_per_face
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Documentation**


test/test_grid.py
```diff
@@ -286,8 +286,8 @@ def test_read_scrip(self):
         xr_grid_u30 = xr.open_dataset(ug_30)
         ux_grid_u30 = ux.Grid(xr_grid_u30)  # tests from ugrid
 
-    def test_build_face_dimension(self):
-        """Tests the construction of the ``Mesh2_face_dimension`` variable."""
+    def test_build_nNodes_per_face(self):
```
hongyuchen1030, line 289: Can we utilize the `Grid.__from_vert__()` to input our written`face_nodes_connectivity` and then verify if our dimensions calculation for that is accurate?

uxarray/grid.py
```diff
@@ -660,11 +660,11 @@ def _populate_lonlat_coord(self):
...
 
-    def _build_face_dimension(self):
-        """Constructs ``Mesh2_face_dimension``, which calculates the dimension
-        of each face in ``Mesh2_face_nodes``"""
+    def _build_nNodes_per_face(self):
+        """Constructs ``nNodes_per_face``, which contains the number of non-
+        fill-value nodes for each face in ``Mesh2_face_nodes``"""
 
-        # Triangular Mesh
+        # Triangular Mesh (No Fill Values)
```
hongyuchen1030, line 667: Why the branching condition for triangular mesh is `if not hasattr(self, "nMaxMesh2_face_nodes")?`

In other words, in the future, is that possible that some input files/input parameters don't have the `nMaxMesh2_face_nodes` attribute and also have some fill values?

philipc2: This is related to what we discussed on Tuesday during our meeting. Since there is an instance where we read in a triangular mesh, it will not have a `nMaxMesh2_face_nodes`. I will fix this in a separate request.

test/test_grid.py
philipc2: That's a great idea, I'll do that.

uxarray/grid.py
hongyuchen1030: I understand what you're talking about the MPAS data.
 
 My concern is: this branching is not very strict and robust. If some inputs(not MPAS triangulars) don't have the nMaxMesh2_face_nodes attribute and also have some fill values, it will break this function.
 
 You can also consider add this as a test case for this function using the `Grid.__from_vert__()`

philipc2: I agree. I am planning to remove this branching with a PR that addresses #311

> nMaxMesh2_face_nodes attribute and also have some fill values

I don't believe that this would ever happen since always have `nMaxMesh2_face_nodes` when fill values are present

rajeeja (approved):

erogluorhan (approved):

PR #317, opened 2023-05-12
CI Update: Drop support for Python 3.7 and 3.8; add 3.11

### LINKED ISSUE #295: Deprecate Python 3.8 and support 3.11
Relevant: https://github.com/NCAR/geocat-comp/issues/381#issue-1638006388

Xarray requires Python 3.9, with Python 3.8 builds only supporting up to  `xarray 2023.1.0`, which may break things in the future. Additionally, we should support Python 3.11.

erogluorhan:
Closes #295 

## Overview

Removes Python 3.7 and 3.8 from and adds 3.11 to setup.py and ci.yml

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**
- [x] Adequate tests are covered when existing tests pass with Python 3.11]


philipc2 (commented):
https://github.com/UXARRAY/uxarray/blob/2ccd302970918e9d07c24e269e37822ea24440b1/setup.py#L26

This should be changed to:
```Python
python_requires='>=3.9',

**Comment by philipc2** [comment]
Relevant Issue from Xarray: https://github.com/pydata/xarray/pull/7353#issue-1474717029

**Review by philipc2 — COMMENTED** [review]
I'll do some digging into the failures. Most likely one of our dependencies is not playing well with 3.11, so we will need to set up a separate 3.11 environment like Xarray excluding any failing packages.

https://github.com/pydata/xarray/blob/964d350a80fe21d4babf939c108986d5fd90a2cf/ci/requirements/environment-py311.yml

https://github.com/pydata/xarray/pull/7469#issue-1552997940

**Comment by philipc2** [comment]
In our CI setup, we use we have `pre_commit` listed in the requirements, which has has different system expectations on [conda-forge](https://anaconda.org/conda-forge/pre_commit) while Xarray and most other packages use `pre-commit` as shown [here ](https://anaconda.org/conda-forge/pre-commit). From what I can tell, they are the same version and are built from the same repository. Unsure if this is causing any issues though

**Comment by philipc2** [comment]
It appears to be a conflict with Numba. Support for Numpy 1.24 and Python 3.11 was added in the `0.57.0` release, which is not available on the conda forge channel.

**Comment by philipc2** [comment]
See https://github.com/conda-forge/numba-feedstock/pull/115

**Comment by erogluorhan** [comment]
> It appears to be a conflict with Numba. Support for Numpy 1.24 and Python 3.11 was added in the `0.57.0` release, which is not available on the conda forge channel which currently hosts `0.56.4`

Good catch!

**Comment by philipc2** [comment]
@erogluorhan 

https://github.com/conda-forge/numba-feedstock/pull/115 was resolved earlier today and after some slight modifications to our `environment.yml` file, everything appears to be working as expected. I believe this should be ready for review now.

File: ci/environment.yml [hunk 1]
```diff
@@ -2,13 +2,13 @@ name: uxarray_build
...
+  - dask
   - numpy
+  - scipy
+  - netcdf4
   - pathlib
   - pre_commit
   - pytest
   - pytest-cov
-  - scipy
-  - xarray
```
erogluorhan, line 14: @philipc2 this file has been in alphabetical order; that's not vital, but we would still not want to break it unless this new order you are making has a functional effect.

philipc2: I'll do a quick re-run in alphabetical order, but it wasn't until I changed the order to have `xarray` `numba` and `dask` at the top of the list that the conda environment built properly.

Let me look into this

philipc2: Appears to be working in alphabetical order, should be good now!

philipc2 (approved):
Looks good!

anissa111:
This should probably also be updated:
https://github.com/UXARRAY/uxarray/blob/bb3e12f5f392e5f39d512cca4cc33cec5d833ccc/docs/installation.rst?plain=1#L52C1-L54

After this and the tests go through I think it'll be good!

erogluorhan:
> This should probably also be updated: https://github.com/UXARRAY/uxarray/blob/bb3e12f5f392e5f39d512cca4cc33cec5d833ccc/docs/installation.rst?plain=1#L52C1-L54
> 
> After this and the tests go through I think it'll be good!

Good catch! Updated now.

erogluorhan:
> > This should probably also be updated: https://github.com/UXARRAY/uxarray/blob/bb3e12f5f392e5f39d512cca4cc33cec5d833ccc/docs/installation.rst?plain=1#L52C1-L54
> > After this and the tests go through I think it'll be good!
> 
> Good catch! Updated now.

Since tests are passing, merging this now per the comment and one existing approval.

PR #333, opened 2023-05-31
Implement Version Attribute 

### LINKED ISSUE #277: Implement Version Method
Implement `ux.__version__()` or similar method

aaronzedwick:
Closes #277 

## Overview

* Adds a version attribute allowing users to view current version

## Expected Usage

```Python
import uxarray as ux

ux.__version__
```

## PR Checklist

**General**
- [x] Add appropriate labels

### DIFF
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -3,6 +3,7 @@ from .helpers import *
 
 
 def __version__():
+    """Returns the version of uxarray currently installed."""
     # Attempt to import the needed modules
     try:
         from importlib.metadata import version as version
@@ -11,7 +12,8 @@ def __version__():
 
     try:
         __version = version("uxarray")
-        print(__version)
-    except AttributeError:
-        # Incase something happens, such as the library isn't installed
+    except NameError:
+        # Placeholder version incase an error occurs, such as the library isn't installed
         __version = "0000"
+
+    return __version
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -1,2 +1,17 @@
 from .grid import *
 from .helpers import *
+
+
+def __version__():
+    # Attempt to import the needed modules
+    try:
+        from importlib.metadata import version as version
+    except ImportError:
+        from importlib_metadata import version as version
+
+    try:
+        __version = version("uxarray")
+        print(__version)
+    except AttributeError:
+        # Incase something happens, such as the library isn't installed
+        __version = "0000"
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -4,7 +4,7 @@ from .helpers import *
 # Attempt to import the needed modules
 try:
     from importlib.metadata import version as version
-except ImportError:
+except Exception:
     from importlib_metadata import version as version
 
 try:
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -1,19 +1,14 @@
 from .grid import *
 from .helpers import *
+# Sets the version of uxarray currently installed
+# Attempt to import the needed modules
+try:
+    from importlib.metadata import version as version
+except ImportError:
+    from importlib_metadata import version as version
 
-
-def __version__():
-    """Returns the version of uxarray currently installed."""
-    # Attempt to import the needed modules
-    try:
-        from importlib.metadata import version as version
-    except ImportError:
-        from importlib_metadata import version as version
-
-    try:
-        __version = version("uxarray")
-    except NameError:
-        # Placeholder version incase an error occurs, such as the library isn't installed
-        __version = "0000"
-
-    return __version
+try:
+    __version = version("uxarray")
+except Exception:
+    # Placeholder version incase an error occurs, such as the library isn't installed
+    __version = "999"
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -8,7 +8,7 @@ except ImportError:
     from importlib_metadata import version as version
 
 try:
-    __version = version("uxarray")
+    __version__ = version("uxarray")
 except Exception:
     # Placeholder version incase an error occurs, such as the library isn't installed
-    __version = "999"
+    __version__ = "999"
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -8,7 +8,7 @@ except Exception:
     from importlib_metadata import version as _version
 
 try:
-    __version__ = version("uxarray")
+    __version__ = _version("uxarray")
 except Exception:
     # Placeholder version incase an error occurs, such as the library isn't installed
     __version__ = "999"
--- a/uxarray/__init__.py
+++ b/uxarray/__init__.py
@@ -3,9 +3,9 @@ from .helpers import *
 # Sets the version of uxarray currently installed
 # Attempt to import the needed modules
 try:
-    from importlib.metadata import version as version
+    from importlib.metadata import version as _version
 except Exception:
-    from importlib_metadata import version as version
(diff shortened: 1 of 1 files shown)


uxarray/__init__.py
```diff
@@ -1,2 +1,17 @@
...
+def __version__():
+    # Attempt to import the needed modules
+    try:
+        from importlib.metadata import version as version
+    except ImportError:
+        from importlib_metadata import version as version
+
+    try:
+        __version = version("uxarray")
+        print(__version)
```
philipc2, line 14: We want to avoid directly printing the version. We instead would like to return a string of the current version.

uxarray/__init__.py
```diff
@@ -1,2 +1,17 @@
...
+    except AttributeError:
```
philipc2, line 15: This should be a `NameError`, since if our library isn't installed, we would have never declared the `version` method above. 

Attempting to execute `version("uxarray")` would then raise a `NameError`

uxarray/__init__.py
```diff
@@ -1,2 +1,17 @@
...
+        # Incase something happens, such as the library isn't installed
```
philipc2, line 16:
```suggestion
        # Placeholder version incase an error occurs, such as the library isn't installed
```

uxarray/__init__.py
```diff
@@ -1,2 +1,17 @@
 from .grid import *
 from .helpers import *
+
+
+def __version__():
```
philipc2, line 5: Even though this is a very small function, please include a simple docstring describing the functionality. It can be simply a single sentence like the example below:

https://github.com/UXARRAY/uxarray/blob/efca32ed3bd8edb0a81edfba18561938b06200db/uxarray/grid.py#L703-L705

philipc2 (changes requested):
Thank you for putting this together so quickly! A couple comments:

uxarray/__init__.py
anissa111, line 5: In my experience, it's more standard to implement `__version__` as an attribute instead of a function. 

I'd point you towards xarray's implementation here: https://github.com/pydata/xarray/blob/9909f90b4781be89e3f3ff7c87893928b3e3be6e/xarray/__init__.py#L41-L52
```python
try:
    from importlib.metadata import version as _version
except ImportError:
    # if the fallback library is missing, we are doomed.
    from importlib_metadata import version as _version

try:
    __version__ = _version("xarray")
except Exception:
    # Local copy or not installed with setuptools.
    # Disable minimum version checks on downstream libraries.
    __version__ = "999"
```

uxarray/__init__.py
anissa111: I think we can actually generalize this to just `except Exception`, as we don't really care what specifically goes wrong, just that something did.

uxarray/__init__.py
```diff
@@ -1,2 +1,17 @@
...
+        from importlib.metadata import version as version
+    except ImportError:
+        from importlib_metadata import version as version
+
+    try:
+        __version = version("uxarray")
+        print(__version)
+    except AttributeError:
+        # Incase something happens, such as the library isn't installed
+        __version = "0000"
```
anissa111, line 17: Actually, I'd suggest a placeholder like 999, as some code that uses ours as a dependency might have minimum version checks, which we don't need to trigger if the only thing wrong here is pulling our version, does that make sense?

uxarray/__init__.py
aaronzedwick: Good point, thanks for all the feedback on everything, I really appreciate it!

anissa111 (changes requested):
Hi Aaron! I've included some thoughts below. Thanks for your contribution (and nice to virtually interact with you)!

uxarray/__init__.py
aaronzedwick: Okay, that does make sense, thanks!

uxarray/__init__.py
aaronzedwick: Yeah, that makes perfect sense, I hadn't thought of that at all, thanks!

uxarray/__init__.py
```diff
@@ -1,2 +1,14 @@
...
+try:
+    from importlib.metadata import version as version
+except ImportError:
+    from importlib_metadata import version as version
+
+try:
+    __version = version("uxarray")
+except Exception:
+    # Placeholder version incase an error occurs, such as the library isn't installed
+    __version = "999"
```
anissa111, line 14: Just a little change to keep the intended `ux.__version__`

```suggestion
try:
    __version__ = version("uxarray")
except Exception:
    # Placeholder version incase an error occurs, such as the library isn't installed
    __version__ = "999"
```

uxarray/__init__.py
```diff
@@ -1,2 +1,14 @@
 from .grid import *
 from .helpers import *
+# Sets the version of uxarray currently installed
+# Attempt to import the needed modules
+try:
+    from importlib.metadata import version as version
+except ImportError:
+    from importlib_metadata import version as version
```
anissa111, line 8:
```suggestion
except Exception:
    from importlib_metadata import version as version
```

Again, I think it'd be better to catch as broadly as possible here, since it doesn't really matter why the first import fails.

You're right, it'll likely be an import error, but there's no reason to not go through the same steps if it's another kind of exception.

aaronzedwick: Good point! Thanks for all the comments, the time is greatly appreciated!

uxarray/__init__.py
```diff
@@ -1,2 +1,14 @@
 from .grid import *
 from .helpers import *
+# Sets the version of uxarray currently installed
+# Attempt to import the needed modules
+try:
+    from importlib.metadata import version as version
+except Exception:
+    from importlib_metadata import version as version
```
anissa111, line 8: Okay, last change before I'll be okay with approving.

```suggestion
try:
    from importlib.metadata import version as _version
except Exception:
    from importlib_metadata import version as _version
```

It will be better to specify this import as an underscored method. Since this is no longer in a function, things imported in this addition will be available to import under the uxarray namespace.

Like it is now, uxarray will appear to have a `ux.version()` and a `ux.__version__`. If we change the import to `ux._version`, it will be more obvious that that's not the one intended for public use. 


I installed this on my machine to demonstrate and this is what I see:
``` ipython
>>> import uxarray as ux
>>> ux.version
<function version at 0x112c01750>
>>> ux.__version__
'2023.5.0'
```

aaronzedwick: That is a very good point, I see what you mean.

anissa111 (approved):
Looks good!

We'll need to make sure this works as expected with the released versions of the package, but I currently have no concerns that this won't behave as expected.

anissa111:
Ah, good catch! I would also suggest editing the title and PR description to reflect the changes :)

uxarray/__init__.py
hongyuchen1030: If we catch an Error, we can just print out the error here instead of assigning ""0000 to __version

anissa111: I think I'm missing what benefit that would add. Sure, it might be useful to have the stack trace for whatever went wrong, but this code will be run when the package is imported and raising an exception on import just because of a failure to pull the version from metadata wouldn't be productive because there's nothing the end user could or should do about it.

Assigning a dummy value allows the import to succeed, `ux.__version__` to still exist but with a value that indicates there was an issue with pulling the version, and is fairly standard practice anyway.

philipc2 (approved):
Looks good Aaron!

PR #342, opened 2023-06-09
Multiprecision unique coords detection through ds

### LINKED ISSUE #318: Incorporating GMPY2 for multiple precision calculation
## Overview

This feature introduces multi-precision calculation support in Project UXArray. It offers advantages such as detecting unique coordinates and exact calculation of intersection points, which are crucial in computational geometry. With gmpy2 library, we can achieve higher precision and accuracy in computations, making our project reliable and robust. 

For more information about the multi-precision and exact computation paradigm, please refers to the following materials
[GMPY2](https://gmpy2.readthedocs.io/en/latest/index.html)
[The Exact Computation Paradigm](https://www.cgal.org/exact.html)

## Expected Usage

```Python
import uxarray as ux

# Provide nodes that are extremely closed to each other
face_nodes_connectivity = [['1.000000000000000000', '1.000000000000000000001', '1.0000000000000000000012'],
 ['1.000000000000000001', '1.000000000000000000002', '1.0000000000000000000013']]

uxgrid =   ux.Grid(face_nodes_connectivity, multi_precision=True, precision= 65,
                        vertices=True,
                        islatlon=False,
                        concave=False)
print(uxgrid.nMesh2_face )# It will still detect two faces without degenration
>>> 2

# Perform any provided grid manipulation and then the results is precise up the the 65 bits
```
## Proposed Progress
**New project structure that incorporates the multi-precision mode**
- [ ] Figure out the API for turning on the multi-precision modes #339 

**Data Read in**
- [x] Support multi-precision when initializing through  #`Grid.__from_vert__()`  #321
- [x] #353

**Grid Pre-processing**
- [x] Precise spherical and cartesian coordinates conversion #326 
- [ ] Build exact Latitude Longitude bound
    - [ ] The customized intervaltree #113 

**Grid Operation**
- [ ] Exact intersection between two great circle arcs (GCR)
- [ ] Exact intersection between a GCR and a constant latitude
- [ ] Exact Non-conservative Zonal Average

**Grid Construction**
- [ ] Exact conservative zonal regrinding.

### 1.  **Q:** From 53 to 60 bit precision - is that necessary to use another library for this? 
    
**A:** The Python float type is limited to 53 bits of precision, which means we can't go beyond that using standard libraries. By integrating gmpy2, designed for multi-precision arithmetic, we can surpass the 53-bit limit and achieve higher precision.

While the PR mentions a precision increase to 60 bits as an example, multi-precision support allows for arbitrary precision settings. We can choose any desired value, such as 100 bits or 200 bits, depending on project requirements. The 60-bit example is just illustrative.

Enabling multi-precision support facilitates not only higher precision calculations but also exact computation. The exact computation is crucial in computational geometry to ensure precise and accurate results, especially when rounding errors and numerical instability could impact the outcome.

This feature allows for higher precision calculations and provides the capability for exact computation. Exact computation is an important paradigm in computational geometry as it ensures precise and accurate results, particularly in cases where rounding errors and numerical instability can affect the outcome.

Incorporating multi-precision support isn't just about adding another package. It means embracing the exact computation paradigm, providing robust analysis for both the geoscience and computational geometry communities.

### 2.  **Q:** Can you say a little more about how this will be used in practice? Normally everything in Uxarray will be in default python precision but sometimes a calculation must be done in increased precision. How do we decide and what has to happen to "convert" to using this library?

**A:**  So by default, our UXArray project uses the python `float` for all calculations. But if users want to achieve the arbitrary-precision, they can set the `multi-precision` flag as `True` and the desired precision when initializing the Grid object. Then all operations in the UXarray will be done with the desired precision.

```python

# The default floating point Grid object
float_grid = ux.Grid(verts) 

# The multiplrecision Grid object that's set to the 100 precision
multi_precision_grid = ux.Grid(verts, multiprecision=True, precision=100)
```

In other words, we never mix the usage of multi-precision and the python `float`, if the grid object is set to default `float` operation, all the operations will be in floating points. If the grid object is set to multi-precision, then all operations will be in arbitrary-precision. And we can always round the multi-precision number back to the `float` precision with different rounding options

With this feature, we can achieve

- The unique coordinate determination: We can distinguish two coordinates even though they're extremely close to each other
- The exact intersection points between GCRs or GCR and constant latitude
- We can also achieve exact computations of conservative zonal re-gridding

### 3.  **Q:** Other than initializing a grid from vertices, what other applications would there be for this? Is there good references/papers about this ideas?

**A:**   The application of the exact computation of the [https://www.cgal.org/exact.html](https://www.cgal.org/exact.html) will be not only useful in determining unique coordinates but also in
    
- Determining is a point really on that GCR? (What if it just 1.0e-20 magnitude away from the GCR?). A single point doesn't seems important here, but what if there'are 200000 points that each one of it is 1.0e-20 from the last one? The error will accumulate

When dealing with large datasets, even small errors can accumulate and lead to significant discrepancies. Using exact computation can help ensure that each point is accurately placed on the GCR.

- Get the exact intersection point between two GCRs: Are you sure such an intersection point is exactly on both GCRs. not just 1.0e-20 away from the 1st GCR? And if I calculate this point from the 1st GCR, it will yield the exact same result as when I calculate it from the 2nd GCR?

Without exact computation, it's possible that the intersection point may be off by a small margin, which can cause issues in certain applications. By using exact computation, you can be sure that the intersection point is exactly on both GCRs and that the same result will be obtained regardless of which GCR is used to calculate it.

If you're interested, you can also refer to the following:
- [Adaptive Precision Floating-Point Arithmetic
and Fast Robust Geometric Predicates](https://people.eecs.berkeley.edu/~jrs/papers/robustr.pdf)
- [Precision and robustness in geometric computations](https://link.springer.com/chapter/10.1007/3-540-63818-0_9)

### 4. **Q:** What happens if you want to do an operation between two grids, like an intersection, and one is default precision and one is arbitrary? As for the performance. Do operations on a grid declared with arbitrary precision take a lot more time?

**A:**  I have tested the performance difference on the coordinates conversion so far, and here're the results summary:

Using the `gmpy2` library, precision can be increased by 2 to 3 orders of magnitude compared to the standard `float64` type. This significant improvement in precision comes at the cost of only 1 to 2 orders of magnitude increase in runtime.

aaronzedwick:
Closes #318 

## Overview

 Changes to __from_ds function to include multi-precision
## Expected Usage

```Python
import uxarray as ux

# Provide nodes that are extremely closed to each other
face_nodes_connectivity = [['1.000000000000000000', '1.000000000000000000001', '1.0000000000000000000012'],
 ['1.000000000000000001', '1.000000000000000000002', '1.0000000000000000000013']]

uxgrid =   ux.Grid(face_nodes_connectivity, multi_precision=True, precision= 65,
                        vertices=True,
                        islatlon=False,
                        concave=False)
print(uxgrid.nMesh2_face )# It will still detect two faces without degenration
>>> 2

# Perform any provided grid manipulation and then the results is precise up the the 65 bits
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**


hongyuchen1030 (commented):
Please see comments above

uxarray/grid.py
```diff
@@ -277,8 +397,31 @@ def __from_ds__(self, dataset):
         else:
             raise RuntimeError("unknown mesh type")
 
+        # set precision flags
+        self._multi_precision = multi_precision
+        self._precision = precision
+
+        # loops through the dataset
+        if self._multi_precision:
+            for varName, varData in dataset.data_vars.items():
```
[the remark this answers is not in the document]
aaronzedwick: Thanks for this Hongyu! I am still not sure why we wouldn't want to change everything in self.ds to multi-precision though. Are there some datasets inside there that can't be converted? Or do we just not use anything other than the node_y and node_x in calculations, so doing so would be a waste of computational power?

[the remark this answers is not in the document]
hongyuchen1030: 1. We only want the multi-precision for floating point numbers. (Since our integer usage is not crazily big, we don't need the multi-precision integer so far)
2. the `self._ds` store many different data, I strongly recommend you take a closer look into this dataset
3. The helper function `convert_to_multi_precision()` function has a limited usage range, please read its docstring for more information

uxarray/grid.py
```diff
@@ -401,22 +404,19 @@ def __from_ds__(self, dataset, multi_precision=False, precision=FLOAT_PRECISION_
...
-                    convertedData = convert_to_multiprecision(
-                        varData.values, str_mode=False, precision=self._precision)
-                    self.ds[varName] = (varData.dims, convertedData)
-                elif np.issubdtype(varData.dtype, np.str):
-                    convertedData = convert_to_multiprecision(
-                        varData.values, str_mode=True, precision=self._precision)
-                    self.ds[varName] = (varData.dims, convertedData)
+            if "Mesh2_node_y" in self.ds and "Mesh2_node_x" in self.ds:
+                y_coords = self.ds["Mesh2_node_y"].values
+                x_coords = self.ds["Mesh2_node_x"].values
```
hongyuchen1030, line 411: Directly cast the `self.ds["Mesh2_node_x"].values` into `convert_to_multiprecision` instead of making another copy

uxarray/grid.py
```diff
@@ -401,22 +404,19 @@ def __from_ds__(self, dataset, multi_precision=False, precision=FLOAT_PRECISION_
...
+
+                converted_y_coords = convert_to_multiprecision(
+                    y_coords, str_mode=False, precision=self._precision)
+                converted_x_coords = convert_to_multiprecision(
+                    x_coords, str_mode=False, precision=self._precision)
+
+                self.ds["Mesh2_node_y"] = converted_y_coords
+                self.ds["Mesh2_node_x"] = converted_x_coords
```
hongyuchen1030, line 419: What if the `self.islatlon == False`? Indicating that the input coordinates is 3D cartesian coordinates?

hongyuchen1030 (commented):
Looks good overall. Thank you for your work. Add some comments about the implementation.

Also, please refer to the PR template provided at the PR creation page, and convert the PR to `ready for review` when you want to be reviewed

uxarray/grid.py
rajeeja: That would be better, Hongyu can review and guide the changes then.

philipc2: @aaronzedwick @rajeeja 

I have changed the base to the `multi_precision` branch now.

test/test_grid.py
```diff
@@ -761,27 +850,55 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+
+class TestMeshLoader(TestCase):
+
+    def test_from_ds(self):
+        exodus_filepath = current_path / "meshfiles" / "exodus" / "outCSne8" / "outCSne8.g"
+
+        xrds_exodus = xr.open_dataset(exodus_filepath)
+
+        # Create the grid with multi-precision off
+        gridOff = ux.Grid(xrds_exodus)
```
hongyuchen1030, line 887: The `ux.Grid(xrds_exodus)` should call the `Grid.__from_ds__()` internally.

Also, please put this initializing from ds testing inside the previous `class testGrid` and when you are testing for the  multi-precision, name your function similar to `def test_from_ds_multiprecision`

test/test_grid.py
```diff
@@ -761,27 +850,55 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+        # Create the grid with multi-precision on
+        gridOn = ux.Grid(xrds_exodus)
+        resultsPrecisionOn = gridOn.__from_ds__(xrds_exodus,
+                                                multi_precision=True,
+                                                precision=128)
+
+        # Check that the nodes are no longer equal
+        firstNodePrecisionOn = resultsPrecisionOn["Mesh2_node_y"].values[0]
+        firstNodePrecisionOff = resultsPrecisionOff["Mesh2_node_y"].values[0]
+        self.assertNotEqual(firstNodePrecisionOn, firstNodePrecisionOff)
```
hongyuchen1030, line 900: You cannot directly compare two multiplrecision values using the `sefl.assertEqual` or other similar functions. Since they're all designed for the floating point number.

Use the `gmpy2.cmp(x,y)` to compare the values. Specifically, if you have two values, A is floating point, B is the `gmpy2.mpfr`, do the following:
Cast the the A into the mpfr: `mpfr(str(A))`
Then call the comparator: `assertTrue(gmpy2.cmp(A, B) != 0)` 

For more information, you can refer to the gmpy2 documentation

hongyuchen1030 (commented):
Only some minor changes require. Also, try addressing the read-doc failing here. After that, this branch can be merged

Please remember to re-request review after you address changes and resolve conversation. (You can also read through the PR template for more guidance)

test/test_grid.py
aaronzedwick: I am not sure what you mean when you say it should call it internally. Isn't that what I am doing in line 888?

hongyuchen1030: Since `ux.Grid(xrds_exodus)` should  already call the `Grid.__from_ds__()` internally. The line 888 is a duplicate call

philipc2:
New branch created #372

PR #354, opened 2023-06-22
DRAFT: Handling Antimeridian Faces

philipc2:
:construction: :construction: :construction: :construction:
_**very much still a draft, please come back soon**_ 
:construction: :construction: :construction: :construction:

Closes #XXX

## Expected Usage

```Python
import uxarray as ux

```

## PR Checklist

**General**
- [ ] An issue is linked created and linked
- [ ] Add appropriate labels

**Testing**

**Documentation**

**Examples**

### DIFF
--- a/test/test_grid.py
+++ b/test/test_grid.py
@@ -785,3 +785,22 @@ class TestConnectivity(TestCase):
         self.assertTrue(
             np.array_equal(res_face_nodes_connectivity,
                            uds._ds["Mesh2_face_nodes"].values))
+
+    def test_build_antimeridian_faces(self):
+        uxgrid = ux.open_grid(self.mpas_filepath)
+
+        faces, edges = uxgrid._build_antimeridian_faces()
+
+        faces = np.array(faces)
+
+        n_nodes_per_face = uxgrid.nNodes_per_face.values[faces]
+
+        x = uxgrid.Mesh2_node_x.values
+        face_nodes = uxgrid.Mesh2_face_nodes.values[faces]
+
+        for face_node, n_node, idx in zip(face_nodes, n_nodes_per_face, faces):
+            edge = np.array(edges[idx])
+            x_face = x[face_node[:n_node]]
+            continue
+
+        pass
--- a/uxarray/core/grid.py
+++ b/uxarray/core/grid.py
@@ -365,6 +365,12 @@ class Grid:
         """
         return self._ds["nNodes_per_face"]
 
+    @property
+    def nEdges_per_face(self):
+        if "nEdges_per_face" not in self._ds:
+            self._build_nEdges_per_face()
+        return self._ds['nEdges_per_face']
+
     # coordinate properties
 
     @property
@@ -991,3 +997,59 @@ class Grid:
             data=nNodes_per_face,
             dims=["nMesh2_face"],
             attrs={"long_name": "number of non-fill value nodes for each face"})
+
+    def _build_nEdges_per_face(self):
+
+        closed = np.ones((self.nMesh2_face, self.nMaxMesh2_face_edges + 1),
+                         dtype=INT_DTYPE) * INT_FILL_VALUE
+
+        closed[:, :-1] = self.Mesh2_face_edges.copy()
+
+        nEdges_per_face = np.argmax(closed == INT_FILL_VALUE, axis=1)
+
+        self._ds["nEdges_per_face"] = xr.DataArray(
+            data=nEdges_per_face,
+            dims=["nMesh2_face"],
+            attrs={"long_name": "number of non-fill value edges for each face"})
+
+    def _build_antimeridian_faces(self, bias="left"):
+
+        node_x = self.Mesh2_node_x.values
+        node_y = self.Mesh2_node_y.values
+
+        face_edges = self.Mesh2_face_edges.values
+
+        edge_nodes = self.Mesh2_edge_nodes.values
+
+        n_edges_per_face = self.nEdges_per_face.values
+
+        antimeridian_face_edges = []
+        antimeridian_faces = []
+
+        for cur_face_idx, (edge_indices, n_edges) in enumerate(
+                zip(face_edges, n_edges_per_face)):
+
+            edges = edge_nodes[edge_indices[:n_edges]]
+            edge_state = []
+
+            for cur_edge_idx, edge in enumerate(edges):
+                v0 = node_x[edge[0]]
+                v1 = node_x[edge[1]]
+
+                if abs(v0 - v1) >= 180:
+
+                    # v0 ---> v1
+                    if v0 <= 180.0 <= v1:
+                        edge_state.append(1)
+
+                    # v0 <--- v1
+                    elif v1 <= 180 <= v0:
+                        edge_state.append(-1)
+
+                else:
+                    edge_state.append(0)
+            if np.any(edge_state):
+                antimeridian_faces.append(cur_face_idx)
+            antimeridian_face_edges.append(edge_state)
+
+        return antimeridian_faces, antimeridian_face_edges


philipc2:
@ifranda 

Can you include a few examples of visualizations you were able to achieve using this PR? I've already put a few examples in the description of this PR, but a few more would be good to include, especially examples using PolyCollection and visualizations with data.

ifranda:
> @ifranda
> 
> Can you include a few examples of visualizations you were able to achieve using this PR? I've already put a few examples in the description of this PR, but a few more would be good to include, especially examples using PolyCollection and visualizations with data.

Definitely!

ifranda:
Let me know if there is anything else that would be helpful

philipc2:
> Let me know if there is anything else that would be helpful

The ones that you added are great! I'll request you as a reviewer, so hopefully we can catch anything that needs to be fixed then!

uxarray/core/dataarray.py
```diff
@@ -97,3 +96,94 @@ def uxgrid(self):
...
+                f"for face-centered data.")
+
+        # face-centered data
+        if self.data.size == self.uxgrid.nMesh2_face:
+            poly_collection = self.uxgrid.to_polycollection(
+                override=override_grid, cache=cache_grid)
+
+            # map data with antimeridian polygons
+            if self.uxgrid.corrected_polygon_shells is not None:
+                data = self.data[self.uxgrid.original_to_corrected_indices]
```
ifranda, line 170:
```suggestion
                data = self.data[self.uxgrid.corrected_shells_to_original_faces]
```
I think this just needs to be updated to reflect the name change

ifranda (changes requested):
Besides the one change I think things are looking good!

ifranda (approved):

rajeeja:
I see this PR has cool pictures!! kudos on the good work - @ifranda and @philipc2 !

Possible to add an example notebook? -  potentially something with geographic national showing some the grid and some temp/etc. data.

minor note: It'd be good to have some more documentation about the introduction of geometry.py (maybe on top of that file). Geometry means very different things in different context.

philipc2:
> Possible to add an example notebook? - potentially something with geographic national showing some the grid and some temp/etc. data.

This is currently in the works as a separate cookbook, not directly included in our usage examples (link [here](https://uxarray.github.io/unstructured-grid-viz-cookbook/notebooks/notebook-template.html)). However a usage example notebook under our documentation detailing the algorithm and implementation would be good. 

> minor note: It'd be good to have some more documentation about the introduction of geometry.py (maybe on top of that file). Geometry means very different things in different context.

I agree, I chose the name `geometry` since we were using `shapely` to convert our unstructured grid into polygons, however I'm open to other names for the module.

rajeeja:
> I agree, I chose the name `geometry` since we were using `shapely` to convert our unstructured grid into polygons, however I'm open to other names for the module. 

What else would this be useful for? any other geometry type operations or purely for viz purposes only? face transformations for visualisation or 2d element utils would be better.

rajeeja (approved):

philipc2:
> > I agree, I chose the name `geometry` since we were using `shapely` to convert our unstructured grid into polygons, however I'm open to other names for the module.
> 
> What else would this be useful for? any other geometry type operations or purely for viz purposes only? face transformations for visualisation or 2d element utils would be better.

I can see it being used to store functionality related to geometric data structures, such as KDTrees, so not necessarily visualization-only functionality.

aaronzedwick (approved):
This look great! Nice work

uxarray/grid/coordinates.py
```diff
@@ -286,3 +286,50 @@ def _populate_lonlat_coord(grid):
             "long_name": "latitude of mesh nodes",
             "units": "degrees_north",
         })
+
+
+def grid_center_lat_lon(ds):
```
erogluorhan, line 291: I didn't understand why this is a user API if it is a SCRIP-specific function. Moreover, it seems like this was already here in the main branch, I don't know why the PR shows this as an added function. Any ideas?

uxarray/grid/geometry.py
```diff
@@ -0,0 +1,199 @@
+import numpy as np
+
+from uxarray.constants import INT_DTYPE, INT_FILL_VALUE
+from uxarray.grid.connectivity import close_face_nodes
+
+
+def grid_to_polygons(grid, correct_antimeridian_polygons=True):
```
erogluorhan, line 7: As mentioned before, I don't think this should be a user API function

erogluorhan (commented):
Thanks for this significant work! Please see my inline comments

uxarray/core/dataarray.py
```diff
@@ -97,3 +96,112 @@ def uxgrid(self):
...
+                        correct_antimeridian_polygons=True):
+        """Constructs a ``spatialpandas.GeoDataFrame`` with a "geometry"
+        column, containing a collection of Shapely Polygons or MultiPolygons
+        representing the geometry of the unstructured grid, and a data column
+        representing a 1D slice of data mapped to each Polygon.
+
+        Parameters
+        override_grid : bool
+            Flag to recompute the ``GeoDataFrame`` if one is already cached
+        cache_grid : bool
```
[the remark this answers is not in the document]
philipc2: I'll update the docstring to better describe what's happening.

`cache_grid` determines whether the geometry (i.e. GeoDataFrame or PolyCollection) that is construct rued should be re-used for subsequent calls. For example, if the user constructs as `GeoDataFrame` with one data variable and then would like to use the same grid to map another data variable, `cache_grid` is set to True by default to avoid uncessary re-runs of `Grid.to_geodataframe` or `Grid.to_polycollection`

`override_grid` is used to determine whether the user wants to re-compute the grid instead of using a cached version.

uxarray/grid/coordinates.py
philipc2: I'll fix this, it must of gotten moved around.

uxarray/grid/geometry.py
philipc2: The reason I chose to keep this as a user API function is that it returns an array of shapely polygons, which can then be manipulated by the user if they have any workflows that require the unstructured grid to be represented in Shapely.

uxarray/__init__.py
```diff
@@ -1,6 +1,8 @@
-# from uxarray.utils.helpers import *
+from uxarray.grid.grid import Grid
+
 from uxarray.constants import (INT_DTYPE, INT_FILL_VALUE)
 from uxarray.grid import Grid
```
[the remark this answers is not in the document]
philipc2: `from uxarray.grid import Grid`

This is what I was aiming for, I'll go through and consistently use this import.

test/test_dataarray.py
```diff
@@ -0,0 +1,17 @@
...
+current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+
+gridfile_ne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30.ug"
+dsfile_var2_ne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_vortex.nc"
+
+dsfiles_mf_ne30 = str(
+    current_path) + "/meshfiles/ugrid/outCSne30/outCSne30_*.nc"
+
+gridfile_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "grid.nc"
+dsfile_v1_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "v1.nc"
```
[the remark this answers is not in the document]
philipc2: I'll write a few DataArray specific test cases!

docs/user_api/index.rst
```diff
@@ -131,6 +140,13 @@ Attributes
    Grid.Mesh2_face_nodes
    Grid.Mesh2_edge_nodes
    Grid.Mesh2_face_edges
+   Grid.polygon_shells
+   Grid.corrected_polygon_shells
+   Grid.corrected_shells_to_original_faces
+   Grid.original_to_corrected_indices
+   Grid.antimeridian_face_indices
```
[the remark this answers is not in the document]
philipc2: I'd like to keep `polygon_shells` and `corrected_polygon_shells` as part of the user API, since accessing these could be used for user's workflows that require representing the unstructured grid in this format, and not a GeoDataFrame or PolyCollection.

`Grid.antimeridian_face_indices` this similarly should stay as part of the User API, since someone may be interested in knowing which indices into Face Nodes are considered to cross or touch the antemeridian.

`Grid.corrected_shells_to_original_faces` and `Grid.original_to_corrected_indices` we can move to the Internal API, however how would you suggest having internal properties?

uxarray/grid/geometry.py
philipc2: Consider a `Grid.to_shapely_polygons` conversion method instead of directly exposing this function to the user.

uxarray/core/dataarray.py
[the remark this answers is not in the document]
erogluorhan: Thanks! Actually the "_grid" in each variable can get confusing; we are not caching or overriding the actual `grid` object but the geodataframe or polycollection that is being generated from that object. We should consider using a better name

erogluorhan (commented):
Thanks for resolving a lot of things! Please see my second review's comments below

philipc2:
> Thanks! Actually the "_grid" in each variable can get confusing; we are not caching or overriding the actual grid object but the geodataframe or polycollection that is being generated from that object. We should consider using a better name

I can rename it to `cache_geometry` instead.

erogluorhan:
> I can rename it to `cache_geometry` instead.

Would it to be confusing to use only "cache" and "override" instead? Or maybe use one of "cache_result", or "cache_output", or "cache_gdf" and "cache_polycollection"?

erogluorhan:
> 



> > Also, decorating several functions with property would eventually lead to a cluttered grid view in IDEs or Jupyter notebooks, which may not be ideal.
> 
> The purpose of decorating functions with properties is to allow for our `Grid` class to provide a better way for the user to access attributes of interest, as was the case with all of our UGRID variables (node_x, face_nodes, etc). I do see the concern in using the properties to construct the GeoDataFrame/PolyCollection, since they currently will keep the polygon shells and other variables in memory.
> 
> I need to look into it, however the memory footprint of these properties should only become an issue if there has been a call to them, since the internal attribute that is accessed by the property is initialized to None.
> 
> ```python
>   @property
>   def antimeridian_face_indices(self):
>       """Index of each face that crosses the antimeridian."""
>       if self._antimeridian_face_indices is None:
>           self._antimeridian_face_indices = _build_antimeridian_face_indices(
>               self)
>       return self._antimeridian_face_indices
> ```
> 
> > So, we could design them as helper functions that can only be called if/when the user actually needs them.
> 
> I already have these set up as helper functions, as seen above in the example. I can modify the construction code for the GeoDataFrame and PolyCollection to use the helper functions directly instead of the properties.
> 
> I'll remove these as properties for now and we can handle exposing these to the user at a later date.

These are all good points, and we should talk about all of these during our later discussion on this. For now, I think we should clean the following leftover from grid.py:

```
self._polygon_shells = None
self._corrected_polygon_shells = None
self.__corrected_shells_to_original_faces = None
```

and maybe other leftovers after these last minutes changes?

philipc2:
> > I can rename it to `cache_geometry` instead.
> 
> Would it to be confusing to use only "cache" and "override" instead? Or maybe use one of "cache_result", or "cache_output", or "cache_gdf" and "cache_polycollection"?

The caching happens at the `Grid` level, not at the DataArray level, so they are stored under the `Grid` object.

philipc2:
> > 
> 
> > > Also, decorating several functions with property would eventually lead to a cluttered grid view in IDEs or Jupyter notebooks, which may not be ideal.
> > 
> > 
> > The purpose of decorating functions with properties is to allow for our `Grid` class to provide a better way for the user to access attributes of interest, as was the case with all of our UGRID variables (node_x, face_nodes, etc). I do see the concern in using the properties to construct the GeoDataFrame/PolyCollection, since they currently will keep the polygon shells and other variables in memory.
> > I need to look into it, however the memory footprint of these properties should only become an issue if there has been a call to them, since the internal attribute that is accessed by the property is initialized to None.
> > ```python
> >   @property
> >   def antimeridian_face_indices(self):
> >       """Index of each face that crosses the antimeridian."""
> >       if self._antimeridian_face_indices is None:
> >           self._antimeridian_face_indices = _build_antimeridian_face_indices(
> >               self)
> >       return self._antimeridian_face_indices
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > So, we could design them as helper functions that can only be called if/when the user actually needs them.
> > 
> > 
> > I already have these set up as helper functions, as seen above in the example. I can modify the construction code for the GeoDataFrame and PolyCollection to use the helper functions directly instead of the properties.
> > I'll remove these as properties for now and we can handle exposing these to the user at a later date.
> 
> These are all good points, and we should talk about all of these during our later discussion on this. For now, I think we should clean the following leftover from grid.py:
> 
> ```
> self._polygon_shells = None
> self._corrected_polygon_shells = None
> self.__corrected_shells_to_original_faces = None
> ```
> 
> and maybe other leftovers after these last minutes changes?

Good catch on the leftover attributes.

uxarray/core/dataarray.py
```diff
@@ -97,3 +96,113 @@ def uxgrid(self):
     @uxgrid.setter
     def uxgrid(self, ugrid_obj):
         self._uxgrid = ugrid_obj
+
+    def to_geodataframe(self,
+                        override_geometry=False,
+                        cache_geometry=True,
```
erogluorhan, line 102: I read your latest comment on the naming of these, and it makes sense to me; however, I think the description of these vars below is well written and self-explanatory, so I still think naming these as `cache` and `override` just like you did in `grid.to_geodataframe()` and `grid.to_polycollection` should be fine.

erogluorhan (approved):
Great work! Approved but please see one inline comment below. Regardless of your response to it, feel free to merge it.

PR #369, opened 2023-07-07
Fix Upstream CI 

### LINKED ISSUE #322: Upstream Dev CI Failing
Tests are failing due to `xarray` not being installed / recognized

philipc2:
Closes #322

## Overview

* Removes `python` as a dependency from `upstream-dev-ci.yml`, which was causing Python to update from `3.10` to the latest version `3.11`, causing the test to fail
* Successful run from this branch can be seen [here](https://github.com/UXARRAY/uxarray/actions/runs/5490666153)


philipc2:
@erogluorhan 

Cross-posting this from https://github.com/NCAR/geocat-comp/issues/435, want to hear your opinion on which solution would be best. 

> It appears that having Python listed explicitly as a dependency in our upstream overrides the version that is set when configuring the environment (i.e. Environment is initialized as Python 3.10, however installing the upstream dependencies bumps it to 3.11).
> 
> After removing it as a dependency from the upstream environment, it executes properly.
> 
> Additionally, instead of the above solution, we could set the test to run on Python 3.11 instead of 3.10 while keeping Python as a dependency, but personally I don't see why it's necessary to keep Python as a dependency for upstream testing.

PR #371, opened 2023-07-11
Numba JIT and Cache Attribute

### LINKED ISSUE #367: JIT Compiler Caching Attribute
**Overview**
JIT isn't working as it should in PyCharm, causing increased load times after the first run of code. JIT should add a few extra seconds the first time it runs, and after that, it should be extremely quick. This isn't happening. This is simply because we aren't caching the machine code properly. What we have is this:
```Python
@njit
def randomFunction()
```

To fix the problem of caching we need to add:
```Python
@njit(cache=True)
def randomFunction()
```
However instead of just setting cache=True on all njit cases, instead a better solution is to expose this option to the user, creating an attribute that allows users to turn caching on and off. Refer to this notebook for more information on this caching problem: https://github.com/rajeeja/python-performance/blob/main/JIT_Speedup_Example.ipynb

aaronzedwick:
Closes #367 

## Overview

Adds attribute that allows users to turn Numba JIT caching on and off, as well as allow them to turn Numba JIT on/off. The default is both are on.

## Expected Usage

```Python
# To turn JIT on
uxarray.enable_jit(True)
# To turn caching on
uxarray.enable_jit_cache(True)

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Documentation**


uxarray/utils/jit_attributes.py
```diff
@@ -0,0 +1,2 @@
+cache = True
+jit_bool = False
```
rajeeja, line 2: I think we don't need a new file, this can go into constants.py

rajeeja (changes requested):

uxarray/utils/jit_attributes.py
rajeeja: Just a note: Our use case here was to study performance, in regular usage one might never turn off caching as it will only help if we frequently modifying the code or have very limited disk space.

aaronzedwick: Do you want me to remove the cache=True attribute then?

rajeeja: > Do you want me to remove the cache=True attribute then?

No, let it be, just the above in a comment and a more detailed comment near the definition

uxarray/core/api.py
```diff
@@ -261,3 +262,27 @@ def open_mfdataset(grid_filename_or_obj: str,
...
+
+
+def enable_jit(enable_jit):
+    """Allows Numba's JIT application to be turned on and off.
+
+    This lets users choose whether they want machine code to be
+    generated to speed up the performance of the code on large files.
+    Default is on
+    """
+    uxarray.utils.constants.ENABLE_JIT = enable_jit
```
philipc2, line 288: I'd consider a slight design change to these functions, since we'd like to avoid needing to pass through a parameter since there are no protections in place (user could pass through anything). The following would be a safer approach. 

```Python
def enable_jit_cache():
    uxarray.utils.constants.ENABLE_JIT_CACHE = True
```
```Python
def disable_jit_cache():
    uxarray.utils.constants.ENABLE_JIT_CACHE = False
```

```Python
def enable_jit():
    uxarray.utils.constants.ENABLE_JIT = True
```

```Python
def disable_jit():
    uxarray.utils.constants.ENABLE_JIT = False
```

uxarray/utils/constants.py
```diff
@@ -3,3 +3,5 @@
 # numpy indexing code is written for np.intp
 INT_DTYPE = np.intp
 INT_FILL_VALUE = np.iinfo(INT_DTYPE).min
+ENABLE_JIT_CACHE = True
+ENABLE_JIT = False
```
philipc2, line 7: Should this instead directly modify:

```Python
config.DISABLE_JIT
```

uxarray/core/api.py
aaronzedwick: That is a good idea

uxarray/utils/constants.py
aaronzedwick: Isn't that what we are doing in the helpers.py and quadaturesDG.py files? We can't do it inside the constants file unless we want to import numba into that file which doesn't make a whole lot of sense.

philipc2: > unless we want to import numba into that file

Didn't consider this, the current implement is good then!

rajeeja (approved):

uxarray/utils/get_quadratureDG.py
```diff
@@ -1,10 +1,11 @@
 import numpy as np
 from numba import njit, config
+from uxarray.utils.constants import ENABLE_JIT_CACHE, ENABLE_JIT
 
-config.DISABLE_JIT = False
+config.DISABLE_JIT = ENABLE_JIT
```
philipc2, line 5: Is there a specific reason why you have the constant named `ENABLE_JIT`, I'd suggest a name like `USE_JIT` for our internal use, since currently `ENABLE_JIT` "enables JIT" when it's set to False.

philipc2 (changes requested):

philipc2 (approved):
Nice work! This is a great addition.

PR #372, opened 2023-07-11
Multiprecision unique coords detection to `__from_ds__`

### LINKED ISSUE #318: Incorporating GMPY2 for multiple precision calculation

philipc2:
Closes #318 

## Overview

 Changes to __from_ds function to include multi-precision
## Expected Usage

```Python
import uxarray as ux

# Provide nodes that are extremely closed to each other
face_nodes_connectivity = [['1.000000000000000000', '1.000000000000000000001', '1.0000000000000000000012'],
 ['1.000000000000000001', '1.000000000000000000002', '1.0000000000000000000013']]

uxgrid =   ux.Grid(face_nodes_connectivity, multi_precision=True, precision= 65,
                        vertices=True,
                        islatlon=False,
                        concave=False)
print(uxgrid.nMesh2_face )# It will still detect two faces without degenration
>>> 2

# Perform any provided grid manipulation and then the results is precise up the the 65 bits
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**


aaronzedwick:
@hongyuchen1030 Could we get this merged if everything looks good?

hongyuchen1030 (approved):
Good Work!

Issue #288, opened 2023-04-20
Update encode_as to switch coordinates


rajeeja:
Requirement: cartesian data to latlon and vice-versa for a grid file.

The current function `encode_as` has an output grid type as the only argument. 
Can we add another argument that can change the grid from cartesian to spherical?

Current implementation:

```
    def encode_as(self, grid_type):
        """Encodes the grid as a new `xarray.Dataset` per grid format supplied
        in the `grid_type` argument.

        Parameters
        ----------
        grid_type : str, required
            Grid type of output dataset.
            Currently supported options are "ugrid", "exodus", and "scrip"

        Returns
        -------
        out_ds : xarray.Dataset
            The output `xarray.Dataset` that is encoded from the this grid.
```

New option:
```
def encode_as(self, grid_type, coord_type):
coord_type by default is the same as the original grid

if specified as something else, we can transform coordinates internally.
```

This would use helper functions such as:

```
node_xyz_to_lonlat_rad
```

erogluorhan:
Just note, `encode_as()` will be deprecated, but `to_xaray()` is the replacement

PR #373, opened 2023-07-11
Exodus loading and units fix

### LINKED ISSUE #360: Cartesian Coordinates Improperly Labeled as Lat/Lon for Exodus Grids
## Overview
Cartesian Coordinates parsed from Exodus grids that are incorrectly labeled with units "degrees", meaning they are interpreted as lat/lon.

## MCVE

Dataset Used: Exodus outCSne8

```Python
# Base data path
base_path = "../../test/meshfiles/exodus/outCSne8/"

# Path to Grid file
grid_path = base_path + "outCSne8.g"

```

```Python
uxgrid.Mesh2_node_x.units
```
image

```Python
uxgrid.Mesh2_node_x.values.min(), uxgrid.Mesh2_node_x.values.max()
```

image

rajeeja:
Closes #360 

## Overview

```Python
Upon loading the mesh convert the cartesian (Assumption: default for every exodus file that is loaded). Our reader converts populates Mesh2_node_x with spherical coordinates as per our convention and Mesh2_node_cart_x would contain the originally loaded coordinates.
```

## PR Checklist

**General**
- [ X] An issue is linked created and linked
- [ ] Add appropriate labels

**Testing**
- [X ] Tests cover all possible logical paths in your function


uxarray/core/grid.py
```diff
@@ -233,6 +233,16 @@ def __from_ds__(self, dataset):
...
             self._ds = _read_exodus(dataset, self.grid_var_names)
+
+            # Assume Exodus was read as cartesian grid and that coordinates are not set by reader, call the latlon setter
+            self._populate_lonlat_coord()
+            if len(self._ds.coords) > 2:
+                # set coordinates
+                ds = self._ds.set_coords(
+                    ["Mesh2_node_x", "Mesh2_node_y", "Mesh2_node_z"])
+            else:
+                ds = self._ds.set_coords(["Mesh2_node_x", "Mesh2_node_y"])
```
philipc2, line 244: `Mesh2_node_x`, `Mesh2_node_y` and `Mesh2_node_z` should be in degrees (lat/lon), so there should never be an instance of `Mesh2_node_z`

If you are populating the Cartesian coordinates, use `Mesh2_node_cart_x`, `Mesh2_node_cart_y`, `Mesh2_node_cart_z`

It appears that you have these flipped throughout this PR.

philipc2 (changes requested):

uxarray/core/grid.py
rajeeja: I'm trying to write Cartesian mesh instead of latlon for exodus. Also referenced here: https://github.com/UXARRAY/uxarray/issues/288

philipc2: That's understood, however if you are reading/writing in Cartesian, the coordinates should be `Mesh2_node_cart_x/y/z`, since it's expected for `Mesh2_node_x/y/z` to always be in degrees.

There should also never exist a `Mesh2_node_z`, since our interpretation of the UGRID conventions defines the lat/lon coordinates are on a 2D plane. `Mesh2_node_cart_z` would be the equivalent in Cartesian.

rajeeja: I don't follow, for the failing test: we're loading a ugrid file lat/lon, and writing exodus, do you mean Mesh2_node_z should not exist? My intent is to always write Cartesian coordinates for exodus output file.

uxarray/io/_exodus.py
```diff
@@ -50,58 +50,54 @@ def _read_exodus(ext_ds, grid_var_names):
...
+                                                  "long_name": "cartesian x",
+                                                  "units": "m",
+                                              })
+            ds["Mesh2_node_y"] = xr.DataArray(data=ext_ds.coord[1],
+                                              dims=["nMesh2_node"],
+                                              attrs={
+                                                  "standard_name": "y",
+                                                  "long_name": "cartesian y",
+                                                  "units": "m",
+                                              })
```
philipc2, line 66: Relating to the discussion above, what I'm referring to is that we shouldn't use `Mesh2_node_x/y/z` for storing Cartesian coordinates, these should exlucisvely be using for coordinates that are in degrees.

`Mesh2_node_cart_x/y/z` should be used when reading/writing in Cartesian coordinates. Pigning @hongyuchen1030 if you have any comments on this also, since you wrote `_populate_lonlat_coord()` and similar functions.

rajeeja: The above lines are corresponding to the read part, we store the cartesian in `Mesh2_node_x/y/z` and later call `_populate_lonlat_coord()` [after read call coordinate fix](https://github.com/UXARRAY/uxarray/pull/373/files#diff-0278cc295f397962b7270dffc8271f5b39394d97520ac04b5632e1fec2204656R235)

philipc2: I see now, the implementation of `_populate_lonlat_coord()` then seems a bit hacky then, since it's purpose here is essentially "correcting" the improper parsing. We should strongly consider the following approach with some alterations to `_populate_lonlat_coord`: 

At the grid reading step (all grid formats, not just Exodus), we should identify whether the coordinates are in degrees or Cartesiain, and then populate the respective `Mesh2_node_x/y` or `Mesh2_node_cart_x/y/z` instead of assigning invalid coordinates to `Mesh2_node_x/y/z` and then correcting them with `_populate_lonlat_coord`. 

`Mesh2_node_x/y`, which are the default coordinates used throughout UXarray, should always be in degrees, specially with the range `([-180, 180], [-90, 90])` to stay consistent. We should never store Cartesiain coordinates in here, even if they will eventually be "corrected" by `_populate_lonlat_coord`

hongyuchen1030: From the beginning of last summer, we already agreed that `Mesh22_node_x/y` should always store the longitude and latitude in degrees. 

But for the range `([-180, 180], [-90, 90])` we haven't had any regulation on our repository yet. My implementation always uses `[0,360)`. I am open to any notation, just keep in mind that the right side of this range should always be opening: we don't want to have two representations for the same degree.

And we want to convert the degree to cartesian as less as possible. We don't want to do casting back and forth repeatly because that might bring impact to the precision as I explained before.

philipc2 (approved):
The implementation here is sound and I'm good with approving these changes. However, it brought to light a few design flaws with our handling of Lat/Lon and Cartesian coordinates at the I/O stage, specifically the "swapping" and conversion of coordinates done through `_populate_lonlat_coord`.

I'll raise and issue for this and we can tackle it in a separate PR. Good work on this!

hongyuchen1030 (commented):
Great works! I just have some confusion about the coordinates storage

uxarray/io/_exodus.py
```diff
@@ -50,58 +50,54 @@ def _read_exodus(ext_ds, grid_var_names):
...
-                attrs={
-                    "standard_name": "latitude",
-                    "long_name": "latitude of mesh nodes",
-                    "units": "degrees_north",
-                })
+            ds["Mesh2_node_x"] = xr.DataArray(data=ext_ds.coord[0],
+                                              dims=["nMesh2_node"],
+                                              attrs={
+                                                  "standard_name": "x",
+                                                  "long_name": "cartesian x",
```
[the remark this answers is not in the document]
philipc2: This will be tackled in #413, since it requires a bit of re-writing of existing functions. However, you are correct.

What's happening right now is that the Exodus reader assumes the coordinates will be Cartesian (x, y, z) and uses `uxarray.grid.coordinates._populate_lonlat_coord` to populate the lat/lon coordinates, however the function is written to expect `Mesh2_node_x` `Mesh2_node_y` and `Mesh2_node_z` to be populated with Cartesian coordinates.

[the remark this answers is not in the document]
hongyuchen1030: Thank you very much for your explanations.

hongyuchen1030 (approved):

PR #377, opened 2023-07-12
o Fixing a bug found in gauss quadrature order 9 weights. 

rajeeja:
Validated with: https://keisan.casio.com/exec/system/1280801905
Too expensive to add a test for it.


philipc2 (approved):
Good catch!

PR #394, opened 2023-07-24
DRAFT: Build_edge_from_nodes method

### LINKED ISSUE #349: Constructing a grid from x and y node coordinates
The simplest form of represented an unstructured grid is through a set of X and Y coordinates, with no added connectivity.

A grid can be constructed and represented in the UGRID conventions in the following ways
* Delaunay Triangulation
* Convex Hull
* Voronoi Diagram 

This can be achieved using [`scipy.spatial`](https://docs.scipy.org/doc/scipy/reference/spatial.html) or similar libraries

aaronzedwick:
Closes #349 

## Overview

Added a function that takes in only nodes and builds the edges using that information. It uses either Delaunay Triangulation or
Voronoi Diagram to achieve this.

## Expected Usage

```Python
import uxarray as ux

exodus = current_path / "meshfiles" / "exodus" / "outCSne8" / "outCSne8.g"

grid= ux.open_grid(exodus)
grid.build_edge_from_nodes(grid_type='delaunay')

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


aaronzedwick:
@philipc2 This is the starting idea I have to implement. It seems to work from what I can see on everything but MPAS files. I am not completely confident on why, but if I remember correctly doesn't MPAS have both Delaunay and Voronoi? I am not sure if that would cause the problem but any thoughts are much appreciated.

philipc2:
> This is the starting idea I have for the implementation

This is right along the lines of the implementation I was considering also. I have a few comments regarding the design of the implementation and a few considerations. I'll add those as a review. 

> I am not completely confident on why, but if I remember correctly doesn't MPAS have both Delaunay and Voronoi? 

It does! By default, constructing an MPAS UXarray `Grid` object represents the Primal (Voronoi) Mesh. Obtaining the Dual (Triangular) mesh is done by passing through `use_dual=True` into the initialization. I believe our usage example showcases how to do this.

aaronzedwick:
Tackling in another PR #442

PR #395, opened 2023-07-25
`Grid` Refactor (Class Methods, UGRID Variable+Dimension Mapping, Consistent Coordinates, Type Hints)

### LINKED ISSUE #363: Use `grid_var_names` everywhere
**Use `grid_var_names` everywhere**
UXARRY uses `grid_var_names` to map mesh topology variable names, but it is not used for all variables currently. For example, `Mesh2_edge_nodes` is not referred to through `grid_var_names`. Not using mapped names becomes an issue when a dataset uses a different topology information variable name from default `Mesh2`.

Use `grid_var_names` for all topology variables.

philipc2:
## Overview

* Rewrite of `Grid` constructor, now has two class methods: `Grid.from_dataset` and `Grid.from_face_verticies` which replace the need for `__from_ds__` and `__from_verts__`
* Elimated the need for `grid_var_names ` internally, replaced with `Grid._ugrid_mapping` that will be populated with the mapping of original variable and dimension names to their UGRID equivalent, used to resolve #363, however implementation will be tackled in a following PR
* LatLon & Cartesian coordinates are now consistently used throughout. There is no longer any places where, for example, `Mesh2_node_x` contains Cartesian X
* Type hints in `Grid` class
* Slight rework of core API functions to reflect `Grid` constructor changes
* Improvements to coordinate conversion helper functions

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


uxarray/core/api.py
```diff
@@ -122,23 +103,14 @@ def open_dataset(grid_filename_or_obj: str,
         stores the actual data set. It is the same ``filename_or_obj`` in
         ``xarray.open_dataset``.
 
-    islatlon : bool, optional
+    ilatlon : bool, optional
```
aaronzedwick, line 106: *latlon

uxarray/grid/grid.py
```diff
@@ -96,211 +93,77 @@ def __init__(self, input_obj, **kwargs):
...
-                    f"one face is passed in.")
-
-        # check if initializing from string
-        # TODO: re-add gridspec initialization when implemented
-        elif isinstance(input_obj, xr.Dataset):
-            self.mesh_type = _parse_grid_type(input_obj)
-            self.__from_ds__(dataset=input_obj)
-        else:
-            raise RuntimeError("Dataset is not a valid input type.")
+    @classmethod
```
aaronzedwick, line 96: Could you explain to me what making this an @classmethod achieves?

philipc2: Sure! The `@classmethod` decorator allows us to have a function that bounded to a Class, not a Object. This allows to make a `Grid.from_dataset` method that can return an initialized version of a `Grid`. It also means that we can call `Grid.from_dataset` before initializing a `Grid`.

uxarray/io/_exodus.py
```diff
@@ -203,30 +236,27 @@ def _encode_exodus(ds, grid_var_names, outfile=None):
...
-        )
+    if "Mesh2_node_cart_x" not in ds:
+        x, y, z = _get_xyz_from_lonlat(ds["Mesh2_node_x"].values,
+                                       ds["Mesh2_node_y"].values)
+        c_data = xr.DataArray([x, y, z])
     else:
         c_data = xr.DataArray([
             ds["Mesh2_node_cart_x"].data.tolist(),
-            ds["Mesh2_node_cart_y"].data.tolist(),
+            ds["Mesh2_node_cart_z"].data.tolist(),
```
philipc2, line 246: Typo

docs/user_api/index.rst
```diff
@@ -127,6 +129,7 @@ Attributes
 .. autosummary::
    :toctree: _autosummary
 
+   Grid.grid_spec
```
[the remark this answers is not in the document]
rajeeja: it is the format now.

uxarray/grid/grid.py
aaronzedwick: I see, gotcha that is useful. Thanks!

test/test_mpas.py
```diff
@@ -110,7 +110,7 @@ def test_set_attrs(self):
         ]
 
         # included attrs: 'sphere_radius', 'mesh_spec' 'on_a_sphere'
-        ds = _read_mpas(self.mpas_xr_ds)
+        ds, _ = _read_mpas(self.mpas_xr_ds)
```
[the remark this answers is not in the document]
philipc2: every read function for our I/O now returns a ugrid data mapping in addition to the dataset, since this was for testing we did not need the ugrid mapping

uxarray/grid/grid.py
```diff
@@ -748,23 +596,26 @@ def compute_face_areas(self, quadrature_rule="triangular", order=4):
...
+
+        if latlon:
+            x = self.Mesh2_node_x.data
+            y = self.Mesh2_node_y.data
+            z = np.zeros(self.nMesh2_node)
+            coords_type = "spherical"  # TODO: should really be called latlon?
+        else:
+            x = self.Mesh2_node_cart_x.data
+            y = self.Mesh2_node_cart_y.data
+            y = self.Mesh2_node_cart_z.data
```
aaronzedwick, line 611: Should this be z = self.Mesh2_node_cart_z.data?

aaronzedwick (commented):
Just a few comments, it looks really well done though!

uxarray/grid/grid.py
```diff
@@ -358,21 +222,21 @@ def __eq__(self, other):
...
-                                other._ds[other.grid_var_names[key]]):
+                        if not self._ds[value].equals(other._ds[key]):
                             return False
         else:
             return False
 
         return True
 
-    def __ne__(self, other):
+    def __ne__(self, other) -> bool:
```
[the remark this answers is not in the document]
philipc2: For typical functions, that's correct. However, `__ne__` in Python represents the `!=` operator. It performs an operator overload allowing us to perform `uxgrid_A != uxgrid_B` for example

Good read [here](https://www.geeksforgeeks.org/operator-overloading-in-python/)

test/test_geometry.py
```diff
@@ -27,10 +27,7 @@ class TestAntimeridian:
     def test_crossing(self):
         verts = [[[-170, 40], [180, 30], [165, 25], [-170, 20]]]
 
-        uxgrid = ux.open_grid(verts,
-                              vertices=True,
-                              islatlon=True,
-                              isconcave=False)
+        uxgrid = ux.open_grid(verts, latlon=True)
```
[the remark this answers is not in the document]
philipc2: There was no use/implementation for it, and when we do implement checks or applications for it, we will most likely go a different route other than passing through an argument like this.

uxarray/grid/grid.py
```diff
@@ -96,211 +93,77 @@ def __init__(self, input_obj, **kwargs):
...
+
+        if grid_spec == "Exodus":
+            grid_ds, ugrid_mapping = _read_exodus(dataset)
+        elif grid_spec == "Scrip":
+            grid_ds, ugrid_mapping = _read_scrip(dataset)
+        elif grid_spec == "UGRID":
+            grid_ds, ugrid_mapping = _read_ugrid(dataset)
+        elif grid_spec == "MPAS":
+            grid_ds, ugrid_mapping = _read_mpas(dataset, use_dual=use_dual)
+        elif grid_spec == "Shapefile":
```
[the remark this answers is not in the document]
philipc2: Agreed, it's mostly a placeholder right now.

uxarray/grid/grid.py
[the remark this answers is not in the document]
aaronzedwick: Oh I didn’t realize that was an overloaded operator, makes sense!

test/test_mpas.py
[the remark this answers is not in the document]
rajeeja: Is this `mapping` test anywhere?

rajeeja (approved):
good work! 

let's merge this for other PR's to go through in this iteration.

test/test_mpas.py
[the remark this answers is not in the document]
philipc2: Currently it's not, however right now it's mostly acting as a placeholder for eventual use. Will write tests when we fully populate it.

erogluorhan:
> good work!
> 
> let's merge this for other PR's to go through in this iteration.

This iteration is almost over, and tomorrow will be a release day for UXarray. I'd like to give this PR a review. We can always get a second release if we want in the same calendar month though.

philipc2:
> > good work!
> > 
> > let's merge this for other PR's to go through in this iteration.
> 
> This iteration is almost over, and tomorrow will be a release day for UXarray. I'd like to give this PR a review. We can always get a second release if we want in the same calendar month though.

That's good with me! There's no rush and I agree that a second, smaller, release is definitely doable.

aaronzedwick (approved):
Looks good to me, I don't have any other comments!

uxarray/core/api.py
```diff
@@ -14,10 +14,7 @@
 
 def open_grid(grid_filename_or_obj: Union[str, Path, xr.DataArray, np.ndarray,
                                           list, tuple],
-              gridspec: Optional[str] = None,
-              vertices: Optional[list] = None,
-              islatlon: Optional[bool] = False,
-              isconcave: Optional[bool] = False,
+              latlon: Optional[bool] = False,
```
erogluorhan, line 17: This newer name looks better; however, similar to another comment in grid.oy, we have already had these self.islatlon in the user API. Renaming it (even though it is a small thing) is actually not backwards-compatible (we may break the existing user workflows with such simple var name changes). Having a DeprecationWarning for a few version releases into the future before such changes is always a good practice.

uxarray/core/api.py
```diff
@@ -32,23 +29,14 @@ def open_grid(grid_filename_or_obj: Union[str, Path, xr.DataArray, np.ndarray,
...
-        Specify if this grid has concave elements (internal checks for this are possible)
-
-    gridspec: str, optional
-        Specifies gridspec
-
-    vertices: bool, optional
-        Whether to create grid from vertices
-
     source_grid: str, optional
         Path or URL to the source grid file. For diagnostic/reporting purposes only.
```
erogluorhan, line 36: Looks like `source_grid` is no longer used. The same "backwards-compatibility" in the previous comment applies to this one. Furthermore, it was discussed to be important for diagnostics purposes in the past.

uxarray/core/api.py
```diff
@@ -122,23 +103,14 @@ def open_dataset(grid_filename_or_obj: str,
...
-    isconcave: bool, optional
-        Specify if this grid has concave elements (internal checks for this are possible)
-
-    gridspec: str, optional
-        Specifies gridspec
-
-    vertices: bool, optional
-        Whether to create grid from vertices
-
     source_grid: str, optional
```
erogluorhan, line 109: Similar things re `source_grid `

uxarray/core/api.py
```diff
@@ -209,18 +175,9 @@ def open_mfdataset(grid_filename_or_obj: str,
...
         Path or URL to the source grid file. For diagnostic/reporting purposes only.
```
erogluorhan, line 182: Similar things re `source_grid`

test/test_api.py
```diff
@@ -53,10 +53,6 @@ def test_open_dataset(self):
                         constants.NNODES_outCSne30)
         nt.assert_equal(len(uxds_var2_ne30.uxgrid._ds.data_vars),
                         constants.DATAVARS_outCSne30)
-        nt.assert_equal(uxds_var2_ne30.uxgrid.source_grid,
-                        str(self.gridfile_ne30))
-        nt.assert_equal(uxds_var2_ne30.source_datasets,
```
erogluorhan, line 58: Why removing this?

test/test_api.py
```diff
@@ -69,9 +65,6 @@ def test_open_mf_dataset(self):
                         constants.NNODES_outCSne30)
         nt.assert_equal(len(uxds_mf_ne30.uxgrid._ds.data_vars),
                         constants.DATAVARS_outCSne30)
-        nt.assert_equal(uxds_mf_ne30.uxgrid.source_grid,
-                        str(self.gridfile_ne30))
-        nt.assert_equal(uxds_mf_ne30.source_datasets, self.dsfiles_mf_ne30)
```
erogluorhan, line 74: Why removing this?

erogluorhan (commented):
This is great work overall. Please see my inline comments though

uxarray/grid/grid.py
```diff
@@ -96,211 +93,77 @@ def __init__(self, input_obj, **kwargs):
...
+            When reading in MPAS formatted datasets, indicates whether to use the Dual Mesh
+        """
+        if not isinstance(dataset, xr.Dataset):
+            raise ValueError("Input must be an xarray.Dataset")
+
+        # determine grid/mesh specification
+        grid_spec = _parse_grid_type(dataset)
+
+        if grid_spec == "Exodus":
+            grid_ds, ugrid_mapping = _read_exodus(dataset)
```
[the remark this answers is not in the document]
philipc2: Agreed, it might be a better choice to have a `ugrid_dim_mapping` that maps the dimensions (nodes, faces, edges) to the respective name in the source dataset if they differ from the ugrid conventions.

The goal of this was to use it within the Dataset and DataArray to rename any node, edge, face dimension to their UGRID equivalent.

uxarray/grid/grid.py
```diff
@@ -45,26 +46,14 @@ class Grid:
...
-            If specified file not found or recognized
+    grid_ds : xr.Dataset
+        ``xarray.Dataset`` encoded in the UGRID conventions
+
+    grid_spec : str, default="UGRID"
+        Original unstructrued grid format (i.e. UGRID, MPAS, etc.)
 
-    Examples
+    ugrid_mapping : dict, default=None
+        mapping of ugrid dimensions and variables to source dataset's conventions
```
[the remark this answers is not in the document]
philipc2: > Technically, this public constructor can be explicitly called by the user to construct a Grid object (without being called by from_dataset() etc) such as:

This is a valid concern, and it brings up a larger discussion regarding how our design is intended to be used by the user.

The main motivation in re-writing our constructor to use class methods is because in it's current state, our constructor for the `Grid` was a "catch all" approach", which quickly became cluttered and was starting to become difficult to maintain.

Our `ux.open_grid` already acts as a "catch all".

To address your point, yes the user can directly call the `Grid` constructor, however I don't see it as being an issue (however we do need to update the docstring to better reflect the following i'm going to mention)

#### Case 1: General Grid Construction
```Python
uxgrid = ux.open_grid(grid_path, other_arguments)
```
#### Case 2: Specialized Grid Construction
Instead of using the "catch-all" method, the user instead want's to directly construct a `Grid` using one of the provided class methods
```Python
uxgrid = Grid.from_dataset()

uxgrid = Grid.from_face_verticies()

uxgrid = Grid.delaunay()

uxgrid = Grid.from_structured()

# etc
```

#### Case 3: Direct Construction from UGRID
By using the class methods, we essentially have the following flow:
1. Class method is called with source data
2. Class method encodes the source data in the UGRID conventions
3. UGRID dataset and variables are passed into the `Grid` constructor and returned

This means that our default `Grid` constructor always assumes the inputs to be in the UGRID conventions, which is in line with our design. So now, instead of needing to handle every possible input, case, etc. in our default Constructor, specialized class methods can be called from our `ux.open_grid` depending the use case. 

```Python
uxgrid = Grid(ugrid_ds)
```

I do agree that some tweaking and modifications need to be done to the default values in here, but let me know what you think of this workflow.

uxarray/grid/grid.py
```diff
@@ -717,7 +562,10 @@ def calculate_total_face_area(self, quadrature_rule="triangular", order=4):
 
         return np.sum(face_areas)
 
-    def compute_face_areas(self, quadrature_rule="triangular", order=4):
+    def compute_face_areas(self,
+                           quadrature_rule: Optional[str] = "triangular",
+                           order: Optional[int] = 4,
+                           latlon: Optional[bool] = True):
```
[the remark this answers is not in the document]
philipc2: The `latlon` flag here is meant to specify whether we want to compute the face areas in lat/lon space or in cartesiain space.

uxarray/io/_connectivity.py
```diff
@@ -0,0 +1,113 @@
+import xarray as xr
```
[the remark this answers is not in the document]
philipc2: Good suggestion, I wasn't too keen on the name either.

uxarray/core/api.py
philipc2: > Having a DeprecationWarning for a few version releases into the future before such changes is always a good practice.

As per our discussion the other day, I'll look into this and add it in.

uxarray/core/api.py
philipc2: I'll introduce it back, I did not consider this when doing a sweep over the code.

test/test_api.py
philipc2: It was removed due to `source_datasets` not existing, but once it's added back I'll reintroduce this check.

test/test_api.py
philipc2: Same as above.

uxarray/core/api.py
philipc2: > Looks like source_grid is no longer used. The same "backwards-compatibility" issue in the previous comment applies to this one. Furthermore, it was discussed to be important for diagnostics purposes in the past.

I'll add a warning in `ux.open_grid` which explains it's removal.

philipc2:
@erogluorhan 

Should be ready for a second pass through. I believe I've addressed most of the concerns, with the major change being the following:

Each grid now has a `_source_dims_dict`, which provides the following mapping {"Original Dimension Name" : "UGRID Dimension Name"}, instead of storing all the connectivity and coordinates.

This is now used in our API functions when opening a dataset to remap the original dimension names to their UGRID equivalent, if they were parsed. It can be specified using the "remap_dims" param.

Let me know what you think of this design, and this actually addresses #363 directly now

uxarray/grid/grid.py
```diff
@@ -223,19 +238,15 @@ def __eq__(self, other) -> bool:
...
-                            return False
-        else:
+        # TODO: fix failing case where grids are the same, but one has more variables constructed
+        try:
+            if self._ds == other._ds:
+                return True
+            else:
+                return False
+        except:
             return False
```
philipc2, line 248: @erogluorhan 

I'd like to hear any suggestions on how to update our `==` operator after the removal of `grid_var_names`. Any suggestions?

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: Yes, this workflow makes sense to me; all is good. 

> The main motivation in re-writing our constructor to use class methods is because in it's current state, our constructor for the Grid was a "catch all" approach", which quickly became cluttered and was starting to become difficult to maintain.

Great change to our code!

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: Got it

test/test_api.py
erogluorhan: Actually `uxds_var2_ne30.source_datasets` exists.

test/test_api.py
erogluorhan: The same, `uxds_var2_ne30.source_datasets` exists.

erogluorhan:
> Each grid now has a `_source_dims_dict`, which provides the following mapping {"Original Dimension Name" : "UGRID Dimension Name"}, instead of storing all the connectivity and coordinates.

Shouldn't it be {"Our standardized UGRID dim name" : "Original Dimension Name from the source"}. Maybe you already meant that?

uxarray/grid/grid.py
erogluorhan: Great question! I think you are concerned about this scenario, right: We get a UGRID input with custom dim names and because of our standardization of those names into our UGRID dim names, they no longer have the same dim names. 

However, regardless from the removal of `grid_var_names `, I think we would return the same result, i.e. `False`, for this scenario. Because comparing `self._ds` to `other._ds`, they'd have different dim names. Am I missing something here?

Considering how to fix this though, I don't have an immediate answer, but preliminary thought: We'd want this scenario to return `True` instead, right? If so, we'd need to check every single dim & var of each of these grids and see if their values are equal. regardless of their names.

erogluorhan (commented):
Going great! Please see my inline comments as well as replies to your prior coments.

uxarray/grid/grid.py
```diff
@@ -37,57 +34,72 @@
...
-        self.__init_grid_var_names__()
+    def __init__(self,
+                 grid_ds: xr.Dataset,
+                 source_grid_spec: Optional[str] = None,
+                 source_dims_dict: Optional[dict] = {}):
 
-        # initialize face_area variable
-        self._face_areas = None
+        # grid spec not provided, check if grid_ds is a minimum representable UGRID dataset
+        if source_grid_spec is None:
```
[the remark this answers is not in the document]
philipc2: > So, grid_ds needs to be strictly UGRID, but the source_grid_spec can be UGRID or any other format (i.e. whatever format user is telling us they created grid_ds from).

Correct, while the `grid_ds` will always be encoded in the UGRID conventions after the appropriate conversions within the designated I/O module, `source_grid_spec` gives us information to what the original grid format was. 

> Regardless of source_grid_spec is None or anything else, shouldn't we still validate grid_ds is UGRID?

Running the minimum validation check makes sense, even if we provide a `source_grid_spec`, I agree

uxarray/grid/grid.py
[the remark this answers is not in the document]
philipc2: > I mean, we shouldn't attempt on tracking the source names for the formats other than UGRID. The code should look like:

The reason for tracking the `source_dims_dict` across all grid formats is for the case where we end back up in our `UxDataset` and `UxDataArray`, the dimensions that we use within our `Grid` will differ from the dimensions used in those datasets, if they originally come in a different format. 

This is what was discussed in #363

uxarray/core/api.py
philipc2: > Thanks, but the current deprecation warning doesn't help with this variable removal still leading to break the older workflows (even the most recent September version).
> 
> What I actually meant was that we shouldn't do such changes, even simple variable renamings, right away with the first PRs. Instead, we should follow such a flow, when considering changes in the user API:
> 
> 1. Create the issue that describes the changes that will happen in the user API (variable renaming, variable removal, function renaming, etc.)
> 2. Create a PR that does not perform the actual changes but only does start throwing DeprecationWarning for such user API, with an information of when the change will occur in the future
> 3. When the time comes to actual change, create the PR that performs the actual change and closes the issue.
> 
> That said, this PR should not do such actual user API changes. We should instead create an issue that describes all such user APi changes of this PR, and perform its tasks in the future.

I'll start doing this moving forward, this is extremely helpful! 

What would you suggest doing for this PR then?

philipc2:
> > Each grid now has a `_source_dims_dict`, which provides the following mapping {"Original Dimension Name" : "UGRID Dimension Name"}, instead of storing all the connectivity and coordinates.
> 
> Shouldn't it be {"Our standardized UGRID dim name" : "Original Dimension Name from the source"}. Maybe you already meant that?

This was originally the case, however since we want to rename the dimensions within our `UxDataset`, we need the mapping described in the comment.

https://docs.xarray.dev/en/stable/generated/xarray.Dataset.rename_dims.html

uxarray/grid/grid.py
philipc2: > Great question! I think you are concerned about this scenario, right: We get a UGRID input with custom dim names and because of our standardization of those names into our UGRID dim names, they no longer have the same dim names.

That scenario is actually not possible, since we should never have any mismatched dimension or variable names now

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: Oh, I just got the actual concern in #363 and so what you are doing here now; thanks a lot for clarifying that! 

So, okay, we are trying to use standardized dim names consistently, i.e. not only in `Grid` but also in `UxDataset` and `UxDataArray`. I have a few questions and comments then:

1. Why only dim names. Shouldn't we also use var names etc. in the standardized way in our `UxDataset` and `UxDataArray` as well?
2. This is good for satisfying #363, but we will need to handle a further thing with our standardized names. Assume the following scenario:
    - Read a UGRID dataset with custom names into `uxds`, so we will be changing those names to our standardized names.
    - Call `uxds.uxgrid.encode_as("ugrid")` that will output the grid in UGRID format as type xarray.Dataset
       - Problem: It will have our standardized names showing up; however, the user expects to see the original ugrid dataset's custom names because the flow was simply: read in ugrid, encode as ugrid; why variable/dim name changes?
3. If you agree with above cases needing to be addressed, then should we separate this dim/var name mapping things from this PR as this is a big task?

uxarray/core/api.py
erogluorhan: As we are agreed on that, we can get this PR go forward as is (with those changes), but in the future, we should better employ the above flow.

philipc2: > but in the future, we should better employ the above flow.

Will keep this in mind! Appreciate the advice.

uxarray/grid/grid.py
erogluorhan: Oh, I see it now, yes! However, any suggestions will be depending on your thoughts for the following quote from an above topic:

>  Problem: It will have our standardized names showing up; however, the user expects to see the original ugrid dataset's custom names because the flow was simply: read in ugrid, encode as ugrid; why variable/dim name changes?
          - To address this, we'd still need to keep an internal `grid_var_names`-like dict. Tho

philipc2: > Oh, I see it now, yes! However, any suggestions will be depending on your thoughts for the following quote from an above topic:

I've pushed a few changes that should cover the case of grids being equal. (a grid with the same coordinates and face node connectivity can be thought of as equivalent, unless we want to check for more in this operator)

test/test_api.py
philipc2: Reintroduced

test/test_api.py
philipc2: Reintroduced

uxarray/grid/grid.py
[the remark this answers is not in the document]
philipc2: > Why only dim names. Shouldn't we also use var names etc. in the standardized way in our UxDataset and UxDataArray as well?

Since the purpose of our UxDataset and UxDataArray is to store _data variables_, there is "standardized" name for those (other than standardizing the dimensions, which is already done"

> remap_dims = True makes sense, but I am not sure we should even consider providing such an optional argument that could take False value. I can't see use cases with remap_dims = False, so should we handle it like always remapping the dims?

I can remove it and make it always remap if we prefer 

> the flow was simply: read in ugrid, encode as ugrid; why variable/dim name changes?

I don't believe this should be something we need to worry about, especially for the following case
1. Read in UGRID
2. Do stuff (i.e. compute `Mesh2_face_edges`, compute face area)
3. Encode as UGRID

If we chose to encode using the original variable and dimension names, there is a high likelyhood that the encoded dataset will have two types of variables/dims (original mixed with new ones from the computed variables).

for the sake of standardization, I think its fair to assume that our `encode_as` functions will follow our definition of an unstructured grid, including the variable and dimension names.

On the topic of our "definition", we should look into better documenting our definition of an unstructured grid, because even though we follow the UGRID conventions, we have been slowly adding our own variables, such as the way we represent Cartesian Coordinates. 

Once this is merged, I'll handle that in #408 

> If you agree with above cases needing to be addressed, then should we separate this dim/var name mapping things from this PR as this is a big task?

This might be worth revisiting, however for the sake of this PR the removal of `grid_var_names` was to deprecate it's existing inconsistent internal usage. We can consider reintroducing a variant down the line if needed.

philipc2:
@aaronzedwick @rajeeja 

Have either of you encountered this error before with Numba? After these changes, the notebook is starting to fail

image

Interestingly, running the same code snippet in testing doesn't cause the issue. Any thoughts?

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: > Since the purpose of our UxDataset and UxDataArray is to store _data variables_, there isn't "standardized" name for those (other than standardizing the dimensions, which is already done".

I was not clear enough about my use of the word "vars". I meant the cases where UxDataset has also the grid topology variables (e.g. latCell, lonCell, cellsOnEdges, etc.) in it, in which it feels a bit awkward to see the dataset with those source grid format's original names while we have the `uxds.uxgrid` have our standardized names (I know we are supposed to always use grid topology info through `.uxgrid`, but it may still get confusing in the user's end, for instance when hovering over the `uxds` in an IDE). 
 
> > remap_dims = True makes sense, but I am not sure we should even consider providing such an optional argument that could take False value. I can't see use cases with remap_dims = False, so should we handle it like always remapping the dims?
> 
> I can remove it and make it always remap if we prefer

If that also makes sense to you, I think that's better.

> > the flow was simply: read in ugrid, encode as ugrid; why variable/dim name changes?
> 
> I don't believe this should be something we need to worry about, especially for the following case
> 
> 1. Read in UGRID
> 2. Do stuff (i.e. compute `Mesh2_face_edges`, compute face area)
> 3. Encode as UGRID
> 
> If we chose to encode using the original variable and dimension names, there is a high likelyhood that the encoded dataset will have two types of variables/dims (original mixed with new ones from the computed variables).
> 
> for the sake of standardization, I think its fair to assume that our `encode_as` functions will follow our definition of an unstructured grid, including the variable and dimension names.
> 
> On the topic of our "definition", we should look into better documenting our definition of an unstructured grid, because even though we follow the UGRID conventions, we have been slowly adding our own variables, such as the way we represent Cartesian Coordinates.
> 
> Once this is merged, I'll handle that in #408
 
Actually, your points make much sense to me. However, the following quote was from #363, so I think we should still take into consideration there might be some user expectations about this:

> "Converting (or standardizing) names would certainly make the matter simple inside of uxarray, though it means that we are losing a bit of the original UGRID information.
Keeping the original information in post-processed outputs may matter in some cases. (I certainly need to keep it, at least for now.) Uxarray does not (or does not need to) have writing IO functions implemented yet, as far as I know, and it is probably not a concern for now."

aaronzedwick:
> @aaronzedwick @rajeeja
> 
> Have either of you encountered this error before with Numba? After these changes, the notebook is starting to fail
> 
> image
> 
> Interestingly, running the same code snippet in testing doesn't cause the issue. Any thoughts?

Interesting. I have never encountered this specific error before. It looks like your last commits removed stuff from the notebook, so maybe that caused it? If it works inside your IDE it seems likely it could be something is wrong with the notebook, although looking at it I can’t see what. Maybe it got corrupted like Hongyu’s?

uxarray/grid/grid.py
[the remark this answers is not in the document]
philipc2: @erogluorhan 

> I meant the cases where UxDataset has also the grid topology variables (e.g. latCell, lonCell, cellsOnEdges, etc.)

It would makes sense to be able to "filter" these out of the UXdataset to prevent duplicates between the Grid and Dataset, so this is worth considering in a future PR

> Actually, your points make much sense to me. However, the following quote was from https://github.com/UXARRAY/uxarray/issues/363, so I think we should still take into consideration there might be some user expectations about this:

Got it! I will give it some thought and raise an issue best addressing how to provide a solution to this.

philipc2:
> > @aaronzedwick @rajeeja
> > Have either of you encountered this error before with Numba? After these changes, the notebook is starting to fail
> > image
> > Interestingly, running the same code snippet in testing doesn't cause the issue. Any thoughts?
> 
> Interesting. I have never encountered this specific error before. It looks like your last commits removed stuff from the notebook, so maybe that caused it? If it works inside your IDE it seems likely it could be something is wrong with the notebook, although looking at it I can’t see what. Maybe it got corrupted like Hongyu’s?

It was failing prior to updating the notebook, so I don't believe that's the issue.

philipc2:
> > @aaronzedwick @rajeeja
> > Have either of you encountered this error before with Numba? After these changes, the notebook is starting to fail
> > image
> > Interestingly, running the same code snippet in testing doesn't cause the issue. Any thoughts?
> 
> Interesting. I have never encountered this specific error before. It looks like your last commits removed stuff from the notebook, so maybe that caused it? If it works inside your IDE it seems likely it could be something is wrong with the notebook, although looking at it I can’t see what. Maybe it got corrupted like Hongyu’s?

It's resolved now. I had to update the notebook to use use `ux.open_grid` instead of using the old `Grid` constructor approach to opening the grid.

uxarray/grid/grid.py
```diff
@@ -7,87 +7,108 @@
...
+                 source_grid_spec: Optional[str] = None,
+                 source_dims_dict: Optional[dict] = {}):
+
+        # grid spec not provided, check if grid_ds is a minimum representable UGRID dataset
+        if source_grid_spec is None:
+            warn(
+                "Attempting to construct a Grid without passing in source_grid_spec. Use of Grid constructor"
+                "is only advised if grid_ds is following the internal unstructured grid definition, including"
+                "variable and dimension names. Using ux.open_grid() or ux.from_dataset() is suggested.",
+                Warning)
```
philipc2, line 89: @erogluorhan 

I've added this warning, is there anything you'd suggest to add to better reflect the changes to the constructor?

erogluorhan: This looks great!

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: Okay, let me create issues for both of these topics; we can follow up on them in the future

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: Once you address this in the code, we should be good to go with this PR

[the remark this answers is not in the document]
philipc2: > Once you address this in the code, we should be good to go with this PR

Should be good now!

uxarray/grid/grid.py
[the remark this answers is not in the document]
erogluorhan: Okay, created #449 and #450 for these

uxarray/grid/grid.py
```diff
@@ -7,87 +7,112 @@
...
+            )  # TODO: elaborate once we have a formal definition
+
+        # grid spec not provided, check if grid_ds is a minimum representable UGRID dataset
+        if source_grid_spec is None:
+            warn(
+                "Attempting to construct a Grid without passing in source_grid_spec. Use of Grid constructor"
+                "is only advised if grid_ds is following the internal unstructured grid definition, including"
+                "variable and dimension names. Using ux.open_grid() or ux.from_dataset() is suggested.",
+                Warning)
+            if _is_minimum_ugrid:
```
erogluorhan, line 99: At this point, it's guaranteed that `_is_minimum_ugrid `, because we passed the exception above at Line 87,

uxarray/grid/grid.py
```diff
@@ -7,87 +7,112 @@
...
-        self._face_areas = None
+    def __init__(self,
+                 grid_ds: xr.Dataset,
+                 source_grid_spec: Optional[str] = None,
+                 source_dims_dict: Optional[dict] = {}):
+
+        # check if inputted dataset is a minimum representable 2D UGRID unstructured grid
+        _is_minimum_ugrid = _validate_minimum_ugrid(grid_ds)
+
+        if not _is_minimum_ugrid:
```
erogluorhan, line 86:
```suggestion
        if not _validate_minimum_ugrid(grid_ds):
```

because `_is_minimum_ugrid ` will not be used any longer (see the below comment in conjunction)

erogluorhan (commented):
Great' almost there. See inline comments, and are you planning to "always remap"? Should I expect that or approve without it?

uxarray/grid/grid.py
philipc2: Should be fixed now!

philipc2:
> and are you planning to "always remap"? Should I expect that or approve without it?

I've made the change to always remap the dims.

uxarray/grid/grid.py
erogluorhan: What I meant was to replace Lines 83-102 with this (note I also rephrased the text of the ValueError):

       ```
        # check if inputted dataset is a minimum representable 2D UGRID unstructured grid
        if not _validate_minimum_ugrid(grid_ds):
            raise ValueError(
                "Use of Grid constructor requires grid_ds to follow the internal unstructured grid definition, 
                  including variable and dimension names. This grid_ds is not satisfying those requirements. If you are 
                  not sure about how to do that, using ux.open_grid() or ux.from_dataset() is suggested."
            )  # TODO: elaborate once we have a formal definition

        # grid spec not provided, check if grid_ds is a minimum representable UGRID dataset
        if source_grid_spec is None:
            warn(
                "Attempting to construct a Grid without passing in source_grid_spec. Use of Grid constructor"
                "is only advised if grid_ds is following the internal unstructured grid definition, including"
                "variable and dimension names. Using ux.open_grid() or ux.from_dataset() is suggested.",
                Warning)
                
           source_grid_spec = "UGRID"
           source_dims_dict = None
```

**Comment by erogluorhan** [comment]
> > and are you planning to "always remap"? Should I expect that or approve without it?
> 
> I've made the change to always remap the dims.

Can't see these changes

File: uxarray/grid/grid.py (hunk 23 above)
Comment (erogluorhan) [reply uxarray/grid/grid.py:99]: Additionally, You had  `source_grid_spec = "UGRID"` but I don't think we should do it since we can't know the type of the source grid unless the user states.

**Comment by philipc2** [comment]
> > > and are you planning to "always remap"? Should I expect that or approve without it?
> > 
> > 
> > I've made the change to always remap the dims.
> 
> Can't see these changes

Whoops! Forgot to push them.

File: uxarray/grid/grid.py (hunk 23 above)
Comment (philipc2) [reply uxarray/grid/grid.py]: > Additionally, You had source_grid_spec = "UGRID" but I don't think we should do it since we can't know the type of the source grid unless the user states.

If it passes the minimum UGRID, should we not set it as ugrid?

Comment (erogluorhan) [reply uxarray/grid/grid.py]: Yes, I think we should not set it as UGRID. Think about the `Grid.from_dataset()` case: In Line 140:

`source_grid_spec = _parse_grid_type(dataset)`

`_parse_grid_type()` is recognizing the grid-spec from the input `dataset`, not our constructed `grid_ds ` i.e. we are setting `source_grid_spec` to the original mesh type, not UGRID. 

Similaryl, in this constructor, if the user prefers to specify what was the original grid spec they created the UGRID from, they will set `source_grid_spec`; i.e. if we are not given that info, we should keep it as `None`.

Hope this helps

File: uxarray/grid/grid.py [hunk 25]
```diff
@@ -7,87 +7,111 @@
...
+        if source_grid_spec is None:
+            warn(
+                "Attempting to construct a Grid without passing in source_grid_spec. Use of Grid constructor"
+                "is only advised if grid_ds is following the internal unstructured grid definition, including"
+                "variable and dimension names. Using ux.open_grid() or ux.from_dataset() is suggested.",
+                Warning)
+
+            # TODO: more checks for validate grid (lat/lon coords, etc)
+            source_grid_spec = "UGRID"
+            source_dims_dict = None
```
erogluorhan, line 101:
```suggestion
```

uxarray/grid/grid.py
```diff
@@ -7,87 +7,111 @@
...
+        # grid spec not provided, check if grid_ds is a minimum representable UGRID dataset
+        if source_grid_spec is None:
+            warn(
+                "Attempting to construct a Grid without passing in source_grid_spec. Use of Grid constructor"
+                "is only advised if grid_ds is following the internal unstructured grid definition, including"
+                "variable and dimension names. Using ux.open_grid() or ux.from_dataset() is suggested.",
+                Warning)
+
+            # TODO: more checks for validate grid (lat/lon coords, etc)
+            source_grid_spec = "UGRID"
```
erogluorhan, line 100:
```suggestion
```

erogluorhan: If `source_grid_spec` is given `None`, fine, we shouldn't do anything. Giving only a warning should be enough (you may want to rephrase the warning text one more time to emphasize `sourcegrid_spec` is for diagnostics and reporting purposes only, hence a warning not an error).

uxarray/grid/grid.py
erogluorhan: We shouldn't do this assignment as well. If user provided it, just keep it. (In fact, we could theoretically use the dim names in this arg to guess the `source_grid_spec` even if it is `None`, but it is not a must and not in this PR's scope either)

erogluorhan (changes requested):
I will approve this, but please run a final `pre-commit run --all-files` command in your local along with addressing the two inline comments below

uxarray/grid/grid.py
philipc2: It does help! Makes a lot of sense! Thank you!

erogluorhan (approved):

PR #396, opened 2023-07-27
DRAFT: Remove Mesh2 From Variables

### LINKED ISSUE #389: Remove `Mesh2` from grid dimensions and variables
### Discussed in https://github.com/UXARRAY/uxarray/discussions/387

<div type='discussions-op-text'>

<sup>Originally posted by **hetland** July 18, 2023</sup>
I would suggest to remove `Mesh2' from the grid dimensions and variables. This is used in a few example of 2D meshes in the ugrid conventions, but these are clearly just examples in the netCDF header, the associated tables showing the variables are probably more appropriate for naming. I am concerned this will be confusing to users, and will make working with uxarray more difficult. I understand this would touch quite a bit of the code, but I think better to shift to a better representation of the conventions now rather than later.</div>

aaronzedwick:
Closes #389 

## Overview

Removes Mesh2 from all grid variables and connectivity

## PR Checklist

**General**
- [x] Add appropriate labels

**Documentation**

### DIFF
--- a/docs/internal_api/index.rst
+++ b/docs/internal_api/index.rst
@@ -101,7 +101,7 @@ Attributes
 .. autosummary::
    :toctree: _autosummary
 
-   Grid._Mesh2_node_z
+   Grid._node_z
 
 Operators
 ---------
--- a/docs/user_api/index.rst
+++ b/docs/user_api/index.rst
@@ -116,21 +116,21 @@ Attributes
 .. autosummary::
    :toctree: _autosummary
 
-   Grid.Mesh2
+   Grid.Grid2D
    Grid.parsed_attrs
-   Grid.nMesh2_node
-   Grid.nMesh2_face
-   Grid.nMesh2_edge
-   Grid.nMaxMesh2_face_nodes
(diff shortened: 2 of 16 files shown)


aaronzedwick:
@philipc2 Most of this works, I am having some trouble with the ugrid one, specifically in the _ugrid.py. I try to convert the variables to remove the Mesh2 part, but when you do that the code has key errors in the preceding code. Any idea how to assign them different names, since the names are stored inside the ugrid file and we aren't creating separate attributes like other file types?

philipc2:
> @philipc2 Most of this works, I am having some trouble with the ugrid one, specifically in the _ugrid.py. I try to convert the variables to remove the Mesh2 part, but when you do that the code has key errors in the preceding code. Any idea how to assign them different names, since the names are stored inside the ugrid file and we aren't creating separate attributes like other file types?

This implementation will require changing how we use the var name dict within our dict object. I'm planning on tackling this issue in #395 somewhat.

philipc2:
Closing this issue and tackling it in #408

PR #406, opened 2023-08-14
BallTree Data Structure for Center and Corner Nodes

### LINKED ISSUE #404: Tree Data Structure for Nearest Neighbor Search

philipc2:
Closes #404 

## Overview

* Introduces a custom `BallTree` data structured which can be constructed using the Corner (`Mesh2_node_x`, `Mesh2_node_y`) and Center (`Mesh2_face_x`, `Mesh2_face_y`) of each face respectively
* Implementation wraps `sklearn.neighbors.BallTree` using the haversine distance metric
* Custom query methods for finding k nearest neighbors and neighbors within some radius 
* Both data structures can either be accessed through the `Grid` class (i.e. `Grid.ball_tree`) or through an external class initialization

## Expected Usage

```Python
import uxarray as ux

# find the nearest corner node neighbor to (180.0, 0.0)
d, ind = uxgrid.ball_tree.query([180.0, 0.0], k=1, tree_type="corner nodes")

# find the nearest center node neighbor to (180.0, 0.0)
d, ind = uxgrid.ball_tree.query([180.0, 0.0], k=1, tree_type="face centers")

# find all corner nodes within a radius of 90.01 of (-180, 0.0)
d, ind = uxgrid.ball_tree.query_radius([-180, 0.0], r=90.01, tree_type="corner nodes")

# find all center nodes within a radius of 90.01 of (-180, 0.0)
d, ind = uxgrid.ball_tree.query_radius([-180, 0.0], r=90.01, tree_type="face centers")


```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


test/test_grid.py
```diff
@@ -778,3 +781,57 @@ def test_build_face_edges_connectivity_fillvalues(self):
         self.assertTrue(
             np.array_equal(res_face_nodes_connectivity,
                            uds._ds["Mesh2_face_nodes"].values))
+
+
+class TestBallTree(TestCase):
+
+    corner_grid_files = [gridfile_CSne30, gridfile_mpas]
+    center_grid_files = [gridfile_mpas]
```
aaronzedwick, line 789: Why don't you need a dataset here?

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Notes
+    -----
+    See `sklearn.neighbors.BallTree <https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.BallTree.html>`__
+    for further information about the wrapped data structure.
+    """
+
+    def __init__(self, grid, node_type='corner'):
+
+        # construct tree from corner nodes
```
aaronzedwick, line 25: What are the different use cases for this? Why would we want a tree from corner nodes vs center nodes?

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+
+        if k < 1 or k > self.n_elements:
+            raise AssertionError(
+                f"The value of k must be greater than 1 and less than the number of elements used to construct "
+                f"the tree ({self.n_elemetns}).")
+
+        xy = _prepare_xy_for_query(xy, use_radians)
+
+        # perform query with distance
```
aaronzedwick, line 95: Just a question on how this distance works. Is it the arc distance on the sphere from the neighbors? Would this also work on a flat mesh that isn't spherical?

test/test_grid.py
```diff
@@ -778,3 +781,57 @@ def test_build_face_edges_connectivity_fillvalues(self):
...
+
+    def test_construction_from_corner_nodes(self):
+        """Tests the construction of the ball tree on corner nodes and performs
+        a sample query."""
+
+        for grid_file in self.corner_grid_files:
+            uxgrid = ux.open_grid(grid_file)
+
+            # performs a sample query
+            d, ind = uxgrid.corner_node_balltree.query([3.0, 3.0], k=3)
```
aaronzedwick, line 799: The queries are all done in lat/lon, do we not really need to worry about cartesian coordinates? I can't think of a use case that would need cartesian coordinates but I figure I would ask.

aaronzedwick (commented):
Trying to understand what is going on better, so just a few questions on how everything works.

test/test_grid.py
philipc2: both the `corner_node_balltree` and `center_node_balltree` are constructed using lat/lon coordinates, and expect lat/lon inputs. We could consider adding a tree that is build from and queries using Cartesiain, but right now it's only lat/lon

uxarray/grid/neighbors.py
philipc2: The distance metric is computed using the [haversine distance formula](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html), which is the great circle distance. 

> Would this also work on a flat mesh that isn't spherical?

I wrote this with the assumption that our grids will represent the surface of a sphere, hence the use of the haversine distance formula. If we have an unstructured grid that isn't on the surface of a sphere, we would need to use a different distance metric.

@erogluorhan @hongyuchen1030 @paullric 

For UXarray, is it fair to assume that we will always be working with grids that represent the surface of a sphere?

uxarray/grid/neighbors.py
philipc2: Both are needed depending on where our data is defined (i.e. on each node or face center)

### Corner Nodes
Used for computing the distance between the corners that make up each face. 

### Center Nodes
Used for computing the distance between the centers of each face

test/test_grid.py
philipc2: Can you elaborate on what you mean?

aaronzedwick: Yeah, so in line 788 you pass gridfile_CSne30, which if I remember is an integration variable of some type, like temperature or something similar, whereas in the center grid files you don't pass the same file. I was wondering why this was the case.

philipc2: gridfile_CSne30 is actually a grid file, there's also a corresponding data variable.

uxarray/grid/neighbors.py
erogluorhan: I tend to say yes since UXarray is dedicated to geoscience data. An alternative could be to provide an optional argument that'd default to haversine distance though.

test/test_grid.py
```diff
@@ -837,27 +840,80 @@ def test_node_face_connectivity_from_files(self):
...
+            face_index_from_sparse_matrix = grid_ux.Mesh2_node_faces.values[i]
+            valid_face_index_from_sparse_matrix = face_index_from_sparse_matrix[
+                face_index_from_sparse_matrix !=
+                grid_ux.Mesh2_node_faces.attrs["_FillValue"]]
+            valid_face_index_from_sparse_matrix.sort()
+            face_index_from_dict = node_face_connectivity[i]
+            face_index_from_dict.sort()
+            self.assertTrue(
+                np.array_equal(valid_face_index_from_sparse_matrix,
+                               face_index_from_dict))
```
erogluorhan, line 865: Changing the indentation from Line 843 to 865 (i.e. removing those from the for loop) seems to be damaging the fat that the for loop is running assertions for multiple grid paths in `grid_paths`.

philipc2: Good catch! Must of been an artifact from a merge conflict.

uxarray/grid/neighbors.py
philipc2: > An alternative could be to provide an optional argument that'd default to haversine distance though.

That's a good idea

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+            if True, use the dual tree formalism for the query: a tree is built for the query points, and the pair of trees is used to efficiently search this space. This can lead to better performance as the number of points grows large.
+        breadth_first : bool, default=False
+            if True, then query the nodes in a breadth-first manner. Otherwise, query the nodes in a depth-first manner.
+        sort_results : bool, default=True
+            if True, then distances and indices of each point are sorted on return, so that the first column contains the closest points. Otherwise, neighbors are returned in an arbitrary order.
+
+        Returns
+        -------
+        d : ndarray of shape (xy.shape[0], k), dtype=double
+            Each entry gives the list of distances to the neighbors of the corresponding point.
```
erogluorhan, line 83: A bit difficult to read this. Did you mean something like "Distance array that keeps the distances of the k-nearest neighbors to the entries from `xy` in each row"

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+        ind : ndarray of shape (xy.shape[0], k), dtype=INT_DTYPE
+            Each entry gives the list of indices of neighbors of the corresponding point.
```
erogluorhan, line 85: Similarly, did you mean something like "Index array that keeps the indices of the k-nearest neighbors to the entries from `xy` in each row"

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+        xy : array_like
+            coordinate pairs in degrees (lon, lat) to query
+        k: int, default=1
+            The number of nearest neighbors to return
+        use_radians : bool, optional
+            if True, queries assuming xy are inputted in radians, not degrees
+        return_distance : bool, optional
+            if True, return a tuple ``(d, i)`` of distances and indices if False, return array i
+        dualtree : bool, default=False
+            if True, use the dual tree formalism for the query: a tree is built for the query points, and the pair of trees is used to efficiently search this space. This can lead to better performance as the number of points grows large.
```
erogluorhan, line 74: Text too long for a single line. Curious how it passed the pre-commit; did we change the pre-commit configs/dependencies recently?

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+        breadth_first : bool, default=False
+            if True, then query the nodes in a breadth-first manner. Otherwise, query the nodes in a depth-first manner.
+        sort_results : bool, default=True
+            if True, then distances and indices of each point are sorted on return, so that the first column contains the closest points. Otherwise, neighbors are returned in an arbitrary order.
```
erogluorhan, line 78: Similarly, text too long for a single line. There are others similar as well; please check them all

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+        sort_results : bool, default=False
+           if True, the distances and indices will be sorted before being returned. If False, the results will not be sorted. If return_distance == False, setting sort_results = True will result in an error.
+
+
+        Returns
+        -------
+        d : ndarray of shape (xy.shape[0], k), dtype=double
+            Each entry gives the list of distances to the neighbors of the corresponding point.
+        ind : ndarray of shape (xy.shape[0], k), dtype=INT_DTYPE
+            Each entry gives the list of indices of neighbors of the corresponding point.
```
erogluorhan, line 153: Similar to the above comments; please consider rephrasing these two descriptions to make them easy to read

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+
+        if k < 1 or k > self.n_elements:
+            raise AssertionError(
+                f"The value of k must be greater than 1 and less than the number of elements used to construct "
+                f"the tree ({self.n_elemetns}).")
+
+        xy = _prepare_xy_for_query(xy, use_radians)
```
erogluorhan, line 93: xy is no longer xy, it is instead yx, after this line. I am not sure if we should name that way (which would end up having concurrent allocations for two variables), but please at least clarify that with a comment.

uxarray/grid/neighbors.py
```diff
@@ -0,0 +1,220 @@
...
+    if xy.shape[1] != 2:
+        raise AssertionError(
+            f"The dimension of each coordinate pair must be two (lon, lat).)")
+
+    # swap X and Y for query
+    xy[:, [0, 1]] = xy[:, [1, 0]]
+
+    # balltree expects units in radians for query
+    if not use_radians:
+        xy = np.deg2rad(xy)
```
erogluorhan, line 218: Something should be wrong here: if not use_radians, then convert to radians?

erogluorhan (commented):
Great work; please address the comments below.

uxarray/grid/neighbors.py
philipc2: `use_radians` indicates whether the user is already providing their `xy` input in radians.

uxarray/grid/neighbors.py
philipc2: should we considering renaming `xy` to `lonlat`, and then having the "yx" be reffered to as `latlon` internally?

uxarray/grid/neighbors.py
philipc2: I was surprised by this also, not sure why it was allowed.

uxarray/grid/grid.py
```diff
@@ -85,13 +89,15 @@ def __init__(self, input_obj, **kwargs):
...
+        # initialize cached data structures (visualization)
         self._gdf = None
         self._poly_collection = None
-        # TODO: fix when adding/exercising gridspec
 
-        # unpack kwargs
-        # sets default values for all kwargs to None
+        # initialize cached data structures (nearest neighbor operations)
+        self._corner_node_balltree = None
+        self._center_node_balltree = None
```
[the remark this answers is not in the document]
philipc2: I've adopted these changes in the following manner, since there is value to allowing queries on either the face centers or corner nodes in a single workflow without needing to reinitialize the `BallTree`
* The grid now has a single `Grid.ball_tree` attribute/property
* Each `BallTree` is now written to store both a corner node and face center wrapped ball tree internally
* The user can indicate which tree to use for queries by passing a `tree_type` 

```Python
# defaults to "corner nodes"
uxgrid.ball_tree.query([180.0, 0.0],k=1)

# explicitly specify "corner nodes"
uxgrid.ball_tree.query([180.0, 0.0],k=1, tree_type="corner nodes")

# explicitly specify "face centers"
uxgrid.ball_tree.query([180.0, 0.0],k=1, tree_type="face centers")

aaronzedwick (approved):
From what I can see it looks really good, I like the combination of the center and corners, that definitely is an improvement on overall usability.

erogluorhan (commented):
This looks much better than the previous version. I still have some ideas about the design and use though, which led me to create PR #432  over this branch due to several changes needed.  That PR addresses all of my following thoughts (so, instead of responding to my inline comments here, please review that PR):

1. This current version has `BallTree` as a `@property` of the `Grid`, but the actual meaningful`BallTree` initialization can be handled only after a `query` or `query_radius` call. i.e. `BallTree.__init__` can do nothing other than some variable inits before knowing the `tree_type` of interest. This does not seem like the best useful workflow, especially for someone who calls only the property of Grid without a query:
   `uxds.uxgrid.ball_tree`
2.  Because of the code branching caused by `count_only` and `return_distance` args in `query` and `query_radius`, they look pretty cluttered and hard to read. 
3. Also, the inline comments below

erogluorhan (approved):
Great work; thanks very much!

PR #438, opened 2023-09-01
Plot API Structure & `UxDataArray.plot.datashade` Method

### LINKED ISSUE #118: .plot() method for Grid Object
Similar to Xarray, directly allow for plotting from a Grid Object (compared to a DataArray).  

**Proposed Functionality** 

- Calling `.plot()` on a `Grid` object with no data variables (only grid info) would render an unfilled mesh
- Calling `.plot()` on a `Grid` object with an argument for data variables, indexing, etc, would render a filled mesh according to the user inputs

PolyMesh Specific 
- Goal is to incorporate my PolyMesh code to handle visualizations, however due to package dependencies starting off with SciPy's or MPL's triangulation might be better (having both options would be good too)
- Any mesh that's constructed during plotting would be stored in the `Grid` object to reduce computation time on repeated plots (ex. plotting multiple data variables on the same mesh)

philipc2:
Closes #118, #467 

## Overview

* Sets up the project structure for implementing plotting methods for `Grid`, `UxDataArray` and `UxDataset`
* - This PR **does NOT** implement plotting methods for all objects, while the accessor is there, the only working plotting method with an implemented function is `UxDataArray` 
* Initial Implementation of `UxDataArray.datashade()`

## Expected Usage

```Python
import uxarray as ux

# raster plot
uxds['v1'].plot.datashade()

# default plotting method 
uxds['v1'].plot()

```

## Expected Visualizations
This PR implements ``UxDataArray.plot.datashade()``, which can be used to quickly visualize face-centered data. 
image

image

## PR Checklist

**General**
- [x] Add appropriate labels

**Documentation**


uxarray/plot/accessor.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @functools.wraps(dataarray_plot.raster)
+    def raster(self,
+               plot_height: Optional[int] = 300,
+               plot_width: Optional[int] = 600,
+               cmap: Optional[str] = "blue",
+               agg: Optional[str] = "mean"):
+        """TODO: Docstring & additional params"""
+        return dataarray_plot.raster(self._uxda, plot_height, plot_width, cmap,
+                                     agg)
```
philipc2, line 54: @erogluorhan 

For the scope of this PR, I've set up the accessor and modules for interfacing with our plotting methods. Additionally, I've also set up a `UxDataArray.raster()` method directly using Datashader to perform the visualization. 

If you are good with moving forward with this PR only including the `UxDataArray.raster()` method, we can start implementing all other methods. I was planning on beginning with our `Grid` plotting methods, however after our discussion about the issues we've encountered, we need to spend some time figuring those out. 

Let me know what you think, planning on getting this PR polished and ready for review today.

uxarray/plot/accessor.py
```diff
@@ -0,0 +1,75 @@
...
+        return dataarray_plot.plot(self._uxda, **kwargs)
+
+    @functools.wraps(dataarray_plot.raster)
+    def raster(self,
+               plot_height: Optional[int] = 300,
+               plot_width: Optional[int] = 600,
+               x_range: Optional[tuple] = (-180, 180),
+               y_range: Optional[tuple] = (-90, 90),
+               cmap: Optional[str] = "inferno",
+               agg: Optional[str] = "mean"):
```
erogluorhan, line 47: I'd suggest we use *args and **kwargs in the beginning and end of the function signature to accept changeable number/definition of arguments:

 `def raster(self, *args, ... our args ..., **kwargs):`

 This could especially be useful to support any changes of args in the upstream (e.g. `datashader.transfer_functions.shade()` in this case).

uxarray/plot/dataarray_plot.py
```diff
@@ -0,0 +1,57 @@
...
+    return raster(uxda, **kwargs)
+
+
+def raster(uxda: UxDataArray,
+           plot_height: Optional[int] = 300,
+           plot_width: Optional[int] = 600,
+           x_range: Optional[tuple] = (-180, 180),
+           y_range: Optional[tuple] = (-90, 90),
+           cmap: Optional[str] = "inferno",
+           agg: Optional[str] = "mean"):
```
erogluorhan, line 21: The same *args **kwargs comment applies to here and everywhere else

uxarray/plot/dataarray_plot.py
```diff
@@ -0,0 +1,57 @@
...
+    x_range, y_range : tuple, optional
+       A tuple representing the bounds inclusive space ``[min, max]`` along
+       the axis.
+    cmap: str, optional
+        Colormap used for shading
+    agg : str, optional
+        Reduction to compute. Default is "mean", but can be one of "mean" or "sum"
+    """
+    import datashader as ds
+    import datashader.transfer_functions as tf
```
erogluorhan, line 37: Would it be a better idea to have `import`s in the `init` so that we import them only once when `.plot` accessor is called (instead of always importing whenever a plottig function is called)?

erogluorhan (commented):
This looks great; and I will approve it soon, but please address the inline comments along with these:
1. Are you foreseeing holoviews functions to be the only dependencies for our plotting functions (i.e. no direct matplotlib func calls anywhere)? What do you think about supporting multiple viz. engines (e.g. bokeh, matplotlib, etc.)? Maybe getting inputs to choose MPL or bokeh engines for `hv.extension()` for each function?
2. We have at least one new dependency `datashader`, which might need to be added into `ci/enviornment.yml` for local dev env to work with `.plot` accessor (maybe auto-installed through other dependencies though?)

aaronzedwick (approved):
Looks good to me, however one question I did have was why there are empty files?

philipc2:
> This looks great; and I will approve it soon, but please address the inline comments along with these:
> 
> 1. Are you foreseeing holoviews functions to be the only dependencies for our plotting functions (i.e. no direct matplotlib func calls anywhere)? What do you think about supporting multiple viz. engines (e.g. bokeh, matplotlib, etc.)? Maybe getting inputs to choose MPL or bokeh engines for `hv.extension()` for each function?

We should support both, and the implementation of those will become more clear once we start to implement plotting functions that can be executed with either backends. I am currently working on putting together a clear API outline of this, similar to https://github.com/UXARRAY/uxarray/discussions/225.


> 2. We have at least one new dependency `datashader`, which might need to be added into `ci/enviornment.yml` for local dev env to work with `.plot` accessor (maybe auto-installed through other dependencies though?)

I will look into this

uxarray/plot/accessor.py
philipc2: Good idea, I'll look into this. My only concern is that for someone who isn't familiar with Datashader, it may not be clear what to pass through into the function. 

Also, the default values for our application (specifically x and y ranges) are different than those that are default in Datashadder.

philipc2:
> Looks good to me, however one question I did have was why there are empty files?

The empty files are placeholders for future plotting functions.

philipc2:
@erogluorhan 

After spending the last few days digging into Datashader and preparing for the ESDS presentation, I belive we should reconsider the naming of our `raster` method, since in reality this is technically performing datashading, not just rasterization.

uxarray/plot/dataarray_plot.py
philipc2: This approach (also in a few of our geometry functions) is how we handle optional dependencies. I can spend some time figuring out a better solution, but this is what has been working for us as of now.

erogluorhan (approved):

PR #439, opened 2023-09-01
Nearest Neighbor Remapping

### LINKED ISSUE #397: Nearest Neighbor Regridding

KDTree needs to be built on the source mesh, then sampled on the target 

load two meshes a and b into uxarray
map a variable say temperature from mesh a onto mesh b.

Note: this functionality will need a function that takes two meshes as arguments! 

**Additional context**
use `scipy.spatial`

philipc2:
Closes #397

## Overview

* Implements the project structure for Remappiing
* Implements the Nearest Neighbor Remapping algorithm using #406 as a foundation for distance calculations
* Supports Source (``UxDataArray`` and ``UxDataset``) to Destination (``Grid``, ``UxDataArray`` and ``UxDataset``)

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


philipc2:
@aaronzedwick 

@rajeeja and I spend some time coworking and discussing the implementing of Nearest Neighbor Regridding. I'll post a comment later today with an explanation of our design choice and what steps need to be done to move forward.

aaronzedwick:
> @aaronzedwick
> 
> @rajeeja and I spend some time coworking and discussing the implementing of Nearest Neighbor Regridding. I'll post a comment later today with an explanation of our design choice and what steps need to be done to move forward.

Sounds good, just ping me, thanks!

philipc2:
@aaronzedwick 

After speaking with Paul and considering our recent `BallTree` implementation, we are moving forward with 4 types of Nearest Neighbor regridding methods, each of which the same except with differences in their Source and Desitation location of the data variable. 

#### Node to Node
Data lies on the corner nodes of the source grid, regrid it to the corner nodes of the destination grid.

#### Node to Face
Data lies on the corner nodes of the source grid, regrid it to the face centers of the destination grid.

#### Face to Node
Data lies on the face centers of the source grid, regrid it to the corner nodes of the destination grid.

#### Face to Face
Data lies on the face centers of the source grid, regrid it to the corner nodes of the destination grid.

For each of these methods, corner nodes refers to `Mesh2_node_x` and `Mesh2_node_y`, while face centers refers to `Mesh2_face_x` and `Mesh2_face_y`

We will continue this after Labor day! Let me know if you have any questions.

aaronzedwick:
@philipc2 Do you have any ideas on the integration of the interpolation algorithm? Any documentation you have that I could look over? For the interpolation, we might have a grid that is say, 10 times finer, do we just get the ten nearest nodes(or centers) and take an average for the single node? That doesn't seem like the most accurate way to handle it. Plus the mesh won't be perfectly sized to be exactly however many times bigger. As well as the nodes not necessarily being accurately positioned. Like for instance next to the shoreline if we take the nearest ten nodes, depending on the mesh you might get very deep shorelines, if that makes sense. Just wondering if you had any thoughts algorithmically. I have been looking at a few different ideas online for interpolation techniques, like using a moving least square or a radial basis function and wanted to know what you were thinking on this front.

philipc2:
> Do you have any ideas on the integration of the interpolation algorithm? Any documentation you have that I could look over?

This page from [xESMF](https://xesmf.readthedocs.io/en/latest/notebooks/Compare_algorithms.html) has some nice visuals and is worth looking into, especially the Notes section. Specifically in this PR, we are roughly tackling the Nearest Neighbor (Source to Destination) and (Destination to Source) methods


> For the interpolation, we might have a grid that is say, 10 times finer, do we just get the ten nearest nodes(or centers) and take an average for the single node? 

Nearest Neighbor doesn't take any averages, it simply finds the closest node and maps the value onto the destination grid. For the case where we are going from a coarse source grid to a finer destination grid, there will be duplicate data points.

> That doesn't seem like the most accurate way to handle it. Plus the mesh won't be perfectly sized to be exactly however many times bigger

There are many interpolation algorithms, and while it may not be the most accurate for some examples, it has it purpose and advantages (i.e. it maintains local maximums and minimums). The mesh size will still be preserved, but as I mentioned above there may be duplicate / missing points depending on what resolutions we are regridding rom. 

> I have been looking at a few different ideas online for interpolation techniques, like using a moving least square or a radial basis function and wanted to know what you were thinking on this front.

We will implemented multiple regridding/interpolation functions, this one is simply just the eaiest to start off with :)

The next one we can implement should be an Inverse Distance Weighting method, which is similar to what you described and what we have implemented here, you can take a look at some information about it [here](https://en.wikipedia.org/wiki/Inverse_distance_weighting)

test/test_remap.py
```diff
@@ -0,0 +1,124 @@
...
+        np.array_equal(source_data_single_dim, destination_data)
+
+        source_data_multi_dim = np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0],
+                                          [7.0, 8.0, 9.0]])
+
+        destination_data = nearest_neighbor(source_grid,
+                                            destination_grid,
+                                            source_data_multi_dim,
+                                            destination_data_mapping="nodes")
+
```
rajeeja, line 54: We should test writing these out `to_netcdf` and make sure the grid and destination data look good.

philipc2: Good idea

philipc2:
We should also draft up a notebook example, with visualizations, of the regridding. Hoping to get a version of #447 in sometime for face center calculations.

philipc2:
@paullric @erogluorhan @brianpm 

After our discussion today, is there any preference between using the term "regrid" or "remap"? 

This PR currently is using "remap" throughout, which would look something like this in our API:

```Python
uxds.nearest_neighbor_remap(destination_grid)
```

However, it does seem like NCL uses the term "regrid". Any thoughts?

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,199 @@
...
+
+    elif coord_type == "cartesian":
+        # TODO: once a cartesian balltree/kdtree is implemented, implement this
+        raise ValueError(
+            f"Nearest Neighbor Remapping using Cartesian coordinates is not yet supported"
+        )
+
+    else:
+        raise ValueError(
+            f"Invalid coord_type. Expected either 'lonlat' or 'artesian', but received {coord_type}"
```
aaronzedwick, line 100: cartesian*

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,199 @@
...
+    source_uxds : UxDataset
+        Source UxDataset for remapping
+    destination_obj : Grid, UxDataArray, UxDataset
+        Destination for remapping
+    destination_data_mapping : str, default="nodes"
+        Location of where to map data, either "nodes" or "face centers"
+    coord_type : str, default="lonlat"
+        Indicates whether to remap using on latlon or cartesiain coordinates
+    """
+
```
aaronzedwick, line 182: Could use some comments here to explain what is happening

test/test_remap.py
```diff
@@ -0,0 +1,124 @@
...
+    def test_nn_remap(self):
+        """Test nearest neighbor remapping.
+
+        Steps:
+        1. Open a grid and a dataset,
+        2. Open the grid to remap dataset in 1
+        3. Remap the dataset in 1 to the grid in 2
+        """
+        # TODO; write better test
```
aaronzedwick, line 67: This should probably be done before merging. Maybe at the very least confirm that `out_da` has data in it? I am not sure how you would actually test if the remapping was done properly honestly.

aaronzedwick (commented):
This looks really good! I just have a few initial comments.

uxarray/remap/nearest_neighbor.py
philipc2: Good catch!

philipc2:
Here's an example of re-mapping a data variable that is node-centered onto the face-centers of the same grid (compared to taking the average between the node values. 

image

philipc2:
A few more examples

image

image

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,200 @@
...
+        return uxds
+
+    # return UxDataArray with remapped variable
+    else:
+        return uxda_remap
+
+
+def _nearest_neighbor_uxds(source_uxds: UxDataset,
+                           destination_obj: Union[Grid, UxDataArray, UxDataset],
+                           destination_data_mapping: str = "nodes",
```
erogluorhan, line 168:
```suggestion
                           data_mapping: str = "nodes",
```

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,200 @@
...
+
+    else:
+        raise ValueError(
+            f"Invalid coord_type. Expected either 'lonlat' or 'Cartesian', but received {coord_type}"
+        )
+
+
+def _nearest_neighbor_uxda(source_uxda: UxDataArray,
+                           destination_obj: Union[Grid, UxDataArray, UxDataset],
+                           destination_data_mapping: str = "nodes",
```
erogluorhan, line 106:
```suggestion
                           data_mapping: str = "nodes",
```

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,200 @@
...
+    elif isinstance(
+            destination_obj,
+        (uxarray.core.dataarray.UxDataArray, uxarray.core.dataset.UxDataset)):
+        destination_grid = destination_obj.uxgrid
+    else:
+        raise ValueError("TODO: Invalid Input")
+
+    # perform remapping
+    destination_data = nearest_neighbor(source_uxda.uxgrid, destination_grid,
+                                        source_uxda.data,
```
erogluorhan, line 143: `.data` leads to an issue in `nearest_neighbor()` when `source_uxda` is a UxDataArray that wraps up a Dask array.

test/test_remap.py
```diff
@@ -0,0 +1,124 @@
...
+gridfile_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "grid.nc"
+dsfile_v1_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "v1.nc"
+dsfile_v2_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "v2.nc"
+dsfile_v3_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "v3.nc"
+
+
+class TestNearestNeighborRemap(TestCase):
+    """Tests for nearest neighbor remapping."""
+
+    def test_remap_to_same_grid_corner_nodes(self):
```
erogluorhan, line 29: I believe we should only be concerned about (and providing) `uxds.nearest_neighbor_remap()` and `uxda.nearest_neighbor_remap()`, not public `uxarray.remap. nearest_neighbor()` that takes in data as type of `np.ndarray` (See related concerns in `nearest_neighbor.py`). Even if it's going to be user API, it should take in `UxDataarray`.

erogluorhan (changes requested):
Please find my comments below:

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,200 @@
...
+from uxarray.grid import Grid
+
+
+def nearest_neighbor(source_grid: Grid,
+                     destination_grid: Grid,
+                     source_data: np.ndarray,
+                     destination_data_mapping: str = "nodes",
+                     coord_type: str = "lonlat") -> np.ndarray:
+    """Nearest Neighbor Remapping between two grids, mapping data that resides
+    on either the corner nodes or face centers on the source grid to the corner
```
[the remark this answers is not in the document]
philipc2: That's a good idea for consistency.

uxarray/remap/nearest_neighbor.py
philipc2: Do you have any suggestions on what would be a better alternative?

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,200 @@
...
+    from uxarray.core.dataarray import UxDataArray
+
+import numpy as np
+
+import uxarray.core.dataarray
+import uxarray.core.dataset
+from uxarray.grid import Grid
+
+
+def nearest_neighbor(source_grid: Grid,
```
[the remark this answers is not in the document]
philipc2: The absolute import, if one choses to do so, would be:
```Python
from uxarray.remap._nearest_neighbor
```

Which, if we keep this internal I think should be okay

uxarray/remap/nearest_neighbor.py
```diff
@@ -0,0 +1,200 @@
...
+    destination_obj : Grid, UxDataArray, UxDataset
+        Destination for remapping
+    destination_data_mapping : str, default="nodes"
+        Location of where to map data, either "nodes" or "face centers"
+    coord_type : str, default="lonlat"
+        Indicates whether to remap using on latlon or Cartesian coordinates for nearest neighbor computations when
+        remapping.
+    """
+
+    # prepare dimensions
```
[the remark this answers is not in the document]
philipc2: I've changed the lines to reflect the `Mesh2` equivalents until we change it with the `Mesh2` removal.

uxarray/remap/nearest_neighbor.py
philipc2: How about changing it to something like:

`remap_to` instead of `destination_data_mapping`, since I don't want to lose the fact that we are remapping "to" something, and the data mapping doesnt reffer to the original data variable mapping.

erogluorhan: `remap_to` sounds very good!

uxarray/remap/nearest_neighbor.py
erogluorhan: `remap_to` seems good for this, too!

uxarray/remap/nearest_neighbor.py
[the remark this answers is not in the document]
erogluorhan: Sounds great!

uxarray/remap/nearest_neighbor.py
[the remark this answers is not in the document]
philipc2: This'll be documented in the user guide I'm working on in #472

uxarray/remap/nearest_neighbor.py
erogluorhan: `.values` instead would guarantee the returned value to be a `np.ndarray` even if xarray was wrapping a Dask array. 

The big concern though in that case is that if the results of a really big data (which maybe wouldn't fit into memory) were being lazy-computed with a Dask array, we'd be forcing the data to be loaded into memory by calling `.values` explicitly. To avoid it, we'd need to write our methods in a way that they'd consume even internally Xarray.Dataarray instead of np.ndarray.

We need to make sure about this.

philipc2: Got it! Thank you for the clarification.

erogluorhan: Not only for this function but also for the whole code, as part of our performance and scalability work, I think we will need to check all our internal/user API functions from this point.

erogluorhan (commented):
I am good with the overall implementation except our ongoing discussion about the consumption of np.ndarray in an internal function. Once we figure that out, I should be good to approve this. Really nice work!

uxarray/remap/nearest_neighbor.py
philipc2: Agreed, I've changed all instances within `UxDataArray` to use `.values` instead of `.data`, however I'll keep this in mind as we're working on our performance and optimization efforts.

philipc2:
> I am good with the overall implementation except our ongoing discussion about the consumption of np.ndarray in an internal function. Once we figure that out, I should be good to approve this. Really nice work!

Thank you! I've made #560 to tackle this.

uxarray/remap/nearest_neighbor.py
erogluorhan: Sounds good!

erogluorhan:
> > I am good with the overall implementation except our ongoing discussion about the consumption of np.ndarray in an internal function. Once we figure that out, I should be good to approve this. Really nice work!
> 
> Thank you! I've made #560 to tackle this.

Sounds good; I am copying the above discussion into that issue for future reference.

erogluorhan (approved):

PR #442, opened 2023-09-06
DRAFT: Construct Grid From Points

### LINKED ISSUE #349: Constructing a grid from x and y node coordinates

aaronzedwick:
Closes #349

## Overview

The goal is to take in points on a sphere, and be able to create the connectivity information using just the points. The two methods to do this are Voronoi and Delaunay Triangulation. For spherical Voronoi we can use scipy's SphericalVoronoi. For the Delaunay Triangulation, first we must project it to a plane, then use scipy's Delaunay Triangulation, and then relate that back to the sphere. The projection can be done using [Stereographic Projection](https://en.wikipedia.org/wiki/Stereographic_projection)
## Expected Usage

```Python
import uxarray as ux

verts = np.array([[0, 0, 1], [0, 0, -1], [1, 0, 0], [0, 1, 0],
                          [0, -1, 0], [-1, 0, 0]])
grid_from_verts = ux.open_grid(verts, method="spherical_voronoi")
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**

**Examples**


philipc2:
Pending #395, once it's in we'll restructure the `from_verticies` into it's respective sections as described in that PR.

test/test_grid.py
```diff
@@ -935,3 +935,12 @@ def test_antimeridian_distance_nodes(self):
...
         pass
+
+
+class TestGridFromVertices(TestCase):
+
+    def test_spherical_voronoi(self):
+        verts = [(0, 0), (45, 45), (-30, 150), (60, -30), (-75, -120), (90, 0),
+                 (-90, 90), (30, -60)]
+        grid_from_verts = ux.open_grid(verts)
+        grid_from_verts.from_vertices(method="spherical_voronoi")
```
rajeeja, line 946: Two scenarios:

1. ux.open_grid(verts, build_connectivity="triangular" or "voronoi") - this automatically returns a grid with elements
2. grid_with_verts_only = ux.open_grid(verts)  -> 

make the 2nd 
`call grid_with_verts_conn = grid_with_verts_only.from_vertices(method="spherical_voronoi")`
we can use the same object as done above? which one is better?

uxarray/grid/grid.py
```diff
@@ -915,3 +917,55 @@ def to_shapely_polygons(self, correct_antimeridian_polygons=True):
...
+            # Convert the normalized vertices back to a NumPy array
+            normalized_verts = np.array(normalized_verts)
+
+            # Perform Spherical Voronoi Construction
+            grid = SphericalVoronoi(normalized_verts)
+
+            # Handle special cases near the antimeridian and poles if necessary
+            # (e.g., split Voronoi cells that cross the antimeridian)
+
+            # TODO: Assign Mesh2 values to the grid
```
rajeeja, line 960: [This](https://stackoverflow.com/questions/68324079/how-to-truly-calculate-a-spherical-voronoi-diagram-using-sf) shows some example how to do it

It might be easier to do this with cartesian grid.

rajeeja (changes requested):

rajeeja:
add a notebook example

test/test_grid.py
aaronzedwick: I think the first one might make the most sense. If we detect only points being sent in, would it make sense to create the connectivity by default?

rajeeja: > I think the first one might make the most sense. If we detect only points being sent in, would it make sense to create the connectivity by default?
+1

aaronzedwick:
@philipc2 Something in your recent commit broke all the test cases?

philipc2:
> @philipc2 Something in your recent commit broke all the test cases?

Let me fix it! I started merging main into it last night, but haven't finished yet

PR #447, opened 2023-09-07
DRAFT: Constructing Face Centers from Polygon Centroids 

philipc2:


philipc2:
@aaronzedwick 

I've put together a very basic implementation using shapely to compute the centroids of each polygon. When you have a chance, can you compare the results to the Cartesiain one?

aaronzedwick:
> @aaronzedwick
> 
> I've put together a very basic implementation using shapely to compute the centroids of each polygon. When you have a chance, can you compare the results to the Cartesiain one?

Converting to Cartesian and averaging seems worse than the latlon averaging from my tests. I can compare them tomorrow when I can checkout this branch. I think the averaging doesn’t work because it is on a sphere, it would work on a flat plane but not on a circle due to the curvature. If this method is accurate that is good, if not maybe we need weighted averages using the arc distances?

rajeeja:
> > @aaronzedwick
> > I've put together a very basic implementation using shapely to compute the centroids of each polygon. When you have a chance, can you compare the results to the Cartesiain one?
> 
> Converting to Cartesian and averaging seems worse than the latlon averaging from my tests. I can compare them tomorrow when I can checkout this branch. I think the averaging doesn’t work because it is on a sphere, it would work on a flat plane but not on a circle due to the curvature. If this method is accurate that is good, if not maybe we need weighted averages using the arc distances? Just some thoughts I have had

can you give concrete numbers? How much worse?

IMO, averaging our 3-d spherical cartesian coordinates (these coordinates are not mapped into 2d plane, so we it should be reasonably accurate) should have the same result

philipc2:
> > > @aaronzedwick
> > > I've put together a very basic implementation using shapely to compute the centroids of each polygon. When you have a chance, can you compare the results to the Cartesiain one?
> > 
> > 
> > Converting to Cartesian and averaging seems worse than the latlon averaging from my tests. I can compare them tomorrow when I can checkout this branch. I think the averaging doesn’t work because it is on a sphere, it would work on a flat plane but not on a circle due to the curvature. If this method is accurate that is good, if not maybe we need weighted averages using the arc distances? Just some thoughts I have had
> 
> can you give concrete numbers? How much worse?
> 
> IMO, averaging our 3-d spherical cartesian coordinates (these coordinates are not mapped into 2d plane, so we it should be reasonably accurate) should have the same result

A quick analysis into the SCRIP gridfile_CSne8 dataset using these calculations, comparing their absolute errors. 

The `face_x` values differ on average by `0.111` degrees and the `face_y` values differ on average by `0.116` degrees.

The maximum difference for `face_x` is `0.315` degrees and `1.056` degrees for `face_y`

Need to look into how centroids are being calculated for any `MultiPolygon` that we have split along the antemeridian.

philipc2:
> degree

This is comparing an expected face x/y in degrees to the computed ones from the grid that I mentioned, not between cartesiain

philipc2:
> > 
> 
> What means 0.315 degree difference (cartesian converted to degrees and cals done on degrees)? Do you know the exact value to say which one is more accurate?

Feel free to take a look at the test case that I pushed last night.

philipc2:
Closing this in favor of #466

PR #466, opened 2023-09-20
Construct `Mesh2_face_x/y` using Cartesian Averaging

### LINKED ISSUE #340: Functions for Constructing `Mesh2_face_x` & `Mesh2_face_y`
Create functions for constructing the UGRID coordinate variables `Mesh2_face_x` and `Mesh2_face_y`, which represent the center longitude and latitude of each face respectively. 

`_build_face_coordinates` or equivalent

aaronzedwick:
Closes #340

## Overview

Constructs the Mesh2_face coordinates for the centroids of faces, using cartesian averaging. If the coordinates are sent in as latlon it will convert to cartesian, do the calculation, and then convert back to latlon.
## Expected Usage

```Python
import uxarray as ux

grid = ux.open_grid(path)
_centroid_from_mean_verts(grid)

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


aaronzedwick:
@philipc2 Do you know any reason why it has an error when converting back to latlon?

uxarray/grid/centroids.py
```diff
@@ -0,0 +1,64 @@
+import xarray as xr
+import numpy as np
+from uxarray.grid.coordinates import node_xyz_to_lonlat_rad, node_lonlat_rad_to_xyz
+
+
+def centroid_from_mean_verts(self, repopulate=True):
```
philipc2, line 6: We should move this into `grid/coordinates.py`

uxarray/grid/centroids.py
```diff
@@ -0,0 +1,64 @@
...
+    latlon = False
+    if "Mesh2_node_z" not in self._ds:
+        latlon = True
+        lon = self.Mesh2_node_x.values
+        lat = self.Mesh2_node_y.values
+        [node_x, node_y, node_z] = node_lonlat_rad_to_xyz([lon, lat])
+    else:
+        node_x = self.Mesh2_node_cart_x.values
+        node_y = self.Mesh2_node_cart_y.values
+        node_z = self.Mesh2_node_cart_z.values
```
philipc2, line 18: We shouldn't need these checks, since directly calling `self.Mesh2_node_cart_x/y/z` should check and construct them on call. 

Example Below

https://github.com/UXARRAY/uxarray/blob/ba113ae96de1d2931a2d83336b8401f7e008ab68/uxarray/grid/grid.py#L356-L366

uxarray/grid/centroids.py
philipc2, line 6: We should move this to `grid/coordinates.py`

philipc2:
> @philipc2 Do you know any reason why it has an error when converting back to latlon?

You may want to use these functions for the conversion

https://github.com/UXARRAY/uxarray/blob/ba113ae96de1d2931a2d83336b8401f7e008ab68/uxarray/grid/coordinates.py#L177

https://github.com/UXARRAY/uxarray/blob/ba113ae96de1d2931a2d83336b8401f7e008ab68/uxarray/grid/coordinates.py#L110

These don't actually populate our variables in `grid._ds`, they simply perform the appropriate conversion.

uxarray/grid/centroids.py
philipc2: This should also be an internal function with a leading `_`

aaronzedwick:
> > @philipc2 Do you know any reason why it has an error when converting back to latlon?
> 
> You may want to use these functions for the conversion
> 
> https://github.com/UXARRAY/uxarray/blob/ba113ae96de1d2931a2d83336b8401f7e008ab68/uxarray/grid/coordinates.py#L177
> 
> https://github.com/UXARRAY/uxarray/blob/ba113ae96de1d2931a2d83336b8401f7e008ab68/uxarray/grid/coordinates.py#L110
> 
> These don't actually populate our variables in `grid._ds`, they simply perform the appropriate conversion.

Thanks! I didn't realize there were two different conversion functions!

uxarray/grid/grid.py
```diff
@@ -410,14 +409,24 @@ def Mesh2_node_cart_y(self) -> xr.DataArray:
...
-        if "Mesh2_face_y" in self._ds:
-            return self._ds["Mesh2_face_y"]
-        else:
-            return None
+        if "Mesh2_face_y" not in self._ds:
+            _centroid_from_mean_verts(self)
+        return self._ds['Mesh2_face_y']
+
+    @property
+    def Mesh2_face_z(self) -> xr.DataArray:
```
aaronzedwick, line 421: Now that I think of it do we actually want this since sometimes you won't have a face_z?

philipc2: Yeah, we don't need `Mesh2_face_z` since this has no meaning with lat/lon coords.

We will however have a `Mesh2_face_cart_z` that's getting implemented in #455

uxarray/grid/coordinates.py
```diff
@@ -235,3 +234,72 @@ def _populate_lonlat_coord(grid):
...
+        if "Mesh2_face_y" not in self._ds or repopulate:
+            self._ds["Mesh2_face_y"] = xr.DataArray(
+                mesh2_face_y,
+                dims=["nMesh2_face"],
+                attrs={"standard_name": "degrees_north"})
+        if "Mesh2_face_z" not in self._ds or repopulate:
+            self._ds["Mesh2_face_z"] = xr.DataArray(
+                mesh2_face_z,
+                dims=["nMesh2_face"],
+                attrs={"standard_name": "elevation"})
```
philipc2, line 305: Let's populate all of these in one shot, so there's no need to have a latlon check.

philipc2, line 305: Change these to `Mesh_face_cart_xyz` once the properties are implemented

test/test_centroids.py
```diff
@@ -0,0 +1,21 @@
...
+class TestCentroids(TestCase):
+
+    def test_centroids_from_mean_verts(self):
+        path = xr.open_dataset(current_path / "meshfiles" / "mpas" / "QU" /
+                               "mesh.QU.1920km.151026.nc")
+
+        # Open the dataset and find the centroids
+        grid = ux.open_grid(path)
+        _centroid_from_mean_verts(grid)
+        self.assertIsNotNone(grid.Mesh2_face_x)
```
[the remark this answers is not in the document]
aaronzedwick: Sure, what would that look like? Since I don't have values to compare, should I just take the mean and compare it? Seems slightly redundant but I am not sure how else to check for validity.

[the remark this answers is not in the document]
philipc2: Yeah, I would make a simple face, one with four known vertices, and then run the algorithm to see if it matches the expect centroid from your manual computations.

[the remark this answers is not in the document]
philipc2: Something like this, from the remapping PR

https://github.com/UXARRAY/uxarray/blob/e3f9e0482e99a0180b09d1087716f3d633009cb2/test/test_remap.py#L26-L57

aaronzedwick:
> Let's actually implement the following properties so we can populate both the latlon and Cartesiain centroid coordinates
> 
> ```python
>     @property
>     def Mesh2_face_cart_x(self) -> xr.DataArray:
>         """Coordinate ``Mesh2_face_cart_x``, which contains the Cartesian x
>         location of each face center in meters.
>         Dimensions (``nMesh2_face``)
>         """
>         if "Mesh2_face_cart_x" not in self._ds:
>             _populate_xyz_centroid_coords()
>         return self._ds["Mesh2_face_cart_x"]
> 
>     @property
>     def Mesh2_face_cart_y(self) -> xr.DataArray:
>         """Coordinate ``Mesh2_face_cart_y``, which contains the Cartesian x
>         location of each face center in meters.
>         Dimensions (``nMesh2_face``)
>         """
>         if "Mesh2_face_cart_x" not in self._ds:
>                 _populate_xyz_centroid_coords()
>         return self._ds["Mesh2_face_cart_y"]
> 
>     @property
>     def Mesh2_face_cart_z(self) -> xr.DataArray:
>         """Coordinate ``Mesh2_face_cart_z``, which contains the Cartesian x
>         location of each face center in meters.
>         Dimensions (``nMesh2_face``)
>         """
>         if "Mesh2_face_cart_z" not in self._ds:
>             _populate_xyz_centroid_coords()
>         return self._ds["Mesh2_face_cart_z"]
> ```

So should I make a separate function for _populate_xyz_centroid_coords() instead of just calling _centroid_from_mean_verts()?

philipc2:
> > Let's actually implement the following properties so we can populate both the latlon and Cartesiain centroid coordinates
> > ```python
> >     @property
> >     def Mesh2_face_cart_x(self) -> xr.DataArray:
> >         """Coordinate ``Mesh2_face_cart_x``, which contains the Cartesian x
> >         location of each face center in meters.
> >         Dimensions (``nMesh2_face``)
> >         """
> >         if "Mesh2_face_cart_x" not in self._ds:
> >             _populate_xyz_centroid_coords()
> >         return self._ds["Mesh2_face_cart_x"]
> > 
> >     @property
> >     def Mesh2_face_cart_y(self) -> xr.DataArray:
> >         """Coordinate ``Mesh2_face_cart_y``, which contains the Cartesian x
> >         location of each face center in meters.
> >         Dimensions (``nMesh2_face``)
> >         """
> >         if "Mesh2_face_cart_x" not in self._ds:
> >                 _populate_xyz_centroid_coords()
> >         return self._ds["Mesh2_face_cart_y"]
> > 
> >     @property
> >     def Mesh2_face_cart_z(self) -> xr.DataArray:
> >         """Coordinate ``Mesh2_face_cart_z``, which contains the Cartesian x
> >         location of each face center in meters.
> >         Dimensions (``nMesh2_face``)
> >         """
> >         if "Mesh2_face_cart_z" not in self._ds:
> >             _populate_xyz_centroid_coords()
> >         return self._ds["Mesh2_face_cart_z"]
> > ```
> 
> So should I make a separate function for _populate_xyz_centroid_coords() instead of just calling _centroid_from_mean_verts()?


We can consider that, however I think a simple `_populate_centroid_coords` would be adequate. I would also write a helper called `_construct_xyz_centroids(node_x, node_y, node_z)` similar to the other helper functions in this module (the `_populate_centroid_coords` would call `_construct_xyz_centroids()` .

philipc2 (changes requested):

uxarray/grid/coordinates.py
```diff
@@ -235,3 +234,69 @@ def _populate_lonlat_coord(grid):
...
+    vertices."""
+
+    node_x = self.Mesh2_node_cart_x.values
+    node_y = self.Mesh2_node_cart_y.values
+    node_z = self.Mesh2_node_cart_z.values
+    face_nodes = self.Mesh2_face_nodes.values
+    nNodes_per_face = self.nNodes_per_face.values
+
+    mesh2_face_x, mesh2_face_y, mesh2_face_z = _construct_xyz_centroids(
+        node_x, node_y, node_z, face_nodes, nNodes_per_face)
```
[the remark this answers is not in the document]
aaronzedwick: Doesn't that defeat the purpose of repopulate though? If they already have latlon and they want to repopulate latlon with our definition of centroid, this wouldn't do that. Same for if we do the same thing for the case when we have cartesian coordinates in our dataset.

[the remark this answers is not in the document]
philipc2: These are the following scenarios that we can end up in:
1. `Mesh2_face_x/y` don't exist
2. `Mesh2_face_x/y` exist, and we want to construct `Mesh2_face_cart_x/y/z` without repopulating the original `Mesh2_face_x/y`
3. `Mesh2_face_x/y` exist, and we want to re-compute them using this algorithm

#### Case 1
Nothing needs to be done here

#### Case 2
For this example, if the user wants the preserve the `Mesh2_face_x/y` that already exist within the `Grid` (which we should by default keep, since most of the time the centroid calculations that are parsed are sound), we should simply perform the Cartesian conversion to populate `Mesh2_face_cart_x/y/z` 

#### Case 2
I'm not 100% sure how we should handle this case, since given the way we have our attributes set up, there's no way to pass through the `repopulate=True` parameter. I think for now, we shouldn't worry about the repopulate tag (we can keep it until a good way comes up to expose it to the user)

uxarray/grid/grid.py
```diff
@@ -375,14 +375,13 @@ def Mesh2_node_cart_x(self) -> xr.DataArray:
     @property
     def Mesh2_face_x(self) -> xr.DataArray:
         """UGRID Coordinate Variable ``Mesh2_face_x``, which contains the
-        longitude of each face center.
+        longitude or cartesian coordinate of each face center.
```
philipc2, line 378: `Mesh2_face_x` can only contain the longitude of each face center, this seems to be left-over from a previous version of this PR. we can revert back to the original docstring.

uxarray/grid/grid.py
```diff
@@ -410,14 +409,46 @@ def Mesh2_node_cart_y(self) -> xr.DataArray:
     @property
     def Mesh2_face_y(self) -> xr.DataArray:
         """UGRID Coordinate Variable ``Mesh2_face_y``, which contains the
-        latitude of each face center.
+        latitude or cartesian coordinate of each face center.
```
philipc2, line 412: Same comment as above

philipc2 (approved):
Fanatic work Aaron! Only two small corrections left. I've also added a test case to compare your computed value to a SCRIP dataset with `face_x/y` values, and they are matching up!

uxarray/grid/coordinates.py
```diff
@@ -235,3 +234,88 @@ def _populate_lonlat_coord(grid):
             "long_name": "latitude of mesh nodes",
             "units": "degrees_north",
         })
+
+
+def _populate_centroid_coord(self, repopulate=False):
```
hongyuchen1030, line 239: Please add a proper description of the limitations of this function and adequate warning. As far as I can tell, this approach can only be applied to some convex polygon (ideally triangle). An exception should be raised if the input polygons are not within the criteria. 

Also, please include the definition of centroid here.

test/test_centroids.py
```diff
@@ -0,0 +1,49 @@
...
+from uxarray.grid.coordinates import _populate_centroid_coord
+
+current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+
+gridfile_CSne8 = current_path / "meshfiles" / "scrip" / "outCSne8" / "outCSne8.nc"
+
+
+class TestCentroids(TestCase):
+
+    def test_centroids_from_mean_verts(self):
```
hongyuchen1030, line 16: Please include testcases that are more complicate:

1. A polygon that has more than 3 vertexes
2. A polygon that one of its vertexes are far away from the other
3. Other irregular shape polygon

hongyuchen1030 (changes requested):

uxarray/grid/coordinates.py
aaronzedwick: Thanks for the review Hongyu! It is my understanding this would work with any convex polygon. Since our definition here is an average of vertices, any convex polygon would be able to be used, if I understand it correctly @philipc2

uxarray/grid/coordinates.py
```diff
@@ -235,3 +234,88 @@ def _populate_lonlat_coord(grid):
...
+
+    # Calculate the mean xyz for all the faces
+    for cur_face_nodes, n_nodes in zip(face_nodes, nNodes_per_face):
+        centroid_x = np.append(centroid_x,
+                               np.mean(node_x[cur_face_nodes[:n_nodes]]))
+        centroid_y = np.append(centroid_y,
+                               np.mean(node_y[cur_face_nodes[:n_nodes]]))
+        centroid_z = np.append(centroid_z,
+                               np.mean(node_z[cur_face_nodes[:n_nodes]]))
+
```
hongyuchen1030, line 320: Without normalization, we can't be sure that these points are on the unit sphere. Please at least call the normalizaiton process here

uxarray/grid/coordinates.py
hongyuchen1030: That's what I was talking about. If we can only confirm this will only be valid within a convex polygon, we should include that in the description. And if the input polygon is not convex, a warning should be raised. Otherwise, someone will get a silence crash when they're trying to apply this function on a concave polygon, which we don't want to see happening.

Another thing that brought up my concern here is that we are working on the spherical surface. So my questions here are:

1. The coordinates averaging works in the 3-dimensional space for "planes" like n-dimensional simplex. Does it work for the spherical surface?

2. Even though we assume it works, we need to add the normalization step for your centroid calculation function (I also added a comment there). After we project it to the spherical surface using this naive method, How can we confirm that the projected centroid (if we assume it's the centroid we want) is still within the polygon on the sphere? Because the edge might arc up on the sphere

philipc2, line 239: Missed this in my review! Please replace all instances of `self` with `grid` throughout this function.

philipc2 (changes requested):

test/test_centroids.py
philipc2: I would say that our test `test_centroids_from_mean_verts_scrip` is pretty comprehensive, since it runs the algorithm on a dataset that already has `Mesh2_face_x/y` and the results match up. This dataset is also made up of quads, which checks off the first point.

hongyuchen1030: > I would say that our test `test_centroids_from_mean_verts_scrip` is pretty comprehensive, since it runs the algorithm on a dataset that already has `Mesh2_face_x/y` and the results match up. This dataset is also made up of quads, which checks off the first point.
> 
> However, additional tests would be good to have.

So does this SCRIP dataset use a real spherical surface presentation (cruved) or an approximate one using triangles?? 

The point is that only averaging the coordinates without normalization (our current implementation), the centroid is not actually on the sphere (Unless that's exactly what we want)

philipc2: > > I would say that our test `test_centroids_from_mean_verts_scrip` is pretty comprehensive, since it runs the algorithm on a dataset that already has `Mesh2_face_x/y` and the results match up. This dataset is also made up of quads, which checks off the first point.
> > However, additional tests would be good to have.
> 
> So does this SCRIP dataset use a real spherical surface presentation (curved) or an approximate one using triangles?
> 
> The point is that only averaging the coordinates without normalization (our current implementation), the centroid is not actually on the sphere (Unless that's exactly what we want)

I believe so

*[SCRIP](https://github.com/SCRIP-Project/SCRIP): A Spherical Coordinate Remapping and Interpolation Package (SCRIP) is a software package used to generate interpolation weights for remapping fields from one grid to another in spherical geometry.  It was developed at the Los Alamos National Laboratory by Philip W. Jones. The SCRIP library was converted from Fortran to ANSI C and is used as the base for remapping operators in CDO.

aaronzedwick: The SCRIP test doesn’t test the accuracy well, because if latlon already exists it doesn’t recreate it using our definition. It just assigns it to what it already is. Really all it does is convert it to get the Cartesian. This is what I was saying about repopulation not doing anything in that regard.

philipc2: > The SCRIP test doesn’t test the accuracy well, because if latlon already exists it doesn’t recreate it using our definition. It just assigns it to what it already is. Really all it does is convert it to get the Cartesian. This is what I was saying about repopulation not doing anything in that regard.

I might be missing something, but are you sure this is the case? I've stepped through the execution of the test case and the population function it does indeed compute `centroid_lon` and `centroid_lat` and then re-populates the data array.

#### `Mesh2_face_x/y` and  `_populate_centroid_coord(uxgrid, repopulate=True)` is called

##### Computation
```Python
# block (1)
    if "Mesh2_face_x" not in self._ds or repopulate:
        # Construct the centroids if there are none stored
        if "Mesh2_face_cart_x" not in self._ds:
            centroid_x, centroid_y, centroid_z = _construct_xyz_centroids(
                node_x, node_y, node_z, face_nodes, nNodes_per_face)

        else:
            # If there are cartesian centroids already use those instead
            centroid_x, centroid_y, centroid_z = self.Mesh2_face_cart_x, self.Mesh2_face_cart_y, self.Mesh2_face_cart_z
# block (2)
        # Convert from xyz to latlon
        centroid_lon, centroid_lat = _get_lonlat_from_xyz(
            centroid_x, centroid_y, centroid_z)
# block (3)
    else:
        # Convert to xyz if there are latlon centroids already stored
        centroid_lon, centroid_lat = self.Mesh2_face_x.values, self.Mesh2_face_y.values
        centroid_x, centroid_y, centroid_z = _get_xyz_from_lonlat(
            centroid_lon, centroid_lat)
```

We enter block (1) since `repopulate` is set to True, which computes our `centroid_x/y/z`

Block (2) is followed and we computed out `centroid_lat/lon` (recomputed, not the original `Mesh2_face_x/y`

We never actually enter block (3)


Any thoughts?

philipc2:
@rajeeja Any thoughts on this PR?

uxarray/grid/coordinates.py
philipc2: > that's what I was talking about. If we can only confirm this will only be valid within a convex polygon

@paullric @rljacob @erogluorhan 

Do we have any sample mesh files that contain concave faces? I'm interested in how often we may encounter concave faces when working with, for example, grids from climate models, since that's our most applicable use case.

test/test_centroids.py
aaronzedwick: Oh yeah, you are right, somehow I forgot about the or statement there, my bad! That’s actually really good then.

rajeeja:
Looks good, should check normalization.

rajeeja (approved):
Good work!

We can revisit our internal variable names in a future PR, after removing Mesh2 prefix.

test/test_centroids.py
philipc2: > Oh yeah, you are right, somehow I forgot about the or statement there, my bad! That’s actually really good then.

All good!

hongyuchen1030: > > > I would say that our test `test_centroids_from_mean_verts_scrip` is pretty comprehensive, since it runs the algorithm on a dataset that already has `Mesh2_face_x/y` and the results match up. This dataset is also made up of quads, which checks off the first point.
> > > However, additional tests would be good to have.
> > 
> > 
> > So does this SCRIP dataset use a real spherical surface presentation (curved) or an approximate one using triangles?
> > The point is that only averaging the coordinates without normalization (our current implementation), the centroid is not actually on the sphere (Unless that's exactly what we want)
> 
> I believe so
> 
> *[SCRIP](https://github.com/SCRIP-Project/SCRIP): A Spherical Coordinate Remapping and Interpolation Package (SCRIP) is a software package used to generate interpolation weights for remapping fields from one grid to another in spherical geometry. It was developed at the Los Alamos National Laboratory by Philip W. Jones. The SCRIP library was converted from Fortran to ANSI C and is used as the base for remapping operators in CDO.

A spherical coordinate system doesn't necessarily mean they are using a **curve** surface here. Many application utilizes triangles to approximate the sphere, which works fine as well, it's just not our case.

Again, in the geometry calculation, directly averaging the cartesian coordinates of a **curved** "triangle" in the 3-D spaces, the resulted "centroid" shouldn't be on the spherical surface. 

However, the normalization process can project this "centroid" to the surface. But I am not sure if this "projected" centroid is the real centroid according to the definition.

The curved surface is way more complicated than the normal plane surface. And it's understandable that we want to stay with the current implementation to expedite the process as well. Just make sure to clarify that in the comment.

uxarray/grid/coordinates.py
```diff
@@ -235,3 +236,96 @@ def _populate_lonlat_coord(grid):
...
+            attrs={"standard_name": "cartesian y"})
+
+        grid._ds["Mesh2_face_cart_z"] = xr.DataArray(
+            centroid_z,
+            dims=["nMesh2_face"],
+            attrs={"standard_name": "cartesian z"})
+
+
+@njit(cache=ENABLE_JIT_CACHE)
+def _construct_xyz_centroids(node_x, node_y, node_z, face_nodes,
```
philipc2, line 310: @aaronzedwick 

Instead of directly adding a review for this, I went ahead and modified this function to work with Numba so we can move forward with getting this in for this iteration.

philipc2 (approved):
Excellent work Aaron!

PR #495, opened 2023-10-04
FutureWarning for `Mesh2` Prefix Removal

### LINKED ISSUE #494: Raise FutureWarning for `Mesh2_` variables
https://github.com/UXARRAY/uxarray/discussions/387 and #389 plan to remove the `Mesh2` prefix from grid dimensions, coordinates, and connectivity variables. We should raise a FutureWarning to let the userbase know.

philipc2:
Closes #494

### DIFF
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -134,7 +134,6 @@ class Grid:
         Only raises the warning once when a effected property is called.
         """
         if not self._mesh2_warning_raised:
-            print("HELLO")
             self._mesh2_warning_raised = True
             warn(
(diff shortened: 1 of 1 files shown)


rajeeja (approved):
Looks good. Quick question: when we remove the prefix, we just remove the calls to this function, right?

Possible to make this a generic deprecation function with Mesh2 and some deprecation message as argument?

philipc2:
> Looks good. Quick question: when we remove the prefix, we just remove the calls to this function, right?

That's the intention.

> Possible to make this a generic deprecation function with Mesh2 and some deprecation message as argument?

I was considering this, however I only want the warning to be raised once, since having it showed up when accessing any `Mesh2_` variable would end up getting really cluttered.

aaronzedwick (approved):

PR #505, opened 2023-10-09
Point Rasterization

### LINKED ISSUE #492: Support Rasterized Point Plots
> Example can be seen [here](https://gist.github.com/philipc2/fa748514107fea8ee09910f706a25cc0)

philipc2:
Closes #492

## Overview

* Implements Point Rasterization for `UxDataArray`

## Expected Usage

```Python
import uxarray as ux

# rasterized plotting, a few examples
uxds['v1'].plot.rasterize()
uxds['v1'].plot.rasterize(backend='bokeh', width=1000, height=5000)
uxds['v1'].plot.rasterize(pixel_ratio=2.0, projection=ccrs.Robinson())


```

## PR Checklist

**General**
- [x] Add appropriate labels

**Documentation**


uxarray/plot/dataarray_plot.py
```diff
@@ -61,3 +71,159 @@ def datashade(uxda: UxDataArray,
...
+    elif uxda.node_centered():
+        # data mapped to face corner coordinates
+        data_mapping = "corner"
+        lon = uxda.uxgrid.Mesh2_node_x.values
+        lat = uxda.uxgrid.Mesh2_node_y.values
+    else:
+        raise ValueError(
+            "Issue with data. It is neither face-centered nor node-centered!")
+
+    data_values = uxda.values
```
erogluorhan, line 105: `data_values` being called only twice in the rest; let us just use `uxda.values` to help increase readability of the code by avoiding defining too many variables.

uxarray/plot/dataarray_plot.py
```diff
@@ -61,3 +71,159 @@ def datashade(uxda: UxDataArray,
...
+                  colorbar: Optional[bool] = True,
+                  cmap: Optional[str] = "Blues",
+                  aggregator: Optional[str] = "mean",
+                  interpolation: Optional[str] = "linear",
+                  npartitions: Optional[int] = 1,
+                  **kwargs):
+
+    if uxda.face_centered():
+        # data mapped to face centroid coordinates
+        data_mapping = "center"
```
erogluorhan, line 93: `data_mapping` is being called only twice in the rest; let us just use `uxda.face_centered()` when needed (it is not an expensive run) to help increase readability of the code by avoiding defining too many variables.

uxarray/plot/dataarray_plot.py
```diff
@@ -61,3 +71,159 @@ def datashade(uxda: UxDataArray,
         _cmap = cmap
 
     return tf.shade(aggregated, cmap=_cmap, **kwargs)
+
+
+def _point_raster(uxda: UxDataArray,
```
erogluorhan, line 76: positional args, i.e.  `*args` here

uxarray/plot/dataarray_plot.py
```diff
@@ -61,3 +71,159 @@ def datashade(uxda: UxDataArray,
...
+        if uxda.uxgrid._corner_points_df_proj[
+                0] is not None and uxda.uxgrid._corner_points_df_proj[
+                    1] == projection:
+            recompute = False
+            points_df = uxda.uxgrid._corner_points_df_proj[0]
+
+    if recompute:
+        # need to recompute points & projection
+        if projection is not None:
+            lon, lat, _ = projection.transform_points(ccrs.PlateCarree(), lon,
```
erogluorhan, line 125: What is the cost of calculating this again and again vs. storing this? e.g. Time and memory costs, respectively, for the 3.75km maybe? **For point rasterization**, execution costs for any of the functions being called in this method were reasonable small as far as I remember, so any caching might be unnecessary.

uxarray/plot/dataarray_plot.py
```diff
@@ -61,3 +71,159 @@ def datashade(uxda: UxDataArray,
...
+        else:
+            uxda.uxgrid._corner_points_df_proj[0] = point_ddf
+            uxda.uxgrid._corner_points_df_proj[1] = projection
+
+    else:
+        # can use existing cached points & projection
+        points_df['var'] = pd.Series(data_values)
+        points = hv.Points(points_df, ['lon', 'lat'])
+
+    if backend == "matplotlib":
```
erogluorhan, line 148: It seems like, when backend is MPL, we are not using the width and height provided. Shouldn't we?

uxarray/core/dataarray.py
```diff
@@ -282,3 +282,11 @@ def integrate(self,
...
 
         return uxda
+
+    def face_centered(self) -> bool:
+        return (self.uxgrid.nMesh2_face == self.shape[-1] and
+                self.uxgrid.nMesh2_node not in self.shape)
+
+    def node_centered(self) -> bool:
+        return (self.uxgrid.nMesh2_node == self.shape[-1] and
+                self.uxgrid.nMesh2_face not in self.shape)
```
erogluorhan, line 292: Let us keep these two functions in the internal API (and not forget to list them under docs/internal/index.rst). 

This is because there might not be any need to have such user API functions since the task they perform is straightforward for the use by just checking the array dimensions and comparing them to grid values (but calculating that info becomes vital for the visualization functions like here)

erogluorhan (commented):
Great overall; please address a few comments though.

uxarray/plot/dataarray_plot.py
philipc2: These timings are a rough benchmark, however this is what I observed.

image

Additionally, the caching here is of the Dask dataframe, not just the transformed coordinates.

uxarray/plot/dataarray_plot.py
philipc2: The MPL backend does not accept width/height arguments. It does, however, accept a `fig_size` which is the closest similar argument.

https://holoviews.org/user_guide/Plotting_with_Matplotlib.html#plot-size

uxarray/core/dataarray.py
philipc2: > This is because there might not be any need to have such user API functions since the task they perform is straightforward for the use by just checking the array dimensions and comparing them to grid values (but calculating that info becomes vital for the visualization functions like here)

While it may not seem like there is a need, I also don't see any harm in exposing it to the user. However, if we still want to keep it internal I can make that change.

uxarray/plot/dataarray_plot.py
philipc2: I'm still a bit hesitant to why we need to include `*args` in our function definition, as I don't see a use-case where having a non-fixed amount of positional arguments would ever be used in any of the functions that we call, unless it's a direct wrapper to a plotting method.

I've removed it from the rasterize function unless we can come up with a valid use case for it.

erogluorhan: Ok, that makes sense, we might likely never need positional args.

uxarray/plot/dataarray_plot.py
erogluorhan: Yeah, that's the Dask dataframe. This chart shows great the exceution times. What's the memory space our cached attributes take?

uxarray/plot/dataarray_plot.py
erogluorhan: Yeah, do you think we should take these optional args in as "hv_width, hv_height", and "mpl_figsize" and document which one is for what backend?

uxarray/core/dataarray.py
erogluorhan: One concern is that the user API doc gets overpopulated and less readable/helpful. So, I'd still ask to keep them internal unless there is explicit request from the user. Would that make sense?

philipc2: That does make sense, but typically the API documentation isn't mean to be "readable", it should be a comprehensive list that someone can quickly search for. 

The readability comes from our usage example and (soon) the user guides.

uxarray/plot/dataarray_plot.py
philipc2: For the 3.75km dataset, the Dask dataframe occupies about 3 * (42mb) for the (lon, lat, data var) when persisted in memory. This is in addition to the list that stores the data frame, along with the projection reference, which are essentially points, so it is minimal .

uxarray/plot/dataarray_plot.py
philipc2: That could work, but I don't like the "hv" or "mpl" prefix in front of these variables. In relativity, any backend-specific arguments should be handled by `**kwargs`. 

We could keep the `height` and `width` the same, include a small blurb in it's docstring, and introduce a default `fig_size` . What do you think?

uxarray/core/dataarray.py
erogluorhan: I agree with the comprehensiveness of the API doc but wouldn't agree with that it's not meant to be readable. One should be able to practically browse user API docs and sometimes read a number of functions (even sometimes in in a sequence just to explore) and shouldn't feel overwhelmed due to too many things. That said, my main point here is that we shouldn't overpopulate our user API with too many functions regardless of if they'd be used or not by the end-user. 

I believe these functions that check if data is face- or node-centered are good internal API functions but not needed in the user API.

philipc2: I see your point now! This makes sense.

I'll add a docstring to them and make them internal

uxarray/plot/dataarray_plot.py
erogluorhan: Is it bit (mb) or byte (MB)? If bit, that seems pretty viable. If byte though, 3 * 42 MB may still seem reasonable but adds to the total memory usage UXarray has (along with the other cached data). I've seen memory usages in the margins of 32 GB while performing such visualizations with 7.5km data, which seems huge. I couldn't say anything further for now since we don't have any detailed analysis, but we should take that into account.

One intermediary solution maybe the `cache=False` optional arg, and if user wants it `True` we would do caching (as we did that with other caching things in the past).

uxarray/plot/dataarray_plot.py
erogluorhan: Sounds good to me.

uxarray/plot/dataarray_plot.py
philipc2: This was in Mega Bytes

I will add a cache parameter to the function.

erogluorhan: Oops, if it is MB, then yeah having an optional arg makes it a bit safer I think.

philipc2: Good suggestion, I've added it.

uxarray/plot/dataarray_plot.py
erogluorhan: Great!

erogluorhan (approved):
Great work!

uxarray/grid/grid.py
```diff
@@ -119,6 +119,9 @@ def __init__(self,
         self._gdf = None
         self._poly_collection = None
         self._line_collection = None
+        self._centroid_points_df_proj = [None, None]
```
rajeeja, line 122: Is this documented?
what is `..df_proj`?

uxarray/grid/grid.py
```diff
@@ -119,6 +119,9 @@ def __init__(self,
...
+        self._corner_points_df_proj = [None, None]
```
rajeeja, line 123: centroid is face centered data and corner is node data, is that correct?

uxarray/plot/dataarray_plot.py
```diff
@@ -61,3 +71,168 @@ def datashade(uxda: UxDataArray,
...
+        if uxda.uxgrid._corner_points_df_proj[
+                0] is not None and uxda.uxgrid._corner_points_df_proj[
+                    1] == projection:
+            recompute = False
+            points_df = uxda.uxgrid._corner_points_df_proj[0]
+
+    if recompute:
+        # need to recompute points and/or projection
+        if projection is not None:
+            lon, lat, _ = projection.transform_points(ccrs.PlateCarree(), lon,
```
rajeeja, line 125: ccrs is one of the options for projection?
`projection: Optional[ccrs] = None,` 

what other options? hard-coded?

rajeeja (changes requested):

rajeeja:
can we support  variables that reside on the corner nodes by averaging and treating as centroidal data for visualization?

uxarray/grid/grid.py
philipc2: Yes

philipc2:
> can we support variables that reside on the corner nodes by averaging and treating as centroidal data for visualization?

This will be added with #514

uxarray/grid/grid.py
philipc2: it's internally used in our point raster for caching, it's not referenced anywhere else

It's a list containing our Point Data Frame and Projection respectively.

uxarray/plot/dataarray_plot.py
philipc2: The typing here is only for hints, so it doesn't restrict us, but it's expected to be a Cartopy Projection.

rajeeja (approved):

PR #510, opened 2023-10-12
Is pole point inside

### LINKED ISSUE #486: Detect if the pole point is inside a convex polygon
A clear and concise description of what the problem is.
A function that can detect if the pole point is inside the given polygon or not

hongyuchen1030:
Closes #486 

## Overview
- Added a new utility function `_is_pole_point_inside_polygon` to determine if a pole point is inside a polygon.
- The function raises a `ValueError` if the pole is neither 'North' nor 'South'.
- This function works for any face as long as it's within a hemisphere
- A warning is issued : To use this function, the given face cannot reach out the other hemisphere. For example, if you
        want to check if the North pole is inside a polygon, the polygon should not reach the south hemisphere.

## Expected Usage

```Python
# This function is part of the internal API, so it's primarily for internal use within the library.

# Example of how the function might be used:

pole = 'North'
face_edge_cart = ... # some numpy array representation
result = _is_pole_point_inside_polygon(pole, face_edge_cart)
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


uxarray/grid/geometry.py
```diff
@@ -242,3 +244,58 @@ def _grid_to_matplotlib_linecollection(grid):
...
+    Warning
+    -----
+    UserWarning
+        To use this function, the given face cannot reach out the other hemisphere. For example, if you
+        want to check if the North pole is inside a polygon, the polygon should not reach the south hemisphere.
+    """
+    POLE_POINTS = {
+        'North': np.array([0.0, 0.0, 1.0]),
+        'South': np.array([0.0, 0.0, -1.0])
+    }
```
philipc2, line 282: These should be declared at the top of this module, not inside of the function.

uxarray/grid/geometry.py
```diff
@@ -242,3 +244,58 @@ def _grid_to_matplotlib_linecollection(grid):
...
+    pole_point = POLE_POINTS[pole]
+    ref_point = node_lonlat_rad_to_xyz(np.array(
+        [0.0, 0.0]))  # We just set the reference point to be on the equator
+
+    GCA = np.array([pole_point, ref_point])
+
+    intersection_count = sum(1 for edge in face_edge_cart
+                             if gca_gca_intersection(GCA, edge).size != 0)
+
+    return intersection_count % 2 == 1
```
philipc2, line 301: If I'm understanding this line correctly, it computes how many of the edges in our inputted face intersect the reference GCA.

Since our reference GCA is an arc between the equator (0, 0) and either pole (0, 90) or (0, -90), can you explain how having an odd number of intersections means that our face contains a pole point?

philipc2: Can we generalize this function to perform any "Point in Face" operator? Or is there something special about the pole points?

philipc2:
For this implementation, do we consider the case where one of the corners of our face is either pole point?

uxarray/grid/geometry.py
hongyuchen1030: Unfortunately, it cannot be generalized right now. "a point in a polygon" problem is a very common topic in computational geometry and the "a point in a polygon on a sphere" is known to be challenging as well 

To generalize this problem, we need to deal with performance issues and precision issues, which are too complicate for our current phase now, And plus we only need to know if the pole point inside a given polygon anyway.

But it's my goal for my next research step to devise a fast and accurate "a point in a polygon" problem. But to expedite our current progress, this is the most suitable solution.

hongyuchen1030:
> For this implementation, do we consider the case where one of the corners of our face is either pole point?

Good catch. So technically "one of the corners of our face is pole point" means the pole point is not inside the polygon. But for our next step, we do want to specify this issue. So I am not so sure about how to handle this case.

We have two choices

1. Let this function be strictly "if pole point *inside*", and for the next step we will manually check if one corner of the face is the pole point

2. Make this function a little bit "relax", if the query point, in our case, the pole point is on the edge, we still consider it as "inside" the polygon.

I wonder what's your opinion about it.

hongyuchen1030:
> > For this implementation, do we consider the case where one of the corners of our face is either pole point?
> 
> Good catch. So technically "one of the corners of our face is pole point" means the pole point is not inside the polygon. But for our next step, we do want to specify this issue. So I am not so sure about how to handle this case.
> 
> We have two choices
> 
> 1. Let this function be strictly "if pole point _inside_", and for the next step we will manually check if one corner of the face is the pole point
> 2. Make this function a little bit "relax", if the query point, in our case, the pole point is on the edge, we still consider it as "inside" the polygon.
> 
> I wonder what's your opinion about it.

After thinking about it, probably the second "relax" solution is more suitable

philipc2:
> > > For this implementation, do we consider the case where one of the corners of our face is either pole point?
> > 
> > 
> > Good catch. So technically "one of the corners of our face is pole point" means the pole point is not inside the polygon. But for our next step, we do want to specify this issue. So I am not so sure about how to handle this case.
> > We have two choices
> > 
> > 1. Let this function be strictly "if pole point _inside_", and for the next step we will manually check if one corner of the face is the pole point
> > 2. Make this function a little bit "relax", if the query point, in our case, the pole point is on the edge, we still consider it as "inside" the polygon.
> > 
> > I wonder what's your opinion about it.
> 
> After thinking about it, probably the second "relax" solution is more suitable

I'm fine with either approach, we could add a parameter that switches between having the point strictly within or to include the case where it's one of the edge nodes.

uxarray/grid/geometry.py
philipc2: > Unfortunately, it cannot be generalized right now. "a point in a polygon" problem is a very common topic in computational geometry and the "a point in a polygon on a sphere" is known to be challenging as well
> 
> To generalize this problem, we need to deal with performance issues and precision issues, which are too complicate for our current phase now, And plus we only need to know if the pole point inside a given polygon anyway.
> 
> But it's my goal for my next research step to devise a fast and accurate "a point in a polygon" problem. But to expedite our current progress, this is the most suitable solution.

Got it! Definitely will be on the look out for that, as having that type of functionality will be extremely useful! Keep up the great work.

uxarray/grid/geometry.py
```diff
@@ -242,3 +249,64 @@ def _grid_to_matplotlib_linecollection(grid):
...
+        want to check if the North pole is inside a polygon, the polygon should not reach the south hemisphere.
+    """
+
+    if pole not in POLE_POINTS:
+        raise ValueError('Pole point must be either "North" or "South"')
+
+    warnings.warn(
+        'To use this function, the given face cannot reach out the other hemisphere. For example, if you '
+        'want to check if the North pole is inside a polygon, the polygon should not reach the south hemisphere.'
+    )
```
philipc2, line 292: Do we need to raise this warning everything we call this function? I think having a blurb in the docstring should suffice.

uxarray/grid/geometry.py
```diff
@@ -242,3 +249,64 @@ def _grid_to_matplotlib_linecollection(grid):
 
     # need transform? consider adding it later if needed
     return LineCollection(lines)
+
+
+def _is_pole_point_inside_polygon(pole, face_edge_cart):
```
philipc2, line 254: I think we can drop the `_is` 
```suggestion
def pole_point_inside_polygon(pole, face_edge_cart):
```

philipc2 (changes requested):
A couple small suggestions! looking good!

uxarray/grid/geometry.py
hongyuchen1030: Good point, I thought about that before. But I am not sure what the users will prefer for the handling so I just adopt a very conservative approach here.

But I am working on the algo to deal with the equator-crossing case now and hope it works out!

uxarray/grid/geometry.py
```diff
@@ -1,7 +1,14 @@
 import numpy as np
-
-from uxarray.constants import INT_DTYPE, INT_FILL_VALUE
+from uxarray.grid.coordinates import node_lonlat_rad_to_xyz
```
philipc2, line 2: This import isn't used anywhere here.

uxarray/grid/geometry.py
```diff
@@ -242,3 +249,113 @@ def _grid_to_matplotlib_linecollection(grid):
...
+    UserWarning
+        Raised if the face contains both pole points.
+    """
+    if pole not in POLE_POINTS:
+        raise ValueError('Pole point must be either "North" or "South"')
+
+    # Classify the polygon's location
+    location = _classify_polygon_location(face_edge_cart)
+    pole_point = POLE_POINTS[pole]
+    ref_point = np.array([1.0, 0.0, 0.0])  # Reference point on the equator
```
philipc2, line 288: I would consider declaring this at the top of the module, similar to the pole points as `EQUATOR`

philipc2 (changes requested):
looking good, once we get the tests passing I think it should be ready to merge.

philipc2 (approved):

Issue #96, opened 2022-08-11
Add Gradients (derivatives)


erogluorhan:
Gradients (derivatives)

_Originally posted by @erogluorhan in https://github.com/UXARRAY/uxarray/discussions/46#discussioncomment-2746622_

erogluorhan:
See this [comment](https://github.com/UXARRAY/uxarray/discussions/46#discussioncomment-2746622) for keeping in mind other functions when implementing this.

Issue #400, opened 2023-08-09
Small Test Dataset for Geometry Conversion Methods


philipc2:
Instead of using the mesh files that we have stored, which still contain hundreds or thousands of faces, a small sample dataset with < 10 faces, with or without antemeridian cases would be useful for testing.

philipc2:
This can be easily achieved once #616 is merged

PR #534, opened 2023-10-24
Implement `UxDataArray.gradient()` and `UxDataArray.difference()`

### LINKED ISSUE #525: Populate `edge_node_distances`
`Grid.Mesh2_edge_node_distances` contains the distance between the two nodes that saddle a given edge. We should be able to populate this variable by parsing over either the `(lat, lon)` or `(x, y, z)` coordinates of the two edges and compute the distance.

philipc2:
Closes #525, #400, #96

## Overview

* Implements the first `gradient` method, which computes the computing the horizontal gradient of a field on each
    cell using values at adjacent cells. from Eq. 22 in Ringler et al. (2010), J. Comput. Phys., code is adapted from https://github.com/theweathermanda/MPAS_utilities/blob/main/mpas_calc_operators.py
* Implements the `difference` method, which computes the absolute difference of a data variable across either the edges, nodes, or faces. Currently implemented across edges.
* Construction of `edge_node_distances` and `edge_face_distances` 
* Introduces a new test dataset comprised of 4 hexagons for testing
## Expected Usage

```Python
import uxarray as ux

# gradient of a face-centered data variable
uxds['v1'].gradient()

# normalizing data
uxds['v1'].gradient(normalize=True)

# difference of a data variable
uxds['v1'].difference(destination='edge')


```


philipc2:
@rajeeja @erogluorhan 

This should be ready for an initial review. I'd appreciate any feedback, especially about testing, since I'm unsure of the best way to determine the "correctness"

philipc2:
Pinging @theweathermanda, since she was the one that first initiated the discussion & algorithm used here. Any feedback would be greatly appreciated!

uxarray/core/dataarray.py
```diff
@@ -353,6 +355,49 @@ def nodal_average(self):
             name=self.name + "_nodal_average" if self.name is not None else None,
         ).rename({"n_node": "n_face"})
 
+    def gradient(
+        self, use_magnitude: Optional[bool] = True, normalize: Optional[bool] = True
+    ):
+        """Computes the gradient of a data variable residing on an unstructured
```
erogluorhan, line 361:
```suggestion
        """Computes the horizontal gradient of a data variable residing on an unstructured
```

erogluorhan (commented):
Code looks good to me. Please address just a few points:

1. Test cases: In addition to the existing ones:
    i. Test if the horizontal gradients works as expected if we have "Time" and "Level" (Actually this is an overall testing need for functionality, but only for this PR; I believe we should test this. Thoughts?)
    ii. Test some actual gradient calculations (not only for uniform data). Maybe we could use a very small data by `from_vertices`?

2. What is the plan with edge-centered and node-centered data? Maybe create an issue for that since this will close #96?

philipc2:
Here's a visual of a mostly-uniform value grid, with a couple extreme values at the face centers. Here the color bar is for the gradient on each edge, with values close to white being small gradients, and values near red as higher.


image

These values appear to be as expected, considering right now this implementation uses the magnitude of the gradient, since we don't have any information about incoming / outgoing edges.

rljacob:
I think the PR description should be updated to say its for gradients on MPAS meshes.

philipc2:
> > I think the PR description should be updated to say its for gradients on MPAS meshes.
> 
> It was based of the MPAS algorithm but the implementation here is for any grid format UXarray supports.

Yep! The algorithm here essentially computes the following
* Difference between face values that saddle a given cell (this becomes an edge-centered data variable)
* The difference is then divided by the distance between the face-centers

I'm working on getting a few more plots, showing the non-mpas examples.

philipc2:
> I think the PR description should be updated to say its for gradients on MPAS meshes.

Here's an example using a subset of our CSne30 mesh in the test files, with normalized gradient values in the range [0, 1]

image

philipc2:
> Code looks good to me. Please address just a few points:
> 
> 1. Test cases: In addition to the existing ones:
>    i. Test if the horizontal gradients works as expected if we have "Time" and "Level" (Actually this is an overall testing need for functionality, but only for this PR; I believe we should test this. Thoughts?)
>    ii. Test some actual gradient calculations (not only for uniform data). Maybe we could use a very small data by `from_vertices`?
> 2. What is the plan with edge-centered and node-centered data? Maybe create an issue for that since this will close [Add Gradients (derivatives) #96](https://github.com/UXARRAY/uxarray/issues/96)?

Thank you for the suggestions!

#### 1 (i)
Should be covered by `test_grad_multi_dim`

#### 1 (ii)
`test_quad_hex` should cover this. It checks to make sure the gradients follow the expected values, zero for edges that don't have a neighbor, and a non-zero value otherwise. For this test, do you have any suggestions on how to test on actual gradient computations without simply reimplmenting what is already in `.gradient()`? 

#### 2

Gradients of node-centered data variables is planned, not sure about the application of edge-centered variables. I can create an issue for node-centered gradients.

philipc2:
A couple more examples of the gradient, including comparison to a difference method. 

#### MPAS Bottom Depth Gradient

image

All cells bordering the continents (i.e. those with no neighbors) have zero values

Zoomed in, we can see the larger gradient values for edges that have a large difference in values (in this case, large differences in depth)

image

#### MPAS Bottom Depth Difference

In addition to the ``uxda.gradient()`` method, I've added a ``uxda.difference()`` method, which performs the same computation, however it doesn't divide by the distance between the cell-centers. After discussing with Paul, this is a good way of "approximating' gradients, (It's used in estimating gradients within cells, e.g., in Barth and Jespersen 1989)

image

Zoomed into the same region, we can observe a similar trend. 

image

erogluorhan (approved):
This is very good, and I am approving it, though you can find my comments to your latest points below:

> #### 1 (ii)
> `test_quad_hex` should cover this. It checks to make sure the gradients follow the expected values, zero for edges that don't have a neighbor, and a non-zero value otherwise. For this test, do you have any suggestions on how to test on actual gradient computations without simply reimplmenting what is already in `.gradient()`?

Testing for zero / nonzero values makes sense despite not testing the actual values. To test for actual values, I was referring to `from_vertices()` thinking if we could create a very small grid with only 4-5 faces where we could assign the data values so that we could calculate differences / gradients for them by hand, and the n test them against our function outputs.

philipc2:
> This is very good, and I am approving it, though you can find my comments to your latest points below:
> 
> > #### 1 (ii)
> > `test_quad_hex` should cover this. It checks to make sure the gradients follow the expected values, zero for edges that don't have a neighbor, and a non-zero value otherwise. For this test, do you have any suggestions on how to test on actual gradient computations without simply reimplmenting what is already in `.gradient()`?
> 
> Testing for zero / nonzero values makes sense despite not testing the actual values. To test for actual values, I was referring to `from_vertices()` thinking if we could create a very small grid with only 4-5 faces where we could assign the data values so that we could calculate differences / gradients for them by hand, and the n test them against our function outputs.

I see! If that's the case, the `quad-hex` example only has 4 faces, with 4 data values associated to each face. I can do some computations by hand and expand that test case. Thank you for the clarification!

philipc2:
@erogluorhan 

I've refined the test case for the `quad-hex` example! Here's a plot of the gradients for reference, which match up with the computations I've done by hand.

image

PR #557, opened 2023-10-27
Accurate dot product and accurate operation utils

### LINKED ISSUE #558: Accurate dot product calculation and related accurate operators

Provide a more accurate dot product operation.
And also employ some common accurate computing tools like `2SUM`, `2ProdFMA`, `3FMA`, `Fast2Sum`, `Fast2Mult`, .etc

hongyuchen1030:
Closes #558 

## Overview

In this PR, we have organized and enhanced the utilities for accurate computing within uxarray. All functions related to accurate computing have been consolidated into a single file, `uxarray.grid.accurate_computing_utils.py,` for better maintainability and clarity. 

Additionally, I have implemented a more precise method for dot product calculations, utilizing Fused Multiply-Add (FMA) operations to improve the accuracy and reliability of the results.
## Expected Usage

```Python
import uxarray.grid.accurate_computing_utils as ac_utils


# Utilizing the accurate dot product function
        v1 = np.array(normalize_in_place([1.0, 0.0, 0.0]), dtype=np.float64)
        v2 = np.array(normalize_in_place([1.0, 0.0, 0.0]), dtype=np.float64)
        fma_dot = ac_utils.dot_fma(v1, v2)
        fma_cross = ac_utils.cross_fma(v1, v2)

# Employing other accurate computing utilities
output = ac_utils.some_accurate_computing_function(param1, param2)


```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


philipc2 (commented):
I think this is a great change.

Can we think of a different module name though, preferably without the `utils` suffix.

Also, I don't think these should be under the `grid` directory, since I can see these being useful elsewhere too (i.e. within a DataArray)

I'll think of a good name and get back to you

hongyuchen1030:
> I think this is a great change.
> 
> Can we think of a different module name though, preferably without the `utils` suffix.
> 
> Also, I don't think these should be under the `grid` directory, since I can see these being useful elsewhere too (i.e. within a DataArray)
> 
> I'll think of a good name and get back to you

Thank you very much for your help. And I just wanted to bring the ideas on the table now and I am also looking for a more appropriate arrangement and name. please let me know if you have any problems

hongyuchen1030:
> How about we move these into a `computing.py` under the `uxarray.utils` directory?
> 
> This would lead to imports like:
> 
> ```python
> from uxarray.utils.computing import dot_fma
> ```

Thank you for your suggestion! However, I am thinking about if we should keep the work "accurate" here? The term accurate refers to this series of floating point operations that provide high accuracy. For example, [accupy](https://github.com/sigma-py/accupy)

philipc2:
> > How about we move these into a `computing.py` under the `uxarray.utils` directory?
> > This would lead to imports like:
> > ```python
> > from uxarray.utils.computing import dot_fma
> > ```
> 
> Thank you for your suggestion! However, I am thinking about if we should keep the word "accurate" here? The term accurate refers to this series of floating point operations that provide high accuracy. For example, [accupy](https://github.com/sigma-py/accupy)

It's typically not suggested to use multiple words in a module name, such as `accurate_computing_utils.py` The "accurate" aspect can be described in our documentation or through usage examples, since most, if not all, of the uses of these functions will still be only used internally.

hongyuchen1030:
> > > How about we move these into a `computing.py` under the `uxarray.utils` directory?
> > > This would lead to imports like:
> > > ```python
> > > from uxarray.utils.computing import dot_fma
> > > ```
> > 
> > 
> > Thank you for your suggestion! However, I am thinking about if we should keep the word "accurate" here? The term accurate refers to this series of floating point operations that provide high accuracy. For example, [accupy](https://github.com/sigma-py/accupy)
> 
> It's typically not suggested to use multiple words in a module name, such as `accurate_computing_utils.py` The "accurate" aspect can be described in our documentation or through usage examples, since most, if not all, of the uses of these functions will still be only used internally.

Thank you for your information, it works for me.

docs/user_api/index.rst
```diff
@@ -343,12 +343,13 @@ Intersections
...
 
-Utils
+Accurate Computing Utils
 -----
 .. autosummary::
    :toctree: _autosummary
 
-   grid.utils.cross_fma
+   grid.accurate_computing_utils.cross_fma
+   grid.accurate_computing_utils.dot_fma
```
philipc2, line 352:
```suggestion
   grid.utils.computing.cross_fma
   grid.utils.computing.dot_fma
```

philipc2 (changes requested):
Looks good to me. Will approve once these are corrected!

philipc2 (approved):
These are some great additions!

PR #574, opened 2023-11-09
Vector Polygon, Mesh, Node Plots, Rasterized Polygon Plots, `GeoDataFrame` Optimizations, Consistent Longitude Range

philipc2:
## Overview

* Introduces the Polygon Rasterization Plot Method through `UxDataArray.plot.rasterize(method="polygon")`
* Introduces the Polygon Vector Plot Method through `UxDataArray.plot.polygons()`
* Introduces the Vector Mesh Plot method through `Grid.plot.mesh()` or `Grid.plot.edges()`
* Introduces the Vector Node Plot for Grid, through `Grid.plot.node_coords`, `Grid.plot.face_coords`, and `Grid.plot.edge_coords`
* Introduces the Vector Node Plot for DataArray, through `UxDataArray.plot.points`

* General improvements & optimizations to GeoDataFrame conversion
  - Most notably, the addition of the `exclude_antimeridian` parameter, which if selected, leads to a significant (~15-20x) performance increase at scale.  
 * Longitude is now always standardized to [-180, 180] 

## Expected Usage

```Python
import uxarray as ux

# rasterized polygon plot 
uxds['v1'].plot.rasterize(method="polygon")

# vector polygon plot (expected to be slow, but high quality)
uxds['v1'].plot.polygons()

# vector mesh plot
uxds.uxgrid.plot.mesh()

# default plot points to mesh
uxds.uxgrid.plot()

```


philipc2:
@erogluorhan 

This PR is pending the merge of https://github.com/holoviz/spatialpandas/pull/132. 

I wanted to mention that the `mesh` plotting added here using the `GeoDataFrame` to obtain the edges/paths, since it doesn't require any additional computation on our end. The performance of this compared to using the connectivity's to construct, for example, a Pandas Dataframe with edges, is much slower. I'll tackle adding a connectivity-informed mesh method with #580

erogluorhan:
A quick comment before reviews:

Providing a `Grid.plot.edges()` functions that would do nothing more than calling `Grid.plot.mesh()` under the hood would be great (when user sees `Grid.plot.nodes()`, they'd naturally expect `.edges()` as well). I think this could be done easily by adding a new access function into `plot/accessor.py` without adding anything into `grid_plot.py`.

philipc2:
@erogluorhan 

This should be ready for some initial testing / reviews. 

> A quick comment before reviews:
> 
> Providing a `Grid.plot.edges()` function that would do nothing more than calling `Grid.plot.mesh()` under the hood would be great (when user sees `Grid.plot.nodes()`, they'd likely expect to see `.edges()` as well).

I was considering this. Essentially another "alias" for mesh, right?

philipc2:
> A quick comment before reviews:
> 
> Providing a `Grid.plot.edges()` function that would do nothing more than calling `Grid.plot.mesh()` under the hood would be great (when user sees `Grid.plot.nodes()`, they'd likely expect to see `.edges()` as well).



> A quick comment before reviews:
> 
> Providing a `Grid.plot.edges()` function that would do nothing more than calling `Grid.plot.mesh()` under the hood would be great (when user sees `Grid.plot.nodes()`, they'd likely expect to see `.edges()` as well).

I've added it. I think this PR is in a good spot if you'd like to give it a test run / initial review. This ended up being much larger than initially intended, but a lot of these additions are essential for the Cookbook.

philipc2:
@erogluorhan 

Additionally, for our `rasterize` method, what do you think of renaming "method" to "element", since we are rasterization polygons, nodes, points, etc. 

Instead of selecting what "method" for rasterization, we select what "element" to rasterize. Does this make more sense?

erogluorhan:
> @erogluorhan
> 
> Additionally, for our `rasterize` method, what do you think of renaming "method" to "element", since we are rasterization polygons, nodes, points, etc.
> 
> Instead of selecting what "method" for rasterization, we select what "element" to rasterize. Does this make more sense?

Even though I see where you're coming, I think I still like `method` more. One more thing is that although the current available set of rasterization methods seem like they're elements-specific only, we will likely have additional rasterization "methods"  in the future that are note directly element-specific, e.g. triangular.

uxarray/plot/dataarray_plot.py
```diff
@@ -179,60 +282,146 @@ def _point_raster(uxda: UxDataArray,
...
+                                colorbar=colorbar,
+                                cmap=cmap,
+                                **kwargs)
+
+
+def nodes(uxda: UxDataArray,
+          backend: Optional[str] = "bokeh",
+          width: Optional[int] = 1000,
+          height: Optional[int] = 500,
+          **kwargs):
```
erogluorhan, line 387: I'd expect to see this function just calling `uxda.uxgrid.plot.nodes()` instead of implementing things itself. 

Its current implementation seems like you're trying to plot not only the Grid nodes' coords but also their corresponding values. I think such a functionality should be implemented with something like `UxDataArray.plot.values()`, and `UxDataArray.plot.nodes()` should only focus on the Grid geometry, hence my expectation with calling `uxda.uxgrid.plot.nodes()`. For future, `UxDataArray.plot.values()` would show node coords and values, or face coords and values (still as a point-plot), depending on data being node-centered or facec-centered, respectively.

philipc2:
> current



> > @erogluorhan
> > Additionally, for our `rasterize` method, what do you think of renaming "method" to "element", since we are rasterization polygons, nodes, points, etc.
> > Instead of selecting what "method" for rasterization, we select what "element" to rasterize. Does this make more sense?
> 
> Even though I see where you're coming, I think I still like `method` more. One more thing is that although the current available set of rasterization methods seem like they're elements-specific only, we will likely have additional rasterization "methods" in the future that are note directly element-specific, e.g. triangular.

One could argue that the "triangles" we are rasterizing are an element (the triangulation of our grid), but I get your point.

uxarray/plot/dataarray_plot.py
philipc2: `uxda.uxgrid.plot.nodes()` is equivalent to `uxgrid.plot.nodes()`, which carries no information about the data variable.

`UxDataArray.plot.values()` sounds like what about default `UxDataArray.plot()` method, where depending on what element our data is mapped to, it returns the appropriate plot.

Having `UxDataArray.plot.values()` return a node plot with face-centered data may be misleading, because the data isn't really mapped to the centroid, it's meant to shade it's corresponding face.

There's an expectation that _any_ plotting method directly called on a `Grid` instance will be data-less (i.e. only the geometry). Similarly, _any_ plot method called directly on a `UxDataArray` will have a data variable. 

### Plot Nodes without Data
```Python
uxds.uxgrid.plot.nodes()
```

### Plot Nodes with Data
```Python
uxds['d_var'].plot.nodes()
```

Maybe its better to think about the case of Edge Plots. For visualizing geometry, we would still have:
```Python
uxds.uxgrid.plot.edges()
```

But how about if we have a data variable on the edge (i.e. gradients). It would simiarly be called
```Python
uxds.['grad'].plot.edges()
```

They are both "edges", but one has a data variable and the other doesn't.

erogluorhan: I agree with that "Having `UxDataArray.plot.values()` return a node plot with face-centered data may be misleading". I also agree with the default `UxDataArray.plot()` method covering a similar need to what I suggested with `UxDataArray.plot.values()`, so let's stick to `UxDataArray.plot()` only. 

However, I don't agree with the idea of  "Plot Nodes with Data" via `uxds['d_var'].plot.nodes()`; I think that'd also be misleading in the user's end (See the next paragraph for an example case). I believe we should consistently use the names "nodes" and "edges" for plotting functions (no what what class, Grid, UxDataArray, or UxDataset, they are called from) for the grid geometry information only (i.e. without any information about the data variable). In my suggestion, `uxds.uxgrid.plot.nodes()` and `uxds['d_var'].plot.nodes()` are and should be exactly equal (Just because not accessing `.uxgrid` accessor, I don't think the end user would expect the difference you propose; I think they'd still expect to explore the grid geometry). 

Having a face-centered data variable and calling `uxds['d_var'].plot.nodes()` would give a `ValueError` in the current implementation, which would be confusing for the end-user since they'd expect to explore the geometry of the Grid that is associated to their UxDataArray.

cc'ing @rajeeja if he has any thoughts.

erogluorhan: Also, re "But how about if we have a data variable on the edge (i.e. gradients)", the default ` uxda.plot()` function should work very well for any of these possibilities, i.e. face-centered, node-centered, edge-centered, and is named very intuitively. One would most likely expect `UxDataArray.plot()` to plot the actual data values.

philipc2: > Having a face-centered data variable and calling uxds['d_var'].plot.nodes() would give a ValueError in the current implementation, which would be confusing for the end-user since they'd expect to explore the geometry of the Grid that is associated to their UxDataArray.

Is this not desired? The way we've written most (if not all) of our functionality is rooted in this separation between `Grid` and `UxDataArray` / `UxDataset` functionality. 

For someone exploring the geometry of the Grid is expected to either have a standalone `Grid` instance or use the `uxda.uxgrid` / `uxds.uxgrid` accessor. This is where our `GridPlotAccessor` comes in, which enables our strictly "geometric" plotting routines through `Grid.plot()` and other methods. 

When interacting with data variables, everything is handled through the `UxDataset` and `UxDataArray`, with most of the calls to the underlying `uxgrid` accessor abstracted away. 

Given this, while our "node" in both the `Grid` and `UxDataArray` case both share the same spatial information, the `uxda.plot.node` method enables the coloring of each node, if applicable (corner node data).

philipc2: > Having a face-centered data variable and calling uxds['d_var'].plot.nodes() would give a ValueError in the current implementation, which would be confusing for the end-user since they'd expect to explore the geometry of the Grid that is associated to their UxDataArray.

The more I think about this, the more it makes sense to split up `plot.nodes()` into `plot.corner_nodes()` and `plot.centroids` or `plot.face_centers()`.

This explicit naming is much more clear, and it would still enable us to plot face-centered data variables respective to the face center coordinates.

I think geometrically it makes sense too, respective to our `Grid`.

philipc2: > Also, re "But how about if we have a data variable on the edge (i.e. gradients)", the default  uxda.plot() function should work very well for any of these possibilities, i.e. face-centered, node-centered, edge-centered, and is named very intuitively. One would most likely expect UxDataArray.plot() to plot the actual data values.

Our default `plot` method will attempt to chose the most appropriate plotting method based on the mapping of the data (additionally the size of the data will dictate whether we default to vector or rasterized approaches)

philipc2:
@erogluorhan 

A quick summary of the current state of the API that we can use for discussing. 

## Preface 
The following example assumes we have a `UxDataset` containing three dummy variables `t2m` (face-centered), `vorticity` (node-centered) and `gradient` (edge-center), all residing on some `Grid` accessed through the `.uxgrid` accessor
```Python
```

## `Grid` Plotting Routines
Plotting methods for visualzing `Grid` geometries

```Python
# default plotting method (mesh/edge plot)
uxds.uxgrid.plot()

# plots the edges of each face (either function call)
uxds.uxgrid.plot.mesh()
uxds.uxgrid.plot.edges()

# plots each node 
uxds.uxgrid.plot.nodes()
```

## `UxDataArray` Plotting Routines

```Python
# default plotting method
uxds['t2m'].plot()

# vector polygons (with antemeridian)
uxds['t2m'].plot.polygons()

# vector polygons (without antemeridian)
uxds['t2m'].plot.polygons(exclude_antimeridian=True)

# rasterized polygons
uxds['t2m'].plot.rasterize(method='polygon')

# vector node plot
uxds['vorticity'].plot.nodes()

# rasterized node (point) plot
uxds['vorticity'].plot.rasterize(method='point')

# rasterized node plot using face centers
uxds['t2m'].plot.rasterize(method='point')

# vector edge plot (NOT IMPLEMENTED)
uxds['gradient'].plot.edges()

# rasterized edge plot (NOT IMPLEMENTED)
uxds['t2m'].plot.rasterize(method='edges')

File: uxarray/plot/dataarray_plot.py (hunk 1 above)
Comment (philipc2) [reply uxarray/plot/dataarray_plot.py:387]: To summarize my thoughts, one should never expect to strictly visualize the "geometry" of the grid if they access they use the `.plot()` method of a `UxDataset` or `UxDataArray`. These will _always_ plot some data variable residing on our unstructured grid (or through an alternative visualuzation method such as computing the trimesh).

**Comment by philipc2** [comment]
I'll also drop this here, which re-iterates the Fluent Interface that we are employing here (aligned with Xarray).

## Scenario
I have a node-centered data variable, but I would like to create a polygon plot.

Since our implementation is Fluent, we would expect the following to raise an error, because our data variable is node-centered, and polygon plots expect face-centered variables.

```Python
>>> uxds['some_var_on_nodes'].plot.polygons()
ValueError: Polygon plotting expects face-centered data
```

The user is first expected to remap / transform their data to map to the correct element. This can be done either with our remapping functionality or by computing the nodal average (as described in #583). While this example is for node centered to face centered data, generally we can expect the user to chain together whatever operations needed to allow for their desired visualization method to work. 

```Python
>>> uxds['some_var_on_nodes'].nodal_average().plot.polygons()
```

I wanted to bring this up, because I don't think our `plot.polygon()` or other methods should remap or transform the data in any way, even if it would be more "continent" for the user. What do you think?

uxarray/plot/dataarray_plot.py
erogluorhan: > To summarize my thoughts, one should never expect to strictly visualize the "geometry" of the grid if they access they use the .plot() method of a UxDataset or UxDataArray. These will always plot some data variable residing on our unstructured grid (or through an approximate visualization method such as computing the trimesh).

Or we should avoid using the same naming for different functionality in the plotting context (or in general). This is not only due to expectations but also due to the high possibility of one's overlooking if there is `.ugrid` or not in their code (e.g. `uxds[`var`].uxgrid.plot.nodes()` looks quite similar to `uxds[`var`].plot.nodes()`, especially in a long workflow, and can easily be overlooked). To make the distinction much clearer, the plotting function with data variable information could be named differently, something like:

`UxDataArray.plot.node_values()`

erogluorhan: > The more I think about this, the more it makes sense to split up `plot.nodes()` into `plot.corner_nodes()` and `plot.centroids` or `plot.face_centers()`.
> 
> This explicit naming is much more clear, and it would still enable us to plot face-centered data variables respective to the face center coordinates.
> 
> I think geometrically it makes sense too, respective to our `Grid`.

I don't agree with this since "nodes" are already the "corner nodes". If we call them "corner nodes", then one would get confused about if there are other types of nodes in the grid geometry, but there is not. Having only "nodes", "edges", and "face_centers" is sufficient IMO.

erogluorhan: > > Having a face-centered data variable and calling uxds['d_var'].plot.nodes() would give a ValueError in the current implementation, which would be confusing for the end-user since they'd expect to explore the geometry of the Grid that is associated to their UxDataArray.
> 
> Is this not desired? The way we've written most (if not all) of our functionality is rooted in this separation between `Grid` and `UxDataArray` / `UxDataset` functionality.
> 
> For someone exploring the geometry of the Grid is expected to either have a standalone `Grid` instance or use the `uxda.uxgrid` / `uxds.uxgrid` accessor. This is where our `GridPlotAccessor` comes in, which enables our strictly "geometric" plotting routines through `Grid.plot()` and other methods.
> 
> When interacting with data variables, everything is handled through the `UxDataset` and `UxDataArray`, with most of the calls to the underlying `uxgrid` accessor abstracted away.
> 
> Given this, while our "node" in both the `Grid` and `UxDataArray` case both share the same spatial information, the `uxda.plot.node` method enables the coloring of each node, if applicable (corner node data).

This makes much sense to me, and I agree with how we mostly differentiate the ` uxda` and `uxda.uxgrid` from each other, but as I explained in the second prior message, using the exact same name (i.e. `.nodes()`) would still confuse.  I'd highly recommend to use a different name than the particular geometric element's name to make the distinction 100% clear.

erogluorhan: > > Also, re "But how about if we have a data variable on the edge (i.e. gradients)", the default  uxda.plot() function should work very well for any of these possibilities, i.e. face-centered, node-centered, edge-centered, and is named very intuitively. One would most likely expect UxDataArray.plot() to plot the actual data values.
> 
> Our default `plot` method will attempt to chose the most appropriate plotting method based on the mapping of the data (additionally the size of the data will dictate whether we default to vector or rasterized approaches)

Yes, and I feel like this is the perfect way of working. However, it'd be fine to have additional functions like `uxda.plot.node_values()` `uxda.plot.edge_values()`, and `uxda.plot.face_values()` (or alternative names that are not the same as Grid's geometric element plotting fucns' names), that would cover parts of this default `plot()` or throw `ValueError` if not applicable. 

Would it make sense?

philipc2: > > The more I think about this, the more it makes sense to split up `plot.nodes()` into `plot.corner_nodes()` and `plot.centroids` or `plot.face_centers()`.
> > This explicit naming is much more clear, and it would still enable us to plot face-centered data variables respective to the face center coordinates.
> > I think geometrically it makes sense too, respective to our `Grid`.
> 
> I don't agree with this since "nodes" are already the "corner nodes". If we call them "corner nodes", then one would get confused about if there are other types of nodes in the grid geometry, but there is not. Having only "nodes", "edges", and "face_centers" is sufficient IMO.

Maybe we should stray away from calling this method "nodes", since like you mentioned we reffer to "nodes" as our "corner nodes".

We have three types of elements that could be interpreted as "nodes":
* `node_lat/lon`: Corner Node Coordinates
* `face_lat/lon`: Face Center Coordinates
* `edge_lat/lon`: Edge Center Coordinates

Would it make more sense to call it `Grid.plot.points(element="nodes")` and specify which element ("nodes", "faces", "edges"). This also lines up with our "Point Rasterization" instead of calling it "Node Rasterization".

philipc2: We could also keep `Grid.plot.nodes` as an alias of `Grid.plot.points(element="nodes")`, while re-writing `UxDataArray.plot.nodes` to `UxDataArray.plot.points` with a similar naming convention.

philipc2: > > Our default `plot` method will attempt to chose the most appropriate plotting method based on the mapping of the data (additionally the size of the data will dictate whether we default to vector or rasterized approaches)
> 
> Yes, and I feel like this is the perfect way of working. However, it'd be fine to have additional functions like `uxda.plot.node_values()` `uxda.plot.edge_values()`, and `uxda.plot.face_values()` (or alternative names that are not the same as Grid's geometric element plotting fucns' names), that would cover parts of this default `plot()` or throw `ValueError` if not applicable.
> 
> Does it make sense?

This does make sense, but personally I'm not a big fan of extending the names with an `_` I'd like to hear what @rajeeja, @rljacob , or anyone else thinks about the `uxgrid` accessor not being enoguh to difference between "grid-only" and "data variable" plotting.

erogluorhan: > Maybe we should stray away from calling this method "nodes", since like you mentioned we reffer to "nodes" as our "corner nodes".
> 
> We have three types of elements that could be interpreted as "nodes":
> 
> * `node_lat/lon`: Corner Node Coordinates
> * `face_lat/lon`: Face Center Coordinates
> * `edge_lat/lon`: Edge Center Coordinates
> 
> Would it make more sense to call it `Grid.plot.points(element="nodes")` and specify which element ("nodes", "faces", "edges"). This also lines up with our "Point Rasterization" instead of calling it "Node Rasterization".

To me (and the UGRID conventions AFAIK), the only one that should actually be interpreted as "nodes" is the first one out of these three. The others should never be interpreted as "nodes", maybe instead "face_centers (or face_coordinates)" and "edge_centers (or edge_coordinates)". That said I would not find `Grid.plot.points(element="nodes")` to be better.

Could you please elaborate on why you'd interprete all these three as "nodes"?

erogluorhan: Additionally, my general understanding is that "nodes" or "vertices" (in different models etc.) are always interpreted for the "corner nodes" and never for anything related to face or edges. Similarly, "face" or "cell" is always used for what we know as face in our Grid. MPAS has "vertices"  of Dual esh corresponding to "face_centrers" of Primal mesh, and vice versa, but that's a special case.

philipc2: > To me (and the UGRID conventions AFAIK), the only one that should actually be interpreted as "nodes" is the first one out of these three. The others should never be interpreted as "nodes", maybe instead "face_centers (or face_coordinates)" and "edge_centers (or edge_coordinates)". That said I would not find Grid.plot.points(element="nodes") to be better.
> 
> Could you please elaborate on why you'd interprete all these three as "nodes"?

I consider a "node", "vertex", or "point" in our context as a tuple of `(lon, lat)`. Each of our `node_lat/lon`, `face_lat/lon` and `edge_lat/lon` are in this form. Below are some snippets from the MPAS specification.

image

image

image

philipc2: One scenario where we may want to plot `node_lat/lon` alongside `face_lat/lon` is when investigating the computed centroid location.

Or similarly, plotting the `edge_lat/lon` with a mesh/edge plot to see if they centroids lie on each edge.

philipc2: For the DataArray case, one may want to plot an `edge` data variable as a note, instead of shading the whole edge, when plotting the gradient for example.

erogluorhan: > > To me (and the UGRID conventions AFAIK), the only one that should actually be interpreted as "nodes" is the first one out of these three. The others should never be interpreted as "nodes", maybe instead "face_centers (or face_coordinates)" and "edge_centers (or edge_coordinates)". That said I would not find Grid.plot.points(element="nodes") to be better.
> > Could you please elaborate on why you'd interprete all these three as "nodes"?
> 
> I consider a "node", "vertex", or "point" in our context as a tuple of `(lon, lat)`. Each of our `node_lat/lon`, `face_lat/lon` and `edge_lat/lon` are in this form. Below are some snippets from the MPAS specification.
> 
> image
> 
> image
> 
> image

I see. I'd agree with you (i.e. naming tuples of lot/lan as "nodes" regardless of them being faces or edges as well) if this was a general-purpose geometry application, but since UXarray is specialized to unstructured grids and in geoscientific unstructured grid applications "node" is used for the "corner of a face", I don't agree with this interpretation (i.e. we should avoid confusing "nodes" with such tuples). 

I think even the MPAS snippets you provided support the fact that "vertices/nodes" are used only for the "corner of cells/faces", not for any face-related or edge-related thing (despite MPAS's special case I mentioned).

For instance, the text of `latCell` and `lonCell` only refers to "cell centers", that of `latEdge` and `lonEdge` only mentions to "edge locations"; however, `latVertex` and `lonVertex` mentions "cell vertices".

When it comes to tuples of such lat/lons, amybe one should call them "node_coords", "edge_coords", "face_coords", not "nodes" or "vertices". 

I am happy to discuss this in a video chat, 1:1 or in a broader group.

philipc2: Calling them "cords" makes much more sense to me.

erogluorhan: > One scenario where we may want to plot the `face_lat/lon` is when we want to investigate the centroid computation. We would plot these as "nodes" along with a `edge` or `polygon` plot
> 
> ```python
> # composition with Holoviews 
> uxgrid.mesh() * uxgrid.points(element='face')
> ```
> 
> Or similarly, plotting the `edge_lat/lon` with a mesh/edge plot to see if they centroids lie on each edge.
> 
> ```python
> uxgrid.mesh() * uxgrid.points(element='edge')
> ```

I'd foresee these as:

`uxgrid.mesh() * uxgrid.face_centers()` and 
`uxgrid.mesh() * uxgrid.edge_centers()` or `uxgrid.mesh() * uxgrid.edge_locations()`

erogluorhan: > Calling them "coords" makes much more sense to me.

Yeah, or "nodes", "face_centers", and "edge_locations" always sounded sufficient to me.

philipc2:
@erogluorhan 

The discussion here was really helpful! I'll collect the main points into a revision tomorrows. I'll ping a re-review once it's ready.

philipc2:
@erogluorhan 

I've made a few changes, I think the design is much better now. Let me know what you think! 

## `Grid` Plotting
Explicit selection of what coordinate to plot
```Python
# geometry of nodes
uxgrid.plot.node_coords()

# geometry of edge centers
uxgrid.plot.edge_coords()

# geometry of facecenters
uxgrid.plot.face_coords()
```

### `UxDataArray` Plotting
The DataArray method takes in a data variable and maps it to whatever dimension the data maps to (no explicit edge, node, face point plots, it's all handled through one "points"  method)
```Python
# map data to node coords 
uxda['node_centered_var'].plot.points()

# map data to edge coords 
uxda['edge_centered_var'].plot.points()

# map data to facecoords 
uxda['face_centered_var'].plot.points()

**Review by erogluorhan — COMMENTED** [review]
Overall, great implementation, and we are really close to there! Please find my comments about naming (still :) ) and a few more inline reviews below:

> @erogluorhan
> 
> I've made a few changes, I think the design is much better now. Let me know what you think!
> 
> ## `Grid` Plotting
> Explicit selection of what coordinate to plot
> 
> ```python
> # geometry of nodes
> uxgrid.plot.node_coords()
> 
> # geometry of edge centers
> uxgrid.plot.edge_coords()
> 
> # geometry of facecenters
> uxgrid.plot.face_coords()
> ```
> 

Even though I like these much better than yesterday's suggestions, I'd still prefer some alternatives as follows:
1. `uxgrid.plot.nodes()` either instead of or in addition to `uxgrid.plot.node_coords()`. This is because, nodes are also geometric elements in our grids just like edges, we have `uxgrid.plot.edges()`, so not having `uxgrid.plot.nodes()` is a bit awkward. Additionally, since this is an explit call to `uxgrid`'s plotting accessor, one can expect this is node coords (i.e. without any data).
2. `face_coords` and `edge_coords` sound okay, but I'd still prefer `face_centers` and `edge_centers` in the naming as "center" already suggests the center point/coordinate of a geometric element and is the mentioned term in the UGRID convention.
3. Maybe, we should consider providing aliases for the above namings to support both conventions?

> ### `UxDataArray` Plotting
> The DataArray method takes in a data variable and maps it to whatever dimension the data maps to (no explicit edge, node, face point plots, it's all handled through one "points" method). The name "Points" now lines up with our "Point Rasterization"
> 
> ```python
> # map data to node coords 
> uxda['node_centered_var'].plot.points()
> 
> # map data to edge coords 
> uxda['edge_centered_var'].plot.points()
> 
> # map data to facecoords 
> uxda['face_centered_var'].plot.points()
> ```

Similarly, I find these okay, but I have a slight concern about whether the user confuses the term "points" with the geometric element "nodes". Maybe docstring will speak to that and help avoid confusions.

File: docs/internal_api/index.rst [hunk 2]
```diff
@@ -143,6 +147,7 @@ Coordinates
    grid.coordinates._populate_lonlat_coord
    grid.coordinates._populate_centroid_coord
    grid.coordinates._construct_xyz_centroids
+   grid.coordinates._set_desired_longitude_range
```
erogluorhan, line 150: Looks like we missed to include some new internal funcs from dataarray_plot.py and grid_plot.py

philipc2: Good catch!

philipc2:
> Similarly, I find these okay, but I have a slight concern about whether the user confuses the term "points" with the geometric element "nodes". Maybe docstring will speak to that and help avoid confusions.

I'll add to the docstring to make it more clear, but I couldn't think of a better alternative.

philipc2:
> Even though I like these much better than yesterday's suggestions, I'd still prefer some alternatives as follows:
> 
> 1. `uxgrid.plot.nodes()` either instead of or in addition to `uxgrid.plot.node_coords()`. This is because, nodes are also geometric elements in our grids just like edges, we have `uxgrid.plot.edges()`, so not having `uxgrid.plot.nodes()` is a bit awkward. Additionally, since this is an explit call to `uxgrid`'s plotting accessor, one can expect this is node coords (i.e. without any data).
> 2. `face_coords` and `edge_coords` sound okay, but I'd still prefer `face_centers` and `edge_centers` in the naming as "center" already suggests the center point/coordinate of a geometric element and is the mentioned term in the UGRID convention.
> 3. Maybe, we should consider providing aliases for the above namings to support both conventions?

I think `nodes`, `edge_centers` and `face_centers` are good choices, especially given that we already have `edges` as a plotting method.

I like the idea of having an alias for both, since either could be preferable to the user.

uxarray/plot/grid_plot.py
```diff
@@ -1,18 +1,175 @@
...
+                           backend: Optional[str] = "bokeh",
+                           width: Optional[int] = 1000,
+                           height: Optional[int] = 500,
+                           **kwargs):
+    """Helper function for plotting coordinates (node, edge, face) as Points
+    with Holoviews."""
+
+    if element == "node":
+        node_lon = uxgrid.node_lon.values
+        if node_lon.max() > 180:
```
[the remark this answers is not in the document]
philipc2: Yeah, I had a first-pass attempt at it yesterday, but ended up scrapping it in favor of what you saw there. I revisited it and now it should be implemented in the `Grid` object. Good suggestion!

uxarray/plot/grid_plot.py
```diff
@@ -1,18 +1,175 @@
...
+
+    return _plot_coords_as_points(element="edge",
+                                  uxgrid=uxgrid,
+                                  backend=backend,
+                                  width=width,
+                                  height=height,
+                                  **kwargs)
+
+
+def _plot_coords_as_points(element,
```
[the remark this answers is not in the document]
philipc2: Thank you!

philipc2:
@erogluorhan 

I've put together an rough draft for a usage example notebook that we could use to showcase all of these additions to the API.

Let me know what you think of the current state of the function definitions & other aspects of this PR, and I can give it some final refinements before requesting others for review.

philipc2:
[Link to the Usage Example](https://uxarray--574.org.readthedocs.build/en/574/examples/006-plotting-api.html)

uxarray/plot/grid_plot.py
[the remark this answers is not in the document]
erogluorhan: Great, thank you!

erogluorhan:
> [Link to the Usage Example](https://uxarray--574.org.readthedocs.build/en/574/examples/006-plotting-api.html)
> 
> I'm especially happy with how [this ](https://uxarray--574.org.readthedocs.build/en/574/examples/006-plotting-api.html#putting-both-together)visual came out, can really see the MPAS structure here.

That plot looks wonderful and will be very helpful for the reader I believe!

erogluorhan (commented):
The new version of this pR looks great, and the usage example is fantastic! I'd like to approve it soon, just after making sure about the following things:

1. PR status and title indicates it's really draft, but I think it's ready for merge, right?
2. It seems like you will still need to push your addressing of some of the review comments that you already agreed to. Is that correct, should I wait for them?
3. Re Usage Example: Please do not forget to give a good mention, if not main call, to the function aliases such as `Grid.plot.nodes()`, `Grid.plot.edges()`, `Grid.plot.face_centers()`, and `Grid.plot.edge_centers()` (and maybe others if any).

erogluorhan:
> 3. Re Usage Example: Please do not forget to give a good mention, if not main call, to the function aliases such as `Grid.plot.nodes()`, `Grid.plot.edges()`, `Grid.plot.face_centers()`, and `Grid.plot.edge_centers()` (and maybe others if any).

EDIT: Looks like you mentioned `plot.nodes()` and `plot.edges()` but there is a typo for `face_centers` (seems like it was copied as `face_coords` again in the "alternative call" text)

philipc2:
> 2\. addressing



> The new version of this pR looks great, and the usage example is fantastic! I'd like to approve it soon, just after making sure about the following things:
> 
> 1. PR status and title indicates it's really draft, but I think it's ready for merge, right?
> 2. It seems like you will still need to push your addressing of some of the review comments that you already agreed to. Is that correct, should I wait for them?
> 3. Re Usage Example: Please do not forget to give a good mention, if not main call, to the function aliases such as `Grid.plot.nodes()`, `Grid.plot.edges()`, `Grid.plot.face_centers()`, and `Grid.plot.edge_centers()` (and maybe others if any).

1. I'll change the title & status after I do some final revisions, and we get @rajeeja to give it a look through.

2. I thought I've addressed them all, but I'll give it another look through

3. Will do!

ci/environment.yml
```diff
@@ -20,8 +20,8 @@ dependencies:
   - scikit-learn
   - scipy
   - shapely
-  - spatialpandas
   - xarray
   - pip:
     - antimeridian
     - pyfma
+    - git+https://github.com/holoviz/spatialpandas.git
```
erogluorhan, line 27: This is until their official release, right?

rajeeja: Testing this, awesome work. Gimme a day, at SC..

docs/internal_api/index.rst
erogluorhan: Is this addressed yet?

ci/environment.yml
philipc2: > This is until their official release, right?

Correct

uxarray/grid/geometry.py
```diff
@@ -12,98 +16,162 @@
...
+                                           grid.n_nodes_per_face.values)
 
-    # list of shapely Polygons representing each face in our grid
-    polygons = Polygons(polygon_shells)
+    antimeridian_face_indices = _build_antimeridian_face_indices(
+        polygon_shells[:, :, 0])
 
-    # handle antimeridian polygons, if any
-    if grid.antimeridian_face_indices is not None and correct_antimeridian_polygons:
+    if grid.n_face > 1000000:
```
erogluorhan, line 64: Another magic number!

ci/environment.yml
erogluorhan: Please open an issue if we haven't done so yet in order not to forget this

erogluorhan:
> 2. I thought I've addressed them all, but I'll give it another look through

I think the only missing one is inter API doc entries, e.g. `_plot_... ` function and maybe others

ci/environment.yml
philipc2: #588

docs/internal_api/index.rst
philipc2: It should be now.

philipc2:
@ifranda 

I've pinged you as a reviewer (no need to give it a deep review), but I'm planning to chat about this PR durring our meeting on Monday, so feel free to give it a look through before then! 

The dataset in the usage example notebook is the same one that you used for your SIParCS work, so it should be familiar.

philipc2:
Everything except for the notebook of this PR is now ready for review. I will be continuing to update the notebook throughout today and tomorrow.

erogluorhan (commented):
Fantastic contribution! Will approve this, though I'd really love to hear thoughts from @rajeeja @hongyuchen1030 @aaronzedwick, if at all possible, on some naming and docs points below (in addition to if you have anything additional to prior convo, @philipc2 ):

1. The function name `UxDataArray.plot.points()`still doesn't feel like the best fit to me (I am a bit concerned if any user confuses this with the geometric element "nodes"). What about `UxDataArray.plot.values()`?
2. Re usage example, 
     1. I feel like we should emphasize  
        - `Grid.plot.nodes()`, `Grid.plot.edge_centers()`, and `Grid.plot.face_centers()` as the main functions, and
        - `Grid.plot.node_coords()`, `Grid.plot.edge_coords()`, and `Grid.plot.face_coords()` as the alternative ways. 
     2. I believe we should have `Grid.plot.nodes()` showcased together with `Grid.plot.mesh/edges()` as part of a potential "Geometric Elements Plotting"  section instead of being listed together with the other "Coordinate Plots"

philipc2:
> Fantastic contribution! Will approve this, though I'd really love to hear thoughts from @rajeeja @hongyuchen1030 @aaronzedwick, if at all possible, on some naming and docs points below (in addition to if you have anything additional to prior convo, @philipc2 ):
> 
> 1. The function name `UxDataArray.plot.points()`still doesn't feel like the best fit to me (I am a bit concerned if any user confuses this with the geometric element "nodes"). What about `UxDataArray.plot.values()`?
> 2. Re usage example,
>    
>    1. I feel like we should emphasize
>       
>       * `Grid.plot.nodes()`, `Grid.plot.edge_centers()`, and `Grid.plot.face_centers()` as the main functions, and
>       * `Grid.plot.node_coords()`, `Grid.plot.edge_coords()`, and `Grid.plot.face_coords()` as the alternative ways.
>    2. I believe we should have `Grid.plot.nodes()` showcased together with `Grid.plot.mesh/edges()` as part of a potential "Geometric Elements Plotting"  section instead of being listed together with the other "Coordinate Plots"

1. I could see `plot.values()` leading to more confusion, since someone may expect the "values" of a face-centered variable to be mapped to polygons, not points. I still like the name points ([it lines up with Holoviews' naming](https://holoviews.org/reference/elements/matplotlib/Points.html)). The only other alternative I could think of is calling it `plot.scatter()`, which is a scatter plot of points with a data variable mapped to it. 
2. These are great suggestions, I'll incorporate these.

erogluorhan:
> 1. I could see `plot.values()` leading to more confusion, since someone may expect the "values" of a face-centered variable to be mapped to polygons, not points. I still like the name points ([it lines up with Holoviews' naming](https://holoviews.org/reference/elements/matplotlib/Points.html)). The only other alternative I could think of is calling it `plot.scatter()`, which is a scatter plot of points with a data variable mapped to it. However [here's a good blurb](https://holoviews.org/reference/elements/matplotlib/Points.html#:~:text=Note%3A%20Although%20the,help(hv.Points).) on why it may be preffered to stick with the name `points`
> 2. These are great suggestions, I'll incorporate these. I'll actually split up the usage example into two, one for visualizing Grid topology, and the other for visualizing data variables.

1. Yes, I get your points with `points`; those make sense. Apparently, there is no one-name-fixes-all-confusions chance here. 
2. Thanks! 

So, I am approving this now and leaving the rest of the discussion about naming etc., if any, to you and the other folks pinged.

erogluorhan (approved):

philipc2:
> > 1. I could see `plot.values()` leading to more confusion, since someone may expect the "values" of a face-centered variable to be mapped to polygons, not points. I still like the name points ([it lines up with Holoviews' naming](https://holoviews.org/reference/elements/matplotlib/Points.html)). The only other alternative I could think of is calling it `plot.scatter()`, which is a scatter plot of points with a data variable mapped to it. However [here's a good blurb](https://holoviews.org/reference/elements/matplotlib/Points.html#:~:text=Note%3A%20Although%20the,help(hv.Points).) on why it may be preffered to stick with the name `points`
> > 2. These are great suggestions, I'll incorporate these. I'll actually split up the usage example into two, one for visualizing Grid topology, and the other for visualizing data variables.
> 
> 1. Yes, I get your points with `points`; those make sense. Apparently, there is no one-name-fixes-all-confusions chance here.
> 2. Thanks!
> 
> So, I am approving this now and leaving the rest of the discussion about naming etc., if any, to you and the other folks pinged. A significant contribution to UXarray!

Thank you so much for all the discussion & suggestions this week, they were extremely helpful in getting this all implemented.

rajeeja (changes requested):
Great work! 

can you add/test plot feature for uxgrid objects from nodes/faces. 
Eg. 3  from this notebook: [here](https://uxarray.readthedocs.io/en/latest/examples/003-area-calc.html#getting-area-of-individual-faces)
and Eg. 4-6 also. I got some errors testing these.

philipc2:
> Great work!
> 
> can you add/test plot feature for uxgrid objects from nodes/faces. Eg. 3 from this notebook: [here](https://uxarray.readthedocs.io/en/latest/examples/003-area-calc.html#getting-area-of-individual-faces) and Eg. 4-6 also. I got some errors testing these.

Looks like these actually broke when the Validate Mesh PR was merged (following docs are generated from that PR) 

https://uxarray--173.org.readthedocs.build/en/173/examples/003-area-calc.html

philipc2:
> Great work!
> 
> can you add/test plot feature for uxgrid objects from nodes/faces. Eg. 3 from this notebook: [here](https://uxarray.readthedocs.io/en/latest/examples/003-area-calc.html#getting-area-of-individual-faces) and Eg. 4-6 also. I got some errors testing these.

Testing the correctness of plots is difficult, but I can add some tests the ensure no errors are raised with execution.

PR #591, opened 2023-11-17
Inverse Distance Weighted Remapping Implementation 

### LINKED ISSUE #443: Inverse Distance Weighted Remapping
Implement Inverse Distance Weighted (IDW) Remapping, with options to select how many nearest neighbors (k) to use for the weighting. 

A good read into the underlying algorithm [here](https://help.rockware.com/rockworks17/WebHelp/gridding_inv_dist.htm#:~:text=Inverse%2DDistance%20is%20one%20of,number%20of%20directionally%20distributed%20neighbors.)

aaronzedwick:
Closes #443

## Overview

Adds inverse distance weighted remapping. This will recalculate the data for each point based on the nearest neighbors. Read more [here](https://help.rockware.com/rockworks17/WebHelp/gridding_inv_dist.htm#:~:text=Inverse%2DDistance%20is%20one%20of,number%20of%20directionally%20distributed%20neighbors.)
## Expected Usage

```Python
import uxarray as ux

dataset = ux.open_dataset(gridfile_geoflow, dsfile_v1_geoflow)
destination_uxds = ux.open_dataset(gridfile_geoflow, dsfile_v1_geoflow)

data_on_face_centers = dataset['v1'].inverse_distance_weighted_remap(destination_uxds.uxgrid)

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


aaronzedwick:
@rajeeja @philipc2 This is ready whenever y'all get to it. I tried to keep the format as close as I could to the nearest neighbor. If you guys had any suggestions on how to make the test cases (especially the first two) check for correctness I would appreciate it. The data didn't change in the nearest neighbor, so asserting the data was correct was simple. Here, if I wanted to do the same I suppose I would have to recalculate it in the test case? So yeah, any thoughts on that, if I should add something, or if it's fine as is, would be appreciated. Thanks!

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -0,0 +1,224 @@
...
+        Source grid that data is mapped from.
+    destination_grid : Grid
+        Destination grid to remap data to.
+    source_data : np.ndarray
+        Data variable to remap.
+    remap_to : str, default="nodes"
+        Location of where to map data, either "nodes" or "face centers".
+    coord_type: str, default="spherical"
+        Coordinate type to use for nearest neighbor query, either "spherical" or "Cartesian".
+    power : float, default=2
```
rajeeja, line 38: add a little blurb about what this power parameter can do?

rajeeja (changes requested):
This is good work and consistent with nearest neighbor implementation. 

Some tests could be more thorough.

balltree is hardcoded for spherical coords and kdtree for cartesian, can this be a variable exposed?

rajeeja:
> Closes #443
> 
> ## Overview
> Adds inverse distance weighted remapping. This will recalculate the data for each point based on the nearest neighbors. Read more [here](https://help.rockware.com/rockworks17/WebHelp/gridding_inv_dist.htm#:~:text=Inverse%2DDistance%20is%20one%20of,number%20of%20directionally%20distributed%20neighbors.)


This is a good [paper](https://gmd.copernicus.org/articles/15/6601/2022/) by @vijaysm @rljacob and @paullric on this subject and something we can visit at a later point.

test/test_remap.py
```diff
@@ -143,3 +143,102 @@ def test_remap_return_types(self):
         # Dataset with four vars: original "psi" and remapped "v1, v2, v3"
         assert isinstance(remap_uxds_to_uxds, UxDataset)
         assert len(remap_uxds_to_uxds.data_vars) == 4
+
+
+class TestInverseDistanceWeightedRemapping(TestCase):
+    """Testing for inverse distance weighted remapping."""
+
+    def test_remap_center_nodes(self):
```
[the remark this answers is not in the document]
aaronzedwick: Hmm, it’s not a bad idea to add those checks. I don’t know if there would be unsupported scenarios, I don’t think there would be though

aaronzedwick:
> This is good work and consistent with nearest neighbor implementation.
> 
> Some tests could be more thorough.
> 
> balltree is hardcoded for spherical coords and kdtree for cartesian, can this be a variable exposed?

Currently the implementation of BallTree is only for spherical cords and KDTree for Cartesian, so exposing them wouldn’t be a good idea since it would create errors. We should implement both coord types in both tree styles. At that time then we could expose the variable.

rajeeja:
> > This is good work and consistent with nearest neighbor implementation.
> > 
> > Some tests could be more thorough.
> > 
> > balltree is hardcoded for spherical coords and kdtree for cartesian, can this be a variable exposed?
> 
> Currently the implementation of BallTree is only for spherical cords and KDTree for Cartesian, so exposing them wouldn’t be a good idea since it would create errors. We should implement both coord types in both tree styles. At that time then we could expose the variable. 

I know, we can handle this in a new PR

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -144,6 +144,13 @@ def _inverse_distance_weighted_remap_uxda(source_uxda: UxDataArray,
         remapping.
     """
 
+    # check dimensions remapped to and from
+    if (source_uxda.dims[-1] == "n_node" and remap_to != "nodes") or \
+            (source_uxda.dims[-1] == "n_face" and remap_to != "face centers"):
+        warnings.warn(
+            f"Your data is stored on {source_uxda.dims[-1]}, but you are remapping to {remap_to}"
+        )
+
```
philipc2, line 153: Could these be used here?

https://github.com/UXARRAY/uxarray/blob/302f42d9de04f9ab280b486df7eacedaf8a10099/uxarray/core/dataarray.py#L349-L362

aaronzedwick: Oh yeah, those will!

uxarray/core/dataset.py
```diff
@@ -345,3 +348,25 @@ def nearest_neighbor_remap(
             Indicates whether to remap using on spherical or cartesian coordinates
         """
         return _nearest_neighbor_uxds(self, destination_obj, remap_to, coord_type)
+
+    def inverse_distance_weighted_remap(
+        self,
+        destination_obj: Union[Grid, UxDataArray, UxDataset],
+        remap_to: str = "nodes",
+        coord_type: str = "spherical",
```
philipc2, line 356: Parameters should match:

https://github.com/UXARRAY/uxarray/blob/086b9e9ede1def5de9c127429846a42f7d5a32fb/uxarray/remap/inverse_distance_weighted.py#L16-L23

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -0,0 +1,269 @@
...
+    return destination_data
+
+
+def _inverse_distance_weighted_remap_uxda(
+    source_uxda: UxDataArray,
+    destination_obj: Union[Grid, UxDataArray, UxDataset],
+    remap_to: str = "nodes",
+    coord_type: str = "spherical",
+):
```
philipc2, line 161: Parameters should match

philipc2 (changes requested):

philipc2:
I've run a test remapping between two meshes, below are results of running the remap from ~6 million face mesh to a ~650 thousand face mesh.

The plots here are of a subset of data too, but the same trends apply for the global mesh. 

image

All the values appear to be close to zero for the IDW remap.

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -0,0 +1,269 @@
...
+
+    if nearest_neighbor_indices.ndim > 1:
+        nearest_neighbor_indices = nearest_neighbor_indices.squeeze()
+
+    weights = 1 / (distances**power + 1e-6)
+    weights /= np.sum(weights)
+
+    destination_data = np.sum(
+        source_data[..., nearest_neighbor_indices] * weights, axis=-1
+    )
```
philipc2, line 151: I'll give this a deep look tomorrow, but I presume the issue lies somewhere here in how we compute our weights.

aaronzedwick: Oh, yeah that is a problem! I will start seeing if I can figure it out too, how did you make those plots? Could you send the code for that over Slack or in a gist or something? Having that for any remapping function would be great for testing.

aaronzedwick: I think I found the problem. Each point gets its 8 nearest neighbors (by default), and calculates a weight based on those. The sum of those 8 should be equal to 1. The current way however takes sum of all the weights in the entire grid, instead of the 8 it should be for a single point, and divides the weights by that, thus making them extremely small, causing near zero values. I just added a loop in the weights calculation to properly make the weights get divided by the correct sum of weights. I will push my changes, could you let me know if this works on your plots?

philipc2: > I think I found the problem. Each point gets its 8 nearest neighbors (by default), and calculates a weight based on those. The sum of those 8 should be equal to 1. The current way however takes sum of all the weights in the entire grid, instead of the 8 it should be for a single point, and divides the weights by that, thus making them extremely small, causing near zero values. I just added a loop in the weights calculation to properly make the weights get divided by the correct sum of weights. I will push my changes, could you let me know if this works on your plots?

Let me know! I'll generate some more plots once you've got it resolved, but this definitely sounds like the issue.

philipc2: For these plots, I've been using the GeoDataFrame approach

```Python
%%time
downsampled_uxds_nn = uxds["refl10cm_max"].isel(Time=2).nearest_neighbor_remap(destination_obj = coarse_grid, remap_to="face centers")
downsampled_uxds_idw = uxds["refl10cm_max"].isel(Time=2).inverse_distance_weighted_remap(destination_obj = coarse_grid, remap_to="face centers")
lon_bounds = (-75, -71)
lat_bounds = (40, 42)
gdf_og = uxds["refl10cm_max"].isel(Time=2).subset.bounding_box(lon_bounds, lat_bounds).to_geodataframe()
gdf_nn = downsampled_uxds_nn.subset.bounding_box(lon_bounds, lat_bounds).to_geodataframe()
gdf_idw = downsampled_uxds_idw.subset.bounding_box(lon_bounds, lat_bounds).to_geodataframe()

(gdf_og.hvplot.polygons(rasterize=True, 
                              c='refl10cm_max', 
                              cmap=cmap, 
                              geo=True, 
                              width=1000, 
                              height=500,
                              grid=True, 
                              projection=projection,
                              title="Original Mesh") + \
    
    
    gdf_nn.hvplot.polygons(rasterize=True, 
                              c='refl10cm_max', 
                              cmap=cmap, 
                              geo=True, 
                              width=1000, 
                              height=500,
                              grid=True, 
                              projection=projection,
                              title = "Nearest Neighbor Remap") + \
 gdf_idw.hvplot.polygons(rasterize=True, 
                                 c='refl10cm_max',
                                 cmap=cmap, 
                                 geo=True, 
                                 width=1000, 
                                 height=500, 
                                 grid=True, 
                                 projection=projection,
                                 title = "IDW Remap")).cols(1).opts(fig_size=200)
```

aaronzedwick: Awesome! Thanks for the code snippet. The changes should be up now, let me know what you find and if the changes create any new problems. Thanks for catching this error!

test/test_remap.py
```diff
@@ -154,9 +154,137 @@ def test_edge_centers_remapping(self):
...
+        ]
+        source_uxds = ux.open_mfdataset(gridfile_geoflow, source_data_paths)
+        destination_uxds = ux.open_dataset(gridfile_CSne30,
+                                           dsfile_vortex_CSne30)
+
+        remap_uxda_to_grid = source_uxds['v1'].inverse_distance_weighted_remap(
+            destination_uxds.uxgrid)
+
+        assert isinstance(remap_uxda_to_grid, UxDataArray)
+
```
rajeeja, line 240: How about adding `assert len(remap_uxda_to_uxda.data_vars) == 1`here and for cases that follow?

test/test_remap.py
```diff
@@ -154,9 +154,137 @@ def test_edge_centers_remapping(self):
...
+        """Tests the ability to remap on edge centers using Inverse Distance
+        Weighted Remapping."""
+
+        # Open source and destination datasets to remap to
+        source_grid = ux.open_dataset(gridfile_geoflow, dsfile_v1_geoflow)
+        destination_grid = ux.open_dataset(mpasfile_QU, mpasfile_QU)
+
+        # Perform remapping to the edge centers of the dataset
+
+        remap_to_edge_centers = source_grid['v1'].inverse_distance_weighted_remap(destination_obj=destination_grid,
```
rajeeja, line 286: where is the variable v1 located in source grid? (It must also be on edge centers?)

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+def _inverse_distance_weighted_remap(
+    source_grid,
+    destination_grid,
+    source_data,
+    remap_to="nodes",
+    coord_type="spherical",
+    power=2,
+    k_neighbors=8,
```
rajeeja, line 23: k_neighbors=8 is on the higher side. Would it make sense to lower it?

test/test_remap.py
aaronzedwick: It doesn't matter where the variable is located, so you can remap from nodes to edges if you want.

uxarray/remap/inverse_distance_weighted.py
aaronzedwick: 8 is the default value from what I found, also 8 isn't that high when most meshes have high numbers of nodes. Also, the point is each neighbor is weighted differently, so as you move further from the destination point each neighbor is weighted as less "important" so to speak in the final calculation of the destination data point.

philipc2: Yeah 8 seems reasonable, as long as we allow the user to customize it and keep it documented, it's all good.

uxarray/remap/inverse_distance_weighted.py
philipc2: image

This looks as expected!

philipc2: Once you address my other comments (the one about passing in more params through a dataset or dataarray), I'll generate more plots and do some additional testing.

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+
+    if nearest_neighbor_indices.ndim > 1:
+        nearest_neighbor_indices = nearest_neighbor_indices.squeeze()
+
+    weights = []
+    for ind in range(len(distances)):
+        weight = 1 / (distances[ind] ** power + 1e-6)
+        weight /= np.sum(weight)
+        weights.append(weight)
```
philipc2, line 150: Does this need to be in a for loop? I'd suggest trying to vectorize it. Should be able to do some tricks using `axis=1` and some numpy functionality. 

Let me know if you need any help!

philipc2 (changes requested):
Looks great, really only one suggestion.

uxarray/remap/inverse_distance_weighted.py
aaronzedwick: @philipc2 Oh yeah, you are right, I can use my original code for the weights if I use an axis on the sum. Thanks for pointing that out!

philipc2 (commented):
Preparing some final figures, with examples of upsampling / downsampling using the implementation here. If all checks out, I'll approve. 

Pinging @erogluorhan incase you want to give it a quick look!

philipc2:
image

#### Nearest Neighbor Reference

image

#### Default IDW

image

The up sampling example (480km to 120km) looks so cool! 

#### IDW (k=1)

When $k=1$ for IDW, I would assume the result to look exactly like nearest neighbor, however this is the result I got

image

uxarray/core/dataarray.py
```diff
@@ -262,6 +265,36 @@ def nearest_neighbor_remap(
 
         return _nearest_neighbor_uxda(self, destination_obj, remap_to, coord_type)
 
+    def inverse_distance_weighted_remap(
+        self,
+        destination_obj: Union[Grid, UxDataArray, UxDataset],
+        remap_to: str = "nodes",
+        coord_type: str = "spherical",
+        power=2,
+        k_neighbors=8,
```
philipc2, line 274:
```suggestion
        k=8,
```

Can we change this throughout to be consistent with the `k` used in our `.query`.

philipc2 (changes requested):

philipc2:
An example of using different power values for up sampling (480 to 120)

image

aaronzedwick:
@philipc2 these are great!! I really appreciate you taking the time to generate these, it’s super cool to see the results and compare them to nearest neighbor! I will also make the k_neighbors change today.

philipc2:
> @philipc2 these are great!! I really appreciate you taking the time to generate these, it’s super cool to see the results and compare them to nearest neighbor! I will also make the k_neighbors change today.

Of course! When we get around to making a usage example notebook & writing the user guide, I'll share the code for these!

erogluorhan:
> > @philipc2 these are great!! I really appreciate you taking the time to generate these, it’s super cool to see the results and compare them to nearest neighbor! I will also make the k_neighbors change today.
> 
> Of course! When we get around to making a usage example notebook & writing the user guide, I'll share the code for these!

These cases look great

rajeeja (approved):
Looks good, notebooks and cool viz. can be in the next PR.

philipc2:
> @philipc2 these are great!! I really appreciate you taking the time to generate these, it’s super cool to see the results and compare them to nearest neighbor! I will also make the k_neighbors change today.

Let me know if there's any other changes. Were you able to modify the $k=1$ case?

aaronzedwick:
> > @philipc2 these are great!! I really appreciate you taking the time to generate these, it’s super cool to see the results and compare them to nearest neighbor! I will also make the k_neighbors change today.
> 
> Let me know if there's any other changes. Were you able to modify the k=1 case?

Right, I am looking at that now. It seems to be due to the weights calculation, it makes all the values equal. It also has an error on some files if `k=1` due to `destination_data = np.sum(
        source_data[..., nearest_neighbor_indices] * weights, axis=-1
    )`
 It causes a ValueError: `ValueError: operands could not be broadcast together with shapes (1,20,6000) (6000,1)`. This only happens when `k=1`. I think this is just because of the weights being calculated, it is meant to have multiple neighbors to do it properly with Numpy. I think the best solution would be to not allow `k=1` to be used. If users want NN they can just use the function we have for that, so I don't see the harm in not allowing them to try and use IDW as a NN function.

philipc2:
> > > @philipc2 these are great!! I really appreciate you taking the time to generate these, it’s super cool to see the results and compare them to nearest neighbor! I will also make the k_neighbors change today.
> > 
> > 
> > Let me know if there's any other changes. Were you able to modify the k=1 case?
> 
> Right, I am looking at that now. It seems to be due to the weights calculation, it makes all the values equal. It also has an error on some files if `k=1` due to `destination_data = np.sum( source_data[..., nearest_neighbor_indices] * weights, axis=-1 )` It causes a ValueError: `ValueError: operands could not be broadcast together with shapes (1,20,6000) (6000,1)`. This only happens when `k=1`. I think this is just because of the weights being calculated, it is meant to have multiple neighbors to do it properly with Numpy. I think the best solution would be to not allow `k=1` to be used. If users want NN they can just use the function we have for that, so I don't see the harm in not allowing them to try and use IDW as a NN function.

Yeah, let's restrict it to only allow values of `k > 1`, with an appropriate error. Good idea!

philipc2 (approved):
Looks great!

PR #596, opened 2023-11-21
DRAFT: Trimesh Visualization, Trimesh & Point Visualization Usage Examples

philipc2:
## Overview

* `UxDataArray.to_trimesh()` and `Grid.to_simplices()` methods for triangulating coordinates for visualization
* `UxDataArray.plot.trimesh()` and `UxDataArray.plot.rasterize(method='trimesh')` plotting methods
* Added supported for geographic projections in Point plots
* Two new usage examples
  * Point Visualization Usage Example
  * Triangulation Visualization Usage Example


uxarray/plot/dataarray_plot.py
```diff
@@ -374,6 +386,72 @@ def _polygon_raster(uxda: UxDataArray,
...
+                    width: Optional[int] = 1000,
+                    height: Optional[int] = 500,
+                    colorbar: Optional[bool] = True,
+                    cmap: Optional[str] = "Blues",
+                    aggregator: Optional[str] = "mean",
+                    interpolation: Optional[str] = "linear",
+                    xlabel: Optional[str] = "Longitude",
+                    ylabel: Optional[str] = "Latitude",
+                    **kwargs):
+    """Implementation of Polygon Rasterization."""
```
erogluorhan, line 404:
```suggestion
    """Implementation of Triangular Mesh Rasterization."""
```

uxarray/core/dataarray.py
```diff
@@ -249,6 +250,65 @@ def to_dataset(self) -> UxDataset:
...
+            raise ValueError(
+                "Data is not mapped to nodes, edge centers, or face centers.")
+
+        if projection is not None:
+            # perform geographic projection
+            lon, lat, _ = projection.transform_points(ccrs.PlateCarree(), lon,
+                                                      lat).T
+
+        # obtain triangle simplices
+        simplices = self.uxgrid.to_simplices(element, projection, override,
```
erogluorhan, line 299: This is not taking the connectivity information in the data into account that ends up creating triangles where there is no existence of faces in the grid, which you also exemplified in the usage example section "Triangulation of a Partial-Coverage Grid" (e.g. land triangles). This is hard to be acceptable because we plot as if there is data. We need to use connectivity to overcome this (you can investigate the implementation in the feature branch [plot_rasterize](https://github.com/UXARRAY/uxarray/blob/plot_rasterize/uxarray/plot/dataarray_plot.py), e.g. `_grid_to_hvTriMesh`, `_create_hvTriMesh` that use connectivity)

erogluorhan (commented):
Love the overall implementation and usage examples, but there is a significant thing that we need to address as mentioned in one of the two inline comments

uxarray/core/dataarray.py
philipc2: Thank you for pointing this out. I think now after you've explained it, it finally clicked with me.

Since we already know the connectivity and mesh structure, instead of running Delaunay triangulation on the set of points, we use triangulate each polygon. This should be much faster than Delaunay as well.

I'll put this PR back to a draft & migrate any point visualization related items (notebook, projection improvements) into a new PR)

erogluorhan: 100%

philipc2:
Closing this PR since it has become out of date. The linked issues and branch will remain up.

PR #628, opened 2023-12-19
UGRID Conformance, Grid Construction from Minimal Grid Variables

philipc2:
### UGRID Conformance & Standardization 

* Introduced in the `uxarray.conventions` directory with `ugrid.py` and `descriptors.py` modules which store the names, dimensions, and attributes associate with all unstructured grid coordinate, connectivity, and descriptor variables. 
* Existing I/O modules (i.e. MPAS, SCRIP, etc.) and other modules have been updated to use the definitions in the `uxarray.conventions` directory as opposed to re-defining them each time in their individual scopes

### Constructing a Grid from a user-defined topology
* Addition of the `Grid.from_topology()` classmethod, which can construct a grid from user-defined grid variables

### Conversion to Xarray
* Added a `Grid.to_xarray()` method for encoding the contents of a `Grid` into an Xarray Dataset.
* `Grid.encode_as()` will begin to be deprecated 

### Quality of Life Changes
* Introduced `Grid.dims`, `Grid.sizes`, `Grid.coordinates`, and `Grid.connectivity` analgolous to Xarray's implementations
* Renamed `Grid.parsed_attrs` to `Grid.attrs` to be consistent with Xarray, with a deprecation warning now raised for `Grid.parsed_attrs`
* `__getitem__` operator for `Grid` to allow for indexing to obtain variables, such as `Grid['face_node_connectivity']`


uxarray/grid/grid.py
```diff
@@ -1064,6 +1214,112 @@ def compute_face_areas(
 
         return self._face_areas, self._face_jacobian
 
+    def to_netcdf(
```
philipc2, line 1217: as per discussion with Orhan, update `Grid.encode_as` to be chained with Xarray functionality (i.e. `to_netcdf` or `to_zarr`)

philipc2: Also, possible name changes to `encode_as`

uxarray/conventions/ugrid.py
```diff
@@ -0,0 +1,323 @@
...
+    "n_max_node_edges",
+    "two",
+]
+
+NODE_DIM = "n_node"
+EDGE_DIM = "n_edge"
+FACE_DIM = "n_face"
+N_MAX_FACE_NODES_DIM = "n_max_face_nodes"
+
+NODE_COORDINATES = ["node_lon", "node_lat"]
```
hongyuchen1030, line 38: Should we also consider the Cartesian Coordinates here or for future use? Specifically, should we name it as `SPHERICAL_NODE_COORDINATES` instead of just `SPHERICAL_NODE_COORDINATES`?

uxarray/conventions/ugrid.py
```diff
@@ -0,0 +1,323 @@
...
+
+EDGE_COORDINATES = ["edge_lon", "edge_lat"]
+
+EDGE_LON_ATTRS = {
+    "standard_name": "longitude",
+    "long name": "Longitude of the center of each edge",
+    "units": "degrees_east",
+}
+
+EDGE_LAT_ATTRS = {
```
hongyuchen1030, line 61: How we calculate the edge center latitude here? Since the GCA is an arc, the maximum/midpoint of the arc might not necessary equals to the average latitude of the two end points' latitude

uxarray/conventions/ugrid.py
```diff
@@ -0,0 +1,323 @@
...
+    "standard_name": "latitude",
+    "long name": "Latitude of the center of each edge",
+    "units": "degrees_north",
+}
+
+FACE_COORDINATES = ["face_lon", "face_lat"]
+
+FACE_LON_ATTRS = {
+    "standard_name": "longitude",
+    "long name": "Longitude of the center of each face",
```
hongyuchen1030, line 71: Same here, how we define the center?

uxarray/conventions/ugrid.py
```diff
@@ -0,0 +1,323 @@
...
+}
+
+# coordinates (cartesian)
+NODE_X_ATTRS = {
+    "standard_name": "x",
+    "long name": "x location of the corner nodes of each face ",
+    "units": "meters",
+}
+
+NODE_Y_ATTRS = {
```
hongyuchen1030, line 88: Does these X, Y,Z all refer to the cartesian coordinates? We better specify that in the long name

uxarray/conventions/ugrid.py
```diff
@@ -0,0 +1,323 @@
...
+# connectivity (edge_)
+EDGE_NODE_CONNECTIVITY_ATTRS = {
+    "cf_role": "edge_node_connectivity",
+    "long name": "Maps every edge to the two nodes that it connects.",
+    "start_index": 0,
+    "dtype": INT_DTYPE,
+}
+EDGE_NODE_CONNECTIVITY_DIMS = ["n_edge", "two"]
+
+EDGE_EDGE_CONNECTIVITY_ATTRS = {
```
hongyuchen1030, line 179: How to define "neighbor" here? Do you mean that if two edges share a vertex?

uxarray/conventions/ugrid.py
philipc2: They do. This is a good idea.

uxarray/conventions/ugrid.py
philipc2: We take the Cartesian average and then convert to lat/lon

https://github.com/UXARRAY/uxarray/blob/d72a5ca3ca0401428795e2cdafa92a5d1ab8e2db/uxarray/grid/coordinates.py#L323-L326

uxarray/conventions/ugrid.py
philipc2: Similar to above, we take the Cartesian average and then convert to lat/lon

https://github.com/UXARRAY/uxarray/blob/d72a5ca3ca0401428795e2cdafa92a5d1ab8e2db/uxarray/grid/coordinates.py#L234-L238

uxarray/conventions/ugrid.py
philipc2: It would be a good idea to mention our method as an attribute.

uxarray/conventions/ugrid.py
philipc2: From the UGRID conventions

> The attribute node_coordinates points to the auxiliary coordinate variables representing the node locations (latitude, longitude, or other spatial coordinates, and optional elevation or other coordinates).

We should be able to set `NODE_COORDINATES` to what you outlined. I'll give this a try.

philipc2: The same can be done for `EDGE_COORDINATES` and `FACE_COORDINATES`

uxarray/conventions/ugrid.py
philipc2: The case where two edges share a vertex is handled by `node_edge_connectivity`

This is the definiion from the MPAS conventions

image

Section 5.2 [here ](https://mpas-dev.github.io/files/documents/MPAS-MeshSpec.pdf)explains goes more into detail. 

I'll look into a better way we should define it, or if there is any other applications other than tangential velocities.

uxarray/conventions/ugrid.py
hongyuchen1030: So I think it will be good to specify in the document that: the centroid returned by this method might not be on the arc

erogluorhan (approved):
This is a fantastic improvement to the code; thanks very much for handling this!

Other than the things @hongyuchen1030  pointed out, I don't have anything to add.

Approved!

philipc2:
> This is a fantastic improvement to the code; thanks very much for handling this!
> 
> Other than the things @hongyuchen1030 pointed out, I don't have anything to add.
> 
> Approved!

Thank you!

Issue #627, opened 2023-12-16
Add Spherical/Cartesian support for BallTree/KDTree


aaronzedwick:
### Proposed new feature or change:

The current implementation uses cartesian for KDTree and spherical for BallTree. Add support so that both tree structures can use either cartesian or spherical.

Issue #617, opened 2023-12-12
BallTree builds from "nodes" instead of "face centers"


philipc2:
### Version

v2023.12.0

### How did you install UXarray?

Source

### What happened?

Attempting to construct a BallTree from "face centers" actually builds the tree from corner nodes.

### What did you expect to happen?

Tree should be built from face centers.

### Can you provide a MCVE to repoduce the bug?

```python
tree = grid.get_ball_tree(tree_type='face centers')
tree.query([0.0, 0.0], k=2)
```

PR #616, opened 2023-12-12
Unstructured Grid Subsetting & Indexing

philipc2:
## Overview

* Slicing across grid dimensions with `Grid.isel()` and overloaded `UxDataArray.isel()`
* Subsetting API through `Grid.subset` and `UxDataArray.subset` accessors 
* Nearest Neighbor Subsetting
* Bounding Circle Subsetting
* Coordinate-Based Bounding Box Subsetting
* Fixes a bug in Tree construction for face-centers

## Expected Usage

[Link to generated usage example](https://uxarray--616.org.readthedocs.build/en/616/examples/009-subsetting.html)

PR #575, opened 2023-11-09
Add Ruff to Pre-Commit Hooks

philipc2:
Closes #576 

# Overview
- Replaces `YAPF` linting with `ruff`, which is significantly faster & more comprehensive 
- Changes throughout to adhere to `ruff` linting

PR #632, opened 2023-12-21
KDTree & BallTree Refactor

### LINKED ISSUE #631: Construct KDTree & BallTree from Edge Centers

When selecting `element='edge centers'`, the tree should be constructed from `edge_lon` and `edge_lat` or `edge_x`, `edge_y`, and `edge_z`

aaronzedwick:
Closes #631 #627 #617 #633

## Overview

Adds seven different features:
1. Our tree structures (KDTree and BallTree) are currently constructed from face centers and nodes. This adds another construction method, creating them from the edge centers of the faces, using `edge_lon` and `edge_lat` or `edge_x,` `edge_y,` and `edge_z`.
2. Currently KDTree is made using cartesian and BallTree uses spherical. This allows either tree to use both coordinate types.
3. Fixes the BallTree error that causes it to build from nodes when face centers is selected
4. Adds a reconstruction option so trees can be reconstructed on a grid. This allows switching between Cartesian and spherical. 
5. Exposes the distance metric to the user, allowing them to pick which they want to use. 
6. Fixes bug that caused an error while using query with `return_distance=False`
7. Changes `tree_type` to `coordinates` and `coordinate_type` to `coordinate_system` for more comprehensive variable names.

## Expected Usage

```Python
import uxarray as ux

center_grid_files = gridfile_mpas

uxgrid = ux.open_grid(self.center_grid_file)

# Performs a sample query
d, ind = uxgrid.get_kd_tree(coordinates="edge centers", coordinate_system="cartesian").query([1.0, 0.0, 1.0], k=1)
d2, ind2 = uxgrid.get_ball_tree(coordinates="edge centers", coordinate_system="spherical", reconstruct=True, distance_metric="minkowski").query([3.0, 3.0], k=3)
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


philipc2:
Can you add a fix to [this](https://github.com/UXARRAY/uxarray/issues/617)? I had this in #616 but this is a more fitting PR to tackle this. 

image

aaronzedwick:
> Can you add a fix to [this](https://github.com/UXARRAY/uxarray/issues/617)? I had this in #616 but this is a more fitting PR to tackle this.
> 
> image

Yup, I saw that and fixed it already, just gotta push my new changes.

philipc2:
This isn't directly related to anything you put together here, but its on the topic of the general refactor.

I think we should consider renaming the `tree_type` parameter throughout, because it doesn't really fit its purpose.

Tree Type can be interpreted as selecting a type of tree, like KD, Ball, or R, but in our case it selects what elements coordinate to use to build it (node, edge, or face)

What do you think of this naming scheme for the parameters?

`coordinates`: One of `node`, `edge`, or `face` (this is consistent with the UGRID conventions, since these coordinates usually referred to as `node_coordinates`, `edge_coordinates`, or `face_coordinates`). 

`coordinate_system`: One of `spherical` or `cartesian`

The function call would look something like this then, which is significantly more readable and understandable at a glance, imo

```Python
uxgrid.get_ball_tree(coordinates='face', coordinate_system='cartesian')
```

I would emphasize in the docstrings that the ball tree and kdtree work exclusively on coordinates, (a.k.a. point data), so there is no connectivity that is used.

aaronzedwick:
I like that a lot! I think it does a much better job describing what is happening. It’s a very good idea, I will add it to the refactor! Any other ideas are welcome, @rajeeja if you have thoughts or suggestions for this PR too I welcome them. My goal is that this can for the most part solidly finish these two tree structures, and we can reformat existing remapping functionality, as well as any future ones, to allow flexible use of either tree in a stable way. 

Also, @philipc2 do you have any ideas about the KDTree taking so long and potentially crashing on larger files? Have you messed around at all with that? I experienced this on the large mesh you gave me for testing. Any thoughts would be helpful as well!

uxarray/grid/neighbors.py
ahijevyc, line 1: In lines 655-656 of the edited version
```
    # swap X and Y for query
    xy[:, [0, 1]] = xy[:, [1, 0]]
```
can have the side effect of reversing the order of the elements when BallTree.query_radius(xy,r) is called.
In other words, the xy argument going into BallTree.query_radius is not the same as what comes out.
```
xy = [-100, 40]
_, _ = uxtree.query_radius(xy, r)
print(xy)
[40, -100]
```
It might be better to store the swapped result in a different variable, `yx` instead of `xy`.
```
    # swap X and Y for query
    yx = xy[:, [1, 0]]
```
and change to `yx` onward. For example, in the rest of the _prepare_xy_for_query function, where _prepare_xy_for_query is called, and where query and query_radius are called (I counted ~8 places).

aaronzedwick: Could we just swap the xy back at the end of the function? I think this is a solution that would work. Thoughts @philipc2?

ahijevyc: Hi @philipc2 , 
xy may be kind of a generic variable name for a location, but the first letter being x suggests to me that x (or longitude) comes first. That is why I suggested a new variable, yx, for when y (i.e.  latitude) is first.

aaronzedwick: Oh, I see that makes sense. Does reversing the variable have any effect on the end user? Or is it fine that it swaps, since we are just getting the indexes and distance?

ahijevyc: When I stored coordinates in variable xy before calling query_radius I needed xy to remain unchanged, as I was using the coordinates later for plotting.  it would reverse the elements every time xy was called. I know typically the user would only call query_radius once but since I was debugging, it was constantly getting flipped.  Either way, I wouldn't want it to be flipped, as I was using the xy coordinates later for plotting.

uxarray/grid/neighbors.py
```diff
@@ -186,25 +266,35 @@ def query_radius(self,
...
 
-            if xyz.shape[0] == 1:
+            if coords.shape[0] == 1:
                 ind = ind.squeeze()
 
             if return_distance:
+                if not in_radians and self.coordinate_type == "spherical":
+                    d = np.rad2deg(d[0])
 
                 return d, ind
```
aaronzedwick, line 299: @philipc2 is there a reason we reverse the return variables compared to sklearn here? We return `d, ind` not `ind, d`? Shouldn't we keep with the established return values of the library used?

uxarray/grid/neighbors.py
aaronzedwick: That makes sense! Thanks for noticing this, it should be simple to fix!

aaronzedwick: @philipc2 why are we swapping the x and y coordinates in the first place?

philipc2: > @philipc2 why are we swapping the x and y coordinates in the first place?

It's expected from scikit-learn when using the haversine distance as a metric

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html

rajeeja (changes requested):
looks good, please complete the test cases by adding appropriate assertions etc.

uxarray/grid/neighbors.py
aaronzedwick: @ahijevyc I tried replicating this, but the original variable xy doesn't seem to change. It shouldn't change the original xy coordinates sent in. It does change the coordinates inside the class, but it is all local to the BallTree class and shouldn't be able to reverse the outside variables you send into the function, to my understanding.  Where have you experienced this bug and could you give me your code so I can try and understand what is going on? Here is my attempt at replication:

```
xy = [-100, 40]
ind = uxgrid.get_ball_tree(tree_type="face centers",
                                          coordinate_type="spherical",
                                          distance_metric="haversine").query_radius(
           xy, r=5, return_distance=False)
print(xy)
```
Result:
`[-100, 40]`

I didn't realize this was only for haversine, so I did add a check to make sure when using other distance metrics like Minkowski it won't undergo this swap, so that is a good catch. But due to it not always occurring, I think changing the variable names doesn't make a whole lot of sense to me personally. Thoughts?

ahijevyc:
Yeah, if the elements are not always swapped then you shouldn't rename xy
to yx.

uxarray/grid/grid.py
```diff
@@ -738,7 +738,11 @@ def face_jacobian(self):
             self._face_areas, self._face_jacobian = self.compute_face_areas()
         return self._face_jacobian
 
-    def get_ball_tree(self, tree_type: Optional[str] = "nodes"):
+    def get_ball_tree(self,
+                      tree_type: Optional[str] = "nodes",
+                      coordinate_type="spherical",
+                      distance_metric="haversine",
```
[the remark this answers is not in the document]
philipc2: Haversine was selected because it accounts for the wrap around the poles, and is built for spherical (lat lon) coords.

uxarray/grid/neighbors.py
philipc2: > Shouldn't we keep with the established return values of the library used?

This is a fair point, I don't remember exactly why I chose to swap the order, other than it may have been more intuitive for me.

We can go ahead and change it to be consistent with SKlearn if there's no objections.

philipc2:
> I like that a lot! I think it does a much better job describing what is happening. It’s a very good idea, I will add it to the refactor! Any other ideas are welcome, @rajeeja if you have thoughts or suggestions for this PR too I welcome them. My goal is that this can for the most part solidly finish these two tree structures, and we can reformat existing remapping functionality, as well as any future ones, to allow flexible use of either tree in a stable way.
> 
> Also, @philipc2 do you have any ideas about the KDTree taking so long and potentially crashing on larger files? Have you messed around at all with that? I experienced this on the large mesh you gave me for testing. Any thoughts would be helpful as well!

> do you have any ideas about the KDTree taking so long and potentially crashing on larger files?

I'm not sure, however this could be worth checking to see if it leads to any improvements. (taken from the [docs](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KDTree.html))

image

uxarray/grid/neighbors.py
ahijevyc: @aaronzedwick - Yeah, I thought that was strange too. I will attempt to write a minimal code example demonstrating that reversal of coords by the function. I'll probably find an error on my part first.

ahijevyc: @aaronzedwick Here's my minimal example. It may be related to the np.flip function but I don't know why. This may be more of a numpy or general python question, rather than uxarray, but could you help me understand what is happening here?
```
import numpy as np
import uxarray

grid_path = "/glade/campaign/mmm/parc/schwartz/MPAS/15-3km_mesh/mpas_init/static.nc"
data_path = '/glade/campaign/mmm/parc/schwartz/HWT2023/mpas/2023053100/post/mem_4/diag.2023-06-01_12.00.00.nc'

uxds = uxarray.open_dataset(
    grid_path, data_path, use_dual=False
)
uxtree = uxds.uxgrid.get_ball_tree(tree_type="face centers")

c = np.flip([[-100, 40]], axis=1)
print(c)
uxtree.query_radius(c, 1)
print(c)
```

```
glade/work/ahijevyc/conda-envs/uxarray/lib/python3.11/site-packages/xarray/core/utils.py:494: FutureWarning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.
  warnings.warn(
[[  40 -100]]
[[-100   40]]
/glade/work/ahijevyc/conda-envs/uxarray/lib/python3.11/site-packages/xarray/core/utils.py:494: FutureWarning: The return type of `Dataset.dims` will be changed to return a set of dimension names in future, in order to be more consistent with `DataArray.dims`. To access a mapping from dimension names to lengths, please use `Dataset.sizes`.
  warnings.warn(

```

uxarray/grid/neighbors.py
aaronzedwick: @philipc2 I see the problem, `query_radius` in SKLearn returns ind, d but `query` returns d, ind. So they aren't consistent across their standards.

uxarray/grid/neighbors.py
ahijevyc: I see you're returning from uxtree.query_radius with a different c. Before it is [40 -100] and after it is [-100 40]. 
Is this expected? That is what I was confused about. Why should the numpy array c change after running uxtree.query_radius?

philipc2: @ahijevyc @aaronzedwick 

This seems like a Python issue!

`c` is being passed in and is being modified in place within this function, I belive

https://github.com/UXARRAY/uxarray/blob/86ac234522410da62d7cc72f57a6fd8183b3700d/uxarray/grid/neighbors.py#L465-L491

We could avoid this by taking a copy of `xy` (a.k.a. our `c`) before passing it into this function from our internal calls.

uxarray/grid/grid.py
```diff
@@ -738,57 +738,83 @@ def face_jacobian(self):
             self._face_areas, self._face_jacobian = self.compute_face_areas()
         return self._face_jacobian
 
-    def get_ball_tree(self, tree_type: Optional[str] = "nodes"):
+    def get_ball_tree(self,
+                      coordinates: Optional[str] = "nodes",
+                      coordinate_system="spherical",
```
philipc2, line 743:
```suggestion
                      coordinate_system: Optional[str]="spherical",
```

uxarray/grid/grid.py
```diff
@@ -738,57 +738,83 @@ def face_jacobian(self):
...
+                      distance_metric="haversine",
```
philipc2, line 744:
```suggestion
                      coordinate_system="spherical",
                      distance_metric: Optional[str]="haversine",
```

uxarray/grid/grid.py
```diff
@@ -738,57 +738,83 @@ def face_jacobian(self):
...
+                      reconstruct=False):
```
philipc2, line 745:
```suggestion
                      reconstruct: bool = False):
```

uxarray/grid/grid.py
```diff
@@ -738,57 +738,83 @@ def face_jacobian(self):
...
+                self._ball_tree.coordinates = coordinates
 
         return self._ball_tree
 
-    def get_kd_tree(self, tree_type: Optional[str] = "nodes"):
+    def get_kd_tree(self,
+                    coordinates: Optional[str] = "nodes",
+                    coordinate_system="cartesian",
+                    distance_metric="minkowski",
+                    reconstruct=False):
```
philipc2, line 785: Same typing as comments above

uxarray/grid/neighbors.py
```diff
@@ -33,86 +38,151 @@ class KDTree:
 
     def __init__(self,
                  grid,
-                 tree_type: Optional[str] = "nodes",
-                 distance_metric="minkowski"):
+                 coordinates: Optional[str] = "nodes",
+                 coordinate_system="cartesian",
+                 distance_metric="minkowski",
+                 reconstruct=False):
```
philipc2, line 44: Same typing as comments above

uxarray/grid/neighbors.py
```diff
@@ -33,86 +38,151 @@ class KDTree:
...
+                                   self._source_grid.node_z.values),
+                                  axis=-1)
+
+            elif self.coordinate_system == "spherical":
+                coords = np.vstack(
+                    (deg2rad(self._source_grid.node_lat.values),
+                     deg2rad(self._source_grid.node_lon.values))).T
+
+            else:
+                raise TypeError
```
philipc2, line 88: Can we populate the various TypeErrors and ValueErrors with appropriate messages?

philipc2 (changes requested):
Looking great! A few small comments.

uxarray/grid/neighbors.py
ahijevyc: @aaronzedwick @philipc2  taking a copy of the original coordinates`xy` (a.k.a. my `c`) before passing it into `_prepare_xy_for_query` sounds messy, but good, if it avoids modifying the original coordiates.

aaronzedwick: @ahijevyc @philipc2 it was something to do with how you swapped them here: `xy[:, [0, 1]] = xy[:, [1, 0]]`. I have replaced the statement with the np.flip that @ahijevyc used, it seems to be fixed just by doing that.

ahijevyc: @aaronzedwick That fixed it for me.

rajeeja:
@aaronzedwick, looks good to go, 
Please resolve the conflicts and conversation

philipc2:
> @aaronzedwick, looks good to go, Please resolve the conflicts and conversation

+1, make sure to re-install your pre-commit (see #575 )

aaronzedwick:
@philipc2 @rajeeja whenever you guys have time, could this get a review from both of you and get merged?

rajeeja (approved):
Looks good,
in the next refactor we can add logging and timing around the critical pieces of the code.

philipc2 (approved):
This looks great Aaron! Thanks for addressing everything!

PR #658, opened 2024-01-12
Compute Edge Centroids

### LINKED ISSUE #589: Compute and Populate Edge Centroid Coordinates

Compute and Populate the following variables, which represent the coordinates of the centroid of each edge.
* `edge_lon`, `edge_lat`
* `edge_x`, `edge_y`, `edge_z` 

This will most likely employ the same Cartesian average scheme used for the face centroid computations.

aaronzedwick:
Closes #589

## Overview

Computes and populates `edge_lon`, `edge_lat`, `edge_x`, `edge_y`, and `edge_z` using a similar method as the face centroids computation used.
## Expected Usage

```Python
import uxarray as ux

test_triangle = np.array([(0, 0, 1), (0, 0, -1), (1, 0, 0)])

# Open the grid and find the centroids
grid = ux.open_grid(test_triangle)
_populate_centroid_coord(grid)

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


test/test_centroids.py
```diff
@@ -62,10 +63,46 @@ def test_centroids_from_mean_verts_scrip(self):
...
+        centroid_x = np.mean(grid.node_x[grid.edge_node_connectivity[0][0:]])
+        centroid_y = np.mean(grid.node_y[grid.edge_node_connectivity[0][0:]])
+        centroid_z = np.mean(grid.node_z[grid.edge_node_connectivity[0][0:]])
+
+        # Test the values of computed first edge centroid and the populated one
+        self.assertEqual(centroid_x, grid.edge_x[0])
+        self.assertEqual(centroid_y, grid.edge_y[0])
+        self.assertEqual(centroid_z, grid.edge_z[0])
+
+    def test_edge_centroids_from_scrip(self):
```
philipc2, line 93: Can you write a test using one of the MPAS datasets (when loaded, they already have the spherical and Cartesian edge coordinates).

You can call `_construct_edge_centroids` after loading the MPAS mesh and then compare the expected to the computed results.

uxarray/grid/coordinates.py
```diff
@@ -319,6 +320,91 @@ def _construct_xyz_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fac
...
+
+        grid._ds["edge_z"] = xr.DataArray(
+            centroid_z, dims=["n_edge"], attrs={"units": "meters"}
+        )
+
+
+def _construct_edge_centroids(node_x, node_y, node_z, edge_nodes_con):
+    """Constructs the xyz centroid coordinate for each edge using Cartesian
+    Averaging."""
+    centroids = np.zeros((3, edge_nodes_con.shape[0]), dtype=np.float64)
```
philipc2, line 387:
```suggestion
    centroids = np.zeros((3, edge_nodes_con.shape[0]), dtype=node_x.dtype)
```

Minor change, but I suggest using the same data type as the node coordinates.

uxarray/grid/coordinates.py
```diff
@@ -319,6 +320,91 @@ def _construct_xyz_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fac
...
+        grid._ds["edge_y"] = xr.DataArray(
+            centroid_y, dims=["n_edge"], attrs={"units": "meters"}
+        )
+
+        grid._ds["edge_z"] = xr.DataArray(
+            centroid_z, dims=["n_edge"], attrs={"units": "meters"}
+        )
+
+
+def _construct_edge_centroids(node_x, node_y, node_z, edge_nodes_con):
```
philipc2, line 384: Can you try decorating this with @njit and see if it's compatible?

Also, for the parts above that have `@njit(cache=ENABLE_JIT_CACHE)`, we shouldn't need to cache the machine code since these functions are run at most once during the life cycle of a `Grid`.

uxarray/grid/coordinates.py
```diff
@@ -319,6 +320,91 @@ def _construct_xyz_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fac
...
+        # Convert to xyz if there are latlon centroids already stored
+        centroid_lon, centroid_lat = grid.edge_lon.values, grid.edge_lat.values
+        centroid_x, centroid_y, centroid_z = _get_xyz_from_lonlat(
+            centroid_lon, centroid_lat
+        )
+
+    # Populate the centroids
+    if "edge_lon" not in grid._ds or repopulate:
+        grid._ds["edge_lon"] = xr.DataArray(
+            centroid_lon, dims=["n_edge"], attrs={"units": "degrees_east"}
```
philipc2, line 364: This section is taken from #628, which is still a WIP, but please populate the "long name" and "standard name" attributes for all the parts here. 

https://github.com/UXARRAY/uxarray/blob/93b8b947e8d5b5442675d8a1c6cebe9d2bc87efc/uxarray/conventions/ugrid.py#L43-L53

uxarray/grid/grid.py
```diff
@@ -14,7 +14,7 @@
 from uxarray.io.utils import _parse_grid_type
 from uxarray.grid.area import get_all_face_area_from_coords
 from uxarray.grid.coordinates import (
-    _populate_centroid_coord,
+    _populate_face_centroids,
```
philipc2, line 17: Looks like you still need to import `_populate_edge_centroids` and add the population to each of our properties.

philipc2 (changes requested):

uxarray/grid/coordinates.py
aaronzedwick: If `node_x` is int32 it seems to cause an error in `node_xyz_to_lonlat_rad` due to numba, this happens in the triangle test case as it is just integer points. So using `np.float64` seems the better route, as it forces the centroids to be compatible with the conversion function.

uxarray/grid/coordinates.py
aaronzedwick: It is compatible, and you are right about the caching, good point!

test/test_centroids.py
aaronzedwick: This test was actually labeled wrong and is using an MPAS file, as the SCRIP file doesn't have any edges on it. I will change the name to reflect this, thanks for noticing this!

uxarray/grid/coordinates.py
aaronzedwick: For the cartesian ones, I see the long name is "TODO", should I leave it like that or label it something such as "x coordinate of the center of each edge"?

philipc2: Yes that works!

uxarray/grid/coordinates.py
philipc2: Interesting, is there any cases where `node_x` is an integer type? 

This is fine with me though, this works.

uxarray/grid/grid.py
```diff
@@ -537,7 +538,7 @@ def edge_lon(self) -> xr.DataArray:
         Dimensions (``n_edge``)
         """
         if "edge_lon" not in self._ds:
-            return None
+            return _populate_edge_centroids(self)
```
philipc2, line 541: We shouldn't be returning this function, we just need to call it

```suggestion
            _populate_edge_centroids(self)
```

Can you make this change to all of the ones you've added?

philipc2 (changes requested):

test/test_centroids.py
```diff
@@ -62,10 +63,46 @@ def test_centroids_from_mean_verts_scrip(self):
...
+        expected_face_x = uxgrid.edge_lon.values
+        expected_face_y = uxgrid.edge_lat.values
+
+        _populate_edge_centroids(uxgrid, repopulate=True)
+
+        computed_face_x = (uxgrid.edge_lon.values + 180) % 360 - 180
+        computed_face_y = uxgrid.edge_lat.values
+
+        nt.assert_array_almost_equal(expected_face_x, computed_face_x)
+        nt.assert_array_almost_equal(expected_face_y, computed_face_y)
```
philipc2, line 108: A couple notes about the variables here
1. It should be `edge_` and not `face_`
2. Use `edge_lon` and `edge_lat` instead of `x` and `y`

PR #664, opened 2024-01-16
Return a list of NumPy arrays when performing a radius query with multiple points

### LINKED ISSUE #663: uxtree.query_radius returns only one list even though multiple points provided
### Version

2023.12.0 main branch pulled Jan 14, 2024

Source

### What happened?

Asked for neighbors within r degrees of n locations on a uxarray grid. uxtree.query_radius only returned one list. I think it corresponded to the first location. 

It should return a list of n lists, one for each location.  Don't return just a list of neighbors for the first location.

```python
import uxarray

grid_path = "/glade/campaign/mmm/parc/schwartz/MPAS/15-3km_mesh/mpas_init/static.nc"
data_path = "/glade/campaign/mmm/parc/schwartz/HWT2023/mpas/2023053100/post/mem_4/diag.2023-06-01_12.00.00.nc"

uxds = uxarray.open_dataset(grid_path, data_path, use_dual=False)
uxtree = uxds.uxgrid.get_ball_tree(coordinates="face centers")
c = [
    [-100, 40],
    [-101, 38],
    [-102, 38],
]
idxs = uxtree.query_radius(c, 0.05)
print(idxs)
```

`[6487403 6487413 6487423 6487405 6487421  120343  120348  120339  120342
  120344  120345 6487422  120338]`
```

### PR DESCRIPTION
**Pull request by philipc2**
Closes #663

## Overview

* Fixes a bug where performing a radius query on multiple points where only a single list is returned
* Now, a list of arrays is returned, since each point we query may have a different number of elements found, meaning we can't have a fixed-size array.

### CODE REVIEW

File: uxarray/grid/neighbors.py [hunk 1]
```diff
@@ -345,24 +345,26 @@ def query_radius(
                 coords, r, return_distance, count_only, sort_results
             )
 
-            ind = np.asarray(ind[0], dtype=INT_DTYPE)
+            ind = [np.asarray(cur_ind, dtype=INT_DTYPE) for cur_ind in ind]
```
ahijevyc, line 348: Do we need to convert to integer? sklearn's KDTree.query_radius already returns integers

uxarray/grid/neighbors.py
```diff
@@ -719,23 +721,26 @@ def query_radius(
...
+            d = [np.asarray(cur_d) for cur_d in d]
 
             if coords.shape[0] == 1:
-                ind = ind.squeeze()
+                ind = ind[0]
+                d = d[0]
 
             if not in_radians and self.coordinate_system == "spherical":
-                d = np.rad2deg(d[0])
+                d = [np.rad2deg(cur_d) for cur_d in d]
```
ahijevyc, line 732: Could we just do `d  = np.rad2deg(d)`?

ahijevyc (commented):
Had a general questions about the need to convert array indices from query_radius to int. And whether one could substitute `d=np.rad2deg(d)` for `d=[np.rad2deg(d0) for d0 in d]` 
But otherwise, the fix for returning a list of NumPy arrays instead of just al single list looks good.

uxarray/grid/neighbors.py
philipc2: `INT_DTYPE` is our standard data type used for indexing, its for consistency.

ahijevyc: seems like a lot of extra typing for consistency.

ahijevyc (approved):
thanks for the response to my comments. I approve changes

PR #728, opened 2024-03-20
Ruff Updates

### LINKED ISSUE #729: Run Ruff Formatting & Linting on Notebooks

When running pre-commit, allow for ruff to run formatting and limiting on notebooks.

philipc2:
Closes #729 

## Overview

* Updates `ruff-pre-commit` to the latest version
* Enabled linting & formatting on Jupyter Notebooks, with certain rules disabled to prevent the removal of `hvplot` imports, which would otherwise led to the notebooks not executing
* Benchmark can be run locally using `asv run --bench mpas_ocean` from within the `uxarray/benchmarks` dir


docs/examples/001-working-with-unstructured-grids.ipynb
erogluorhan, line 1: In this file, the older block as follows seems better than what Ruff does:

```
foo = ux.UxDataArray(
    data = np.random.random(grid.n_face),
    dims = ["n_face"],
    uxgrid = grid
)
```

Re what Ruff did, I didn't like that
- It started the function name in the new line instead of just next to the equals sign (This behavior occurs in several other files below but will not mention repetitively)
- Grouped `dims=["n_face"], uxgrid=grid)` into the same line, which may easily lead the reader to miss the second var in any case

ruff.toml
```diff
@@ -1,3 +1,10 @@
 extend-exclude = [
 "test",
 "benchmarks"]
+
+[lint]
+ignore-init-module-imports = false
+
+[lint.per-file-ignores]
+"docs/*" = ["E402", "F401"]
+"uxarray/plot/*" = ["E402", "F401"]
```
erogluorhan, line 10: These are Flake8 codes, right?

erogluorhan (commented):
Overall, I feel like it does a good job, but there are a couple things I didn't like; in the comments below

ruff.toml
philipc2: The ones starting with "F" are flake, and "E" are pycodestyle.

https://docs.astral.sh/ruff/rules/#pyflakes-f

docs/examples/001-working-with-unstructured-grids.ipynb
philipc2: Agreed, the previous one looks much better. Let me look into what rule is doing this.

philipc2: I haven't been able to narrow down what rule is causing this, but this is most likely one of the `pep8` rules from [`pycodestyle`](https://docs.astral.sh/ruff/rules/#pycodestyle-e-w)

I personally think it looks fine. Below is an enable from Xarray's documentation of a very similar example. 


```Python
data = np.random.rand(4, 3)

locs = ["IA", "IL", "IN"]

times = pd.date_range("2000-01-01", periods=4)

foo = xr.DataArray(data, coords=[times, locs], dims=["time", "space"])

foo
```

philipc2: I'll continue looking into it if you'd like @erogluorhan, otherwise I'm okay with moving forward with this.

erogluorhan: I am fine with that; we can always revisit styling and apply something else if we have issues. Thanks for looking into it!

erogluorhan (approved):

PR #731, opened 2024-03-20
Fix CI Failures 

### LINKED ISSUE #723: CI Failing after Dask Release
As of `dask 2024.3.0`, our `TestPlot.test_clabel` test has started to fail

philipc2:
Closes #723

## Overview

* Modifications to Point Raster functionality to conform to new Dask changes. 
* Removes some old pins that weren't in `enviroment.yml`

### DIFF
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -143,8 +143,14 @@ class Grid:
         self._gdf_exclude_am = None
         self._poly_collection = None
         self._line_collection = None
-        self._centroid_points_df_proj = [None, None]
-        self._corner_points_df_proj = [None, None]
+
+        # caching for point viz
+        self._face_points_ref = {"df": None, "projection": None}
+        self._edge_points_ref = {"df": None, "projection": None}
+        self._node_points_ref = {"df": None, "projection": None}
+
+        # self._centroid_points_df_proj = [None, None]
+        # self._corner_points_df_proj = [None, None]
         self._raster_data_id = None
 
         # initialize cached data structures (nearest neighbor operations)
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -143,14 +143,6 @@ class Grid:
         self._gdf_exclude_am = None
         self._poly_collection = None
         self._line_collection = None
-
-        # caching for point viz
-        self._face_points_ref = {"df": None, "projection": None}
-        self._edge_points_ref = {"df": None, "projection": None}
-        self._node_points_ref = {"df": None, "projection": None}
-
-        # self._centroid_points_df_proj = [None, None]
-        # self._corner_points_df_proj = [None, None]
         self._raster_data_id = None
 
         # initialize cached data structures (nearest neighbor operations)
--- a/uxarray/plot/dataarray_plot.py
+++ b/uxarray/plot/dataarray_plot.py
@@ -214,38 +214,31 @@ def _point_raster(
 
     if uxda._face_centered():
         # data mapped to face centroid coordinates
-        lon = uxda.uxgrid.face_lon.values
-        lat = uxda.uxgrid.face_lat.values
+        lon, lat = uxda.uxgrid.face_lon.values, uxda.uxgrid.face_lat.values
+        data_mapping = "face"
     elif uxda._node_centered():
         # data mapped to face corner coordinates
-        lon = uxda.uxgrid.node_lon.values
-        lat = uxda.uxgrid.node_lat.values
+        lon, lat = uxda.uxgrid.node_lon.values, uxda.uxgrid.node_lat.values
+        data_mapping = "node"
     elif uxda._edge_centered():
         # data mapped to face corner coordinates
-        lon = uxda.uxgrid.edge_lon.values
-        lat = uxda.uxgrid.edge_lat.values
+        lon, lat = uxda.uxgrid.edge_lon.values, uxda.uxgrid.edge_lat.values
+        data_mapping = "edge"
     else:
         raise ValueError(
             f"The Dimension of Data Variable {uxda.name} is not Node or Face centered."
         )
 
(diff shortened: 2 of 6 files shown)


erogluorhan (commented):
An actual solution, if possible, rather than pinning would be ideal. Let me know if you explored anything that might be related to the actual issue; I am happy to look into this a bit.

philipc2:
> An actual solution, if possible, rather than pinning would be ideal. Let me know if you explored anything that might be related to the actual issue; I am happy to look into this a bit.

I can continue to do some digging into it, the failure is as follows:

> ../../../miniconda3/envs/uxarray_build/lib/python3.10/site-packages/datashader/data_libraries/dask.py:147: AssertionError

It appears to be something with Datashader and Dask. I'll continue looking into it and try to reproduce it locally.

philipc2:
This seems to be because of the user of `dask-expr` in the recent release for Dataframes, which breaks something with the way that we are caching (and honestly, we should investigate better ways to "cache" as opposed to just keeping a reference in the grid)
https://docs.dask.org/en/stable/changelog.html#v2024-3-0

Considering how quick point rasterization is, I don't think we need to cache, so I've removed it and cleaned up the point raster. 

image

image

Any thoughts?

uxarray/grid/grid.py
```diff
@@ -143,8 +143,6 @@ def __init__(
         self._gdf_exclude_am = None
         self._poly_collection = None
         self._line_collection = None
-        self._centroid_points_df_proj = [None, None]
-        self._corner_points_df_proj = [None, None]
```
erogluorhan, line 147: These were not used at all, right?

uxarray/plot/dataarray_plot.py
```diff
@@ -214,65 +213,29 @@ def _point_raster(
...
-    if recompute:
-        # need to recompute points and/or projection
-        if projection is not None:
-            lon, lat, _ = projection.transform_points(ccrs.PlateCarree(), lon, lat).T
-
-        point_dict = {"lon": lon, "lat": lat, "var": uxda.values}
-
-        # Construct Dask DataFrame
-        point_ddf = dd.from_dict(data=point_dict, npartitions=npartitions)
+    if projection is not None:
```
erogluorhan, line 228: This change seems to be beyond the scope; please clarify

uxarray/plot/dataarray_plot.py
```diff
@@ -214,65 +213,29 @@ def _point_raster(
...
+        # apply projection to coordinates
+        lon, lat, _ = projection.transform_points(ccrs.PlateCarree(), lon, lat).T
 
-        hv.extension("bokeh")
-        points = hv.Points(point_ddf, ["lon", "lat"]).opts(size=size)
+    # this will be fixed in #733
```
erogluorhan, line 232: Please clarify

uxarray/plot/dataarray_plot.py
```diff
@@ -214,65 +213,29 @@ def _point_raster(
...
 
-        # cache computed points & projection
-        if cache:
-            if uxda._face_centered() == "center":
-                uxda.uxgrid._centroid_points_df_proj[0] = point_ddf
-                uxda.uxgrid._centroid_points_df_proj[1] = projection
-            else:
-                uxda.uxgrid._corner_points_df_proj[0] = point_ddf
-                uxda.uxgrid._corner_points_df_proj[1] = projection
+    point_dict = {"lon": lon, "lat": lat, "var": uxda.data}
```
erogluorhan, line 235: Please clarify

erogluorhan (commented):
Thanks for this! However, changes in this PR seem to be beyond this PR's scope, or it is hard to understand all those changes were necessary for the Dask issue and if they are not changing the code flow significantly. Please clarify

philipc2:
> Thanks for this! However, changes in this PR seem to be beyond this PR's scope, or it is hard to understand all those changes were necessary for the Dask issue and if they are not changing the code flow significantly. Please clarify

Thanks for the review! I'll add some comments describing the sections of the code. However, these aren't new. The changes made here are, for the time being, remove our "caching" approach for the point rasters. When we use a cached dataframe, attempting to swap the data variable for the points leads to the Dask error that we were getting.

The difference in performance with or without using caching is comparable as seen in the Viz cookbook,.

Additionally, there are probably better ways for caching compared to the way we were doing it, which worked fine for standard arrays, but not so much for Dask.

Let me know what you think, I submitted an issue (#739) that will describe this in detail.

uxarray/plot/dataarray_plot.py
philipc2: #733 will prevent the need to always call `hv.extension`, instead it will only be called whenever we swap from one backend to another.

uxarray/plot/dataarray_plot.py
philipc2: There's no changes to the functionality, other than this PR does not cache the results due to the issues with Dask.

#739 will address this in the future, after some research into best ways to handle storing these dask objects.

uxarray/plot/dataarray_plot.py
philipc2: As per other comments, this removes the caching, since our approach using the Dask dataframe is leading to issues.

uxarray/plot/dataarray_plot.py
erogluorhan: I am so sorry; I missed your [yesterday's comment](https://github.com/UXARRAY/uxarray/pull/731#issuecomment-2013472875), which clarifies everything, before giving this review.

philipc2: No worries!

erogluorhan (approved):
So sorry for missing your [yesterday's comment](https://github.com/UXARRAY/uxarray/pull/731#issuecomment-2013472875) and asking clarifications. Everything is clarified now.

PR #733, opened 2024-03-20
Compare and Swap Holoviews Extension

### LINKED ISSUE #734: Repeated calls to `hv.extension` in plotting routines
When calling most plotting methods, we always set `hv.extension()` to the desired backend, even if it has already been set before.

philipc2:
Closes #734 

## Overview

* Removes duplicate and redundant calls to `hv.extension` by keeping track of the previous backend and only calling `hv.extension` when a swap occurs.


uxarray/plot/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class HoloviewsBackend:
+    """Utility class to compare and set holoviews backends for
+    visualization."""
+
+    def __init__(self):
+        self.backend = None
+
+    def compare_and_set(self, backend: str):
```
erogluorhan, line 11: While this function makes sense to me, a property-based implementation and its usage would make more sense I think, how about something like:

```
class HoloviewsExtension:
   def __init__(self):
        self._backend = None
   @property
     def backend(self): 
         return self._backend
       
     @backend.setter 
     def backend(self, backend): 
          if backend not in ["bokeh", "matplotlib"]:
            raise ValueError(
                f"Unsupported backend. Expected one of ['bokeh', 'matplotlib'], but received {backend}"
            )

        if backend != self._backend:
            hv.extension(backend)
            self._backend = backend

hv_extension_ref = HoloviewsExtension()
```

and then usage would look like:

```
from uxarray.plot.utils import hv_extension_ref

hv_extension_ref.backend("matplotlib")
```

1. The name `hv_extension_ref` still itches; maybe just "hv_extension" or something else?
2. The words "extension" and "backend" can be swapped if desired in the above code.

erogluorhan (commented):
Thanks very much for handling this! Overall I like it and will approve soon. Just couple things:
1. See the inline comment]
2. I am wondering how the cookbook would render with this. Did you (or if not, could you) install UXarray from this branch in a local copy of the cookbook and see how its plots rendering? Though, this might not even be needed, I just checked the usage examples rendered in this PR, and I think the issues seem resolved?

uxarray/plot/utils.py
philipc2: I haven't rendered the cookbook using this branch yet, but the usage examples in our documentation are only showing the backend logo when the first plot is generated or a backend is switched.

image


The suggestion for a property-based implementation looks really clean, but I want to hear your feedback on the implementation below, which wouldn't need to use a setter.

> The name hv_extension_ref still itches; maybe just "hv_extension" or something else?

> The words "extension" and "backend" can be swapped if desired in the above code.

Yeah, I think using `backend` instead of `extension` is much more clear.

Maybe instead of directly importing the object, we could access it through the module like so, with the following naming change

Inside of uxarray.plot.utils
```Python
...
# same implementation, renamed `compare_and_set` to `assign`
backend = HoloviewsExtension()
```

```Python
from uxarray.plot.utils import plotting_backend

uxarray.plot.utils.backend.assign("matplotlib")
```

This way, we don't need to worry about importing the object and can operate directly through the module.

erogluorhan: > I haven't rendered the cookbook using this branch yet, but the usage examples in our documentation are only showing the backend logo when the first plot is generated or a backend is switched.
> 
> image

This is great! And all of the plots that are adjacently happening with the same backend are rendering properly as I've observed, right? If you have also confirmed this, maybe there is no need to look into the cookbook.

erogluorhan: > The suggestion for a property-based implementation looks really clean, but I want to hear your feedback on the implementation below, which wouldn't need to use a setter.
> 
> > The name hv_extension_ref still itches; maybe just "hv_extension" or something else?
> 
> > The words "extension" and "backend" can be swapped if desired in the above code.
> 
> Yeah, I think using `backend` instead of `extension` is much more clear.
> 
> Maybe instead of directly importing the object, we could access it through the module like so, with the following naming change
> 
> Inside of uxarray.plot.utils
> 
> ```python
> ...
> # same implementation, renamed `compare_and_set` to `assign`
> backend = HoloviewsExtension()
> ```
> 
> Inside of plotting methods
> 
> ```python
> from uxarray.plot.utils import plotting_backend
> 
> uxarray.plot.utils.backend.assign("matplotlib")
> ```
> 
> This way, we don't need to worry about importing the object and can operate directly through the module.

You meant:

```
from uxarray.plot.utils import plotting_backend
 
plotting_backend.assign("matplotlib")
```
right? 

Yeah, this would also work; I am fine with either way. An additional reason for me to suggest a property was that if anytime user wants to return what backend they were on (i.e. using the property getter), but this maynot be a need or can it still be handled via holoviz's native functionality?

philipc2: Whoops, I meant the following:

```Python
# import plotting utils
import uxarray.plot.utils

# directly access object through module instead of importing object
uxarray.plot.utils.backend.assign("matplotlib")
```

erogluorhan: Ok, this looks great!

erogluorhan (approved):
Looks good to me

PR #741, opened 2024-03-25
Structured Test Grid and Plotting Improvements

philipc2:
## Overview

* Added a structured grid example mesh generated using `xESMF`
* Changed the threshold to use Vector + Antemeridian plots from 100,000 elements to 10,000, otherwise default will be Rasterized with No Antemeridian, for face-centered data
* Added support for edge-centered plotting through the `.plot` default method
* Antemeridian polygons are now excluded by default when calling Polygon visualization methods (raster and vector)
* Removed a warning on visual artifacts, which aren't present / reproducible, so there is no need for the warning.


philipc2:
@erogluorhan 

As we discussed today, should we make the default `.plot` method always rasterize our point data?

Issue #730, opened 2024-03-20
Optimize Face Bounds Construction


philipc2:
### Proposed new feature or change:

Optimize the performance when constructing `Grid.face_bounds`

erogluorhan:
Re @philipc2's benchmarking:
> The GeoFlow grid (12s execution time) has 6000 nodes and 3840 faces.

That's extremely slow for such a small grid (though relatable as the algorithm iterates over each face, and each edge of that face). Taking into account bigger grids might be used, the execution times might make the things impractical, and that makes me think that we should consider either to optimize before a release, or provide caution admonitions, if possible, in the documentation to let the user clearly know the extreme performance bottleneck.

Thoughts?

_Originally posted by @erogluorhan in https://github.com/UXARRAY/uxarray/pull/692#pullrequestreview-1963724660_

erogluorhan:
Some for-loop-based code needs to be replaced with vectorized ones the issue #346. Though, as part of here (if #346 will take too long), should we still find some intermediate (though not ideal) solutions to make our code look better I believe (rather than using a for loop and list appends), e.g. something like the following?:

`face_edges_connectivity_cartesian = np.asarray(list(map(node_lonlat_rad_to_xyz, face_edges_connectivity_lonlat)))`

_Originally posted by @erogluorhan in https://github.com/UXARRAY/uxarray/pull/692#discussion_r1541423946_

philipc2:
We should also investigate the performance of lower-level helper functions that are called throughout. 
* `_pole_point_inside_polygon`
* `extreme_gca_latitude`
* `_insert_pt_in_latlonbox`
* `_get_cartesian_face_edge_nodes`
* `_get_lonlat_rad_face_edge_nodes`

hongyuchen1030:
Hi @erogluorhan @philipc2,

Since the `zonal_mean` feature is ready for review, can we prioritize the optimization for latlon bounds next? This optimization is the building block for our `zonal_mean`, a fundamental, core feature that many users are eagerly waiting to use.

Thank you!

erogluorhan:
Hey @hongyuchen1030, sure!

Issue #202, opened 2023-01-10
The compatibility between Numba and Python vectorization


hongyuchen1030:
I noticed that for the helper function in `uxarray/helpers.py`, we are using the [Numba](https://numba.readthedocs.io/en/stable/user/overview.html), which, according to its documentation, prefer codes writing using "non python" styles (like loop) and lack supports for some features (like nested array)

`Numba generates optimized machine code from pure Python code using the [LLVM compiler infrastructure](http://llvm.org/). With a few simple annotations, array-oriented and math-heavy Python code can be just-in-time optimized to performance similar as C, C++ and Fortran, without having to switch languages or Python interpreters.`

However, at our upper levels like "uxarray/grid.py", we are using the python vectorization to boost our performance: map functions, Ndarray manipulation, and so on, which seems like the opposite direction of the Numba.

I wonder if these two methods can be compatible with each other, in other words, will the mixing of these two styles slow down the performance?

erogluorhan:
This is a great question. Let us investigate this. Pinging @anissa111 @rajeeja

rajeeja:
> In general, if you vectorize numpy, you end up running most things in the C layer and numba doesn't give you a computation speedup. However, numba will remove a *lot* of temporary memory copies, and reduce intermediate memory allocations. This brings a giant speedup usually.
> 
> Since you've already made the choice to choose numba, I would optimize for readability and easy development  first. And then use numba on the hot paths.

@dcherian I mostly agree with your comments and would like to keep numba as I have seen good improvements in the _integrate_ functionality.

There were a few versioning and compatibility issues with numba that caused a bit of pain.

@hongyuchen1030 do you think we can match the optimization that numba provides with the changes you propose?

hongyuchen1030:
> > In general, if you vectorize numpy, you end up running most things in the C layer and numba doesn't give you a computation speedup. However, numba will remove a _lot_ of temporary memory copies, and reduce intermediate memory allocations. This brings a giant speedup usually.
> > Since you've already made the choice to choose numba, I would optimize for readability and easy development  first. And then use numba on the hot paths.
> 
> @dcherian I mostly agree with your comments and would like to keep numba as I have seen good improvements in the _integrate_ functionality.
> 
> There were a few versioning and compatibility issues with numba that caused a bit of pain.
> 
> @hongyuchen1030 do you think we can match the optimization that numba provides with the changes you propose?

[...]

PR #748, opened 2024-04-03
Vectorize Coordinate Conversion Functions

### LINKED ISSUE #346: Support `Numpy.ndArray` for Node 2D/3D Coordinate Conversion Functions
Enable `node_xyz_to_lonlat_rad`, `node_lonlat_rad_to_xyz`, `normalize_in_place` take `np.array()` as input and output

**Link to original functionality implementation elsewhere, if applicable**
Originigaly:
1. It only takes `list` as input and the output is also `list`
**Links to other examples, if applicable**

Modify the current `node_xyz_to_lonlat_rad`, `node_lonlat_rad_to_xyz`, `normalize_in_place` helper functions
such that: 1. it takes `np.array()` or `list` as input. 2. The output is in the `np.array()` format

**Additional context**

philipc2:
Closes #346 

## Overview

* Vectorized implementations of `_lonlat_rad_to_xyz`, `_xyz_to_lonlat_rad`, and `_xyz_to_lonlat_deg`
* Updates to existing coordinate population functions to use the vectorized versions.

## Usage Example
```Python
import uxarray as ux 
import numpy as np
from uxarray.grid.coordinates import _lonlat_rad_to_xyz, _xyz_to_lonlat_rad

# assume an initialized grid
uxgrid = ux.Grid()

lon_rad = np.deg2rad(uxgrid.node_lon)
lat_rad = np.deg2rad(uxgrid.node_lat)

# conversion is vectorized across the entire arrays of coordinates
x, y, z = _lonlat_rad_to_xyz(lon_rad, lat_rad)


```


philipc2:
Hi @hongyuchen1030 

I'm almost done with vectorizing all the functions. The only issue I'm running into is one of the point inside polygon tests is now failing (`test_pole_point_inside_polygon_from_vertice_cross`)

I stepped through the function, and was able to determine that the issue somewhere in this portion of the `_pole_point_inside_polygon` method. 


```Python
        return (
            _check_intersection(ref_edge_north, north_edges)
            + _check_intersection(ref_edge_south, south_edges)
        ) % 2 != 0
```

With the changes I've made here, the sum of the intersections is 2, however prior to the change the number of intersections is 1. 

Do you know what might be causing this? Could it be a precision issue?

hongyuchen1030:
> Hi @hongyuchen1030
> 
> I'm almost done with vectorizing all the functions. The only issue I'm running into is one of the point inside polygon tests is now failing (`test_pole_point_inside_polygon_from_vertice_cross`)
> 
> I stepped through the function, and was able to determine that the issue somewhere in this portion of the `_pole_point_inside_polygon` method.
> 
> ```python
>         return (
>             _check_intersection(ref_edge_north, north_edges)
>             + _check_intersection(ref_edge_south, south_edges)
>         ) % 2 != 0
> ```
> 
> With the changes I've made here, the sum of the intersections is 2, however prior to the change the number of intersections is 1.
> 
> Do you know what might be causing this? Could it be a precision issue? Every other test is passing except for this one.

Thank you very much for your work! I feel like the problem might be a combination of precision and the coordinate conversion style for this test case.

So the test case face tests a very extreme case: The face goes across the equator while still encompassing the North Pole. The face may be referred to as the "bigger face" on the sphere, while it should refer to the "smaller face". I can look into this tomorrow and see what I can find out!

But if you feel this test case is unrealistically complicate for our users, you can feel free to remove it!

philipc2:
> > Hi @hongyuchen1030
> > I'm almost done with vectorizing all the functions. The only issue I'm running into is one of the point inside polygon tests is now failing (`test_pole_point_inside_polygon_from_vertice_cross`)
> > I stepped through the function, and was able to determine that the issue somewhere in this portion of the `_pole_point_inside_polygon` method.
> > ```python
> >         return (
> >             _check_intersection(ref_edge_north, north_edges)
> >             + _check_intersection(ref_edge_south, south_edges)
> >         ) % 2 != 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With the changes I've made here, the sum of the intersections is 2, however prior to the change the number of intersections is 1.
> > Do you know what might be causing this? Could it be a precision issue? Every other test is passing except for this one.
> 
> Thank you very much for your work! I feel like the problem might be a combination of precision and the coordinate conversion style for this test case. I tried to change our error tolerance to 10-3 but it's still failing, so I don't think that's the case (at least not the only factor)
> 
> So the test case face tests a very extreme case: The face goes across the equator while still encompassing the North Pole. The face may be referred to as the "bigger face" on the sphere, while it should refer to the "smaller face". I can look into this tomorrow and see what I can find out!
> 
> But if you feel this test case is unrealistically complicate for our users, you can feel free to remove it!

Thanks for the response! If you'd like to look into it, that would be greatly appreciated. I'll be requesting reviews soon, so if we don't get to fixing this test, we can comment it out and put together an issue for this edge case.

uxarray/grid/coordinates.py
hongyuchen1030, line 1: https://github.com/UXARRAY/uxarray/blob/edb1b794eab0003328e2cf98ac0d60eed87210de/uxarray/grid/coordinates.py#L55

Something I found here:
the line `lat = np.where(np.abs(z) > 1.0 - ERROR_TOLERANCE, np.sign(z) * np.pi / 2, lat)` will turn lat from a scalar to an size1 `np.ndarray`

For example, before this line, `lat = 0.64052231`, after the line `lat = array(0.64052231)`, is this something we're expecting?

hongyuchen1030:
Hi @philipc2 , I think I have figured out why it will break, it's related to the coordinates conversion style

in my current `node_xyz_to_lonlat_rad` it has the following lines, you can tell it will make the pole point presentation uniform in a way such that northpole=[0.0, 0.5 * np.pi], southpole=[0.0, -0.5 * np.pi]. However, there's no such uniform function in you branch. All my functions is built on top of the uniform presentation of the north pole and south pole.

So when in your branch, one of the south pole was presented as [3.14,-0.5 * np.pi ], and it cannot be handled by my function.
```python
    if np.absolute(dz) < (1.0 - ERROR_TOLERANCE):
        d_lon_rad = math.atan2(dy, dx)
        d_lat_rad = np.arcsin(dz)

        if d_lon_rad < 0.0:
            d_lon_rad += 2.0 * np.pi
    elif dz > 0.0:
        d_lon_rad = 0.0
        d_lat_rad = 0.5 * np.pi
    else:
        d_lon_rad = 0.0
        d_lat_rad = -0.5 * np.pi
```

After I add an extra `lon = 0.0` here below in your `_xyz_to_lonlat_rad`, it passes. I

```python
    if scalar:
        if np.abs(z) > 1.0 - ERROR_TOLERANCE:
            lat = np.sign(z) * np.pi / 2
            lon = 0.0
```

I feel like this will be a good test case for the new coordinates conversion function.

uxarray/grid/coordinates.py
philipc2: `lat = 0.64052231` and `lat = array(0.64052231)` are both treated the same by `numpy`. 

https://stackoverflow.com/questions/9814226/how-can-i-get-the-value-stored-in-a-0-dimensional-numpy-array

philipc2:
> Hi @philipc2 , I think I have figured out why it will break, it's related to the coordinates conversion style
> 
> in my current `node_xyz_to_lonlat_rad` it has the following lines, you can tell it will make the pole point presentation uniform in a way such that northpole=[0.0, 0.5 * np.pi], southpole=[0.0, -0.5 * np.pi]. However, there's no such uniform function in you branch. All my functions is built on top of the uniform presentation of the north pole and south pole.
> 
> So when in your branch, one of the south pole was presented as [3.14,-0.5 * np.pi ], and it cannot be handled by my function.
> 
> ```python
>     if np.absolute(dz) < (1.0 - ERROR_TOLERANCE):
>         d_lon_rad = math.atan2(dy, dx)
>         d_lat_rad = np.arcsin(dz)
> 
>         if d_lon_rad < 0.0:
>             d_lon_rad += 2.0 * np.pi
>     elif dz > 0.0:
>         d_lon_rad = 0.0
>         d_lat_rad = 0.5 * np.pi
>     else:
>         d_lon_rad = 0.0
>         d_lat_rad = -0.5 * np.pi
> ```
> 
> After I add an extra `lon = 0.0` here below in your `_xyz_to_lonlat_rad`, it passes. I
> 
> ```python
>     if scalar:
>         if np.abs(z) > 1.0 - ERROR_TOLERANCE:
>             lat = np.sign(z) * np.pi / 2
>             lon = 0.0
> ```
> 
> I feel like this will be a good test case for the new coordinates conversion function.

Thank you for the explanation! It's working for me locally now too.

uxarray/grid/coordinates.py
hongyuchen1030: The reason I discovered it because the return is `[lon, array(lat)]`, for example when I call the `_node_xyz_to_lonlat_rad` on an array of vertices, I will sometimes get `[[3.14, array(-1.57)],[0, array(1.57)],[1, array(0.29)]]` as the return. The dimensions don't look uniform for the longitude and latitude.

philipc2: image

If we convert that 2D matrix into a numpy array, it correctly represents the scalar values.

Even when we don't convert to numpy, this still shouldn't cause any issues since a 0-dim numpy array is treated as a scalar.

For example, using the math library:

image

philipc2: I've actually mostly left the original implementations the same, so there isn't any issues with using the lists. 

For example, some of the functions would have a `gca_pt` (or something similar) that would be a list of cartesiain coordinates: `[x, y, z]`. We can still use these new functions with minimal changes.

```Python
gca_pt = [0, 1, 0]

# unpack a list for parameters
lon, lat = _xyz_to_lonlat_rad(*gca_pt)

# equivalent to 
lon, lat = _xyz_to_lonlat_rad(gca_pt[0], gca_pt[1], gca_pt[2])
```

Even if one of these returns a `np.array(0)` or something similar, it hasn't broken anything and is still being treated as a scalar.

philipc2:
For combability, the coordinate conversion functions here are all written in pure Numpy, no longer depending on Numba in certain cases.

The main reason for this is that many Numpy functions (such as `linalg.norm`) when used with Numba don't support optional parameters.

Once Rachel joins us this summer, we can do some further Numba exploration as was described in #202

philipc2:
Some preliminary performance analysis of running the code vectorized compared to iteratively. 

image

image

About a 150x improvement for Cartesian to Spherical and 50x for Spherical to Cartesian.

erogluorhan:
Performance improvements sound great; I am excited to review it as soon as I can (aiming at today)

test/test_geometry.py
```diff
@@ -441,7 +447,7 @@ def test_populate_bounds_normal(self):
 
         # Convert everything into radians
         vertices_rad = np.radians(vertices_lonlat)
-        vertices_cart = [node_lonlat_rad_to_xyz(v) for v in vertices_rad]
+        vertices_cart = [_lonlat_rad_to_xyz(*v) for v in vertices_rad]
```
erogluorhan, line 450: Is it intentional to keep the for-loop? If not, calling the vectorized function still within a for loop does not look elegant and maybe confusing for the reader. There are many uses similar to this throughout the code.

erogluorhan (commented):
This is looking great; just one comment below

test/test_geometry.py
philipc2: Yes, I intentionally left the code in the test cases iterative for the sake of making the tests compatible with this new implementation.

If you'd like, I can go ahead and try to update all of the existing calls in the test cases, but I didn't see it as a priority. 

I've also left some instances in our code-base untouched for the same reasons, especially in Hongyu's helper functions, since those would require more than a simple plug-in replacement to be vectorized

#730 will aim to replace these in our code base, at least for the helpers related to the face bound computations.

uxarray/grid/coordinates.py
```diff
@@ -207,6 +141,25 @@ def _populate_lonlat_coord(grid):
     )
 
 
+def _populate_node_xyz(grid) -> None:
+    """Populates the Cartesiain node coordinates of a Grid (`node_x`, `node_y`
+    and `node_z`)"""
+
+    node_lon_rad = np.deg2rad(grid.node_lon.values)
+    node_lat_rad = np.deg2rad(grid.node_lat.values)
+    x, y, z = _lonlat_rad_to_xyz(node_lon_rad, node_lat_rad)
```
philipc2, line 150: @erogluorhan 

Relating to your comment, I've used the vectorized implementation when populating coordinates. This will lead to significant performance increases when constructing coordinates that were not present in the original grid.

uxarray/grid/arcs.py
```diff
@@ -58,9 +60,13 @@ def point_within_gca(pt, gca_cart, is_directed=False):
...
-    pt_lonlat = node_xyz_to_lonlat_rad(_to_list(pt))
-    GCRv0_lonlat = node_xyz_to_lonlat_rad(_to_list(gca_cart[0]))
-    GCRv1_lonlat = node_xyz_to_lonlat_rad(_to_list(gca_cart[1]))
+    pt_lonlat = np.array(_xyz_to_lonlat_rad(pt[0], pt[1], pt[2]))
+    GCRv0_lonlat = np.array(
+        _xyz_to_lonlat_rad(gca_cart[0][0], gca_cart[0][1], gca_cart[0][2])
+    )
+    GCRv1_lonlat = np.array(
+        _xyz_to_lonlat_rad(gca_cart[1][0], gca_cart[1][1], gca_cart[1][2])
+    )
```
philipc2, line 69: For example here, I've kept the "scalar" implementation, since making this entire helper function vectorized would require a significant re-write.

rajeeja:
> For combability, the coordinate conversion functions here are all written in pure Numpy, no longer depending on Numba in certain cases.
> 
> The main reason for this is that many Numpy functions (such as `linalg.norm`) when used with Numba don't support optional parameters.
> 
> Once Rachel joins us this summer, we can do some further Numba exploration as was described in #202

Good work. 

After some profiling and understanding the number of calls to certain functions - I believe in introducing some numba decorator again, that were removed from this PR - specifically the one's in area calculation; it might speedup things. 

Getting things compatible and decorating all function->function->function with Numba can be a pain. It is known that Numba struggles with functions that use dynamic data structures.

rajeeja:
> Some preliminary performance analysis of running the code vectorized compared to iteratively.
> 
> image
> 
> image
> 
> About a 150x improvement for Cartesian to Spherical and 50x for Spherical to Cartesian.

Is my understanding correct: Removing numba and using numpy vectorization caused this?

rajeeja:
There is some numba in grid/ geometry, neighbors and connectivity - all mostly non-nested functions. A note for #202

rajeeja (approved):

philipc2:
> There is some numba in grid/ geometry, neighbors and connectivity - all mostly non-nested functions. A note for #202

Thanks for that! Can you reference that in the issue directly too?

philipc2:
> > Some preliminary performance analysis of running the code vectorized compared to iteratively.
> > image
> > image
> > About a 150x improvement for Cartesian to Spherical and 50x for Spherical to Cartesian.
> 
> Is my understanding correct: Removing numba and using numpy vectorization caused this?

Correct, the vectorization timings can be achieved when running this function on the entire arrays, as opposed to on each value itteratievely. 

```Python
x, y, z = _lonlat_rad_to_xyz(lon_rad, lat_rad)
```

philipc2:
> > For combability, the coordinate conversion functions here are all written in pure Numpy, no longer depending on Numba in certain cases.
> > The main reason for this is that many Numpy functions (such as `linalg.norm`) when used with Numba don't support optional parameters.
> > Once Rachel joins us this summer, we can do some further Numba exploration as was described in #202
> 
> Good work.
> 
> After some profiling and understanding the number of calls to certain functions - I believe in introducing some numba decorator again, that were removed from this PR - specifically the one's in area calculation; it might speedup things.
> 
> Getting things compatible and decorating all function->function->function with Numba can be a pain. It is known that Numba struggles with functions that use dynamic data structures.

Yeah, I'm hoping we can explore this more this summer, since there is still room for improvement with the performance.

uxarray/grid/coordinates.py
erogluorhan: Yes, this is what I meant for the test cases. Thanks!

erogluorhan: Relatedly, do we have a lot of other places in our (non-test) code, which is not in the scope of #730 either, that we will need to switch to vectorized call?

test/test_geometry.py
erogluorhan: Thanks for the clarification! 

#730 sounds good to me regarding the code-base changes. Actually, not only Hongyu's helpers, but I believe we should aim to change all of such calls to vectorized calls for optimization purposes, and I think if it is too much for this PR, it can be handled s part of Rachel's optimization efforts.

The same applies to test case changes; I think we need to address them either as part of this PR or create an issue for that and deal with it soon. I believe the changes would even help with the tests' execution times in our CI action runs, right? I also think the changes should be straightforward to still keep testing for arrays rather than scalars, similar to what you did [here](https://github.com/UXARRAY/uxarray/pull/748/files#r1600845378), right?

uxarray/grid/coordinates.py
philipc2: > Relatedly, do we have a lot of other places in our (non-test) code, which is not in the scope of #730 either, that we will need to switch to vectorized call?

No, there only places we'd need to update the calls to be vectorized are
* Zonal average helper functions (`integrate.py`)
* Face area helper functions (`area.py`)
* Arc helper functions (`arc.py`)

uxarray/grid/arcs.py
erogluorhan: That's okay for the scope of this PR, assuming this will be handled in #730 or something similar, right?

philipc2: Correct

test/test_geometry.py
philipc2: > https://github.com/UXARRAY/uxarray/issues/730 sounds good to me regarding the code-base changes. Actually, not only Hongyu's helpers, but I believe we should aim to change all of such calls to vectorized calls for optimization purposes, and I think if it is too much for this PR, it can be handled as part of Rachel's optimization efforts.

Agreed, the following modules are the ones I was able to find that still call these functions iteratively. 

* Zonal average helper functions (`integrate.py`)
* Face area helper functions (`area.py`)
* Arc helper functions (`arc.py`)

> The same applies to test case changes; I think we need to address them either as part of this PR or create an issue for that and deal with it soon. BTW, the changes would even help with the tests' execution times in our CI action runs, right? I also think the changes should be straightforward to still keep testing for arrays rather than scalars, similar to what you did [here](https://github.com/UXARRAY/uxarray/pull/748/files#r1600845378), right?

I can make changes to the test cases where appropriate. It should be a simple fix.

erogluorhan: Sounds good; let me know once you are done, then we can push this forward

test/test_geometry.py
```diff
@@ -470,7 +477,7 @@ def test_populate_bounds_antimeridian(self):
 
         # Convert everything into radians
         vertices_rad = np.radians(vertices_lonlat)
-        vertices_cart = [node_lonlat_rad_to_xyz(v) for v in vertices_rad]
+        vertices_cart = np.vstack([_lonlat_rad_to_xyz(vertices_rad[:, 0], vertices_rad[:, 1])]).T
```
philipc2, line 480: @erogluorhan 

I've updated the ones that could be easily vectorized. Some of the others would require significant re-writes of the test cases, which is beyond the scope of this PR, and many of them are iterating over very small numbers of points (i.e. 4, 5, 6), so performance is not a concern.

erogluorhan: That sounds great! Please open an issue for them for future reference even if we will cover them together with the face bounds optimization in a single PR. Regardless, I am approving this now. Great work!

erogluorhan (approved):
Great work!

PR #760, opened 2024-04-09
Support Xarray Plotting Routines

### LINKED ISSUE #759: Support Xarray Plotting Routines

As of right now, Xarray's plotting methods (i.e. `plot.hist()`) are not accessible with the way the custom `plot` accessor was written.

We should still support Xarray's functionality, even though some of them may break with unstructured grids.

philipc2:
Closes #759

## Overview

* Adds support to call Xarray plotting methods through the `UxDataArray` and `UxDataset` plotting accessors

## Expected Usage

```Python
import uxarray as ux

# call Xarray Plotting Methods
uxds['var'].plot.hist()
uxds['var'].plot.line()
# and so on

```


philipc2:
Examples

```Python
uxds['v1'][0][0].plot.hist(bins=60, figsize=(10, 5))
```

image

```Python
uxds['v1'][0].mean("n_node").plot.line("o--", figsize=(10, 5))
```

image

philipc2:
Update the `.plot` accessor docstring to mention the use of Xarray functionality.

philipc2:
I have fixed the `matplotlib` backend issue we were encountering. Previously, if we would call `backend='matplotlib'` for any of our HoloViz functionality, it would set the backend used in the `matplotlib` package to `agg`, which differs from the default one in `xarray`, which is set to `"module://matplotlib_inline.backend_inline"`

With the changes, we the `xarray` methods no longer have any issues when `backend='matplotlib'` is set for any of our HoloViz functionality

```Python
# holoviews with a matplotlib backend
uxds['t2m'].plot(backend='matplotlib')

# xarray method, would not render before, but with the change it works
uxds['t2m'].plot.hist()
```

PR #808, opened 2024-06-03
Implement `Grid.open_file` method support non xarray formats including Shapefile/Geojson

### LINKED ISSUE #635: Implement open_file method of Grid class, support  non xarray formats Shapefile/Geojson

Add reader shapefile/geojson file
There is post about some existing options [here](https://gis.stackexchange.com/questions/113799/reading-shapefile-in-python)

UXarray requires only certain aspects from the file - coords, connectivity, so implementing something from scratch would be the way to do. Details about the format is [here](https://www.esri.com/content/dam/esrisites/sitecore-archive/Files/Pdfs/library/whitepapers/pdfs/shapefile.pdf)

rajeeja:
Closes #635 

Support Polygon and Multipolygon


rajeeja:
Hey @philipc2 it seems like  INT_FILL_VALUE is going over allowable value for a C long causing the overflow error. [ci failure](https://github.com/UXARRAY/uxarray/actions/runs/9359919487/job/25764444068?pr=808#step:8:58) change logic or fix this?

uxarray/core/api.py
```diff
@@ -80,6 +82,11 @@ def open_grid(
         # construct Grid from face vertices
         uxgrid = Grid.from_face_vertices(grid_filename_or_obj, latlon=latlon)
 
+    # # check to see if this is a shapefile
+    # elif extension[1] == ".shp":
+    #     print("This is a shapefile")
+    #     uxgrid = Grid.from_shapefile(grid_filename_or_obj)
+
```
philipc2, line 89: Make sure to update this logic or remove the comments.

philipc2:
I'm wondering if this implementation can be generalized as a `Grid..from_geopandas` or similar method as discussed in https://github.com/UXARRAY/uxarray/issues/635#issuecomment-1907005496

rajeeja:
> I'm wondering if this implementation can be generalized as a `Grid.from_geopandas()` or similar method as discussed in https://github.com/UXARRAY/uxarray/issues/635#issuecomment-1907005496

This makes sense, I will try to generalize it and at least add a test case for geojson also. It would have to be seen how the reader object is different when reading different kinds of format that are supported by geopandas: We may support a limited subset and we can specify what we support in test cases.

Thanks so much for the review. I will push the changes shortly

philipc2:
> will



> > I'm wondering if this implementation can be generalized as a `Grid.from_geopandas()` or similar method as discussed in [#635 (comment)](https://github.com/UXARRAY/uxarray/issues/635#issuecomment-1907005496)
> 
> This makes sense, I will try to generalize it and at least add a test case for geojson also. It would have to be seen how the reader object is different when reading different kinds of format that are supported by geopandas: We may support a limited subset and we can specify what we support in test cases.
> 
> Thanks so much for the review. I will push the changes shortly

Maybe for our `Grid`, it would make the most sense to have a `Grid.from_geodataframe()` method, where the user can provide a [`GeoDataFrame` ](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.html) that contains the geometry we will read in. 

The workflow would require the user to use the [`geopandas.read_file()`](https://geopandas.org/en/stable/docs/reference/api/geopandas.read_file.html#geopandas.read_file) or similar methods to first obtain a `GeoDataFrame` and then pass it into a `Grid`


```Python
import geopandas as gp
import uxarray as ux

# load in a shape (or other) file
gdf = gp.read_file(...)

# create a Grid from the GeoDataFrame
uxgrid = ux.Grid.from_geodataframe(gdf)
```

Pinging @kmsampson for any thoughts.

test/meshfiles/geojson/sample_chicago_buildings.geojson
```diff
@@ -0,0 +1,17 @@
+{
```
erogluorhan, line 1: This is a general comment about all the files this PR adds, not only this one: There are several folders and files of different extensions, which many of our developers and users might not be familiar with. For future developer's reference, would it make sense to write a README under each of the `shp` and `geojson` folders that would explain some things such as:
- What is each file extension (.geojson, .cpg, dbf, .shp, .shx) for?
- What are groups of files (5poly, multipoly, etc.) for?

erogluorhan (commented):
Thanks for this! Overall implementation looks good to me, but please see a few comments I have below

uxarray/io/_geopandas.py
```diff
@@ -0,0 +1,267 @@
...
+    node_index = 0
+
+    for _, row in gdf.iterrows():
+        geometry = row["geometry"]
+        if geometry.geom_type == "Polygon":
+            node_lat, node_lon, connectivity, node_index = _read_polygon(
+                geometry, node_lat, node_lon, connectivity, node_index
+            )
+        elif geometry.geom_type == "MultiPolygon":
+            node_lat, node_lon, connectivity, node_index = _read_multipolygon(
```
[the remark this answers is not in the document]
rajeeja: sometimes they are very high 100's as geojson and shapefiles have 2D shapes covering boundaries.

test/meshfiles/geojson/sample_chicago_buildings.geojson
rajeeja: done.

uxarray/io/_geopandas.py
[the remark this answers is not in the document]
erogluorhan: Then, maybe a few ideas to optimize this:

- If you know the end size of the numpy arrays, node_lon and node_lat, create a numpy array in that size upfront, then perform append 
- Otherwise, append to a list rather than numpy array. At the very end of the operation, convert list to numpy array

What do you think?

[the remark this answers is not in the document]
rajeeja: > Then, maybe a few ideas to optimize this:
> 
> * If you know the end size of the numpy arrays, node_lon and node_lat, create a numpy array in that size upfront, then perform append
> * Otherwise, append to a list rather than numpy array. At the very end of the operation, convert list to numpy array
> 
> What do you think?

Using lists as  list is an amortized O(1), thanks for this suggestion I have just pushed changes.

uxarray/grid/grid.py
```diff
@@ -231,6 +231,24 @@ def from_dataset(
 
         return cls(grid_ds, source_grid_spec, source_dims_dict)
 
+    @classmethod
+    def from_geodataframe(
```
philipc2, line 235:
```suggestion
    def from_geometry_file()
```

This signature might be more appropriate, since we aren't explicity passing in a GeoDataFrame to this function. Instead, it's a file containing the geometry in a format that is then read in as a GeoDataFrame. 

This signature might be more appropriate, since we aren't passing in a GeoDataFrame at this step, instead it's a file

uxarray/grid/grid.py
```diff
@@ -231,6 +231,24 @@ def from_dataset(
...
+        cls,
+        filename: str,
+        **kwargs,
+    ):
+        """Constructs a ``Grid`` object from a shapefile.
```
philipc2, line 240: This should be updated.

uxarray/grid/grid.py
```diff
@@ -231,6 +231,24 @@ def from_dataset(
...
+
+        Parameters
+        ----------
+        filename : str
+           shapefile filename
```
philipc2, line 245: Mention that any data that is supported by `geopandas.read_file()` is supported here.

https://geopandas.org/en/stable/docs/reference/api/geopandas.read_file.html#geopandas-read-file

uxarray/grid/grid.py
```diff
@@ -231,6 +231,24 @@ def from_dataset(
...
+        """
+        source_grid_spec = "Shapefile"
```
philipc2, line 247: This should be generalized, either to `"Geometry File (format)"` or simply just `(format)`

uxarray/core/api.py
```diff
@@ -61,6 +61,8 @@ def open_grid(
     >>> uxgrid = ux.open_grid("grid_filename.g")
     """
 
+    # extension = os.path.splitext(grid_filename_or_obj)
+
     if "source_grid" in kwargs.keys():
```
philipc2, line 66: We can extend the API to support the different geometry files using `open_grid` and `open_dataset` in a separate PR, unless you'd like to do so in this one.

philipc2 (changes requested):
Nice work, a few comments:

philipc2 (commented):
We should consider a basic User Guide notebook for this too. I can provide the `quad-hexagon` grid in multiple different formats (i.e. shape, geoparquet, etc.) to showcase this functionality.

uxarray/core/api.py
rajeeja: how about a new PR for this and the notebook that you mention in the next comment?

rajeeja:
> We should consider a basic User Guide notebook for this too. I can provide the `quad-hexagon` grid in multiple different formats (i.e. shape, geoparquet, etc.) to showcase this functionality.

I have some mesh files as a part of this, we want to be careful with how many we bundle as a part of the repo.

uxarray/grid/grid.py
rajeeja: `from_geometry_file` is too generic - that would mean any geometry, quick one `from_geopandas_geometry`?

uxarray/grid/grid.py
rajeeja: thanks for the catch, fixed. please review.

uxarray/grid/grid.py
rajeeja: I gave it specific name based on extensions from filename - isn't that better?

erogluorhan (approved):
This is great addition; thanks a lot!

erogluorhan:
> > We should consider a basic User Guide notebook for this too. I can provide the `quad-hexagon` grid in multiple different formats (i.e. shape, geoparquet, etc.) to showcase this functionality.
> 
> I have some mesh files as a part of this, we want to be careful with how many we bundle as a part of the repo.

Maybe using online files rather than having them all hosted in the repo?

philipc2 (changes requested):
Please update the title & linked issue to reflect the generalization changes that were made.

philipc2:
> > > We should consider a basic User Guide notebook for this too. I can provide the `quad-hexagon` grid in multiple different formats (i.e. shape, geoparquet, etc.) to showcase this functionality.
> > 
> > 
> > I have some mesh files as a part of this, we want to be careful with how many we bundle as a part of the repo.
> 
> Maybe using online files rather than having them all hosted in the repo?

Do we have an issue open for this topic in general?

philipc2 (changes requested):

uxarray/grid/grid.py
rajeeja: > Good point.
> 
> Can we instead create a `Grid.from_file()`, which can branch either to Xarray or GeoPandas. It can take a parameter `backend` which can either be `xarray` or `geopandas` for reading a geometry file.
> 
> For reference: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.from_file.html
> 
> Any other ideas @erogluorhan

I like a more generic` from_file` with a backend; xarray doesn't have a from_file, it does have a `from_dict` which we don't need for unstructured grid. @erogluorhan should I make this change?

uxarray/grid/grid.py
```diff
@@ -232,6 +234,52 @@ def from_dataset(
...
         return cls(grid_ds, source_grid_spec, source_dims_dict)
 
+    @classmethod
+    def from_file(
+        cls,
+        filename: str,
+        backend: Optional[str] = "geopandas",
+        **kwargs,
+    ):
+        """Constructs a ``Grid`` object from a using geopandas read_file.
```
philipc2, line 244: This should now reflect the new changes.

""Attempts to read a grid file using either Xarray or GeoPandas and constructs a ``Grid`` object.

uxarray/grid/grid.py
```diff
@@ -232,6 +234,52 @@ def from_dataset(
...
+            if str(filename).endswith(".shp"):
+                source_grid_spec = "Shapefile"
+            elif str(filename).endswith(".geojson"):
+                source_grid_spec = "GeoJSON"
+            else:
+                source_grid_spec = "OtherGeoFormat"
+
+            grid_ds, source_dims_dict = _read_geodataframe(filename)
+
+        elif backend == "netcdf":
```
philipc2, line 275: Backend should be `xarray`, not `netcdf`

uxarray/grid/grid.py
```diff
@@ -232,6 +234,52 @@ def from_dataset(
...
+        filename: str,
+        backend: Optional[str] = "geopandas",
+        **kwargs,
+    ):
+        """Constructs a ``Grid`` object from a using geopandas read_file.
+
+        Parameters
+        ----------
+        filename : str
+            Path to shapefile or geojson filename
```
philipc2, line 249:
```suggestion
            Path to grid file

**Review by philipc2 — CHANGES_REQUESTED** [review]
A few comments. Please update the PR title & linked issue as well.

**Comment by rajeeja** [comment]
> > > > We should consider a basic User Guide notebook for this too. I can provide the `quad-hexagon` grid in multiple different formats (i.e. shape, geoparquet, etc.) to showcase this functionality.
> > > 
> > > 
> > > I have some mesh files as a part of this, we want to be careful with how many we bundle as a part of the repo.
> > 
> > 
> > Maybe using online files rather than having them all hosted in the repo?
> 
> Do we have an issue open for this topic in general?

#946 would cover this also, not a explicit issue for this.

File: uxarray/core/api.py [hunk 12]
```diff
@@ -61,6 +61,8 @@ def open_grid(
     >>> uxgrid = ux.open_grid("grid_filename.g")
     """
 
+    # extension = os.path.splitext(grid_filename_or_obj)
```
philipc2, line 64:
```suggestion
```

philipc2 (approved):
Looks good! Other than the one comment, feel free to merge it once it's been resolved.

PR #811, opened 2024-06-11
Compute Grid Centerpoint using Welzl's algorithm

rajeeja:


uxarray/grid/coordinates.py
```diff
@@ -329,18 +329,18 @@ def haversine_distance(point1, point2):
...
     point2 : tuple
-        A tuple containing the latitude and longitude of the second point.
+        A tuple containing the lon and lat of the second point.
 
     Returns
     -------
     float
         The distance between the two points on the unit sphere.
     """
     R = 1.0  # Radius of the Earth assumed to be 1 (unit sphere)
```
hongyuchen1030, line 341: The `R` is never used here. And is not needed as well

uxarray/grid/coordinates.py
```diff
@@ -227,9 +228,86 @@ def _populate_face_centroids(grid, repopulate=False):
...
+
+        grid._ds["face_z_ctrpt"] = xr.DataArray(
+            ctrpt_z, dims=[ugrid.FACE_DIM], attrs=ugrid.FACE_Z_ATTRS
+        )
+
+
 def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_face):
     """Constructs the xyz centroid coordinate for each face using Cartesian
-    Averaging."""
+    Averaging.
```
hongyuchen1030, line 291: Can this function be vectorized without the for loop?

uxarray/grid/coordinates.py
```diff
@@ -244,6 +322,211 @@ def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fa
     return _normalize_xyz(centroid_x, centroid_y, centroid_z)
 
 
+def haversine_distance(point1, point2):
```
hongyuchen1030, line 325: please use the _angle_of_two_vector function to avoid the trigonometric functions, which might introduce great precision problem

uxarray/grid/coordinates.py
```diff
@@ -244,6 +322,211 @@ def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fa
...
+    dlon = lon2 - lon1
+
+    a = np.sin(dlat / 2) ** 2 + np.cos(lat1) * np.cos(lat2) * np.sin(dlon / 2) ** 2
+    c = 2 * np.arctan2(np.sqrt(a), np.sqrt(1 - a))
+
+    distance = R * c
+    return distance
+
+
+def circle_from_two_points(p1, p2):
```
hongyuchen1030, line 355: How about the wrap-around case where the great circle goes across the anti-meridian? Can you explain how you treat that case?

Also, this looks like a duplication of the haversine_distance, especially if we use the `_angle_of_two_vector ` function.

philipc2:
I'm not a big fan of the `_ctrpt` approach to the properties. These values are still, for example, `face_lon` and `face_lat`, they simply use a different algorithm to compute them.

Maybe we should consider the following design?
- Have the default `face_xyz` or `face_latlon` values be either what was parsed from a dataset OR the existing Cartesian averaging 
- Introduce a grid-level `Grid.populate_face_coordinates()` function (similar to the internal ones that we have to allow the user to re-populate or set the desired algorithm they'd like to use for the construction.

This would make the workflow look something like the following:
```Python
# get the value of face_lon without needing to specify an algorithm, will use either the stored value or cart avg
uxgrid.face_lon

# I want to explicitly set the algorithm to be Welzl
uxgrid.populate_face_coordinates(method='welzl')

# value will now be populated using your approach
uxgrid.face_lon

# I want to re-populate again using cartesiain averaging
uxgrid.populate_face_coordinates(method='cartesian average')

# value will now be populated using artesian average
uxgrid.face_lon
```


This allows us to not need to define any new properties and to better stick to the UGRID conventions. What do you think?

uxarray/grid/coordinates.py
rajeeja: removed that fun.

uxarray/grid/coordinates.py
rajeeja: done, thanks!!

rajeeja:
> I'm not a big fan of the `_ctrpt` approach to the properties. These values are still, for example, `face_lon` and `face_lat`, they simply use a different algorithm to compute them.
> 
> Maybe we should consider the following design?
> 
> * Have the default `face_xyz` or `face_latlon` values be either what was parsed from a dataset OR the existing Cartesian averaging
> * Introduce a grid-level `Grid.populate_face_coordinates()` function (similar to the internal ones that we have to allow the user to re-populate or set the desired algorithm they'd like to use for the construction.
> 
> This would make the workflow look something like the following:
> 
> ```python
> # get the value of face_lon without needing to specify an algorithm, will use either the stored value or cart avg
> uxgrid.face_lon
> 
> # I want to explicitly set the algorithm to be Welzl
> uxgrid.populate_face_coordinates(method='welzl')
> 
> # value will now be populated using your approach
> uxgrid.face_lon
> 
> # I want to re-populate again using cartesiain averaging
> uxgrid.populate_face_coordinates(method='cartesian average')
> 
> # value will now be populated using artesian average
> uxgrid.face_lon
> ```
> 
> This allows us to not need to define any new properties and to better stick to the UGRID conventions. What do you think?

During my testing and sometimes in testing the face geometry both centerpoint and centroid might be needed. When working with a mesh I wanted to check how much did one deviate from the other and if one or the other made more sense. 

We might be able to get both with the way you propose also, but with two calls to `populate` one with either options, having both available to the grid object at once might be better. 

We can get another name for `ctrpt`, I don't like it also:)

uxarray/grid/coordinates.py
```diff
@@ -227,9 +115,86 @@ def _populate_face_centroids(grid, repopulate=False):
...
+
+    # Check if the centerpoints are already populated
+    if "face_lon_ctrpt" not in grid._ds or repopulate:
+        # Construct the centerpoints if there are none stored
+        if "face_x_ctrpt" not in grid._ds:
+            centerpoint_lon, centerpoint_lat = _construct_face_centerpoints(
+                node_lon, node_lat, face_nodes, n_nodes_per_face
+            )
+    # get the cartesian coordinates of the centerpoints
+    ctrpt_x, ctrpt_y, ctrpt_z = _lonlat_rad_to_xyz(centerpoint_lon, centerpoint_lat)
```
hongyuchen1030, line 151: Can we just call the `_poplate_xyz` for the entire `Grid` first?

uxarray/grid/coordinates.py
hongyuchen1030: This should be able to be vectorized in a way similar to the following codes
```python
    # Mask to ignore invalid indices
    mask = np.arange(face_nodes.shape[1])[None, :] < n_nodes_per_face[:, None]

    # Calculate centroids
    centroid_x = np.sum(node_x[face_nodes] * mask, axis=1) / n_nodes_per_face
    centroid_y = np.sum(node_y[face_nodes] * mask, axis=1) / n_nodes_per_face
    centroid_z = np.sum(node_z[face_nodes] * mask, axis=1) / n_nodes_per_face

    return _normalize_xyz(centroid_x, centroid_y, centroid_z)
 ```

File: uxarray/grid/coordinates.py [hunk 6]
```diff
@@ -244,6 +209,186 @@ def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fa
...
+    -------
+    tuple
+        A tuple containing the center (as a tuple of lon and lat) and the radius of the circle.
+    """
+    center_lon = (p1[0] + p2[0]) / 2
+    center_lat = (p1[1] + p2[1]) / 2
+    center = (center_lon, center_lat)
+    v1, v2 = (np.array(_lonlat_rad_to_xyz(*np.radians(p))) for p in (p1, p2))
+    distance = _angle_of_2_vectors(v1, v2)
+    radius = distance / 2
```
hongyuchen1030, line 232: Is the radius the distance on the sphere, or the distance on a plane?

uxarray/grid/coordinates.py
rajeeja: > This should be able to be vectorized in a way similar to the following codes
> 
> ```python
>     # Mask to ignore invalid indices
>     mask = np.arange(face_nodes.shape[1])[None, :] < n_nodes_per_face[:, None]
> 
>     # Calculate centroids
>     centroid_x = np.sum(node_x[face_nodes] * mask, axis=1) / n_nodes_per_face
>     centroid_y = np.sum(node_y[face_nodes] * mask, axis=1) / n_nodes_per_face
>     centroid_z = np.sum(node_z[face_nodes] * mask, axis=1) / n_nodes_per_face
> 
>     return _normalize_xyz(centroid_x, centroid_y, centroid_z)
> ```
I don't see the code snipped you provided in our repo, where is it from?
This PR is not about face_centroids, in anycase this change can be made if it is not already a part of some other PR.
For this PR, `_construct_face_centerpoints` is hard to vectorize given the call to `smallest_enclosing_circle` because it operates on individual points and not entire set. We can handle it in another PR as it might involve using the `concurrent` package. Another solution is doing this smallest_encolsing_circle for the entire grid at once and it surely complicates the algorithm.

uxarray/grid/coordinates.py
rajeeja: > Is the radius the distance on the sphere, or the distance on a plane?

welzl is typically for a set of points doesn't have to be on a sphere, the algorithm doesn't force anything on the points it works with. Here, I'm making it work as if on a sphere and calculating the distances on sphere instead of a plane, so we just need Haversine distance, I check that `_angle_of_2_vectors` is same as Haversine distance.

hongyuchen1030: When comes to the spherical surface, you can use the modified version called [Bounding Sphere](https://en.wikipedia.org/wiki/Bounding_sphere), it's the 3-D version of the welzl's algorithmn

rajeeja: > When comes to the spherical surface, you can use the modified version called [Bounding Sphere](https://en.wikipedia.org/wiki/Bounding_sphere), it's the 3-D version of the welzl's algorithmn

Hey, I was thinking about this and was half way through my implementation - made a rookie mistake of not thinking through before implentation --

 the current algorithm is correct - it is modified for 2d spherical geometry and the distance is computed using the great circle distance. We don't need 3D sphere for this.

philipc2:
> > I'm not a big fan of the `_ctrpt` approach to the properties. These values are still, for example, `face_lon` and `face_lat`, they simply use a different algorithm to compute them.
> > Maybe we should consider the following design?
> > 
> > * Have the default `face_xyz` or `face_latlon` values be either what was parsed from a dataset OR the existing Cartesian averaging
> > * Introduce a grid-level `Grid.populate_face_coordinates()` function (similar to the internal ones that we have to allow the user to re-populate or set the desired algorithm they'd like to use for the construction.
> > 
> > This would make the workflow look something like the following:
> > ```python
> > # get the value of face_lon without needing to specify an algorithm, will use either the stored value or cart avg
> > uxgrid.face_lon
> > 
> > # I want to explicitly set the algorithm to be Welzl
> > uxgrid.populate_face_coordinates(method='welzl')
> > 
> > # value will now be populated using your approach
> > uxgrid.face_lon
> > 
> > # I want to re-populate again using cartesiain averaging
> > uxgrid.populate_face_coordinates(method='cartesian average')
> > 
> > # value will now be populated using artesian average
> > uxgrid.face_lon
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This allows us to not need to define any new properties and to better stick to the UGRID conventions. What do you think?
> 
> During my testing and sometimes in testing the face geometry both centerpoint and centroid might be needed. When working with a mesh I wanted to check how much did one deviate from the other and if one or the other made more sense.
> 
> We might be able to get both with the way you propose also, but with two calls to `populate` one with either options, having both available to the grid object at once might be better.
> 
> We can get another name for `ctrpt`, I don't like it also:)


My main concern with breaking up the different types of coordinates in separate attributes is that it'll add extra overhead for us to ensure that the coordinates we read match the ones that we want to store, not to mention needing to redefine / extent the UGRID conventions further past what we've already done. Even with this (and say some other method down the line), this could end up looking like:
- `Grid.face_lon`
- `Grid.face_lon_centerpoint`
- `Grid.face_lon_some_other_defenition`


Consider the case where two UGRID (or any other format) grid files are loaded into UXarray. If we move forward with a split attribute approach, we'd need to ensure that the coordinates we are reading either go into `face_lat/lon` or `face_lat/lon_centerpoints`. There's also no easy way to determine what method each dataset used to compute the centroids at the loading step without parsing for any specific attributes in the file (if they exist), since this is not outlined in the UGRID conventions. 

I'm still in favor of keeping `face_lon` and `face_lat` and general variables for storing some coordinate that represents the center/midpoint/centroid etc. of the face. This does limit us to only storing one type of "center" coordinate at a time, but ensures that we don't restrict us to strictly defining the type of definition for the center. 

@paullric @rljacob Is there ever a sceneiro where we would want to have more than one definition of a "center" coordinate attached to a grid at a time? 

Alternatively,

uxarray/grid/coordinates.py
```diff
@@ -227,9 +115,86 @@ def _populate_face_centroids(grid, repopulate=False):
         )
 
 
+def _populate_face_centerpoints(grid, repopulate=False):
```
philipc2, line 118: Can we add a benchmark to our `mpas_ocean` suite that benchmarks the execution time for this function across the 480km and 120km resolutions? 

Something like what @aaronzedwick put together in #870

rajeeja: It is slow, looking to see optimization possibilities.

rljacob:
Why do we need 2 ways to calculate the center of a face?   I see no reason to use them both at once.

rajeeja:
> Why do we need 2 ways to calculate the center of a face? I see no reason to use them both at once.

originally I wanted to compare both and hence keep both, but it makes more sense to just have one attribute and keep the `grid` object clean. I committed the change to have only one.

hongyuchen1030:
> Member



> Why do we need 2 ways to calculate the center of a face? I see no reason to use them both at once.

I think the key idea of Welzl's algorithm is not just to find the center point but to compute a bounding circle. This has several advantages:

1. **Storage Efficiency**: Only two variables are needed—the center point and the radius—reducing storage requirements.
2. **Geometric Integrity**: A bounding circle inherently accounts for spherical geometry, including curvature and pole points, as well as the antimeridian.

However, there are some disadvantages compared to using a latitude-longitude box:

- **Loose Fit for Slim Faces**: A bounding circle can result in a larger radius, which is not as tight when bounding "slim" faces.

Therefore, providing both the bounding circle and the lat-lon box offers users a more comprehensive solution.

rajeeja:
> > Member
> 
> > Why do we need 2 ways to calculate the center of a face? I see no reason to use them both at once.
> 
> I think the key idea of Welzl's algorithm is not just to find the center point but to compute a bounding circle. This has several advantages:
> 
> 1. **Storage Efficiency**: Only two variables are needed—the center point and the radius—reducing storage requirements.
> 2. **Geometric Integrity**: A bounding circle inherently accounts for spherical geometry, including curvature and pole points, as well as the antimeridian.
> 
> However, there are some disadvantages compared to using a latitude-longitude box:
> 
> * **Loose Fit for Slim Faces**: A bounding circle can result in a larger radius, which is not as tight when bounding "slim" faces.
> 
> Therefore, providing both the bounding circle and the lat-lon box offers users a more comprehensive solution.

Nice summary, Welzl being iterative is computationally a little expensive - but has a clear place as you suggest above.
Just that now:  they would both be stored in the same variable in the `Grid` object.  External users can get centerpoint or centroid by specifying something like

`
uxgrid.compute_face_center(method = "welzl"). # or average
`

rajeeja:
@philipc2 any ideas about asv failure:

https://github.com/UXARRAY/uxarray/actions/runs/10568661430/job/29280049250?pr=811#step:11:5

philipc2:
> @philipc2 any ideas about asv failure:
> 
> https://github.com/UXARRAY/uxarray/actions/runs/10568661430/job/29280049250?pr=811#step:11:5

Appears to be an issue with this part of the workflow. Looking into it now. 

https://github.com/UXARRAY/uxarray/blob/38ae0017e1b08cbaf6b9c32a7b4ee9c2d2dc1401/.github/workflows/asv-benchmarking-pr.yml#L23-L33

uxarray/grid/coordinates.py
rajeeja: > It is slow, looking to see optimization possibilities.

was able to fix Numba causes issues with recursive functions, this is much more efficient ~5-10x faster.

benchmarks/mpas_ocean.py
```diff
@@ -164,3 +164,29 @@ def time_nearest_neighbor_remapping(self):
 
     def time_inverse_distance_weighted_remapping(self):
         self.uxds_480["bottomDepth"].remap.inverse_distance_weighted(self.uxds_120.uxgrid)
+
+from uxarray.grid.coordinates import _construct_face_centerpoints, _construct_face_centroids
```
philipc2, line 168: I believe that this import statement is causing the `mpas_ocean` benchmarks to crash for previous runs.


This may also explain the issues you were having locally. I'd put this line inside of the `ConstructFaceLatLon` and it should fix it

image

philipc2: Did you mean to resolve this?

rajeeja: > Did you mean to resolve this?

benchmarks seems to run, but profiler doesn't find it still:
```

asv run --bench ConstructFaceLatLon.time_welzl --python=same

Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Discovering benchmarks
· Running 2 total benchmarks (1 commits * 1 environments * 2 benchmarks)
[ 0.00%] ·· Benchmarking existing-py_opt_homebrew_anaconda3_envs_ux_bin_python
[25.00%] ··· Running (mpas_ocean.ConstructFaceLatLon.time_welzl--)..
[75.00%] ··· mpas_ocean.ConstructFaceLatLon.time_welzl                                          ok
[75.00%] ··· ============ ===========
              resolution             
             ------------ -----------
                480km       156±6ms  
                120km      2.54±0.1s 
             ============ ===========

[100.00%] ··· quad_hexagon.ConstructFaceLatLon.time_welzl                                   490±3ms
```

```
asv profile mpas_ocean.ConstructFaceLatLon.time_welzl --gui=snakeviz  --python=same
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Discovering benchmarks
· 'mpas_ocean.ConstructFaceLatLon.time_welzl' benchmark not found
```

philipc2, line 168: This line needs to be moved either within the `class ConstructFaceLatLon` or inside of the `time_welzl`  method for the benchmarks to run correctly.

rajeeja: > This line needs to be moved either within the `class ConstructFaceLatLon` or inside of the `time_welzl` method for the benchmarks to run correctly.

All benchmarks run to completion:
`asv run --bench mpas_ocean.ConstructTreeStructures.time_kd_tree     --python=same`, 

the problem is profiling, eg:

`asv profile mpas_ocean.ConstructTreeStructures.time_kd_tree` 
doesn't work, something to do with passing arguments `resolution etc.` keeps getting error  
`mpas_ocean.ConstructTreeStructures.time_kd_tree benchmark not found`

philipc2: > ## ASV Benchmarking
> Benchmark Comparison Results
> Benchmarks that have improved:
> 
> Change	Before [[82e46b8](https://github.com/UXARRAY/uxarray/commit/82e46b8c79bb0107753398ddab096c5f6eb9b718)]	After [[9f952f6](https://github.com/UXARRAY/uxarray/commit/9f952f6fff44f428c382d0c51bc64d6ebd9cf536)]	Ratio	Benchmark (Parameter)
> failed	1.51±0.01s	n/a	mpas_ocean.ConstructFaceLatLon.time_welzl('120km')
> failed	95.5±0.1ms	n/a	mpas_ocean.ConstructFaceLatLon.time_welzl('480km')
> -	408M	354M	0.87	mpas_ocean.Integrate.peakmem_integrate('480km')
> Benchmarks that have stayed the same:
> 
> Change	Before [[82e46b8](https://github.com/UXARRAY/uxarray/commit/82e46b8c79bb0107753398ddab096c5f6eb9b718)]	After [[9f952f6](https://github.com/UXARRAY/uxarray/commit/9f952f6fff44f428c382d0c51bc64d6ebd9cf536)]	Ratio	Benchmark (Parameter)
> 379M	380M	1.00	face_bounds.FaceBounds.peakmem_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/mpas/QU/oQU480.231010.nc'))
> 380M	381M	1.00	face_bounds.FaceBounds.peakmem_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/scrip/outCSne8/outCSne8.nc'))
> 383M	384M	1.00	face_bounds.FaceBounds.peakmem_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/ugrid/geoflow-small/grid.nc'))
> 381M	382M	1.00	face_bounds.FaceBounds.peakmem_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/ugrid/quad-hexagon/grid.nc'))
> 1.36±0.02s	1.35±0s	1.00	face_bounds.FaceBounds.time_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/mpas/QU/oQU480.231010.nc'))
> 176±3ms	177±1ms	1.01	face_bounds.FaceBounds.time_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/scrip/outCSne8/outCSne8.nc'))
> 1.64±0.01s	1.65±0.02s	1.00	face_bounds.FaceBounds.time_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/ugrid/geoflow-small/grid.nc'))
> 7.63±0.08ms	7.83±0.2ms	1.03	face_bounds.FaceBounds.time_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/ugrid/quad-hexagon/grid.nc'))
> 1.68±0.01s	1.70±0.01s	1.01	import.Imports.timeraw_import_uxarray
> 621±4ms	623±4ms	1.00	mpas_ocean.ConnectivityConstruction.time_face_face_connectivity('120km')
> 39.3±0.3ms	39.6±0.4ms	1.01	mpas_ocean.ConnectivityConstruction.time_face_face_connectivity('480km')
> 1.80±0.02ms	1.98±0.2ms	~1.10	mpas_ocean.ConnectivityConstruction.time_n_nodes_per_face('120km')
> 474±7μs	476±7μs	1.00	mpas_ocean.ConnectivityConstruction.time_n_nodes_per_face('480km')
> 361±7ms	356±3ms	0.99	mpas_ocean.ConstructFaceLatLon.time_cartesian_averaging('120km')
> 23.4±0.3ms	22.9±0.2ms	0.98	mpas_ocean.ConstructFaceLatLon.time_cartesian_averaging('480km')
> 1.19±0μs	1.23±0.01μs	1.04	mpas_ocean.ConstructTreeStructures.time_ball_tree('120km')
> 305±5ns	303±4ns	1.00	mpas_ocean.ConstructTreeStructures.time_ball_tree('480km')
> 794±5ns	873±1ns	1.10	mpas_ocean.ConstructTreeStructures.time_kd_tree('120km')
> 280±2ns	291±4ns	1.04	mpas_ocean.ConstructTreeStructures.time_kd_tree('480km')
> 402M	399M	0.99	mpas_ocean.GeoDataFrame.peakmem_to_geodataframe('120km', False)
> 390M	387M	0.99	mpas_ocean.GeoDataFrame.peakmem_to_geodataframe('120km', True)
> 362M	362M	1.00	mpas_ocean.GeoDataFrame.peakmem_to_geodataframe('480km', False)
> 360M	361M	1.00	mpas_ocean.GeoDataFrame.peakmem_to_geodataframe('480km', True)
> 1.06±0s	1.04±0.01s	0.98	mpas_ocean.GeoDataFrame.time_to_geodataframe('120km', False)
> 57.4±0.5ms	56.9±0.3ms	0.99	mpas_ocean.GeoDataFrame.time_to_geodataframe('120km', True)
> 77.7±0.8ms	77.3±1ms	1.00	mpas_ocean.GeoDataFrame.time_to_geodataframe('480km', False)
> 5.53±0.2ms	5.22±0.09ms	0.94	mpas_ocean.GeoDataFrame.time_to_geodataframe('480km', True)
> 263M	263M	1.00	mpas_ocean.Gradient.peakmem_gradient('120km')
> 240M	240M	1.00	mpas_ocean.Gradient.peakmem_gradient('480km')
> 2.66±0.01ms	2.74±0.05ms	1.03	mpas_ocean.Gradient.time_gradient('120km')
> 280±0.6μs	284±6μs	1.01	mpas_ocean.Gradient.time_gradient('480km')
> 370M	370M	1.00	mpas_ocean.Integrate.peakmem_integrate('120km')
> 175±1ms	176±0.6ms	1.00	mpas_ocean.Integrate.time_integrate('120km')
> 11.7±0.03ms	12.2±0.04ms	1.04	mpas_ocean.Integrate.time_integrate('480km')
> 343±2ms	343±2ms	1.00	mpas_ocean.MatplotlibConversion.time_dataarray_to_polycollection('120km', 'exclude')
> 346±2ms	343±2ms	0.99	mpas_ocean.MatplotlibConversion.time_dataarray_to_polycollection('120km', 'include')
> 346±3ms	342±2ms	0.99	mpas_ocean.MatplotlibConversion.time_dataarray_to_polycollection('120km', 'split')
> 21.6±0.3ms	22.3±0.6ms	1.03	mpas_ocean.MatplotlibConversion.time_dataarray_to_polycollection('480km', 'exclude')
> 21.6±0.1ms	21.5±0.2ms	1.00	mpas_ocean.MatplotlibConversion.time_dataarray_to_polycollection('480km', 'include')
> 21.4±0.1ms	21.6±0.1ms	1.01	mpas_ocean.MatplotlibConversion.time_dataarray_to_polycollection('480km', 'split')
> 54.7±0.2ms	54.8±0.2ms	1.00	mpas_ocean.RemapDownsample.time_inverse_distance_weighted_remapping
> 44.2±0.03ms	44.4±0.2ms	1.00	mpas_ocean.RemapDownsample.time_nearest_neighbor_remapping
> 361±0.8ms	361±1ms	1.00	mpas_ocean.RemapUpsample.time_inverse_distance_weighted_remapping
> 263±0.3ms	263±1ms	1.00	mpas_ocean.RemapUpsample.time_nearest_neighbor_remapping
> failed	failed	n/a	quad_hexagon.ConstructFaceLatLon.time_welzl
> 240M	237M	0.99	quad_hexagon.QuadHexagon.peakmem_open_dataset
> 236M	236M	1.00	quad_hexagon.QuadHexagon.peakmem_open_grid
> 6.35±0.1ms	6.56±0.2ms	1.03	quad_hexagon.QuadHexagon.time_open_dataset
> 5.43±0.07ms	5.50±0.02ms	1.01	quad_hexagon.QuadHexagon.time_open_grid
> 236M	236M	1.00	quad_hexagon.QuadHexagon2.peakmem_open_grid2
> 6.29±0.03ms	6.51±0.1ms	1.03	quad_hexagon.QuadHexagon2.time_open_dataset
> 5.39±0.04ms	5.58±0.02ms	1.03	quad_hexagon.QuadHexagon2.time_open_grid

@rajeeja 

The results look correct now (compared to what we were getting before). Does running `asv profile` still occur for you locally?

rajeeja: Yes, bench works. profile doesn't - any function with arguments in mpas_ocean file

uxarray/grid/grid.py
```diff
@@ -347,6 +348,28 @@ def validate(self):
...
+            other supported method is Welzl's algorithm, takes value "welzl"
+
+
+        Usage
+        -----
+        >>> import uxarray as ux
+        >>> uxgrid = ux.open_grid("GRID_FILE_NAME")
+        >>> face_lat = uxgrid.construct_face_center(method="welzl")
+        """
+        if method == "average":
```
philipc2, line 366: Should we be more specific and call it "cartesian average" ?

uxarray/grid/grid.py
```diff
@@ -347,6 +348,28 @@ def validate(self):
...
+            _populate_face_centroids(self, repopulate=True)
+        elif method == "welzl":
+            _populate_face_centerpoints(self, repopulate=True)
+        else:
+            raise ValueError("unknown method for face center calculation")
```
philipc2, line 371: Can we change the error to the following? 
```
f"Unknown method for face center calculation. Expected one of ["cartesian average", "welzl"] but received {method}"
```

uxarray/grid/utils.py
```diff
@@ -423,3 +428,168 @@ def _get_lonlat_rad_face_edge_nodes(
     face_edges_lonlat_rad[valid_mask, 1] = node_lat_rad[valid_edges]
 
     return face_edges_lonlat_rad.reshape(n_face, n_max_face_edges, 2, 2)
+
+
+def _xyz_to_lonlat_rad(
```
philipc2, line 433: Wondering why these were moved from the `coordinates` module?

uxarray/grid/grid.py
```diff
@@ -347,6 +348,28 @@ def validate(self):
         else:
             raise RuntimeError("Mesh validation failed.")
 
+    def compute_face_center(self, method="average"):
```
philipc2, line 351: This may be more appropriate: 

```suggestion
    def construct_face_centers(self, method="average"):
```
or the following

```suggestion
    def construct_face_latlon(self, method="average"):
```

Any thoughts?

uxarray/grid/grid.py
```diff
@@ -347,6 +348,28 @@ def validate(self):
...
+        """Constructs a face_lon and face_lat.
+
+        Parameters
+        ----------
+        method : str, default="average"
+            other supported method is Welzl's algorithm, takes value "welzl"
```
philipc2, line 357: Include the supported methods in the parameter description.

uxarray/grid/grid.py
```diff
@@ -347,6 +348,28 @@ def validate(self):
         else:
             raise RuntimeError("Mesh validation failed.")
 
+    def compute_face_center(self, method="average"):
+        """Constructs a face_lon and face_lat.
```
philipc2, line 352: Can we make this a bit more descriptive? Should mention that this method provides the user direct control to select what type of method is used for constructing the face centers, as opposed to the default option chosen by the grid properties.

Can also mention this can be used to override / re-compute stored/parsed values.

philipc2 (changes requested):

uxarray/grid/utils.py
rajeeja: cyclical dependencies issue

uxarray/grid/grid.py
rajeeja: yes, cartesian is better

uxarray/grid/coordinates.py
```diff
@@ -296,6 +150,266 @@ def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fa
     return _normalize_xyz(centroid_x, centroid_y, centroid_z)
 
 
+def _populate_face_centerpoints(grid, repopulate=False):
```
erogluorhan, line 153: I think we used the term `cache` instead of `repopulate` for similar cases ; should we use the same here? @philipc2 can have a better view on this

uxarray/grid/coordinates.py
```diff
@@ -296,6 +150,266 @@ def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fa
...
+    circle that encloses a set of points. It is here adapted to work on a unit
+    sphere. Also, this algorithm cannot be guaranteed to work on concave
+    polygons.
+
+    Parameters
+    ----------
+    grid : Grid
+        The grid containing the nodes and faces.
+    repopulate : bool, optional
+        Bool used to turn on/off repopulating the face coordinates of the centerpoints, default is False.
```
erogluorhan, line 165: Please also provide information about the returned objects in the docstrings

erogluorhan (commented):
The code looks good to me. Please address just a few simple comments below

uxarray/grid/coordinates.py
philipc2: `repopulate` is the correct term in the context of this function (as we've done with the other internal construction methods), since these check if there already exist some coordinate, with `repopulate=True` indicated to ignore any cached values.

This method will always cache the computed result.

uxarray/grid/grid.py
```diff
@@ -348,6 +349,35 @@ def validate(self):
         else:
             raise RuntimeError("Mesh validation failed.")
 
+    def construct_face_centers(self, method="cartesian average"):
+        """Constructs face centers, this method provides users direct control
+        of the method for constructing the face centers, the default method is
+        "cartesian average", but a more efficient method is "welzl" that is
+        based on the recursive Welzl algorithm. It must be noted that this
+        method can override the parsed/recompute the original parsed face
+        centers.
```
philipc2, line 358: Should mention that it constructs `face_lon` and `face_lat`

philipc2 (changes requested):

erogluorhan (commented):
The code looks good to me. The only thing remaining was the reduced code coverage, but apparently codecov can't track test cases written for the `njit`-decorated functions. After we figure a path forward with that, I am happy to approve this.

uxarray/grid/grid.py
```diff
@@ -348,6 +349,39 @@ def validate(self):
         else:
             raise RuntimeError("Mesh validation failed.")
 
+    def construct_face_centers(self, method="cartesian average"):
+        """Constructs face centers, this method provides users direct control
+        of the method for constructing the face centers, the default method is
+        "cartesian average", but a more efficient method is "welzl" that is
```
philipc2, line 355: Is "efficient" the right word here?

rajeeja: > Is "efficient" the right word here?

accurate is right.

uxarray/grid/utils.py
```diff
@@ -423,3 +428,168 @@ def _get_lonlat_rad_face_edge_nodes(
...
+    return lon, lat
+
+
+@njit
+def _normalize_xyz_scalar(x: float, y: float, z: float):
+    denom = np.linalg.norm(np.asarray(np.array([x, y, z]), dtype=np.float64), ord=2)
+    x_norm = x / denom
+    y_norm = y / denom
+    z_norm = z / denom
+    return x_norm, y_norm, z_norm
```
philipc2, line 595: I'm still a bit hesitant in moving these. Can you try moving them back into `coordinates.py` and provide a bit more info on the cyclical dependencies issue?

rajeeja: all tests should fail now, please check if there is a better fix. I thought those better belonged to grid/utils.py as they can run as a utility and are not tied to other infra.

```
#     _xyz_to_lonlat_rad,
#     _lonlat_rad_to_xyz,
#     _xyz_to_lonlat_deg,
#     _normalize_xyz,
```

rajeeja:
Once, we resolve this circular dependency issue. I will disable NUMBA to check codecov - it might increase the percentage coverage [issue](https://community.codecov.com/t/numba-jitted-methods-are-not-captured-by-codecov/2649)

erogluorhan:
> Once, we resolve this circular dependency issue. I will disable NUMBA to check codecov - it might increase the percentage coverage [issue](https://community.codecov.com/t/numba-jitted-methods-are-not-captured-by-codecov/2649)

How about doing this in such a way: In the beginning of a case that tests a njit-decorated function, disable numba, in the end, enable it back? This helps us notice right away that whenever we see this kind of disable-enable pairs, that is testing a njit-decorated function.

rajeeja:
> > Once, we resolve this circular dependency issue. I will disable NUMBA to check codecov - it might increase the percentage coverage [issue](https://community.codecov.com/t/numba-jitted-methods-are-not-captured-by-codecov/2649)
> 
> How about doing this in such a way: In the beginning of each indidvual case that tests a njit-decorated function, disable numba, in the end, enable it back? This helps us notice right away that whenever we see this kind of disable-enable pairs, that is a case for testing a njit-decorated function.

I removed all the numba stuff on my local and the coverage (85% total and 95% for coordinates.py - see pic below) is considerably higher for sha: `dba53dc8` which is before I introduced circular dependency. 

Screenshot 2024-09-17 at 17 37 01

uxarray/grid/utils.py
rajeeja: > I'm still a bit hesitant in moving these. Can you try moving them back into `coordinates.py` and provide a bit more info on the cyclical dependencies issue?

coordinates.py needs "from uxarray.grid.arcs import _angle_of_2_vectors"
arcs.py needs stuff from coordinates.py - 

from uxarray.grid.coordinates import (
    _xyz_to_lonlat_rad_scalar,
    _normalize_xyz_scalar,
)

benchmarks/quad_hexagon.py
```diff
@@ -34,3 +34,39 @@ def time_open_dataset(self):
...
+    #     return ux.open_grid(grid_path)
+
+    def peakmem_open_grid2(self):
+        """Peak memory usage of a `Grid`"""
+        uxgrid = ux.open_grid(grid_path)
+
+
+    def time_open_dataset(self):
+        """Time to open a `UxDataset`"""
+        ux.open_dataset(grid_path, data_path)
```
philipc2, line 72: These shouldn't be necessary considering the benchmark in `mpas_ocean`

benchmarks/mpas_ocean.py
```diff
@@ -144,3 +144,34 @@ def time_inverse_distance_weighted_remapping(self):
 class HoleEdgeIndices(DatasetBenchmark):
     def time_construct_hole_edge_indices(self, resolution):
         ux.grid.geometry._construct_hole_edge_indices(self.uxds.uxgrid.edge_face_connectivity)
+
+
+class ConstructFaceLatLon:
```
philipc2, line 149:
```suggestion
class ConstructFaceLatLon(GridBenchmark):
```

benchmarks/mpas_ocean.py
```diff
@@ -144,3 +144,34 @@ def time_inverse_distance_weighted_remapping(self):
...
+
+    param_names = ['resolution']
+    params = ['480km', '120km']
+
+    def setup(self, resolution):
+        self.uxgrid = ux.open_grid(file_path_dict[resolution][0])
+
+    def teardown(self, resolution):
+        del self.uxgrid
```
philipc2, line 158:
```suggestion

```

uxarray/grid/coordinates.py
```diff
@@ -1,15 +1,14 @@
...
 import numpy as np
-
+import math
 import warnings
 
-from uxarray.constants import ERROR_TOLERANCE
 from uxarray.conventions import ugrid
-
-from typing import Union
+from uxarray.grid.arcs import _angle_of_2_vectors
```
philipc2, line 7: This appears to be the root cause of the circular import.

To avoid this, can you try using the following in this module when calling this function:

`uxarray.grid.arcs._angle_of_2_vectors`

uxarray/grid/utils.py
```diff
@@ -4,70 +4,6 @@
...
-            )
-    else:
-        raise ValueError(
-            f"Data type {grid_var.dtype} not supported" f"for grid variables"
-        )
-
-    # replace all zeros with a fill value
-    grid_var[fill_val_idx] = new_fill
-
-    return grid_var
```
philipc2, line 68: This shouldn't be moved either. Please refer to the above comment if the circular import issues still persist.

philipc2 (changes requested):

uxarray/grid/coordinates.py
rajeeja: > This appears to be the root cause of the circular import.
> 
> To avoid this, can you try using the following in this module when calling this function:
> 
> `uxarray.grid.arcs._angle_of_2_vectors`

Doesn't work. Numba doesn't like uxarray...

uxarray/grid/utils.py
rajeeja: I checked, it is not used, what is used is in connectivity

philipc2 (approved):
Great work!

PR #819, opened 2024-06-20
Remapping Overwrites Original Data

### LINKED ISSUE #809: Remapping Overwrites Original Data in Dataset
### Version

v2024.05.0

Conda

### What happened?

When using IDW remapping, it rewrote the original dataset variable with the new remapped one. I assume this also occurs with NN remapping.

I was expecting it to return to the provided variable a dataset, and leave my original data untouched.

_No response_

aaronzedwick:
Closes #809 

## Overview

Fixes the issue of IDW and NN remapping overwriting the destination datasets, so that now the destination dataset/grid/dataarray is left alone, and the assigned variable used when calling the function is the only thing written to.
## Expected Usage

```Python
import uxarray as ux

source = ux.open_dataset(grid_path, data_path)
# Destination stays unchanged
destination = ux.open_dataset(grid_path, data_path)

# Remapping results stored in remap_to
remap_to = source['variable'].remap.inverse_distance_weighted(destination_obj=destination)

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**


rajeeja (approved):

rajeeja:
Just to clarify:
`remap_to = source['variable'].remap.inverse_distance_weighted(destination_obj=destination)
`

Assumes that destitionation doesn't have `variable` and if it does?

aaronzedwick:
> Just to clarify: `remap_to = source['variable'].remap.inverse_distance_weighted(destination_obj=destination) `
> 
> Assumes that destitionation doesn't have `variable` and if it does?

It shouldn’t matter if it does or doesn’t. Destination is just what structure we are remapping the data variable to. The ‘remap_to’ holds the results of the remapping. That’s the point of this fix, to make it so if destination has the variable, it isn’t replaced by the remap and is left untouched.

rajeeja:
> > Just to clarify: `remap_to = source['variable'].remap.inverse_distance_weighted(destination_obj=destination) `
> > Assumes that destitionation doesn't have `variable` and if it does?
> 
> It shouldn’t matter if it does or doesn’t. Destination is just what structure we are remapping the data variable to. The ‘remap_to’ holds the results of the remapping. That’s the point of this fix, to make it so if destination has the variable, it isn’t replaced by the remap and is left untouched.

ok, so for the case: source['temp'] and destination['temp'] are already present and if source['temp'] is mapped to destination, what happens? where is the source['temp'] written to in the destination?

aaronzedwick:
Nowhere in destination. It is written to the provided variable to the left of the equal sign, in this case ‘remap_to’.

rajeeja:
> Nowhere in destination. It is written to the provided variable to the left of the equal sign, in this case ‘remap_to’.

ok, remap_to doesn't have a grid associated? or does it have a copy of the destination grid ?

aaronzedwick:
> > Nowhere in destination. It is written to the provided variable to the left of the equal sign, in this case ‘remap_to’.
> 
> ok, remap_to doesn't have a grid associated? or does it have a copy of the destination grid ?

It does, it has a copy.

rajeeja:
> > > Nowhere in destination. It is written to the provided variable to the left of the equal sign, in this case ‘remap_to’.
> > 
> > ok, remap_to doesn't have a grid associated? or does it have a copy of the destination grid ?
> 
> It does, it has a copy. 

OK, what is the call if I don't want to create a new grid and use destination mesh object for remapping?

aaronzedwick:
> > > > Nowhere in destination. It is written to the provided variable to the left of the equal sign, in this case ‘remap_to’.
> > > 
> > > 
> > > ok, remap_to doesn't have a grid associated? or does it have a copy of the destination grid ?
> > 
> > 
> > It does, it has a copy.
> 
> OK, what is the call if I don't want to create a new grid and use destination mesh object for remapping?

You can’t. Philip told me we shouldn’t allow that. Although maybe I misunderstood what he was communicating. It wouldn’t be hard to add an overwrite condition though if we wanted.

philipc2:
After giving this some thought, I think the way we use the `destination_obj` is very awkward, unclear, and breaks out fluent interface that we have with `UxDataArray` and `UxDataset`

I think we should require a `destination_grid` instead, and always return a new `UxDataArray`. Remapping is the process of taking data from one grid and mapping it to some destination grid. Having the "destination" be a `UxDataArray` or `UxDataset` is not clear.

Something like the following would be great. `destination_obj` as an optional parameter, starting the deprecation process in this PR.

```Python

# source dataset
source_uxds= ux.open_dataset(grid_path, data_path)

# destination gird (the example you had opens a dataset, which isn't the most common workflow)
destination_grid = ux.open_grid(grid_path)

# remapping will always return a new `UxDataArray` 
t2m_remapped = uxds['t2m'].remap.nearest_neighbor(destination_grid=destination_grid)

# we dont support remapping an entire dataset, but it would look something like this
uxds_remapped = uxds.remap.nearest_neighbor(destination_grid=destination_grid)
```

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -231,12 +231,13 @@ def _inverse_distance_weighted_remap_uxda(
     )
     # add remapped variable to existing UxDataset
     if isinstance(destination_obj, uxarray.core.dataset.UxDataset):
-        destination_obj[source_uxda.name] = uxda_remap
-        return destination_obj
+        uxds = destination_obj.copy()
```
philipc2, line 234: See my comment below, but we do not want to do this.

philipc2: > I do remember you mentioning only allowing a grid, I figured I would fix that in a different PR and get this patched before the notebook is pushed. But I can tackle it here,

Would be nice to tackle it here. Let me know if you need any help next week.

> Is there ever a case where I want to take and add a second variable to a data array using a remap?

The nice thing about a Data Array is that it only stores a single data variable. A Dataset is used for one or more data variables. 

This does bring up the point of being able to do a "Dataset-wide" remap, i.e. apply the remapping on every variable in a `UxDataset`

> How would I be able to take and say add two remapped variables from two different datasets together in one data array?

We'd use the typically Xarray interface to merge one or more data arrays (or datasets) into a new dataset. 

https://docs.xarray.dev/en/stable/user-guide/combining.html#merge

aaronzedwick:
@philipc2 How does something like this look? Once we finish the deprecation period we can remove the redundant stuff of course, but I thought this was a good way to let everything function as it should for now.

philipc2:
> @philipc2 How does something like this look? Once we finish the deprecation period we can remove the redundant stuff of course, but I thought this was a good way to let everything function as it should for now.

This is exactly what I was thinking, it looks great!

Giving it a review now.

philipc2 (approved):

PR #831, opened 2024-07-02
ASV Benchmarking Pull Request Workflow

philipc2:
# Overview

Introduces a Github Actions Workflow for running ASV Benchmarks on Pull Requests, getting compared to the `main` branch.

The workflow is triggered using the `run-benchmark` label. 

image

By default, the workflow is skipped unless it is queued using the label above. 

image

The workflow can be re-queued by removing and adding the label. 

image

Once the workflow is complete, the results are posted (or updated) by a Github Bot. 

New benchmarks, including those that test new functionality, can be run using this bot as well. They will fail for older versions that raise an error, but that is expected. 

Example screenshot of the output:

image

### DIFF
--- a/.github/workflows/asv-benchmarking-pr.yml
+++ b/.github/workflows/asv-benchmarking-pr.yml
@@ -1,0 +1,107 @@
+name: ASV Benchmarking (PR)
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
+  workflow_dispatch:
+
+jobs:
+  benchmark:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'run-benchmark') && github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
+    name: Linux
+    runs-on: ubuntu-latest
+    env:
+      ASV_DIR: "./benchmarks"
+      CONDA_ENV_FILE: ci/environment.yml
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+
+      - name: Set up Conda environment
+        uses: mamba-org/setup-micromamba@v1
+        with:
+          environment-file: ${{env.CONDA_ENV_FILE}}
+          cache-environment: true
+          environment-name: uxarray_build
+          cache-environment-key: "${{runner.os}}-${{runner.arch}}-py${{env.PYTHON_VERSION}}-${{env.TODAY}}-${{hashFiles(env.CONDA_ENV_FILE)}}-benchmark"
+          create-args: >-
+            asv
+            build
+            mamba
+
+      - name: Run Benchmarks
+        shell: bash -l {0}
+        id: benchmark
+        run: |
+          set -x
+          # ID this runner
+          asv machine --yes
+          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Contender: ${GITHUB_SHA} (${{ github.event.pull_request.head.label }})"
+          # Run benchmarks for current commit against base
+          ASV_OPTIONS="--split --show-stderr"
+          asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA}
+          # Save compare results
+          asv compare --split ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} > asv_compare_results.txt
+        working-directory: ${{ env.ASV_DIR }}
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: asv-benchmark-results-${{ runner.os }}
+          path: |
+            ${{ env.ASV_DIR }}/results
+            ${{ env.ASV_DIR }}/asv_compare_results.txt
+
+      - name: Post or update result comment
+        id: comment
+        uses: actions/github-script@v6
+        with:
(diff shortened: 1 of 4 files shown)


philipc2:
An example of this workflow in action can be seen in this dummy PR on my fork.

https://github.com/philipc2/uxarray/pull/2

ci/environment.yml
```diff
@@ -1,22 +1,27 @@
...
   - gmpy2
+  - h5netcdf
+  - hdf5
   - holoviews
   - matplotlib-base
   - matplotlib-inline
   - netcdf4
   - numba
-  - numpy
+  - numpy<2
```
rajeeja, line 18: 2.0 was just released, any particular reason this won't be compatible with the latest versions going forward.

benchmarks/quad_hexagon.py
```diff
@@ -27,9 +27,9 @@ def time_open_dataset(self):
         """Time to open a `UxDataset`"""
         ux.open_dataset(grid_path, data_path)
 
-    def mem_open_dataset(self):
-        """Memory occupied by a `UxDataset`"""
-        return ux.open_dataset(grid_path, data_path)
+    # def mem_open_dataset(self):
```
rajeeja, line 30: commented for a reason?
same for others.

benchmarks/quad_hexagon.py
```diff
@@ -14,9 +14,9 @@ def time_open_grid(self):
         """Time to open a `Grid`"""
         ux.open_grid(grid_path)
```
rajeeja, line 15: how does this measure time etc.?

.github/workflows/asv-benchmarking-pr.yml
```diff
@@ -0,0 +1,107 @@
...
+          # Save compare results
+          asv compare --split ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} > asv_compare_results.txt
+        working-directory: ${{ env.ASV_DIR }}
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: asv-benchmark-results-${{ runner.os }}
+          path: |
+            ${{ env.ASV_DIR }}/results
```
rajeeja, line 56: A few nice to have requests: 

1. Table broken down into various segments - mem, open time etc. 
2. Maintain another repo for keeping a record a results. Something where we can just based on commit sha track what changes led to the changes in performance.
3. A way to get the performance difference between two dates from the point number 2 mentioned above.

rajeeja (changes requested):
This is excellent and very timely for our project.
We can tackle some of my comments in another future PR.

The table units of before and after and the function calls look a little verbose.

ci/environment.yml
philipc2: We had some isolated, but reproducible issues with the most recent release for some of our visualizations tasks. Our Conda and PyPi releases have it pinned, so this updates it since I left that out last release. 

https://github.com/UXARRAY/uxarray/blob/3b27d3f3ccf1d4a84a28f6538b2a91f79e051160/pyproject.toml#L36

benchmarks/quad_hexagon.py
philipc2: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#timing

.github/workflows/asv-benchmarking-pr.yml
philipc2: > Table broken down into various segments - mem, open time etc.

This is something that we could do. There is a `--sort` flag that could sort the results based on name or ratio. 

> Maintain another repo for keeping a record a results. Something where we can just based on commit sha track what changes led to the changes in performance.

We already do this with our package benchmark (which is our other automated workflow). It keeps track of a record of results for each commit that is part of the main branch.

https://github.com/UXARRAY/uxarray-asv

For example, looking at the web page for one of the benchmarks, each of the points represents a run for a certain commit.

https://uxarray.github.io/uxarray-asv/#face_bounds.FaceBounds.peakmem_face_bounds

I don't think it's worth keeping the results from this PR, other than what is already done with the artifact, since these are mostly to get an idea of what the performance impact of a given PR would be. When (or if) the changes get merged, our other workflow will record the new commit on main and store it in our `uxarray-asv` repo. 

> A way to get the performance difference between two dates from the point number 2 mentioned above.

This might be a helpful reference: https://asv.readthedocs.io/en/stable/commands.html#asv-continuous

benchmarks/quad_hexagon.py
philipc2: This benchmark was giving me a lot of trouble. Regardless of what changes I made to the environment or configuration, it would always throw and error. 

This may be something to do with what is written here:

https://asv.readthedocs.io/en/stable/writing_benchmarks.html#memory

However, this benchmark works fine for our other workflow.

philipc2:
> This is excellent and very timely for our project. We can tackle some of my comments in another future PR.
> 
> The table units of before and after and the function calls look a little verbose.

Thanks for the review! Yeah, the table can definitely be trimmed a bit to look cleaner. I didn't do any processing of the data, simply returned the result of the `asv compare` function call.

I'll definitely look into at least removing the `Change` column, since that's not being populated at all and is just taking up space.

erogluorhan (approved):
This is incredibly useful! Despite some issues (e.g. the Change column), I am fine with getting this merged and fixing those issues in a separate PR since this will not break any functionality and instead will be usable for other PRs immediately after the merge.

PR #868, opened 2024-07-29
HTML Repr for `Grid`, `UxDataArray` and `UxDataset`

### LINKED ISSUE #237: Custom `__repr__` & `_repr_html_` methods for `ux.Dataset`
The inherited `__repr__` and `_repr_html_` methods from `xarray.Dataset` do not display any information about the grid that the data lies on. Below are examples of both methods taken from the `Xarray` gallery

### `__repr__` Output
```Python
>>> print(ds)
```
```
<xarray.Dataset>
Dimensions:   (time: 731, location: 3)
Coordinates:
  * time      (time) datetime64[ns] 2000-01-01 2000-01-02 ... 2001-12-31
  * location  (location) <U2 'IA' 'IN' 'IL'
Data variables:
    tmin      (time, location) float64 -8.037 -1.788 -3.932 ... -1.346 -4.544
    tmax      (time, location) float64 12.98 3.31 6.779 ... 6.636 3.343 3.805
```

### `_repr_html_` Output (Jupyter Cell)
```Python
>>> # html output in a jupyter cell 
>>> ds
```
Screenshot 2023-03-11 194329

### Intended Implementation
* A section for Grid Topology Dimensions & Variables should be added to display information about the underlying grid. 
* Dimensions and Coordinate should be renamed to Data Dimensions and Data Coordinates respectively to specify they are for the data, not grid 
* Custom `formatting.py` and `formatting_html.py` implementations will be necessary, but much can be reused

philipc2:
Closes #237, #869

## Overview

* Introduces HTML Reprs for the `Grid`, `UxDataArray` and `UxDataset` classes to match the style of Xarray
* Grid information is now shown when viewing a `UxDataArray` and `UxDataset`

### Grid

image
 

### Default View (Grid is Collapsed)
image

### Toggled View (Grid information is visible) 

image

By default all grid variables are collapsed, but can be uncollapsed by clicking them (for example with connectivity)

image

## UxDataArray

image


philipc2:
This addition should bring us much closer to being considered a "true" Xarray extension, since now each Grid will be visible in a view that is familiar to Xarray users.

rajeeja:
Great and much needed addition, our codecov goes down, can't think of a way to test this, can you?
If some change in future breaks it, we won't be able to catch it.

philipc2:
> Great and much needed addition, our codecov goes down, can't think of a way to test this, can you? If some change in future breaks it, we won't be able to catch it.

I can try to write a test for this, however I'm not sure how we can properly test the functionality of a `repr` through our testing. Any thoughts?

rajeeja:
> > Great and much needed addition, our codecov goes down, can't think of a way to test this, can you? If some change in future breaks it, we won't be able to catch it.
> 
> I can try to write a test for this, however I'm not sure how we can properly test the functionality of a `repr` through our testing. Any thoughts?

For each Grid, UxDataArray, UxDataset; something like:
```

class TestMyClass(unittest.TestCase):
    def test_repr(self):
        obj = MyClass("test")
        self.assertEqual(repr(obj), "MyClass('test')")
        
    def test_repr_with_number(self):
        obj = MyClass(42)
        self.assertEqual(repr(obj), "MyClass(42)")
        
    def test_repr_with_list(self):
        obj = MyClass([1, 2, 3])
        self.assertEqual(repr(obj), "MyClass([1, 2, 3])")
```

philipc2:
> > > Great and much needed addition, our codecov goes down, can't think of a way to test this, can you? If some change in future breaks it, we won't be able to catch it.
> > 
> > 
> > I can try to write a test for this, however I'm not sure how we can properly test the functionality of a `repr` through our testing. Any thoughts?
> 
> For each Grid, UxDataArray, UxDataset; something like:
> 
> ```
> 
> class TestMyClass(unittest.TestCase):
>     def test_repr(self):
>         obj = MyClass("test")
>         self.assertEqual(repr(obj), "MyClass('test')")
>         
>     def test_repr_with_number(self):
>         obj = MyClass(42)
>         self.assertEqual(repr(obj), "MyClass(42)")
>         
>     def test_repr_with_list(self):
>         obj = MyClass([1, 2, 3])
>         self.assertEqual(repr(obj), "MyClass([1, 2, 3])")
> ```

I'll give this a try!

erogluorhan (approved):
This is a great addition; thanks! Other than the test @rajeeja suggested, I am fine with this PR.

Issue #572, opened 2023-11-08
Performance Comparison Between BallTree and KDTree


aaronzedwick:
### Proposed new feature or change:

Do a performance comparison between the balltree and kdtree implementation in UXarray and compare the two using a large scale mesh. Doing this in a Juypter notebook would probably be ideal.

aaronzedwick:
@philipc2 if you could get me that mesh you mentioned for this that would be great. Also, any advice on the best ways to make this comparison and different things I should test would be appreciated.

philipc2:
> @philipc2 if you could get me that mesh you mentioned for this that would be great. Also, any advice on the best ways to make this comparison and different things I should test would be appreciated.

They're on the NCAR HPC right now, I'll get them downloaded and uploaded to google drive today.

philipc2:
Initially, a simple notebook comparing:
1. Execution time of Construction of each Tree
2. Execution time of some number of queries for each tree

This doesn't have to be formal, but I'm extremely interested to see how it performs. We can convert some of this into a Remapping usage example eventually.

aaronzedwick:
Awesome, thank you!

aaronzedwick:
> > @philipc2 if you could get me that mesh you mentioned for this that would be great. Also, any advice on the best ways to make this comparison and different things I should test would be appreciated.
> 
> They're on the NCAR HPC right now, I'll get them downloaded and uploaded to google drive today.

Could you share the google drive link whenever it's up?

philipc2:
> > > @philipc2 if you could get me that mesh you mentioned for this that would be great. Also, any advice on the best ways to make this comparison and different things I should test would be appreciated.
> > 
> > 
> > They're on the NCAR HPC right now, I'll get them downloaded and uploaded to google drive today.
> 
> Could you share the google drive link whenever it's up?

I'm waiting to hear back from @erogluorhan whether I can share this dataset with non-NCAR collaborators. In the meantime, I can get you two others s as a comparison, but they wont be as large

rajeeja:
Use ASV benchmarks for showcasing this.

aaronzedwick:
Completed by #870

PR #870, opened 2024-07-31
ASV Benchmarks for Remapping and Tree Construction

### LINKED ISSUE #852: Create ASV benchmarks for Remapping

Make ASV benchmark for NN and IDW remapping.

aaronzedwick:
Closes #852 #572

## Overview

Adds benchmarks for nearest neighbor and inverse distance weighted remapping, as well benchmarks for KD and Ball Tree construction.

## PR Checklist

**General**
- [x] Add appropriate labels

### DIFF
--- a/benchmarks/mpas_ocean.py
+++ b/benchmarks/mpas_ocean.py
@@ -134,18 +134,33 @@ class ConstructTreeStructures:
         self.uxds.uxgrid.get_ball_tree()
 
 
-class Remapping:
-    param_names = ['resolution']
-    params = ['480km', '120km']
+class RemapDownsample:
 
-    def setup(self, resolution):
-        self.uxds = ux.open_dataset(file_path_dict[resolution][0], file_path_dict[resolution][1])
+    def setup(self):
+        self.uxds_120 = ux.open_dataset(file_path_dict['120km'][0], file_path_dict['120km'][1])
+        self.uxds_480 = ux.open_dataset(file_path_dict['480km'][0], file_path_dict['480km'][1])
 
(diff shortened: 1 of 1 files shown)


benchmarks/mpas_ocean.py
```diff
@@ -115,3 +115,37 @@ def teardown(self, resolution, periodic_elements):
...
+        self.uxds = ux.open_dataset(file_path_dict[resolution][0], file_path_dict[resolution][1])
+
+    def teardown(self, resolution):
+        del self.uxds
+
+    def time_nearest_neighbor_remapping(self, resolution):
+        self.uxds["bottomDepth"].remap.nearest_neighbor(self.uxds.uxgrid)
+
+    def time_inverse_distance_weighted_remapping(self, resolution):
+        self.uxds["bottomDepth"].remap.inverse_distance_weighted(self.uxds.uxgrid)
```
philipc2, line 151: Looks like the remapping here is from the same grid back onto itself, which may not be the best thing we want to benchmark. Might be a little tricky to handle the `480` to `120` and `120` to `480` benchmarks from the same class.

Maybe split it into two classes:
- `RemapDownsample` (this would be the `120` to `480`)
- `RemapUpsample` (this would be the `480` to `120`) 

The `setup` method can then open both grids.

philipc2 (changes requested):

benchmarks/mpas_ocean.py
aaronzedwick: That's a good idea!

philipc2 (approved):
These are some great additions!

PR #785, opened 2024-05-10
Non-Conservative Zonal Mean

philipc2:
## Overview

- Implement Non-Conservative Zonal Mean. 
- Bug fixes for `_get_zonal_faces_weight_at_constLat` algorithm
- Bug fixes for `gca_constlat_intersection` algorithm
- Bug fixes for `_pole_point_inside_polygon` algorithm
## Expected Usage

```Python
import uxarray as ux

grid_path = "/path/to/grid.nc"
data_path = "/path/to/data.nc"

uxds = ux.open_dataset(grid_path, data_path)

# compute the zonal average for the default start_lat_deg=-90, end_lat_deg=90, step_deg=5
zonal_result = uxds['data_var'].zonal_mean()

# Compute the zonal mean for a single latitude:
zonal_result = uxds["t2m"].zonal_mean(30.0)
```

## PR Checklist

**General**
- [x] An issue is linked created and linked
- [x] Add appropriate labels
- [x] Filled out Overview and Expected Usage (if applicable) sections

**Testing**
- [x] Adequate tests are created if there is new functionality
- [x] Tests cover all possible logical paths in your function
- [x] Tests are not too basic (such as simply calling a function and nothing else)

**Documentation**
- [x] Docstrings have been added to all new functions
- [x] Docstrings have updated with any function changes
- [x] Internal functions have a preceding underscore (`_`) and have been added to `docs/internal_api/index.rst`
- [x] User functions have been added to `docs/user_api/index.rst`

**Examples**
- [x] Any new notebook examples added to `docs/examples/` folder
- [x] Clear the output of all cells before committing
- [x] New notebook files added to `docs/examples.rst` toctree
- [ ] New notebook files added to new entry in `docs/gallery.yml` with appropriate thumbnail photo in `docs/_static/thumbnails/`

PR #877, opened 2024-08-05
Optimize Face Bounds Computation 

### LINKED ISSUE #730: Optimize Face Bounds Construction

philipc2:
Closes #730

## Overview

* Decorated Numpy functions (`allclose`, `isclose`, `cross`, `dot`) that were used by helper functions when building the face bounds
* Added a global option to enable or disable fma instructions

## Expected Usage

```Python
import uxarray as ux

```

### DIFF
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -39,7 +39,7 @@
 
     // List of branches to benchmark. If not provided, defaults to "master"
     // (for git) or "default" (for mercurial).
-    "branches": ["main"], // for git
+    "branches": ["philipc2/optimize-bounds"], // for git
     // "branches": ["default"],    // for mercurial
 
     // The DVCS being used.  If not set, it will be automatically
--- a/benchmarks/asv.conf.json
+++ b/benchmarks/asv.conf.json
@@ -39,7 +39,7 @@
 
     // List of branches to benchmark. If not provided, defaults to "master"
     // (for git) or "default" (for mercurial).
-    "branches": ["philipc2/optimize-bounds"], // for git
+    "branches": ["main"], // for git
     // "branches": ["default"],    // for mercurial
 
     // The DVCS being used.  If not set, it will be automatically
--- a/benchmarks/face_bounds.py
+++ b/benchmarks/face_bounds.py
@@ -10,8 +10,6 @@ grid_geoflow = current_path / "test" / "meshfiles" / "ugrid" / "geoflow-small" /
 grid_scrip = current_path / "test" / "meshfiles" / "scrip" / "outCSne8" / "outCSne8.nc"
 grid_mpas= current_path / "test" / "meshfiles" / "mpas" / "QU" / "oQU480.231010.nc"
 
-
-
 class FaceBounds:
 
     params = [grid_quad_hex, grid_geoflow, grid_scrip, grid_mpas]
--- a/benchmarks/mpas_ocean.py
+++ b/benchmarks/mpas_ocean.py
@@ -164,36 +164,3 @@ class RemapUpsample:
 
     def time_inverse_distance_weighted_remapping(self):
         self.uxds_480["bottomDepth"].remap.inverse_distance_weighted(self.uxds_120.uxgrid)
-
-from uxarray.grid.utils import (
-    _get_cartesian_face_edge_nodes,
-    _get_lonlat_rad_face_edge_nodes,
-)
-
-class FaceBoundsHelpers:
-    param_names = ['resolution']
-    params = ['480km', '120km']
-
-
-    def setup(self, resolution):
-        self.uxgrid = ux.open_grid(file_path_dict[resolution][0])
-
-    def teardown(self, resolution):
-        del self.uxgrid
-
-    def time_get_get_cartesian_face_edge_nodes(self, resolution):
-        _get_cartesian_face_edge_nodes(self.uxgrid.face_node_connectivity.values,
-                                       self.uxgrid.n_face,
-                                       self.uxgrid.n_max_face_edges,
-                                       self.uxgrid.node_x.values,
-                                       self.uxgrid.node_y.values,
-                                       self.uxgrid.node_z.values)
-
-    def time_get_lonlat_rad_face_edge_nodes(self, resolution):
-        _get_lonlat_rad_face_edge_nodes(
-            self.uxgrid.face_node_connectivity.values,
-            self.uxgrid.n_face,
-            self.uxgrid.n_max_face_edges,
-            self.uxgrid.node_lon.values,
-            self.uxgrid.node_lat.values,
-        )
--- a/benchmarks/mpas_ocean.py
+++ b/benchmarks/mpas_ocean.py
@@ -164,3 +164,36 @@ class RemapUpsample:
 
     def time_inverse_distance_weighted_remapping(self):
         self.uxds_480["bottomDepth"].remap.inverse_distance_weighted(self.uxds_120.uxgrid)
+
+from uxarray.grid.utils import (
+    _get_cartesian_face_edge_nodes,
+    _get_lonlat_rad_face_edge_nodes,
+)
+
+class FaceBoundsHelpers:
+    param_names = ['resolution']
+    params = ['480km', '120km']
+
+
+    def setup(self, resolution):
+        self.uxgrid = ux.open_grid(file_path_dict[resolution][0])
+
+    def teardown(self, resolution):
+        del self.uxgrid
+
+    def time_get_get_cartesian_face_edge_nodes(self, resolution):
+        _get_cartesian_face_edge_nodes(self.uxgrid.face_node_connectivity.values,
+                                       self.uxgrid.n_face,
+                                       self.uxgrid.n_max_face_edges,
+                                       self.uxgrid.node_x.values,
+                                       self.uxgrid.node_y.values,
+                                       self.uxgrid.node_z.values)
+
+    def time_get_lonlat_rad_face_edge_nodes(self, resolution):
+        _get_lonlat_rad_face_edge_nodes(
+            self.uxgrid.face_node_connectivity.values,
+            self.uxgrid.n_face,
+            self.uxgrid.n_max_face_edges,
+            self.uxgrid.node_lon.values,
+            self.uxgrid.node_lat.values,
+        )
--- a/test/test_arcs.py
+++ b/test/test_arcs.py
@@ -36,7 +36,7 @@ class TestIntersectionPoint(TestCase):
         ]
         pt_same_lon_in = _lonlat_rad_to_xyz(0.0, 0.0)
         with self.assertRaises(ValueError):
-            point_within_gca(pt_same_lon_in, gcr_180degree_cart)
+            point_within_gca(np.asarray(pt_same_lon_in), np.asarray(gcr_180degree_cart))
 
         gcr_180degree_cart = [
             _lonlat_rad_to_xyz(0.0, np.pi / 2.0),
@@ -45,7 +45,7 @@ class TestIntersectionPoint(TestCase):
 
         pt_same_lon_in = _lonlat_rad_to_xyz(0.0, 0.0)
         with self.assertRaises(ValueError):
-            point_within_gca(pt_same_lon_in, gcr_180degree_cart)
+            point_within_gca(np.asarray(pt_same_lon_in), np.asarray(gcr_180degree_cart))
 
         # Test when the point and the GCA all have the same longitude
         gcr_same_lon_cart = [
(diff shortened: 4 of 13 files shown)


philipc2:
@hongyuchen1030 

For the first step in optimizing, I've looked at decorating some of the commonly called functions with `numba`, and used a `numba` implementation of `isclose`. Looks to be about a 15-20% performance improvement to begin with.

Next step will be to eliminate any unnecessary coordinate conversion calls, such as `_xyz_to_lonlat_rad`

This might look like us providing both the `latlon_rad` and `cartesian` points in the function call to avoid needing to convert. Something like:

```Python
 def _point_within_gca(pt_latlon,
                       pt_xyz,
                       gca_a_latlon,
                       gca_a_xyz,
                       gca_b_latlon,
                       gca_b_xyz,
                       is_directed=False):
```

Hopefully once we've moved these conversions out of each function call, we can see if the remaining logic can be decorated with `numba`.

philipc2:
@hongyuchen1030 


https://github.com/UXARRAY/uxarray/blob/1790e0e30abf6b8869c35747d6393af03bccac88/uxarray/grid/arcs.py#L77

I had a question about this and other parts of the code where the Cartesian coordinates are converted to latlon in radians. Are we able to use the `node_lon` and `node_lat` that are part of the `Grid` (in degrees) instead of needing to run the conversion? Additionally, is radians here just a preference or can we use degrees?

hongyuchen1030:
> Next step will be to eliminate any unnecessary coordinate conversion calls, such as `_xyz_to_lonlat_rad`
>
> This might look like us providing both the `latlon_rad` and `cartesian` points in the function call to avoid needing to convert. Something like:
> 
> ```python
>  def _point_within_gca(pt_latlon,
>                        pt_xyz,
>                        gca_a_latlon,
>                        gca_a_xyz,
>                        gca_b_latlon,
>                        gca_b_xyz,
>                        is_directed=False):
> ```

I have also been considering this design today. It will be highly beneficial to call the coordinate conversion only once, both for performance and accuracy.

We can extract all functions that take node coordinates as input and ensure that both Cartesian and spherical coordinates are required as parameters.

> I had a question about this and other parts of the code where the Cartesian coordinates are converted to latlon in radians. Are we able to use the `node_lon` and `node_lat` that are part of the Grid (in degrees) instead of needing to run the conversion? Additionally, is radians here just a preference or can we use degrees?

That is another reason why I have been advocating for better documentation and conventions for the coordinate system.

Currently, we have the following coordinate systems:
1. Spherical coordinates in degrees for users.
2. Spherical coordinates in radians because the trigonometry functions in `numpy` use radians  (`np.sin`, `np.cos`, etc.).
3. Cartesian coordinate system with normalization.
4. Cartesian coordinate system without normalization (we still haven't confirmed if the node data in MPAS or other unnormalized coordinates are equal to the `normalized_coordinates * earth_radius` yet).

Before we attempt to rewrite everything, we must ensure which coordinate systems we are using across the entire project.

philipc2:
> I have also been considering this design today. It will be highly beneficial to call the coordinate conversion only once, both for performance and accuracy.

Agreed, was thinking of this as well.

> That is another reason why I have been advocating for better documentation and conventions for the coordinate system.

We have some documentation here in our user guide. https://uxarray.readthedocs.io/en/latest/user-guide/representation.html

Any suggestions on how we could improve this? 

> Currently, we have the following coordinate systems:

Wondering if it would make sense to include a `node_lon_rad` and `node_lat_rad` extension to how we represent our coordinates. I've seen some other grids store the coordinates both in degrees and radians, so this might be a worth extension (and would help with some of the conversions we are doing).

Hopefully the normalization inconsistencies will be resolved with #878

hongyuchen1030:
> We have some documentation here in our user guide. https://uxarray.readthedocs.io/en/latest/user-guide/representation.html
> 
> Any suggestions on how we could improve this?
image

1. I understand we already specify the convention underneath; however, the `node_lat` has the description specifying the units while others do not. It would be great to keep them uniform. Additionally, placing the note before the detailed description would enhance clarity.

2. Just a note: I think we didn't use the convention of [-180, 180] until we updated the coordinates conversion. The old conversion function used the convention of [0, 360). All my algorithms are based on [0, 2π) as well. I highly recommend using [0, 2π) or [0, 360) since it provides an easier way to handle the antimeridian faces. This also necessitates a thorough re-check after any major functions are updated.

3. It might also be beneficial to specify that we have a unique definition for the antimeridian points and the pole points.

> Wondering if it would make sense to include a `node_lon_rad` and `node_lat_rad` extension to how we represent our coordinates. I've seen some other grids store the coordinates both in degrees and radians, so this might be a worth extension (and would help with some of the conversions we are doing).

I think what we've discovered so far is that the degrees are solely for the users, while all calculations use radians since the `numpy` trigonometry functions operate in radians. I'm always open to utilizing storage space to reduce time complexity.

philipc2:
> I understand we already specify the convention underneath; however, the node_lat has the description specifying the units while others do not. It would be great to keep them uniform. Additionally, placing the note before the detailed description would enhance clarity.

That's a good catch! 

> Just a note: I think we didn't use the convention of [-180, 180] until we updated the coordinates conversion. The old conversion function used the convention of [0, 360). All my algorithms are based on [0, 2π) as well. I highly recommend using [0, 2π) or [0, 360) since it provides an easier way to handle the antimeridian faces. This also necessitates a thorough re-check after any major functions are updated.

We use -180 to 180 since it makes certain visualizations tasks (i.e. projections and the use of the `antimeridian` package) simpler. Most of our datasets are also already in the -180 to 180 range.

> I think what we've discovered so far is that the degrees are solely for the users, while all calculations use radians since the numpy trigonometry functions operate in radians. I'm always open to utilizing storage space to reduce time complexity.

This isn't fully the case. Many of our internal functions (i.e. BallTree) use degrees. 

 

> numpy trigonometry functions operate in radians

One option could be to use Scipy's degree implementations, but I'm not sure how much this implementation differs from Numpy's  

https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.sindg.html#scipy.special.sindg

philipc2:
@hongyuchen1030 

Would you mind explaining this code snippet for me?

https://github.com/UXARRAY/uxarray/blob/1790e0e30abf6b8869c35747d6393af03bccac88/uxarray/grid/arcs.py#L288-L292

hongyuchen1030:
> @hongyuchen1030
> 
> Would you mind explaining this code snippet for me?
> 
> https://github.com/UXARRAY/uxarray/blob/1790e0e30abf6b8869c35747d6393af03bccac88/uxarray/grid/arcs.py#L288-L292

Sure, I wonder if you want to hop into a quick call?

hongyuchen1030:
> We use -180 to 180 since it makes certain visualization tasks (i.e., projections and the use of the antimeridian package) simpler. Most of our datasets are also already in the -180 to 180 range.

I understand the benefits of using the -180 to 180 range for visualization tasks and consistency with existing datasets. My algorithms do not expose the longitude range to users, nor do they modify existing longitudes, so this shouldn't pose a significant issue. However, I wanted to bring this up because if we require functions to take `node_lat` and `node_lon` from the `grid`, we may need to quickly check if longitude conversion is necessary. This note could be helpful for troubleshooting future bugs.

> This isn't fully the case. Many of our internal functions (i.e., BallTree) use degrees.

Is this because we directly use the `grid.node_lat` and `grid.node_lon` without calling the `numpy` trigonometric functions?

> One option could be to use Scipy's degree implementations, but I'm not sure how much this implementation differs from Numpy's.

I compared the source code for Scipy's `sindg` and Numpy's `sin`. Here’s what I found:

1. Numpy’s `sin` is a wrapper for the compiled `c_sin`, with the source code for `s_sin` available [here](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/fpu/s_sinf.S;h=1529fb1f2452cefa1fdf482f8fa9165ffef4c197;hb=HEAD).
2. Scipy’s `sindg` was implemented recently in this [`.h` file](https://github.com/scipy/scipy/blob/94532e74b902b569bfad504866cb53720c5f4f31/scipy/special/xsf/cephes/sindg.h#L5), and it still converts some variables into radians.

I haven't delved deeply into the internal workings of both functions, but we could perform a simple benchmark to compare their running times and compatibility if needed.

philipc2:
@hongyuchen1030 

After some further optimizations, this is the point that I've gotten to. Looks like the `pyfma` implementations still contribute to a pretty significant slowdown, taking up the majority of the execution time. 

## FMA Disabled 
~5.73s Execution Time 
image


## FMA Enabled
~14.1s Execution Time 

image

philipc2:
A few more refinements and with `fma` disabled, able to get down for 3.74s

image

philipc2:
image

From the benchmarks run above, looks like almost a 8-10x improvement in performance across the board with these changes. 

I was also able to compute the bounds on a 30km MPAS atmosphere grid (655k Faces, 1,310,000 nodes) in about 40 seconds, which is a significant improvement and wouldn't of been possible without these optimizations.

hongyuchen1030:
> @hongyuchen1030
> 
> After some further optimizations, this is the point that I've gotten to. Looks like the `pyfma` implementations still contribute to a pretty significant slowdown, taking up the majority of the execution time.
> 
> ## FMA Disabled
> ~5.73s Execution Time image
> 
> ## FMA Enabled
> ~14.1s Execution Time
> 
> image
> 
> Do you think anything from here could possible act as an alternative to using `pyfma`?
> 
> https://docs.scipy.org/doc/scipy/reference/linalg.blas.html

I've been giving this case a lot of thought, considering my research findings and the intricacies of floating point operations.

Given the technical details involved with floating point operations, I won't delve into too much detail to avoid overwhelming you. Here's a summary of my thoughts:

1. **Maintain Current Approach**: We can keep the current implementation as it is, but set `is_FMA_enable` to `False` globally and use the basic `np.cross`. We'll document this choice, informing users that they can opt for a more accurate method by enabling FMA.
2. **Adjust Error Tolerance**: The immediate task would be to reset the error tolerance for intersection point snapping. This warrants a separate issue to be opened specifically for this. And since we are not allowed to use any compiled files in our project code base, we cannot do much about it now.

### Rationale:

1. **Algorithm Accuracy**: While CDO uses Kahan's Algorithm with FMA, my research indicates there's a more accurate method available. This method works with and without FMA, though the non-FMA version requires more flops. We can explore this choice further as we develop in Python.
2. **Floating Point Operations**: Standard floating point operations generally produce acceptable results. We can monitor their reliability by calculating the condition number. I'm working on an adaptive algorithm to address this in the future.
3. **Task Complexity**: As you might have noticed, all these floating point operations in Python inevitably lead to compiled C. Although all these packages use the same compiled C function, their binding and wrapping are different, making them difficult to connect properly.

### Conclusion:

Since FMA isn't currently ready for the Python package, we can focus on the general case first and handle the ill-conditioned cases later. The package should primarily use normal floating point operations, and when an ill-conditioned case arises, it's not overly time-consuming to apply the more accurate operations to this subset.

If you have any questions or need further clarification, please don't hesitate to ask. I'm more than happy to discuss and explain everything.

philipc2:
> Maintain Current Approach: We can keep the current implementation as it is, but set is_FMA_enable to False globally and use the basic np.cross. We'll document this choice, informing users that they can opt for a more accurate method by enabling FMA.

I think this option makes the most sense to me and would fit the expectations of our users also. 

> And since we are not allowed to use any compiled files in our project code base, we cannot do much about it now.

I wouldn't say that we are **_not allowed_**, it's more something that we haven't explored yet. It might be worth considering. 

I appreciate the write up and glad that we're making some good improvements both to the performance and accuracy of our operators.

philipc2:
Merging this so that we can continue our work with #785 

Overall, was able to get about a 10-12x performance improvement by using `numba` without major rewrites to the algorithm. There's still some more optimizations we could do, but this would require significant rewrites, which should not be a priority right now.

hongyuchen1030:
> Merging this so that we can continue our work with #785
> 
> Overall, was able to get about a 10-12x performance improvement by using `numba` without major rewrites to the algorithm. There's still some more optimizations we could do, but this would require significant rewrites, which should not be a priority right now.

Thank you very much for your excellent jobs! And I will open another issue and PR about our rewriting decisions for the helper functions.

philipc2:
> > Merging this so that we can continue our work with #785
> > Overall, was able to get about a 10-12x performance improvement by using `numba` without major rewrites to the algorithm. There's still some more optimizations we could do, but this would require significant rewrites, which should not be a priority right now.
> 
> Thank you very much for your excellent jobs! And I will open another issue and PR about our rewriting decisions for the helper functions.

Sounds good!

PR #878, opened 2024-08-06
Normalization of Parsed Cartesian Coordinates 

### LINKED ISSUE #872: Inconsistent Normalization of Cartesian Coordinates in `node_x`, `node_y`, and `node_z`
### Version

v2024.06

Conda

### What happened?

Our project performs geometric operations on unstructured grids on a unit sphere, which means all points on the sphere should be normalized. The normalized Cartesian coordinates are stored in the `node_x`, `node_y`, and `node_z` variables.

However, while in most cases the `node_x`, `node_y`, and `node_z` variables are normalized as they should be, when reading in the MPAS data, these variables are not normalized. This inconsistency leads to breakdowns in subsequent operations.

 the `node_x`, `node_y`, and `node_z` variables to be consistently normalized across all datasets, including the MPAS data.

It is crucial to emphasize and document that  `node_x`, `node_y`, and `node_z`must always represent normalized points on the sphere. Ensuring these variables are normalized as required is essential for the robustness and reliability of our project.

_No response_

philipc2:
Closes #872 

## Overview

* Adds a helper `uxarray.grid.validation._check_normalization()` to check if the Cartesian coordinates stored under a `Grid` are normalized. 
* Adds `Grid.normalize_cartesian_coordinates()` to normalize coordinates if they were initially stored as non-normalized. 

## Expected Usage

```Python
import uxarray as ux

# normalize coordinates
uxds.uxgrid.normalize_cartesian_coordinates()

```


uxarray/grid/validation.py
```diff
@@ -65,3 +65,19 @@ def _check_area(self):
...
+        if not (grid.node_x**2 + grid.node_y**2 + grid.node_z**2).all():
+            return False
+    if "edge_x" in grid._ds:
+        if not (grid.edge_x**2 + grid.edge_y**2 + grid.edge_z**2).all():
+            return False
+    if "face_x" in grid._ds:
+        if not (grid.face_x**2 + grid.face_y**2 + grid.face_z**2).all():
+            return False
+
+    return True
```
philipc2, line 83: @hongyuchen1030 

Wondering if this is an appropriate way to check for normalization?

hongyuchen1030: Yeah this is the most fundamental way to checked if the coordinates are normalized. However, as I mentioned before, it will run all the nodes, leading to potential performance slow down

philipc2:
@hongyuchen1030 

Do you recall which MPAS dataset had the non-normalized Cartesian coordinates? The two we have in testing both are already normalized.

philipc2:
@hongyuchen1030 

We should expect the coordinates to be in the `[-1, 1]` range as well correct?

hongyuchen1030:
> @hongyuchen1030
> 
> We should expect the coordinates to be in the `[-1, 1]` range as well correct?

Yes, correct.

And unfortunately, I don't remember which dataset I ran into, but I will like you know once I saw it.

uxarray/grid/validation.py
```diff
@@ -65,3 +65,25 @@ def _check_area(self):
...
         print("-No face area is close to zero.")
         return True
+
+
+def _check_normalization(grid):
+    """Checks whether all the cartesiain coordinates are normalized."""
+
+    if "node_x" in grid._ds:
+        if not (
+            np.isclose((grid.node_x**2 + grid.node_y**2 + grid.node_z**2), 1.0)
```
philipc2, line 75: @hongyuchen1030 

I've updated the check here to account for small floating point differences when checking if the sum of squares adds to one. Does this look good as a check, or should I include anything else?

hongyuchen1030: Thanks for your work!
1. I feel like we should set the error tolerance for the isclose function
2. As I mentioned before, this kind of check will loop over the entire grid and do the following: 

Step1: Calculate grid.node_x**2 + grid.node_y**2 + grid.node_z**2 and see if their results are one

Step2: if the results are not one, Calculate grid.node_x**2 + grid.node_y**2 + grid.node_z**2 again so it can know the norm for the following normalization step.

Can we just store the norm from Step1 until we either: 1. the step 1 return true 2. Or we finish the step 2 normalization?

rajeeja: > Thanks for your work!
> 
> 1. I feel like we should set the error tolerance for the isclose function
np.close has:
rtol (relative tolerance): Default is 1e-05
atol (absolute tolerance): Default is 1e-08

what do you wnat to set it to? we can set it as:

`result = np.isclose(result, 1, rtol=1e-06, atol=1e-09)`

philipc2: > As I mentioned before, this kind of check will loop over the entire grid and do the following:

Great suggestion and thanks for the reminder, forgot to include this.

philipc2: > I feel like we should set the error tolerance for the isclose function

Should we use the global error tolerance? https://github.com/UXARRAY/uxarray/blob/ff9ec76e5ce3422a8af7005e843fcec0a75155d5/uxarray/constants.py#L13

rajeeja: > > I feel like we should set the error tolerance for the isclose function
> 
> Should we use the global error tolerance?
> 
> https://github.com/UXARRAY/uxarray/blob/ff9ec76e5ce3422a8af7005e843fcec0a75155d5/uxarray/constants.py#L13

+1

philipc2: @hongyuchen1030 @rajeeja 

Added the error tolerance and cached flag indicated if we've already run the computation.

test/test_grid.py
```diff
@@ -952,7 +952,38 @@ def test_populate_bounds_GCA_mix(self):
         face_bounds = bounds_xarray.values
         nt.assert_allclose(grid.bounds.values, expected_bounds, atol=ERROR_TOLERANCE)
 
-    def test_opti_bounds(self):
-        import uxarray
-        uxgrid = ux.open_grid(gridfile_CSne8)
-        bounds = uxgrid.bounds
+    def test_populate_bounds_MPAS(self):
+        xrds = xr.open_dataset(self.gridfile_mpas)
+        uxgrid = ux.Grid.from_dataset(xrds, use_dual=True)
```
rajeeja, line 957: is something missing here? 
why does it `pass` at the end?

which grid has non normalized cart coords for testing? it would good to document it.

rajeeja (changes requested):

test/test_grid.py
philipc2: The `pass` is most likely left over from the initial writing of the test case.

I believe this was from a merge conflict that I was resolving. Will fix this.

philipc2:
@hongyuchen1030 @amberchen122 

This should be ready for review.

erogluorhan (approved):
This looks good to me with the application of global error tolerance to `isclose` checks.

rajeeja (approved):

PR #859, opened 2024-07-19
Dual Mesh Construction

aaronzedwick:
Closes #825 

## Overview

Constructs the dual mesh of a grid using `face_centers` and `node_face_connectivity`
## Expected Usage

```Python
import uxarray as ux

grid_path = "/path/to/grid.nc"

grid = ux.open_grid(grid_path)

dual = grid.compute_dual()

```

## PR Checklist

**General**
- [x] An issue is linked created and linked
- [x] Add appropriate labels
- [x] Filled out Overview and Expected Usage (if applicable) sections

**Testing**
- [x] Adequate tests are created if there is new functionality
- [x] Tests cover all possible logical paths in your function
- [x] Tests are not too basic (such as simply calling a function and nothing else)

**Documentation**
- [x] Docstrings have been added to all new functions
- [x] Docstrings have updated with any function changes
- [x] Internal functions have a preceding underscore (`_`) and have been added to `docs/internal_api/index.rst`
- [x] User functions have been added to `docs/user_api/index.rst`

PR #879, opened 2024-08-06
Merge Duplicate Node Indices

### LINKED ISSUE #865: Handling Duplicate Node Indices
Some of our testing grids include duplicates of certain nodes, meaning that there existing multiple nodes that share the same latitude and longitude location, but each have their own index within our connectivity and data variables.

### Example
Consider the following simple grid (

`node_lon = (-180, 0, -180, 180)`
`node_lat = (-90, 0, -90, 90)`
`data_on_nodes = [1, 2, 1, 3]`

Indices 0 and 1 contain the same latitude, longitude, and data value. This causes issues when constructing certain connectivity's (such as `node_face`, since some nodes will only belong to one face.

### Affected Grids
The following grid files in our testing suite each have duplicate nodes. 
`geos-cs/c1/test-c12.native.nc4`
`ugrid/geoflow-small/grid.nc`
`esmf/ne30/ne30pg3.grid.nc`

All of these are some form of a cube-sphere (or variation) 

### Proposed Solution
Instead of deleting the duplicates (which would require a complete re-ordering of connectivity and to modify the size of all node-centered data variables), a simple masking and merging of node indices at the connectivity level should do the trick.

### Example

Let's say node indicies `[0, 5, 10, 23]` all share the same data value and lat/lon coordinates. We can itterate over the `face_node`, `edge_node` and `node_node` connectivity arrays and set the values of `[5, 10, 23]` to point to `0`. This would allow us to leave the data arrays untouched and do a single pass over our connectivity's arrays with a direct replacement, without needing to reshape them in any way.

philipc2:
Closes #865

## Overview

* Adds the `merge_duplicate_node_indices()` function that replaces the index of nodes that have the exact latitude and longitude to only reference one node.

## Expected Usage

```Python
import uxarray as ux

# inplace
uxgrid.merge_duplicate_node_indices()

# return a new grid
new_grid = uxgrid.merge_duplicate_node_indices(inplace=False)

```

### DIFF
--- a/test/test_grid.py
+++ b/test/test_grid.py
@@ -31,6 +31,7 @@ gridfile_CSne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne3
 gridfile_fesom = current_path / "meshfiles" / "ugrid" / "fesom" / "fesom.mesh.diag.nc"
 gridfile_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "grid.nc"
 gridfile_mpas = current_path / 'meshfiles' / "mpas" / "QU" / 'mesh.QU.1920km.151026.nc'
+gridfile_geos_cs = current_path / "meshfiles" / "geos-cs" / "c12" / "test-c12.native.nc4"
 
 dsfile_vortex_CSne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_vortex.nc"
 dsfile_var2_CSne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_var2.nc"
@@ -658,8 +659,8 @@ class TestConnectivity(TestCase):
 
         # construct edge nodes
         edge_nodes_output, _, _ = _build_edge_node_connectivity(mpas_grid_ux.face_node_connectivity.values,
-                                                          mpas_grid_ux.n_face,
-                                                          mpas_grid_ux.n_max_face_nodes)
+                                                                mpas_grid_ux.n_face,
+                                                                mpas_grid_ux.n_max_face_nodes)
 
         # _populate_face_edge_connectivity(mpas_grid_ux)
         # edge_nodes_output = mpas_grid_ux._ds['edge_node_connectivity'].values
@@ -932,6 +933,7 @@ class TestClassMethods(TestCase):
 
 class TestLatlonBounds(TestCase):
     gridfile_mpas = current_path / "meshfiles" / "mpas" / "QU" / "oQU480.231010.nc"
+
     def test_populate_bounds_GCA_mix(self):
         face_1 = [[10.0, 60.0], [10.0, 10.0], [50.0, 10.0], [50.0, 60.0]]
         face_2 = [[350, 60.0], [350, 10.0], [50.0, 10.0], [50.0, 60.0]]
@@ -941,11 +943,10 @@ class TestLatlonBounds(TestCase):
         faces = [face_1, face_2, face_3, face_4]
 
         # Hand calculated bounds for the above faces in radians
-        expected_bounds = [[[0.17453293, 1.07370494],[0.17453293, 0.87266463]],
-                           [[0.17453293, 1.10714872],[6.10865238, 0.87266463]],
-                           [[1.04719755, 1.57079633],[3.66519143, 0.52359878]],
-                           [[1.04719755,1.57079633],[0.,         6.28318531]]]
-
+        expected_bounds = [[[0.17453293, 1.07370494], [0.17453293, 0.87266463]],
+                           [[0.17453293, 1.10714872], [6.10865238, 0.87266463]],
+                           [[1.04719755, 1.57079633], [3.66519143, 0.52359878]],
+                           [[1.04719755, 1.57079633], [0., 6.28318531]]]
 
         grid = ux.Grid.from_face_vertices(faces, latlon=True)
         bounds_xarray = grid.bounds
--- a/test/test_grid.py
+++ b/test/test_grid.py
@@ -983,3 +983,12 @@ class TestDualMesh(TestCase):
 
         # Assert the faces are the same
         nt.assert_equal(dual.face_node_connectivity.values,  mpas_dual.face_node_connectivity.values)
+
+
+    def test_duplicate(self):
+        uxgrid = ux.open_grid(gridfile_geos)
+        uxgrid.merge_duplicate_node_indices()
+
+        uxgrid.compute_dual()
+
+        pass
--- a/test/test_duplicate_nodes.py
+++ b/test/test_duplicate_nodes.py
@@ -1,0 +1,26 @@
+import uxarray as ux
+
+import os
+from pathlib import Path
+
+current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+
+import numpy.testing as nt
+
+gridfile_geos_cs = current_path / "meshfiles" / "geos-cs" / "c12" / "test-c12.native.nc4"
+
+
+
+def test_merge_nodes_geos():
+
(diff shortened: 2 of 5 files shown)


aaronzedwick:
Do you know what the problem was with the previous implementation? If I remember it was still having problems?

philipc2:
> Do you know what the problem was with the previous implementation? If I remember it was still having problems?

I haven't been able to pinpoint the problem, but everything appears to be working correctly here according to the tests.

philipc2:
@aaronzedwick 

I'm going to merge these changes into the branch used in #859 to see the compatibility.

philipc2:
@aaronzedwick 

May you take a look at the case using the `geos` dataset? Looks like something is breaking in the `construct_faces` function call.

rajeeja:
what tests are failing? we may want to issue a warning on how and what is supported.
Supporting all formats may be harder and might break with corner cases.

philipc2:
> what tests are failing? we may want to issue a warning on how and what is supported. Supporting all formats may be harder and might break with corner cases.

Something with the Dual Mesh Construction on the Cube Sphere grid is breaking. It's complaining about a dimension miss match when assigned the faces. 

I don't believe it's an issue with how we are merging ndoes.

aaronzedwick:
> @aaronzedwick
> 
> May you take a look at the case using the `geos` dataset? Looks like something is breaking in the `construct_faces` function call.

Oh yeah, I know the problem. Let me fix it for you.

aaronzedwick:
@philipc2 There you go!

philipc2:
> @philipc2 There you go!

Thanks for fixing this so quickly!

philipc2:
@aaronzedwick 

There still appear to be some underlying issues, at least on the Cube Sphere grid.

Running the following executes, however when I try to plot it I get the following error:

```Python
uxgrid.merge_duplicate_node_indices(inplace=True)
dual = uxgrid.compute_dual()
```

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[18], line 1
----> 1 dual.plot()

File ~/anaconda3/envs/dev-uxarray/lib/python3.11/site-packages/uxarray/plot/accessor.py:30, in GridPlotAccessor.__call__(self, **kwargs)
     29 def __call__(self, **kwargs) -> Any:
---> 30     return grid_plot.plot(self._uxgrid, **kwargs)

File ~/anaconda3/envs/dev-uxarray/lib/python3.11/site-packages/uxarray/plot/grid_plot.py:16, in plot(grid, **kwargs)
     14 def plot(grid: Grid, **kwargs):
     15     """Default Plotting Method for Grid."""
---> 16     return mesh(grid, **kwargs)

File ~/anaconda3/envs/dev-uxarray/lib/python3.11/site-packages/uxarray/plot/grid_plot.py:43, in mesh(uxgrid, backend, exclude_antimeridian, width, height, xlabel, ylabel, **kwargs)
     19 def mesh(
     20     uxgrid: Grid,
     21     backend: Optional[str] = "bokeh",
   (...)
     27     **kwargs,
     28 ):
     29     """Vector Line Plot of the edges that make up each face.
     30 
     31     Parameters
   (...)
     40         Plot Width for Bokeh Backend
     41     """
---> 43     gdf = uxgrid.to_geodataframe(exclude_antimeridian=exclude_antimeridian)
     45     hv_paths = hv.Path(gdf)
     47     uxarray.plot.utils.backend.assign(backend=backend)

File ~/anaconda3/envs/dev-uxarray/lib/python3.11/site-packages/uxarray/grid/grid.py:1304, in Grid.to_geodataframe(self, override, cache, exclude_antimeridian)
   1301     return self._gdf
   1303 # construct a geodataframe with the faces stored as polygons as the geometry
-> 1304 gdf = _grid_to_polygon_geodataframe(
   1305     self, exclude_antimeridian=exclude_antimeridian
   1306 )
   1308 # cache computed geodataframe
   1309 if cache:

File ~/anaconda3/envs/dev-uxarray/lib/python3.11/site-packages/uxarray/grid/geometry.py:95, in _grid_to_polygon_geodataframe(grid, exclude_antimeridian)
     91 def _grid_to_polygon_geodataframe(grid, exclude_antimeridian):
     92     """Converts the faces of a ``Grid`` into a ``spatialpandas.GeoDataFrame``
     93     with a geometry column of polygons."""
---> 95     polygon_shells = _build_polygon_shells(
     96         grid.node_lon.values,
     97         grid.node_lat.values,
     98         grid.face_node_connectivity.values,
     99         grid.n_face,
    100         grid.n_max_face_nodes,
    101         grid.n_nodes_per_face.values,
    102     )
    104     antimeridian_face_indices = _build_antimeridian_face_indices(
    105         polygon_shells[:, :, 0]
    106     )
    108     if grid.n_face > GDF_POLYGON_THRESHOLD:

File ~/anaconda3/envs/dev-uxarray/lib/python3.11/site-packages/uxarray/grid/geometry.py:82, in _build_polygon_shells(node_lon, node_lat, face_node_connectivity, n_face, n_max_face_nodes, n_nodes_per_face, projection)
     77     node_lon = lonlat_proj[:, 0]
     78     node_lat = lonlat_proj[:, 1]
     80 polygon_shells = (
     81     np.array(
---> 82         [node_lon[closed_face_nodes], node_lat[closed_face_nodes]], dtype=np.float32
     83     )
     84     .swapaxes(0, 1)
     85     .swapaxes(1, 2)
     86 )
     88 return polygon_shells

IndexError: index -9223372036854775808 is out of bounds for axis 0 with size 864


Some of the entries in the `face_node_connectivity` are now incorrect. There are some entries that are an entire row of fill values. 

The number of entirely zero entries corresponds to how many duplicates were found.

image

Instead of having rows of full fill values, we should ensure that each duplicate has the same entry in the connectivity.

aaronzedwick:
@philipc2 I am not exactly sure what's happening, but I don't believe its a problem with my code. The only reason you are getting that error is because the connectivity that you are passing in is broken I think. Comment out the line in construct_faces that says if `n_edges[i] < 3: continue` and then compute dual will run, and the plot is still incorrect. All those fill value entries look to only have one node for some reason around which it is trying to make a dual face. Something with the `node_face_connectivity` I suppose?

philipc2:
> @philipc2 I am not exactly sure what's happening, but I don't believe its a problem with my code. The only reason you are getting that error is because the connectivity that you are passing in is broken I think. Comment out the line in construct_faces that says if `n_edges[i] < 3: continue` and then compute dual will run, and the plot is still incorrect. All those fill value entries look to only have one node for some reason around which it is trying to make a dual face. Something with the `node_face_connectivity` I suppose?

Yeah that seems to be the case, I'll look into it!

philipc2:
> @philipc2 I am not exactly sure what's happening, but I don't believe its a problem with my code. The only reason you are getting that error is because the connectivity that you are passing in is broken I think. Comment out the line in construct_faces that says if `n_edges[i] < 3: continue` and then compute dual will run, and the plot is still incorrect. All those fill value entries look to only have one node for some reason around which it is trying to make a dual face. Something with the `node_face_connectivity` I suppose?

Something still appears to be off, I'm getting the same error now (even though the connectivity appears to have been corrected) 

image

aaronzedwick:
Can you push the changes that fixed the connectivity?

test/test_grid.py
```diff
@@ -982,3 +983,12 @@ def test_dual_mesh_mpas(self):
 
         # Assert the faces are the same
         nt.assert_equal(dual.face_node_connectivity.values,  mpas_dual.face_node_connectivity.values)
+
+
+    def test_duplicate(self):
+        uxgrid = ux.open_grid(gridfile_geos)
+        uxgrid.merge_duplicate_node_indices()
```
rajeeja, line 990: add an assert to check that `compute_dual` fails or produces incorrect stuff without this operation.

rajeeja (changes requested):
just one minor test and this is good go.

PR #896, opened 2024-08-16
Fix conn when orig_fv is None

rajeeja:


aaronzedwick (approved):

rajeeja:
we can ignore the failures. seen this earlier.

philipc2 (approved):
Thanks for the fix!

PR #899, opened 2024-08-22
Check Grid for Partial Spherical Coverage

### LINKED ISSUE #898: Function to Check For Meshes With Holes

Function that returns whether or not a mesh has holes in it. Something like this:

`Grid.contains_holes()`

Returns True if any holes exists and False if not.

aaronzedwick:
Closes #898 

## Overview

Allows for a grid to be checked for holes.
## Expected Usage

```Python
import uxarray as ux

uxds = ux.open_grid(grid_path)

uxds.contains_holes()
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


rajeeja (approved):
looks good adds closes #141

hongyuchen1030:
```python
def _mesh_contains_holes(edge_face_connectivity):
    """Check if a mesh has holes in it."""

    # If an edge only has one face saddling it than the mesh has holes in it
    return np.any(edge_face_connectivity[:, 1] == INT_FILL_VALUE)
```

This function is basically checking if the `edge_face_connectivity` has a **FILL VALUE**, which means it has fewer nodes than other faces in the mesh files, **it doesn't mean it has a hole.**

aaronzedwick:
> ```python
> def _mesh_contains_holes(edge_face_connectivity):
>     """Check if a mesh has holes in it."""
> 
>     # If an edge only has one face saddling it than the mesh has holes in it
>     return np.any(edge_face_connectivity[:, 1] == INT_FILL_VALUE)
> ```
> 
> This function is basically checking if the `edge_face_connectivity` has a **FILL VALUE**, which means it has fewer nodes than other faces in the mesh files, **it doesn't mean it has a hole.**

Doesn’t this connective hold for each edge what face saddles it? If it is the case that an edge has only one face and the other side of the edge has a fill value, that means it lacks a face there and therefore there is a hole in the mesh? For each edge it can only ever have 2 faces that saddle it, and if there is only one face saddling it, the logic goes that there must be a hole. Is that assumption incorrect?

hongyuchen1030:
> > ```python
> > def _mesh_contains_holes(edge_face_connectivity):
> >     """Check if a mesh has holes in it."""
> > 
> >     # If an edge only has one face saddling it than the mesh has holes in it
> >     return np.any(edge_face_connectivity[:, 1] == INT_FILL_VALUE)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This function is basically checking if the `edge_face_connectivity` has a **FILL VALUE**, which means it has fewer nodes than other faces in the mesh files, **it doesn't mean it has a hole.**
> 
> Doesn’t this connective hold for each edge what face saddles it? If it is the case that an edge has only one face and the other side of the edge has a fill value, that means it lacks a face there and therefore there is a hole in the mesh? For each edge it can only ever have 2 faces that saddle it, and if there is only one face saddling it, the logic goes that there must be a hole. Is that assumption incorrect?

Oh, so you want to see if a given **mesh file** contains holes, instead of a single face has holes inside it? 
So yeah, your implementation is propably correct for this purpose then. 

However, why we need to run this check over the entire grid?

So basically what happens is: you check the ENTIRE grid to see if it has a hole or not, which will still at least needs an O(log_2 N_Edges) time complexity. And then what will happen if we detect there's at least one hole in the face? What's the following procedure? Are you planning to do another query about where the hole is?

aaronzedwick:
> > > ```python
> > > def _mesh_contains_holes(edge_face_connectivity):
> > >     """Check if a mesh has holes in it."""
> > > 
> > >     # If an edge only has one face saddling it than the mesh has holes in it
> > >     return np.any(edge_face_connectivity[:, 1] == INT_FILL_VALUE)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This function is basically checking if the `edge_face_connectivity` has a **FILL VALUE**, which means it has fewer nodes than other faces in the mesh files, **it doesn't mean it has a hole.**
> > 
> > 
> > Doesn’t this connective hold for each edge what face saddles it? If it is the case that an edge has only one face and the other side of the edge has a fill value, that means it lacks a face there and therefore there is a hole in the mesh? For each edge it can only ever have 2 faces that saddle it, and if there is only one face saddling it, the logic goes that there must be a hole. Is that assumption incorrect?
> 
> Oh, so you want to see if a given **mesh file** contains holes, instead of a single face has holes inside it? So yeah, your implementation is propably correct for this purpose then.
> 
> However, why we need to run this check over the entire grid?
> 
> So basically what happens is: you check the ENTIRE grid to see if it has a hole or not, which will still at least needs an O(log_2 N_Edges) time complexity. And then what will happen if we detect there's at least one hole in the face? What's the following procedure? Are you planning to do another query about where the hole is?

Exactly. The specific use case for this is within the dual mesh. If the grid has holes (like say it doesn’t include landmasses) then the dual mesh of a data array would give incorrect results, as some face centers and nodes won’t be used in the dual mesh (imagine if there was a line of squares in a row, no dual mesh could be made from just that and therefore all data stored there would be lost). Also, for bilinear remapping, if the mesh has holes in it, the construction of the dual mesh needs to be handled differently, in accordance to what Paul does in his implementation. So inside the bilinear remapping a call to this function must be made to determine if a simply dual mesh will work, or if a different approach to constructing the dual must be taken. I hope that helps clear things up!

hongyuchen1030:
> > > > ```python
> > > > def _mesh_contains_holes(edge_face_connectivity):
> > > >     """Check if a mesh has holes in it."""
> > > > 
> > > >     # If an edge only has one face saddling it than the mesh has holes in it
> > > >     return np.any(edge_face_connectivity[:, 1] == INT_FILL_VALUE)
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > This function is basically checking if the `edge_face_connectivity` has a **FILL VALUE**, which means it has fewer nodes than other faces in the mesh files, **it doesn't mean it has a hole.**
> > > 
> > > 
> > > Doesn’t this connective hold for each edge what face saddles it? If it is the case that an edge has only one face and the other side of the edge has a fill value, that means it lacks a face there and therefore there is a hole in the mesh? For each edge it can only ever have 2 faces that saddle it, and if there is only one face saddling it, the logic goes that there must be a hole. Is that assumption incorrect?
> > 
> > 
> > Oh, so you want to see if a given **mesh file** contains holes, instead of a single face has holes inside it? So yeah, your implementation is propably correct for this purpose then.
> > However, why we need to run this check over the entire grid?
> > So basically what happens is: you check the ENTIRE grid to see if it has a hole or not, which will still at least needs an O(log_2 N_Edges) time complexity. And then what will happen if we detect there's at least one hole in the face? What's the following procedure? Are you planning to do another query about where the hole is?
> 
> Exactly. The specific use case for this is within the dual mesh. If the grid has holes (like say it doesn’t include landmasses) then the dual mesh of a data array would give incorrect results, as some face centers and nodes won’t be used in the dual mesh (imagine if there was a line of squares in a row, no dual mesh could be made from just that and therefore all data stored there would be lost). Also, for bilinear remapping, if the mesh has holes in it, the construction of the dual mesh needs to be handled differently, in accordance to what Paul does in his implementation. So inside the bilinear remapping a call to this function must be made to determine if a global dual mesh will work, or if a different local approach to constructing the dual must be taken. I hope that helps clear things up!

So yes, my concern was true. After we call this boolean function to check the entire grid and return False, we need to check the entire grid again to find where the holes are missing. So two same traverse for the entire grid, which is very computationally expensive.

Why don't we just return the clean-up meshes or identify the missing faces during the re-mapping process?

During the `zonal_mean` implementation, we encountered a similar problem and it can be solved using the sweep-line algorithm without checking if such grid contains a missing face

philipc2 (commented):
@aaronzedwick 

If all we are interested in knowing is whether there are empty patches with no geometry, such as continents, we should be able to check the total face area of the unit sphere.

If the area is less than 4pi, it means that not the entire surface is covered with geometries.

philipc2:
I'd also avoid referring to this behavior as "holes", as it can be confused with what we've discussed. We are essentially just checking if the entire sphere is covered with our grid. A grid with partial coverage, such as the grids used in ocean models, only cover a fraction of the sphere.

hongyuchen1030:
> I'd also avoid referring to this behavior as "holes", as it can be confused with what we've discussed. We are essentially just checking if the entire sphere is covered with our grid. A grid with partial coverage, such as the grids used in ocean models, only cover a fraction of the sphere.

And partial grid should not impact the re-gridding behavior as well.

aaronzedwick:
> > I'd also avoid referring to this behavior as "holes", as it can be confused with what we've discussed. We are essentially just checking if the entire sphere is covered with our grid. A grid with partial coverage, such as the grids used in ocean models, only cover a fraction of the sphere.
> 
> And partial grid should not impact the re-gridding behavior as well.

https://github.com/ClimateGlobalChange/tempestremap/blob/6f1b7839a919482606caee6ca5ffb114f2241e29/src/LinearRemapFV.cpp#L2996

Paul does a check for it in his code, unless I am misunderstanding what he does there.

philipc2:
> > > I'd also avoid referring to this behavior as "holes", as it can be confused with what we've discussed. We are essentially just checking if the entire sphere is covered with our grid. A grid with partial coverage, such as the grids used in ocean models, only cover a fraction of the sphere.
> > 
> > And partial grid should not impact the re-gridding behavior as well.
> 
> https://github.com/ClimateGlobalChange/tempestremap/blob/6f1b7839a919482606caee6ca5ffb114f2241e29/src/LinearRemapFV.cpp#L2996
> 
> Paul does a check for it in his code, unless I am misunderstanding what he does there. Unless he is checking for holes inside a face but looking at his code for `DoesMeshContain` holes it seems he does a similar thing as me. Also, @philipc2 the face area idea would work but seems to me that it would take much longer as you need to calculate the area for each face.

That's a fair point. However if we already have the face area, it would be an O(n) operation.

hongyuchen1030:
> > > I'd also avoid referring to this behavior as "holes", as it can be confused with what we've discussed. We are essentially just checking if the entire sphere is covered with our grid. A grid with partial coverage, such as the grids used in ocean models, only cover a fraction of the sphere.
> > 
> > 
> > And partial grid should not impact the re-gridding behavior as well.
> 
> https://github.com/ClimateGlobalChange/tempestremap/blob/6f1b7839a919482606caee6ca5ffb114f2241e29/src/LinearRemapFV.cpp#L2996
> 
> Paul does a check for it in his code, unless I am misunderstanding what he does there. Unless he is checking for holes inside a face but looking at his code for `DoesMeshContain` holes it seems he does a similar thing as me. Also, @philipc2 the face area idea would work but seems to me that it would take much longer as you need to calculate the area for each face.

I understand that you are trying to follow what the tempest extreme does here. My point is: 
1. The tempest extreme is written in C/C++ while Uxarray is written in python. Also the API and data structure are different as well.
2. Can you write down the sudo codes for the entire re-mapping procedure so we can look into what's the ideal way to get it done in the Uxarray.

These two points are very important and introduce a recent important discussion for our developing process.

uxarray/grid/validation.py
```diff
@@ -65,3 +67,11 @@ def _check_area(self):
...
         print("-No face area is close to zero.")
         return True
+
+
+@njit(cache=True)
+def _mesh_contains_holes(edge_face_connectivity):
+    """Check if a mesh has holes in it."""
+
+    # If an edge only has one face saddling it than the mesh has holes in it
+    return np.any(edge_face_connectivity[:, 1] == INT_FILL_VALUE)
```
philipc2, line 77: I don't see an issue with this check, as the reasoning that @aaronzedwick explained is sound. If there exists any edge that does not have two neighboring faces, then there must exist some "empty" area adjacent to that edge. Illustration below for reference: the edges marked in red each neighbor an empty region, which means that we can't construct an edge between the centers of the two faces indicated in the `edge_face_connectivity`

image

This causes issues with the Dual Mesh construction, since node-centered data variables become face-centered. However, there will be no faces constructed centered around these values. Given our interpretation of the UGRID conventions, we have no way to construct a face for this data, meaning that the data is essentially "lost" from the dual mesh construction process. 

Also, I will suggest caching the result once we've run it, so that we don't need to run the repeated searches over the `edge_face_connectivity`. Something as simple as a `Grid._partial_coverage` attribute would suffice.

hongyuchen1030: > I don't see an issue with this check, as the reasoning that @aaronzedwick explained is sound. If there exists any edge that does not have two neighboring faces, then there must exist some "empty" area adjacent to that edge. Illustration below for reference: the edges marked in red each neighbor an empty region, which means that we can't construct an edge between the centers of the two faces indicated in the `edge_face_connectivity`

We have already gone through this part in the previous discussion above. I understand what you mean by "missing a face" here. My current concern is: after we have examined the entire grid to check if there's a partial coverage, what's the next step to deal with the missing faces? Do we need to rerun a similar check to identify which faces don't have neighbors? So basically two O(N) operations?


>Also, I will suggest caching the result once we've run it, so that we don't need to run the repeated searches over the `edge_face_connectivity`. Something as simple as a `Grid._partial_coverage ` attribute would suffice.

This is a good point. As we agreed in the long conversation from the `zonal_mean` PR, we need to provide the sudo codes for the algorithm. So that we can better adapt it to the UXarray API and the python vectorization styles.

hongyuchen1030:
And just some information from Computational Geometry, we normally assign the uncovered area a face index on the spherical surface as well. And when we said is neighboring with this face index, we know it is next by the uncovered area. Hope this point will help.

uxarray/grid/validation.py
philipc2: > My current concern is: after we have examined the entire grid to check if there's a partial coverage, what's the next step to deal with the missing faces? Do we need to rerun a similar check to identify which faces don't have neighbors? So basically two O(N) operations?

This is a good point that I overlooked. We should be keeping track of which indices are considered to border an empty area, which can be achieved by removing the `np.any` call
```Python
edge_face_connectivity[:, 1] == INT_FILL_VALUE
```

Once we have this, we need to identify which nodes (indicated in black) will need to be removed from the new grid that is constructure from the Dual Grid. 

image

The number of faces in the new `Grid` should equal the original number of nodes (`n_node`) minus the number of nodes that share one or more of the "invalid" edges shown above.

philipc2:
> And just some information from Computational Geometry, we normally assign the uncovered area a face index on the spherical surface as well. And when we said is neighboring with this face index, we know it is next by the uncovered area. Hope this point will help.

This does make sense, however the meshes that we work with that have a partial spherical coverage (i.e. ocean meshes from climate models) don't assign an index to the uncovered area. 

image

> And when we said is neighboring with this face index, we know it is next by the uncovered area.

This is roughly what we are trying to achieve with the checks added by @aaronzedwick here. Since we don't have any explicit definitions of the geometry of the uncovered areas, the only way we are able to determine if a cell neighbors an empty region is by performing a search over the `edge_face_connectivity`. An index of `INT_FILL_VALUE` in this connectivity array indicates that a given edge is saddled between a valid face, and an empty region.

uxarray/grid/validation.py
hongyuchen1030:
```python
edge_face_connectivity[:, 1] == INT_FILL_VALUE
```

This sounds good. If we need to check if there's a partial coverage, don't just return a boolean value, but instead directly return where is the partial coverage. And if the return is `None`, then we know is all covered.


> The number of faces in the new Grid should equal the original number of nodes (n_node) minus the number of nodes that share one or more of the "invalid" edges shown above.

It sounds good to me though I haven't thought deep into this yet. But I will keep you updated if anything other things comes into my head.

aaronzedwick: Sounds perfect to me, @philipc2 @hongyuchen1030 thanks for the suggestions on this one! Let me know if you have anymore thoughts and I will get the changes mentioned here implemented!

aaronzedwick:
> > And just some information from Computational Geometry, we normally assign the uncovered area a face index on the spherical surface as well. And when we said is neighboring with this face index, we know it is next by the uncovered area. Hope this point will help.
> 
> This does make sense, however the meshes that we work with that have a partial spherical coverage (i.e. ocean meshes from climate models) don't assign an index to the uncovered area.
> 
> image
> 
> > And when we said is neighboring with this face index, we know it is next by the uncovered area.
> 
> This is roughly what we are trying to achieve with the checks added by @aaronzedwick here. Since we don't have any explicit definitions of the geometry of the uncovered areas, the only way we are able to determine if a cell neighbors an empty region is by performing a search over the `edge_face_connectivity`. An index of `INT_FILL_VALUE` in this connectivity array indicates that a given edge is saddled between a valid face, and an empty region.

Oddly enough, even without the current checks for the data, the dual mesh construction does work with a partially covered data array. I am not sure exactly how the data array is constructed inside UXarray, but I was confused by this.

image

hongyuchen1030:
> > > And just some information from Computational Geometry, we normally assign the uncovered area a face index on the spherical surface as well. And when we said is neighboring with this face index, we know it is next by the uncovered area. Hope this point will help.
> > 
> > 
> > This does make sense, however the meshes that we work with that have a partial spherical coverage (i.e. ocean meshes from climate models) don't assign an index to the uncovered area.
> > image
> > > And when we said is neighboring with this face index, we know it is next by the uncovered area.
> > 
> > 
> > This is roughly what we are trying to achieve with the checks added by @aaronzedwick here. Since we don't have any explicit definitions of the geometry of the uncovered areas, the only way we are able to determine if a cell neighbors an empty region is by performing a search over the `edge_face_connectivity`. An index of `INT_FILL_VALUE` in this connectivity array indicates that a given edge is saddled between a valid face, and an empty region.
> 
> Oddly enough, even without the current checks for the data, the dual mesh construction does work with a partially covered data array. I am not sure exactly how the data array is constructed inside UXarray, but I was confused by this.
> 
> image

Is it possible that it uses the `FILL_VALUE` as the face center value? Have you looked into the output value?

aaronzedwick:
@hongyuchen1030 It doesn't seem to, no, the values in the data array are all actually data points without any fill values, and the number of values matches the number of faces.

philipc2:
> @hongyuchen1030 It doesn't seem to, no, the values in the data array are all actual data points without any fill values, and the number of values matches the number of faces.

Seconding this. `INT_FILL_VALUE` is `int64`, while the type of the data variable isn't constant (however it is typically a `float`)

uxarray/grid/grid.py
```diff
@@ -904,6 +905,18 @@ def face_jacobian(self):
...
+        """Check the grid for faces that are next a hole inside a partial grid.
+
+        Returns
+        """
+        if "index_holes" not in self._ds:
+            self._ds["index_holes"] = _index_faces_on_partial_grid(
+                self.edge_face_connectivity.values
+            )
+        return self._ds["index_holes"]
+
```
philipc2, line 919: I like this implementation a lot, but I think we can maybe use a better name for this. For reference, we have the following which is the indices of all the faces that cross the antemeridian. 

https://github.com/UXARRAY/uxarray/blob/778e05e27af90a27e4ea81d84e137a33b6064a7a/uxarray/grid/grid.py#L846-L851

I think `hole_edge_indices` would be appropriate. We should also explicitly define a hole as a region of the grid that is not covered by any geometry. 

I'd also suggestion modifying the docstring to describe what variable is stored within the property, like we've done with the others.

philipc2 (changes requested):

uxarray/grid/grid.py
aaronzedwick: Where do you think an appropriate place to construct this would be? I don't think validation is a good place for it anymore. Connectivity doesn't really make that much sense either. Geometry maybe?

rajeeja: > Where do you think an appropriate place to construct this would be? I don't think validation is a good place for it anymore. Connectivity doesn't really make that much sense either. Geometry maybe?

+1 for geometry

philipc2 (commented):
May we include a basic benchmark in the `mpas_ocean` file for computing `Grid.hole_edge_indices`

aaronzedwick:
> May we include a basic benchmark in the `mpas_ocean` file for computing `Grid.hole_edge_indices`

Yeah, that's a good idea!

uxarray/grid/grid.py
```diff
@@ -904,6 +905,16 @@ def face_jacobian(self):
             _ = self.face_areas
         return self._face_jacobian
 
+    @property
+    def hole_edge_indices(self):
```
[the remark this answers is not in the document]
aaronzedwick: What would the dimensions of this be?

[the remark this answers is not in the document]
rajeeja: > What would the dimensions of this be?

how about just using `n_edge` and call it `MAX_..`? since we wouldn't know the dimension beforehand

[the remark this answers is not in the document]
philipc2: > What would the dimensions of this be?

The dimension should always be `n_edge`

benchmarks/mpas_ocean.py
```diff
@@ -3,6 +3,7 @@
 from pathlib import Path
 
 import uxarray as ux
+from uxarray.grid.geometry import _construct_hole_edge_indices
```
philipc2, line 6: We need to add this import within the `HoleEdgeIndices` scope, since it's causing our other benchmarks to fail.

See the discussion in #811

philipc2 (changes requested):

philipc2 (approved):
The implementation here looks great and the execution time is quick! Thanks for putting this together.

Issue #919, opened 2024-08-29
Missing version info


kafitzgerald:
### Version

latest version from conda-forge

### How did you install UXarray?

Conda

### What happened?

I was expecting `ux.__version__` to return the current version info for the package.  Instead, I get `'0.0.0'`.

It looks like only the Conda version is impacted (works fine when installing via pip).

### What did you expect to happen?

See above.  

### Can you provide a MCVE to repoduce the bug?

```python
import uxarray as ux
ux.__version__
```

philipc2:
Hi @kafitzgerald 

Thanks for catching this! Since it only appears to be an issue with `conda` installs, I wonder if something might be configured incorrectly in our [feedstock repo](https://github.com/conda-forge/uxarray-feedstock).

Comparing to Xarray's feedstock, this immediately jumped out as a difference:

https://github.com/conda-forge/xarray-feedstock/blob/25fb1d1c91dc951102425b2293d1120192ba9677/recipe/meta.yaml#L1-L2

Maybe the lack of `{% set name = "uxarray" %}` leads to `importlib.metadata` producing an incorrect `0.0.0` version is some form of fallback version?

kafitzgerald:
I suspect it might be that you're missing setuptools_scm from your host requirements in the meta.yml.

philipc2:
After those changes, we're now getting `999` as our version number. Any thoughts @kafitzgerald ?

philipc2:
#921 has fixed the issue!

![image](https://github.com/user-attachments/assets/043eaadc-bfe8-427b-8702-7ecb2d15d4ef)

PR #921, opened 2024-08-29
Add files to support git archives

kafitzgerald:
Relates to #919

## Overview
Adds files to support using setuptools-scm on git archives to hopefully address the Conda versioning snafu.  These should prompt some metadata to be populated when archives are created allowing for things like the version info to be available when building from an archive. 

Related docs here: https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives

The contents differ slightly from what xarray does based on the current guidance in the docs.

## PR Checklist

**General**
- [ ] Add appropriate labels


philipc2 (approved):
Thanks for this!

PR #941, opened 2024-09-09
Apply function to points within circular neighborhood  

### LINKED ISSUE #930: Apply a neighborhood filter with radius r to all elements of UxDataArray

Apply a neighborhood filter within a circular radius r (in degrees) to a UxDataset or UxDataArray.

Overview

This is kind of like uxarray.UxDataArray.inverse_distance_weighted_remap , but the neighborhood is defined by distance, not a number of nearest neighbors. This is ideally suited for a variable resolution mesh, in which a constant of neighbors doesn't have a constant-sized neighborhood. 

This function would share the same logic as uxarray.UxDataArray.subset.bounding_circle to select grid elements in a circular neighborhood, but apply it to all elements in grid, not just one center_coordinate.

Ideally, the filter function func would be a user-defined, but at a minimum would be capable of calculating a neighborhood-maximum. Other possibilities are np.mean, min, np.median and np.percentile.

Expected Usage

import numpy as np

import uxarray

grid_path = "/glade/campaign/mmm/wmr/weiwang/cps/irma3/2020/tk707_conus/init.nc"
data_path = "/glade/campaign/mmm/wmr/weiwang/cps/irma3/mp6/tk707/diag.2017-09-07_09.00.00.nc"
uxds = uxarray.open_mfdataset(
    grid_path,
    data_path
)

# Trim domain
lon_bounds = (-74, -64)
lat_bounds = (18, 24)
uxda = uxds["refl10cm_max"].isel(Time=0).subset.bounding_box(lon_bounds, lat_bounds)
uxda

# this is how you use this function to get the maximum value within a 0.25-deg neighborhood of each element.
uxda_smoothed_mean = uxda.apply_neighorhood_filter(func=max, r=0.25)

ahijevyc:
Apply a neighborhood filter within a circular radius r to a UxDataset or UxDataArray.

Closes #930

## Overview

This is kind of like `uxarray.UxDataArray.inverse_distance_weighted_remap` , but the neighborhood is defined by distance, not a number of nearest neighbors. This is ideally suited for a variable resolution mesh, in which a constant of neighbors doesn't have a constant sized neighborhood. Another difference is that this neighborhood filter does not weight data by inverse distance.

Just like `uxarray.UxDataArray.subset.bounding_circle` this function uses `ball_tree.query_radius` to select grid elements in a circular neighborhood, but this function finds the neighborhood for all elements in grid, not just one center_coordinate.

The filter function `func` may be a user-defined function, but uses `np.mean` by default. It could be `min`, `max`, `np.median`. It can even use functions that require additional arguments, like `np.percentile` if you supply the argument(s) with `functools.partial` (see below)

## Expected Usage

```Python
from functools import partial
import numpy as np
import uxarray

grid_path = "/glade/campaign/mmm/wmr/weiwang/cps/irma3/2020/tk707_conus/init.nc"
data_path = "/glade/campaign/mmm/wmr/weiwang/cps/irma3/mp6/tk707/diag.2017-09-07_09.00.00.nc"
uxds = uxarray.open_mfdataset(
    grid_path,
    data_path
)

# Trim domain
lon_bounds = (-74, -64)
lat_bounds = (18, 24)
uxda = uxds["refl10cm_max"].isel(Time=0).subset.bounding_box(lon_bounds, lat_bounds)

# this is how you use this function to smooth with 0.25-deg filter.
uxda_mean = uxda.neighborhood_filter(func=np.mean, r=0.25)


# this is another way to use this function with np.percentile
uxda_max = uxda.neighborhood_filter(func=partial(np.percentile, q=90), r=0.25)

(uxda.plot.rasterize() + uxda_mean.plot.rasterize() + uxda_max.plot.rasterize()).cols(1)
```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**
- [x] Internal functions have a preceding underscore (`_`); `_neighborhood_filter` is internal to `uxarray/grid/neighbors.py`
- [x] User functions added to `docs/api.rst` (the split user/internal api files no longer exist)

**Examples**


uxarray/core/dataarray.py
```diff
@@ -1104,3 +1103,116 @@ def _slice_from_grid(self, sliced_grid):
...
+        elif self._edge_centered():
+            data_mapping = "edge centers"
+        else:
+            raise ValueError(
+                "Data_mapping is not face, node, or edge. Could not define data_mapping."
+            )
+
+        # reconstruct because the cached tree could be built from
+        # face centers, edge centers or nodes.
+        tree = self.uxgrid.get_ball_tree(coordinates=data_mapping, reconstruct=True)
```
philipc2, line 1139: @aaronzedwick 

We should probably fix this logic in `get_ball_tree()`, since we shouldn't need to manually set `reconstruct=False` 

```Python
        if self._ball_tree is None or reconstruct:
            self._ball_tree = BallTree(
                self,
                coordinates=coordinates,
                distance_metric=distance_metric,
                coordinate_system=coordinate_system,
                reconstruct=reconstruct,
            )
        else:
            if coordinates != self._ball_tree._coordinates:
                self._ball_tree.coordinates = coordinates
```

The `coordinates != self._ball_tree._coordinates` check should be included in the first `if`

ahijevyc: That makes sense. So, move the coordinates check to the if-clause like this?

```
                if (
                    self._ball_tree is None
                    or coordinates != self._ball_tree._coordinates
                    or reconstruct
                ):

                    self._ball_tree = BallTree(
                        self,
                        coordinates=coordinates,
                        distance_metric=distance_metric,
                        coordinate_system=coordinate_system,
                        reconstruct=reconstruct,
                    )

```

What if the coordinate_system is different? Should that also require tree reconstruction?

ahijevyc: Whatever logic is fixed in `Grid.get_ball_tree` should be applied to [`Grid.get_kdtree`](https://github.com/UXARRAY/uxarray/blob/ff9ec76e5ce3422a8af7005e843fcec0a75155d5/uxarray/grid/grid.py#L1238) also.

uxarray/core/dataarray.py
```diff
@@ -1104,3 +1103,116 @@ def _slice_from_grid(self, sliced_grid):
...
+                    self.uxgrid.edge_y.values,
+                    self.uxgrid.edge_z.values,
+                )
+            else:
+                raise ValueError(
+                    f"Invalid data_mapping. Expected 'nodes', 'edge centers', or 'face centers', "
+                    f"but received: {data_mapping}"
+                )
+
+            dest_coords = np.c_[x, y, z]
```
philipc2, line 1192: Would [`np.vstack`](https://numpy.org/doc/stable/reference/generated/numpy.column_stack.html#numpy.column_stack) be more appropriate here?

uxarray/core/dataarray.py
```diff
@@ -1104,3 +1103,116 @@ def _slice_from_grid(self, sliced_grid):
             dims=self.dims,
             attrs=self.attrs,
         )
+
+    def neighborhood_filter(
```
philipc2, line 1107: This implementation looks great! May we move the bulk of the logic into the `uxarray.grid.neighbors` module and call that helper from here?

We can keep the data-mapping checks here, and anything related to constructing and returining the final data array but the bulk of the computations would go inside a helper in the module mentioned above.

uxarray/core/dataset.py
```diff
@@ -338,6 +339,40 @@ def to_array(self) -> UxDataArray:
...
+        r: float = 1.0,
+    ):
+        """Neighborhood function implementation for ``UxDataset``.
+        Parameters
+        ---------
+        func : Callable = np.mean
+            Apply this function to neighborhood
+        r : float, default=1.
+            Radius of neighborhood
+        """
```
philipc2, line 354: Include the same parameter description as the `UxDataArray` implementation.

> r : float, default=1.
>             Radius of neighborhood. For spherical coordinates, the radius is in units of degrees,
>             and for cartesian coordinates, the radius is in meters.

philipc2 (dismissed):
A few initial comments:

uxarray/core/dataarray.py
ahijevyc: I changed to np.vstack

uxarray/core/dataarray.py
ahijevyc: I wouldn't know how to implement that, but I am happy to defer to you.

uxarray/core/dataarray.py
ahijevyc: checking coordinate system also (coordinate_system is not a hidden variable of _ball_tree; it has no underscore):
```Python
                if (
                    self._ball_tree is None
                    or coordinates != self._ball_tree._coordinates
                    or coordinate_system != self._ball_tree.coordinate_system
                    or reconstruct
                ):

                    self._ball_tree = BallTree(
                        self,
                        coordinates=coordinates,
                        distance_metric=distance_metric,
                        coordinate_system=coordinate_system,
                        reconstruct=reconstruct,
                    )
```

philipc2:
HI @ahijevyc 

Apologies for not getting to this PR earlier. 

Looking at the implementation here, it looks great. It does however bring to light a possible need for us to consider a better, more streamlined, approach to handling these types of groupings and then applying some function on the result.

I mention this because of our [Topological Aggregations](https://uxarray.readthedocs.io/en/latest/user-guide/topological-aggregations.html). For this family of functions, we have distinct methods (i.e. `topological_mean()`), which looking back at doesn't seem like the preferred approach, especially if we plan to implement groupings like the neighborhood one and perhaps other spatial ones.

Very generally speaking, these functions essentially: 
1. Group unstructured grid elements based on some condition/algorithm. Here we use the KD/BallTree to determine the candidate elements, while in the topological aggregations we use the connectivity information
2. Apply some function to the grouping (i.e. `mean()`)
3. Store the results back on the unstructured grid element (node, edge, or face)


I wonder if this would be a good opportunity to extend the inherited `.groupby()` method from Xarray to support these spatial groupings. 

I'm not sure of calling these approaches "kernels" is appropriate, but for the sake of this example, we could provide spatial kernels the user could pass into `groupby()` and then perform aggregations directly on the result. This feels much more in line with Xarray's design philosophy. 

```Python
# radial neighborhood of r=0.25
uxds['t2m'].groupby(kernel=ux.BoundingCircle(r=0.25)).mean()

# 2 deg by 2 deg bounding box 
uxds['t2m'].groupby(kernel=ux.BoundingBox(dlon=2, dlat=2))

# group the nodes that surround each face and find the maximum
uxds['node_centered_var'].groupby(kernel=ux.FaceNode()).max()

# this is equivalent to the following in the current release
uxds['node_centered_var'].topological_max(destination='face')
```

I'll ping @aaronzedwick and @erogluorhan for their thoughts on this. I personally really like the design above and think that it aligns well with the overall design.

aaronzedwick:
> HI @ahijevyc
> 
> Apologies for not getting to this PR earlier.
> 
> Looking at the implementation here, it looks great. It does however bring to light a possible need for us to consider a better, more streamlined, approach to handling these types of groupings and then applying some function on the result.
> 
> I mention this because of our [Topological Aggregations](https://uxarray.readthedocs.io/en/latest/user-guide/topological-aggregations.html). For this family of functions, we have distinct methods (i.e. `topological_mean()`), which looking back at doesn't seem like the preferred approach, especially if we plan to implement groupings like the neighborhood one and perhaps other spatial ones.
> 
> Very generally speaking, these functions essentially:
> 
> 1. Group unstructured grid elements based on some condition/algorithm. Here we use the KD/BallTree to determine the candidate elements, while in the topological aggregations we use the connectivity information
> 2. Apply some function to the grouping (i.e. `mean()`)
> 3. Store the results back on the unstructured grid element (node, edge, or face)
> 
> I wonder if this would be a good opportunity to extend the inherited `.groupby()` method from Xarray to support these spatial groupings.
> 
> I'm not sure of calling these approaches "kernels" is appropriate, but for the sake of this example, we could provide spatial kernels the user could pass into `groupby()` and then perform aggregations directly on the result. This feels much more in line with Xarray's design philosophy.
> 
> ```python
> # radial neighborhood of r=0.25
> uxds['t2m'].groupby(kernel=ux.BoundingCircle(r=0.25)).mean()
> 
> # 2 deg by 2 deg bounding box 
> uxds['t2m'].groupby(kernel=ux.BoundingBox(dlon=2, dlat=2))
> 
> # group the nodes that surround each face and find the maximum
> uxds['node_centered_var'].groupby(kernel=ux.FaceNode()).max()
> 
> # this is equivalent to the following in the current release
> uxds['node_centered_var'].topological_max(destination='face')
> ```
> 
> I'll ping @aaronzedwick and @erogluorhan for their thoughts on this. I personally really like the design above and think that it aligns well with the overall design.

That is interesting. So, are you suggesting changing the way we do aggregations entirely? So this would affect the reduction PR I am working on then. Perhaps that PR could implement that change if you wish. I am fine with this, it sounds like it would be intuitive.

philipc2:
> You suggesting changing the way we do aggregations entirely?

We should be changing the API for it, since the current one is not sustainable if we plan to implement many times of spatial grouping methods. 

The underlying functionality would remain mostly unchanged. 

> Perhaps this PR could implement that change if you wish.

If we do decide to do this, I'll open up a separate PR starting with the topological aggregations.

aaronzedwick:
> > You suggesting changing the way we do aggregations entirely?
> 
> We should be changing the API for it, since the current one is not sustainable if we plan to implement many times of spatial grouping methods.
> 
> The underlying functionality would remain mostly unchanged.
> 
> > Perhaps this PR could implement that change if you wish.
> 
> If we do decide to do this, I'll open up a separate PR starting with the topological aggregations.

So would the reductions PR be obsolete?

philipc2:
> > > You suggesting changing the way we do aggregations entirely?
> > 
> > 
> > We should be changing the API for it, since the current one is not sustainable if we plan to implement many times of spatial grouping methods.
> > The underlying functionality would remain mostly unchanged.
> > > Perhaps this PR could implement that change if you wish.
> > 
> > 
> > If we do decide to do this, I'll open up a separate PR starting with the topological aggregations.
> 
> So would the reductions PR be obsolete?

No. The underlying implementation would remain the same, since we would still need those implemented. 

This would just provide a different interface for it, with a more "Xarray-like" interface.

aaronzedwick:
> > > > You suggesting changing the way we do aggregations entirely?
> > > 
> > > 
> > > We should be changing the API for it, since the current one is not sustainable if we plan to implement many times of spatial grouping methods.
> > > The underlying functionality would remain mostly unchanged.
> > > > Perhaps this PR could implement that change if you wish.
> > > 
> > > 
> > > If we do decide to do this, I'll open up a separate PR starting with the topological aggregations.
> > 
> > 
> > So would the reductions PR be obsolete?
> 
> No. The underlying implementation would remain the same, since we would still need those implemented.
> 
> This would just provide a different interface for it, with a more "Xarray-like" interface.

Ah, okay, I see. That makes sense, thanks for the clarification!

erogluorhan:
pre-commit.ci autofix

rajeeja:
@ahijevyc are you still looking into this? Let me know if not, I can takeover..

ahijevyc:
> @ahijevyc are you still looking into this? Let me know if not, I can takeover..
I appreciate that. No I am not still looking at this. It would take a while for me to spin up again.

erogluorhan:
@ahijevyc your last message was a typo I think as it doesn't show any response. Please let us know if you are still looking into it. Otherwise, let @rajeeja take this one since this is an important functionality.

ahijevyc:
@erogluorhan @rajeeja I forgot that I assigned myself to this! Please take it over, Rajeev, and thank you for that.

rajeeja:
@ahijevyc — since GitHub blocks adding the PR author as a reviewer, could you take a look when you get a chance? Key changes since your last review round:

**Bug fixes:**
- `np.empty` → `np.full(…, np.nan)` in `neighbors.py` — empty neighborhoods now correctly return NaN instead of arbitrary memory
- Auto-transpose for non-last grid dim — `(time, n_face)` arrays now work without the caller having to manually transpose
- `raise ValueError` instead of bare `assert` for dimension mismatch (cleaner user-facing error)
- `_copy(data=…, deep=False)` in the filter path — skips a redundant deep copy of the source data, shares the same uxgrid reference

**API compliance / xarray parity:**
- Investigated whether `neighborhood_filter` should use `groupby`: conclusion is **no**. Neighborhoods are *overlapping* spatial balls — a single element belongs to many neighborhoods. xarray's `groupby` requires *non-overlapping* partitions. The direct-method pattern is consistent with `azimuthal_mean` and `zonal_mean`.
- Philip has been removed as a reviewer; @erogluorhan has been added.

**Docs:**
- Expanded user-guide notebook (30 cells): covers face-, node-, and edge-centered data; multi-dim (time × space); xarray chaining (`.where()`, `.groupby()`); radius sweep comparison; `ux.tutorial` datasets throughout (no hardcoded file paths)
- Added `Examples` and `See Also` sections to both `UxDataArray.neighborhood_filter` and `UxDataset.neighborhood_filter` docstrings

Happy to discuss further or adjust anything before merge!

docs/user-guide/neighborhood-filter.ipynb
ahijevyc, line 1: > If the search radius is so small that *no* neighbor is found for a given element,
> the result for that element is `NaN` rather than an uninitialized garbage value.
> In practice this only happens for extremely small radii on very coarse grids; the
> filter always includes the queried element itself, so `r = 0` is safe.
> 
I don't understand this comment. Even with a very coarse grid and an extremely small radius, I don't see how a NaN could be returned. After all, the filter always includes the queried element itself.

ahijevyc (commented):
Thanks for carrying this across the finish line! I just had a comment or two about the assertion that NaN could be returned for a coarse grid and small radius. I don't think this could occur in practice, as the filter should at least return the values at the original grid points even if the radius is zero. And `grid.neighbors.BallTree.query_radius` already makes sure the radius is not negative.

uxarray/grid/neighbors.py
```diff
@@ -1129,3 +1131,111 @@ def _construct_edge_face_distances(face_lon, face_lat, edge_faces):
...
+    tree = grid.get_ball_tree(coordinates=data_mapping)
+
+    coordinate_system = tree.coordinate_system
+
+    dest_coords = _get_element_coords(grid, data_mapping, coordinate_system)
+
+    neighbor_indices = tree.query_radius(dest_coords, r=r)
+
+    # Initialize with NaN so that any element whose neighborhood is empty
+    # (e.g. an isolated point queried with a very small radius) yields NaN
```
ahijevyc: Initializing with NaN is a good defensive idea. But I don't think the reasoning about an isolated point queried with a very small radius is strictly correct. The filter should return the original values even if the radius is zero.

ahijevyc (commented):
looks great

docs/user-guide/neighborhood-filter.ipynb
rajeeja: Thanks, good find — fixed.

You're right: `r=0`, `1e-12`, and `0.001` all return the input unchanged with zero NaNs, and `query_radius` rejects `r < 0`, so an empty neighborhood isn't reachable. Renamed the section to "Radius Edge Cases" and rewrote it to say that, with the NaN allocation described as defensive-only rather than expected behavior. The demo cell now asserts `np.allclose(filtered_r0.values, da_coarse.values)` instead of printing a NaN count.

uxarray/grid/neighbors.py
rajeeja: Thanks, good find — fixed. You're right that the filter always returns the original values even at `r = 0`, so the empty-neighborhood case isn't reachable; I kept `np.full(NaN)` but reworded the comment to say it's purely defensive against `np.empty`, and replaced `test_empty_neighborhood_returns_nan` with a test asserting what actually holds (`r=0` → unchanged, `r=360` → global mean, `r<0` → raises).

uxarray/core/dataarray.py
```diff
@@ -2179,6 +2181,108 @@ def get_dual(self):
 
         return uxda
 
+    def neighborhood_filter(
+        self,
+        func: Callable = np.mean,
+        r: float = 1.0,
+    ) -> UxDataArray:
+        """Apply a neighborhood filter, replacing the value at each grid
```
erogluorhan: Might use a little rephrasing: "replace" might be misleading here since function returns a new data array and actually doesn't change the original data?

uxarray/grid/neighbors.py
```diff
@@ -1129,3 +1131,142 @@ def _construct_edge_face_distances(face_lon, face_lat, edge_faces):
...
+    # Allocate with NaN rather than ``np.empty`` purely as a defensive measure:
+    # if a neighborhood were ever empty, the result would be an obvious NaN
+    # instead of uninitialized garbage memory. In practice this cannot happen,
+    # since ``query_radius`` rejects negative ``r`` and every element is its own
+    # neighbor at distance 0, so even ``r = 0`` returns the original values.
+    destination_data = np.full(data.shape, np.nan)
+
+    # Apply func along the last (grid) axis only, so any extra leading
+    # dimensions (e.g. time) are preserved rather than being collapsed.
+    for i, idx in enumerate(neighbor_indices):
```
erogluorhan: This will iterate through every single geometric element, e.g. faces, right? It looks very costly for km-scale data. Was there any consideration of vectorization for this?

cc: @cmdupuis3 for your thoughts

uxarray/grid/neighbors.py
```diff
@@ -1129,3 +1131,142 @@ def _construct_edge_face_distances(face_lon, face_lat, edge_faces):
...
+        if len(idx):
+            try:
+                destination_data[..., i] = func(data[..., idx], axis=-1)
```
erogluorhan: From what I can tell on this line, every single iteration will trigger rebuilding and executing the dask task graph from scratch because this line returns a lazy dask scalar when data is dask-backed and assigning that into a numpy array index.

I believe this could be avoided if `uxdataarray.py/neighborhood_filter` did an upfront `.compute()` for data (and actually that function already claims "lazy (dask-backed) data is computed eagerly and the result is always NumPy-backed").

cc: @cmdupuis3 you dealt with bundled `.compute()`s through xarray API in #1560, it might not directly related to here,  but you may still have some thoughts.

erogluorhan (commented):
Overall this looks great, but I have a couple concerns in the low-level implementation for which @cmdupuis3 could be helpful as well (added as a reviewer). See below

uxarray/core/dataarray.py
```diff
@@ -2179,6 +2181,108 @@ def get_dual(self):
...
+
+        # Ensure the grid dimension is the last axis, transposing if necessary.
+        # This mirrors the behaviour of UxDataset.neighborhood_filter so that
+        # calling the method directly on a (time, n_face) UxDataArray works.
+        needs_transpose = self.dims[-1] != grid_dim
+        uxda_work = self.transpose(..., grid_dim) if needs_transpose else self
+
+        destination_data = _neighborhood_filter(
+            self.uxgrid, uxda_work.data, data_mapping, func=func, r=r
+        )
```
cmdupuis3: This seems like a good spot for a ufunc, since you're already passing functions around

uxarray/grid/neighbors.py
cmdupuis3: Looks like it. Just at a glance, this block seems like it's basically a ufunc implementation.

uxarray/grid/neighbors.py
cmdupuis3: Something you could do is make a boolean mask out of `neighbor_indices` as keys, then mask it to `neighbor_masked`, then you could delete the whole conditional/try/except block and iterate over only the valid entries. If you really need the full-scale `destination_data`, you can use the mask to reshape the output.

cmdupuis3:
I'm working on a chunk-friendly optimized version based on gufuncs, I'll add my own PR off this one and y'all can merge it. The one caveat is that we'd have to hand-enumerate the specific reductions we want to be able to run optimally, I think otherwise we can offer a fallback option but the performance would suffer. I don't think that would happen that often in practice though.

rajeeja:
> I'm working on a chunk-friendly optimized version based on gufuncs, I'll add my own PR off this one and y'all can merge it. The one caveat is that we'd have to hand-enumerate the specific reductions we want to be able to run optimally, I think otherwise we can offer a fallback option but the performance would suffer. I don't think that would happen that often in practice though.

Makes sense, thanks for finding this out, that code would lose out-of-core/parallel benefits; most of the times this would be one of the steps in the bigger pipeline and if others are nice dask ops that would be a good win. I'll keep an eye on the followup to this.

cmdupuis3:
@ahijevyc @rajeeja I have a draft PR sitting in the fork, I'll keep working on it there. I'm not totally satisfied with the level of complexity. Using a functional approach inside vectorized gufuncs makes it hard to get performance and good readability at once.

cmdupuis3:
pre-commit.ci autofix

cmdupuis3:
I have some more changes to consider on my `cmd/941` branch. The gist is that I wanted to get away from passing numpy functions to the neighborhood filter, because in order to get the vectorized kernels working, you'd basically have to have a dictionary of numpy functions to vectorized kernels, and the API would be sort of a lie.

Instead, my API proposal is that we have all the named kernels be methods. So, we can call the vectorized kernels by name without mystifying what's actually running, and have `nb.reduce(func)` be the catch-all for external kernels.

This has the added advantage that the kernels are now separable from the neighborhood construction, so you can store a neighborhood and call multiple kernels on it rather than constructing a new neighborhood each time.

On the other hand, it raises the possibility of having multiple neighborhoodsesesssses, so I renamed them to be singular as objects.

rajeeja:
> I have some more changes to consider on my `cmd/941` branch. The gist is that I wanted to get away from passing numpy functions to the neighborhood filter, because in order to get the vectorized kernels working, you'd basically have to have a dictionary of numpy functions to vectorized kernels, and the API would be sort of a lie.
> 
> Instead, my API proposal is that we have all the named kernels be methods. So, we can call the vectorized kernels by name without mystifying what's actually running, and have `nb.reduce(func)` be the catch-all for external kernels.
> 
> This has the added advantage that the kernels are now separable from the neighborhood construction, so you can store a neighborhood and call multiple kernels on it rather than constructing a new neighborhood each time.
> 
> On the other hand, it raises the possibility of having multiple neighborhoodsesesssses, so I renamed them to be singular as objects.

I like this design, it is simpler , less duplication and more pythonic. The whole _filter wasn't really needed. One question - Do we really need _BoundNeighborhoodReductions as an ABC, or can the common reduction logic be expressed through a simpler composition/delegation pattern?

cmdupuis3:
I kind of think there should be a way to unify all three classes somehow, but I haven't found it yet. I can try some more things and let you know.

cmdupuis3:
@rajeeja Alright, I refactored it a bit and got rid of the ABC (although spiritually it still basically is one). I attempted taking a compositional approach, but there's no nice solution that doesn't clutter up the API or duplicate all the reduction methods, or have some other drawbacks.

cmdupuis3:
pre-commit.ci autofix

rajeeja (approved):

cmdupuis3 (approved):
Approving, but a good amount now has my fingers on it, so I'd like to defer for another approval.

@erogluorhan ?

dylannelson (approved):
Created an env from the branch, ran all tests, ran the notebook locally, and all checks out.

The only thing discovered while going through the code was how NaN's are handled, and how a "skipna" option for functions like `mean` may be useful for a future implementation. 

Generated a quick notebook to test what happens to NA's as `r` increases when `.mean` is applied to `.neighborhood(r=r, on="face centers")`:

Image

This is likely out of the scope of what's being worked on here, or just may be how functions like this are inconvenienced by NaN's, so I'm approving with no blockers. Thanks!

erogluorhan:
@ahijevyc , now that this has been merged, please feel free to test it on main branch or via our next version when released and let us know what you think.

ahijevyc:
Thank you for all your consideration and effort on this feature. It sounds like it will be useful _and_ fast. I look forward to testing it, and will let you know if I run into issues.

PR #957, opened 2024-09-15
Template Benchmarks 

### LINKED ISSUE #959: Create Template Classes for ASV Benchmarks
There is a significant amount of duplicated code when creating ASV benchmarks, specifically in defining the parameters, setup, and teardown functions. 

Create a template class that can be inherited on a module level for benchmarks. This behavior should be adopted for future benchmarks.

philipc2:
Closes #959

## Overview

* Creates template benchmarks for the `mpas_ocean` benchmark module, which can be re-used to avoid needing to always create `setup`, `teardown` and other configurations
* Templates `DatasetBenchmark` and `GridBenchmark` are used for benchmarks that require a `UxDataset` or `Grid` respectively.

### DIFF
--- a/benchmarks/mpas_ocean.py
+++ b/benchmarks/mpas_ocean.py
@@ -29,7 +29,8 @@ file_path_dict = {"480km": [current_path / grid_filename_480, current_path / dat
 
 
 class DatasetBenchmark:
-    """Class used as a template for benchmarks in this module."""
+    """Class used as a template for benchmarks requiring a ``UxDataset`` in
+    this module across both resolutions."""
     param_names = ['resolution',]
     params = [['480km', '120km'],]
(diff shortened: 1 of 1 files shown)


erogluorhan (approved):

rajeeja:
seems to work, thanks for the cleanup. 
Does this also fix [asv profile issue](https://github.com/UXARRAY/uxarray-asv/issues/2)?

philipc2:
> seems to work, thanks for the cleanup. Does this also fix [asv profile issue](https://github.com/UXARRAY/uxarray-asv/issues/2)?

Unfortunately no, still trying to figure out how to call parametrized benchmarks. Might raise an issue on the `asv` repo

Issue #924, opened 2024-08-30
Use same code style for inverse_distance_weighted.py and nearest_neighbor.py 


ahijevyc:
Under `uxarray.remap`, `inverse_distance_weighted.py` and `nearest_neighbor.py` have a lot of the same code but have different documentation style. Their formatting styles should probably be brought back into agreement.

There is a lot of repeated code also. Perhaps that can be just in one place and reused. 

I put in a pull request #923 for a neighborhood filter `apply_func` that started with the code in these functions as a template, so there was a lot of copying and pasting. If the shared code was in one place it could serve 3 different functions, `inverse_distance_weighted.py` and `nearest_neighbor.py` and the potential new function `apply_func`.

aaronzedwick:
I like this idea a lot, especially coming up with a condensed version of the reused functions. I can tackle this most likely this week.

aaronzedwick:
Thinking about this more, I don't know what code we could reuse as a function. Although similar, the nearest neighbor and inverse distance functions use completely different parameters for calculation things. So the things we could reuse would only save a few extra lines, and just add an extra file to the code base without too much reason.

ahijevyc:
Thanks for looking at this. I was thinking of the similarity of lines 63-138 in `inverse_distance_weighted.py` and lines 45-128 in `nearest_neighbor.py`. There differences in the order of `elif`-clauses and the names of some variables, but otherwise, they are functionally equivalent. In `nearest_neighbor.py` the variables are named `cart_x, cart_y, cart_z` while in `inverse_distance_weighted.py` they are named `x, y, z`.

Plus, there a lot of similarities in the function descriptions but one has argument type hints (`nearest_neighbor.py`) and the other has no type hints `inverse_distance_weighted.py`.  The `Parameters:` sections also have single-character or whitespace differences at the end of lines.

I found this comparison was easy using gvim -d or gvimdiff.

aaronzedwick:
> Thanks for looking at this. I was thinking of the similarity of lines 63-138 in `inverse_distance_weighted.py` and lines 45-128 in `nearest_neighbor.py`. There differences in the order of `elif`-clauses and the names of some variables, but otherwise, they are functionally equivalent. In `nearest_neighbor.py` the variables are named `cart_x, cart_y, cart_z` while in `inverse_distance_weighted.py` they are named `x, y, z`.
> 
[...]

PR #974, opened 2024-09-26
Remapping Code Coverage, Modularize Reused Code, and Documentation Updates

### LINKED ISSUE #963: Improve testing coverage for `uxarray.remap`

aaronzedwick:
Closes #963 #924 

## Overview

Adds tests to get a more complete code coverage. Takes reused code from `nearest_neighbor` and `inverse_distance` and creates one general function for them to call, this should cut down on code and make it easier to create new remapping functions. Updates documentation so the remapping functions are consistent with each other. Also fixes a bug I found when trying to switch from cartesian or spherical after performing one remap.

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


test/test_remap.py
```diff
@@ -142,11 +147,74 @@ def test_overwrite(self):
...
+    def test_source_data_remap(self):
+        """Test the remapping of all source data positions."""
+
+        # Open source and destination datasets to remap to
+        source_uxds = ux.open_dataset(mpasfile_QU, mpasfile_QU)
+        destination_dataset = ux.open_dataset(gridfile_geoflow, dsfile_v1_geoflow)
+
+        # Remap from `face_centers`
+        face_centers = source_uxds['latCell'].remap.nearest_neighbor(
+            destination_grid=destination_dataset.uxgrid,
```
ahijevyc, line 164: Use `destination_grid` from above change instead of `destination_dataset.uxgrid`.

test/test_remap.py
```diff
@@ -142,11 +147,74 @@ def test_overwrite(self):
...
 
         # Assert the remapped data is different from the original data
         assert not np.array_equal(destination_dataset['v1'], remap_to_edge_centers)
 
+    def test_source_data_remap(self):
+        """Test the remapping of all source data positions."""
+
+        # Open source and destination datasets to remap to
+        source_uxds = ux.open_dataset(mpasfile_QU, mpasfile_QU)
+        destination_dataset = ux.open_dataset(gridfile_geoflow, dsfile_v1_geoflow)
```
ahijevyc, line 160: Use ux.open_grid. 
```
destination_grid = ux.open_grid(dsfile_v1_geoflow)
```

test/test_remap.py
```diff
@@ -142,11 +147,74 @@ def test_overwrite(self):
...
+
+        # Remap from `face_centers`
+        face_centers = source_uxds['latCell'].remap.nearest_neighbor(
+            destination_grid=destination_dataset.uxgrid,
+            remap_to="nodes"
+        )
+
+        # Remap from `nodes`
+        nodes = source_uxds['latVertex'].remap.nearest_neighbor(
+            destination_grid=destination_dataset.uxgrid,
```
ahijevyc, line 170: Use destination_grid from above change instead of destination_dataset.uxgrid.

test/test_remap.py
```diff
@@ -142,11 +147,74 @@ def test_overwrite(self):
...
+            remap_to="nodes"
+        )
+
+        # Remap from `edges`
+        edges = source_uxds['angleEdge'].remap.nearest_neighbor(
+            destination_grid=destination_dataset.uxgrid,
```
ahijevyc, line 176: Use destination_grid from above change instead of destination_dataset.uxgrid.

test/test_remap.py
```diff
@@ -142,11 +147,74 @@ def test_overwrite(self):
...
+            source_uxds['v1'].remap.nearest_neighbor(
+                destination_grid=destination_dataset.uxgrid,
+                remap_to="test", coord_type='spherical'
+            )
+        with nt.assert_raises(ValueError):
+            source_uxds['v1'].remap.nearest_neighbor(
+                destination_grid=destination_dataset.uxgrid,
+                remap_to="test", coord_type="cartesian"
+            )
+
```
ahijevyc, line 203: Switch from `ux.open_dataset` to `ux.open_grid` where possible,  replacing `destination_dataset.uxgrid` with `destination_grid` afterwards.

test/test_remap.py
```diff
@@ -257,8 +331,93 @@ def test_overwrite(self):
...
+        # Remap from `nodes`
+        nodes = source_uxds['latVertex'].remap.inverse_distance_weighted(
+            destination_grid=destination_dataset.uxgrid,
+            remap_to="nodes"
+        )
+
+        # Remap from `edges`
+        edges = source_uxds['angleEdge'].remap.inverse_distance_weighted(
+            destination_grid=destination_dataset.uxgrid,
+            remap_to="nodes"
```
ahijevyc, line 363: Switch from ux.open_dataset to ux.open_grid where possible, replacing destination_dataset.uxgrid with destination_grid afterwards.

ahijevyc (commented):
Switch from `ux.open_dataset` to `ux.open_grid` where possible, replacing `destination_dataset.uxgrid` with `destination_grid` afterwards.

Can we avoid hard-coding `reconstruct=True` when getting the Ball Tree?
This probably needs to wait for `Grid.get_ball_tree` to be enhanced.

uxarray/remap/utils.py
```diff
@@ -0,0 +1,132 @@
...
+                destination_grid.edge_lat.values,
+            )
+        else:
+            raise ValueError(
+                f"Invalid remap_to. Expected 'nodes', 'edge centers', or 'face centers', "
+                f"but received: {remap_to}"
+            )
+
+        _source_tree = source_grid.get_ball_tree(
+            coordinates=source_data_mapping, reconstruct=True
```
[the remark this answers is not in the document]
aaronzedwick: Thanks for the review! The `Grid.get_ball_tree` does indeed check and uses the cached version if it's constructed. However, if passed the reconstructed argument, it will override this and ignore the cache. This is needed here because if a user opens a `UxDataset`, and remaps using Cartesian, but then decides to use the same `UxDataset` to remap with spherical, the coordinates used to construct the tree will be incorrect and the query will fail. You can see this in the test cases locally if you remove `reconstruct=True`. 

Also, it could affect the results even if the coordinates aren't changed. The tree is constructed from the grid, so if it's constructed on the nodes on the first remap, and then you remap another variable with the same object that is on the faces, it won't construct the tree on the faces, but use the cached nodes version.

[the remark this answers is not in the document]
philipc2: This is how we handle caching for our `GeoDataFrame` construction. This may be overkill for what we need here though.

https://github.com/UXARRAY/uxarray/blob/510e29901b898f0554c6054b0367bcb65be99411/uxarray/grid/grid.py#L1567-L1689

ahijevyc (approved):
Great switch to open_grid. Thanks for looking into necessity of reconstruct=True

uxarray/remap/utils.py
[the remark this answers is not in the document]
ahijevyc: It does seem like a lot, whenone could just force reconstruct=True.

rajeeja (approved):

PR #982, opened 2024-10-01
Use Miniforge after Mambaforge sunset 

philipc2:
# Overview

> As of July 24, 2024, mambaforge and miniforge are identical, and mambaforge is deprecated. For more details see https://github.com/conda-forge/miniforge?tab=readme-ov-file, which recommends users switch to Miniforge3 immediately.


aaronzedwick (approved):
Thanks for doing this so quickly!

rajeeja (approved):
Sunrise for Miniforge3 🌞

PR #989, opened 2024-10-04
DRAFT: Fast Cross Sections 

philipc2:
Closes #XXX

## Expected Usage

```Python
import uxarray as ux

```

### DIFF
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -65,6 +65,11 @@ from uxarray.grid.neighbors import (
     _populate_edge_node_distances,
 )
 
+from uxarray.grid.intersections import (
+    constant_lat_intersections,
+    constant_lon_intersections,
+)
+
 from spatialpandas import GeoDataFrame
 
 from uxarray.plot.accessor import GridPlotAccessor
@@ -1878,3 +1883,43 @@ class Grid:
             raise ValueError(
                 "Indexing must be along a grid dimension: ('n_node', 'n_edge', 'n_face')"
             )
+
+    def get_edges_at_constant_latitude(self, lat):
+        """TODO:"""
+        lat = np.deg2rad(lat)
+        edge_node_connectivity = self.edge_node_connectivity.values
+        edge_node_x = self.node_x.values[edge_node_connectivity]
+        edge_node_y = self.node_y.values[edge_node_connectivity]
+        edge_node_z = self.node_z.values[edge_node_connectivity]
+
+        edges, _ = constant_lat_intersections(
+            lat, edge_node_x, edge_node_y, edge_node_z, self.n_edge
+        )
+
+        return edges
+
+    def get_edges_at_constant_longitude(self, lon):
+        """TODO:"""
+        lon = np.deg2rad(lon)
+        edge_node_connectivity = self.edge_node_connectivity.values
+        edge_node_x = self.node_x.values[edge_node_connectivity]
+        edge_node_y = self.node_y.values[edge_node_connectivity]
+        edge_node_z = self.node_z.values[edge_node_connectivity]
+
+        edges, _ = constant_lon_intersections(
+            lon, edge_node_x, edge_node_y, edge_node_z, self.n_edge
+        )
+
+        return edges
+
+    def get_faces_at_constant_latitude(self, lat):
+        """TODO:"""
+        lat = np.deg2rad(lat)
+        edges = self.get_edges_at_constant_latitude(lat)
+        return self.edge_face_connectivity[edges].data.ravel()
+
+    def get_faces_at_constant_longitude(self, lon):
+        """TODO:"""
+        lon = np.deg2rad(lon)
+        edges = self.get_edges_at_constant_longitude(lon)
+        return self.edge_face_connectivity[edges].data.ravel()
--- a/uxarray/grid/intersections.py
+++ b/uxarray/grid/intersections.py
@@ -1,10 +1,113 @@
 import numpy as np
-from uxarray.constants import MACHINE_EPSILON, ERROR_TOLERANCE
+from uxarray.constants import MACHINE_EPSILON, ERROR_TOLERANCE, INT_DTYPE
 from uxarray.grid.utils import _newton_raphson_solver_for_gca_constLat
 from uxarray.grid.arcs import point_within_gca, extreme_gca_latitude, in_between
 import platform
 import warnings
 from uxarray.utils.computing import cross_fma, allclose, dot, cross, norm
+
+
+from numba import njit
+
+
+@njit
+def constant_lat_intersections(lat, edge_node_x, edge_node_y, edge_node_z, n_edge):
+    """Determine which edges intersect a constant line of latitude on a sphere.
+
+    Parameters
+    ----------
+    lat:
+        Constant latitude value in radians.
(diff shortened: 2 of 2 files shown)


test/test_cross_sections.py
```diff
@@ -0,0 +1,70 @@
...
+        grid_all_four = uxgrid.constant_latitude_cross_section(lat=0.0)
+
+        assert grid_all_four.n_face == 4
+
+        with pytest.raises(ValueError):
+            # no intersections found at this line
+            uxgrid.constant_latitude_cross_section(lat=10.0)
+
+
+    def test_constant_lat_cross_section_uxds(self):
```
hongyuchen1030, line 53: Please add some test case when lattitude is around pole area,

philipc2: I've added tests for both poles for a cube-sphere grid.

philipc2:
Here's a figure putting the Github Actions results into perspective

image.

erogluorhan:
Could you clarify the addition of `docs/user-guide/grid.nc`? (I couldn't see its use in the user guide doc, and also even if it will be used, can it be used from the "test/meshfiles" dir?)

erogluorhan:
Please clarify if `test/meshfiles/ugrid/quad-hexagon/grid.nc` is being replaced with a different grid now.

philipc2:
@erogluorhan 

The grid the same. I've dropped the `edge_node_connectivity` that was stored there before when I originally created the grid, the connectivity was invalid.

philipc2:
> Could you clarify the addition of `docs/user-guide/grid.nc`? (I couldn't see its use in the user guide doc, and also even if it will be used, can it be used from the "test/meshfiles" dir?)

That should not be there. Good catch.

uxarray/grid/grid.py
```diff
@@ -1912,3 +1935,129 @@ def isel(self, **dim_kwargs):
...
+            raise ValueError(f"No intersections found at lat={lat}.")
+
+        grid_at_constant_lat = self.isel(n_face=faces)
+
+        if return_face_indices:
+            return grid_at_constant_lat, faces
+        else:
+            return grid_at_constant_lat
+
+    def get_edges_at_constant_latitude(self, lat, method="fast"):
```
erogluorhan, line 2000: How about `edges_at_constant_latitude`?

uxarray/grid/grid.py
```diff
@@ -1912,3 +1935,129 @@ def isel(self, **dim_kwargs):
...
+            edges = fast_constant_lat_intersections(
+                lat, self.edge_node_z.values, self.n_edge
+            )
+        elif method == "accurate":
+            raise NotImplementedError("Accurate method not yet implemented.")
+        else:
+            raise ValueError(f"Invalid method: {method}.")
+        return edges.squeeze()
+
+    def get_faces_at_constant_latitude(self, lat, method="fast"):
```
erogluorhan, line 2034: How about `faces_at_constant_latitude`?

uxarray/core/dataarray.py
```diff
@@ -1121,3 +1121,36 @@ def _slice_from_grid(self, sliced_grid):
             dims=self.dims,
             attrs=self.attrs,
         )
+
+    def constant_latitude_cross_section(self, lat: float, method="fast"):
```
erogluorhan, line 1125: How about `cross_section_at_constant_latitude` or `cross_section_at_latitude`?

uxarray/grid/grid.py
```diff
@@ -1912,3 +1935,129 @@ def isel(self, **dim_kwargs):
             raise ValueError(
                 "Indexing must be along a grid dimension: ('n_node', 'n_edge', 'n_face')"
             )
+
+    def constant_latitude_cross_section(
```
erogluorhan, line 1939: How about `cross_section_at_constant_latitude`?

uxarray/grid/intersections.py
```diff
@@ -7,6 +7,48 @@
 from uxarray.utils.computing import cross_fma, allclose, dot, cross, norm
 
 
+from numba import njit, prange
+
+
+@njit(parallel=True, nogil=True, cache=True)
+def fast_constant_lat_intersections(lat, edge_node_z, n_edge):
```
erogluorhan, line 14: A bit skeptical about having "fast" in the function name; could we find an alternative way to emphasize likewise? Also, as before, I think I find it more intuitive to start the function name with the biggest objective of the function, e.g. something like `intersections_at_constant_latitude`?

uxarray/grid/grid.py
philipc2: That naming scheme has been using for our Grid properties, such as `face_node_connectivity` and `node_lon`, which remain static once they are computed. Properties also don't accept any arguments.

Since the user provides a `lat` argument here, the `get_` seemed appropriate. Thoughts?

uxarray/grid/grid.py
philipc2: Same as above.

uxarray/core/dataarray.py
philipc2: Something analogous to our plotting and remapping accessors?

```Python
Grid.cross_section.constant_latitude()
Grid.cross_section.between_constant_latitudes()
Grid.cross_section.constant_longitude()
Grid.cross_section.between_constant_longitudes()
Grid.cross_section.gca_gca()
```

uxarray/grid/intersections.py
philipc2: > A bit skeptical about having "fast" in the function name; could we find an alternative way to emphasize likewise?

Since this function will be paired alongside @hongyuchen1030 accurate algorithm, there is a need to emphasize the difference. I was envisioning:
* `fast_constant_lat_intersections`
* `accurate_constant_lat_intersections`

Open to any other ideas though.

uxarray/core/dataarray.py
erogluorhan: Exactly, but I don't want to make it a priority for this PR as we can always design it over time and if it will require a long thought process

uxarray/grid/grid.py
erogluorhan: Ok, that makes sense about Grid; thanks!

uxarray/grid/intersections.py
erogluorhan: I was expecting to see the distinction between algorithms being exphasized through an argument in a single function signature, rather than having separate functions.

erogluorhan: For instance `intersections_at_constant_latitude(method=`fast`)`. Fast or approximate for the first, and accurate for the second would make sense. @hongyuchen1030 what do you also think?

uxarray/core/dataarray.py
philipc2: Done.

uxarray/grid/intersections.py
philipc2: > I was expecting to see the distinction between algorithms being exphasized through an argument in a single function signature, rather than having separate functions.

This doesn't work well with Numba, which is the main reason I prefixed the function.

erogluorhan: I see it better now; thanks for clarifying!

erogluorhan (approved):
Great work!

PR #818, opened 2024-06-18
Pass coords to UxDataArray

ahijevyc:
Pass along coordinates of UxDataArray when slicing from grid.

Pass self.coords to as coords argument in core/dataarray.py method _slice_from_grid

Otherwise, _slice_from_grid loses coordinate labels (the index values).

I figure if dims, name, and attrs arguments are passed, coords should be too.

I tested uxda.subset.bounding_box(lon_bounds, lat_bounds) on a UxDataArray with and without the pull request. The results are below. The number of Indexes changes from zero to three.

Without passing the coords argument in _slice_from_grid, the result has zero Indexes
```
xarray.UxDataArray'updraft_helicity_max'valid_time: 24mem: 5initial_time: 2n_face: 19327
array([[[[0.00000000e+00, 0.00000000e+00, 0.00000000e+00, ...,
          2.17450754e-04, 0.00000000e+00, 0.00000000e+00],
         [6.97366288e-03, 4.58992086e-03, 6.88771158e-03, ...,
          0.00000000e+00, 0.00000000e+00, 0.00000000e+00]],

        [[1.94994651e-03, 2.29698606e-03, 1.26716075e-03, ...,
          6.46118497e-05, 3.72463410e-06, 0.00000000e+00],
         [0.00000000e+00, 0.00000000e+00, 0.00000000e+00, ...,
          2.02913233e-03, 3.39804428e-05, 8.73764220e-05]],

        [[8.91156844e-04, 1.32420380e-03, 3.09247524e-03, ...,
          1.14122406e-04, 2.35418309e-04, 2.08011316e-03],
         [1.75224006e-04, 1.12577909e-04, 1.69947656e-04, ...,
          2.02424005e-02, 2.01683957e-03, 0.00000000e+00]],

        [[2.08048197e-03, 1.04307907e-03, 8.50136974e-04, ...,
          1.31235253e-02, 1.25350822e-02, 5.65142557e-03],
         [0.00000000e+00, 0.00000000e+00, 0.00000000e+00, ...,
          3.22981328e-02, 4.79420833e-03, 9.13940486e-04]],

...
        [[0.00000000e+00, 0.00000000e+00, 0.00000000e+00, ...,
          0.00000000e+00, 0.00000000e+00, 0.00000000e+00],
         [7.28812627e-03, 5.85812889e-03, 1.00370627e-02, ...,
          1.39404610e-02, 4.06184839e-03, 1.46363070e-02]],

[...]

PR #1003, opened 2024-10-09
Preserve Coordinates when Indexing UxDataArray

### LINKED ISSUE #945: `_slice_from_grid` may need to use coords from a sliced dataarray
### Version

v2024.08.2

Conda

### What happened?

`_slice_from_grid` in UxDataArray create a new subset dataarray using coordinates info from the original dataarray as shown in the lines below:
https://github.com/UXARRAY/uxarray/blob/f71e11bef0fabe3ac74767767ad0509cef5cf504/uxarray/core/dataarray.py#L1099-L1106

This does not cause an issue when the original dataarray does not have coordinates that are sliced, but the constructor fails otherwise with a conflicting size for dimension. Probably a better or a right approach is creating a new sliced dataarray with coordinates from the sliced `d_var` not from the original `self`.

A new sliced (subset) dataarray needs to be created with a correctly sliced coordinate variables, not from the original unsliced coordinate variables.

_No response_

philipc2:
Closes #945

## Overview

* Preserves coordinates when indexing a `UxDataArray`


ahijevyc (commented):
Looks good. I tested the change and tried to replicate the issue I had #818 with the uxarray.UxDataArray.Indexes disappearing but had no problems. 

However, given this fix, I can think of two other places a similar change might be needed:
 https://github.com/UXARRAY/uxarray/blob/ec5cc8f279ae7026e5f42ab0570105cd6861ced7/uxarray/remap/nearest_neighbor.py#L110

and 

https://github.com/UXARRAY/uxarray/blob/ec5cc8f279ae7026e5f42ab0570105cd6861ced7/uxarray/remap/inverse_distance_weighted.py#L151

philipc2:
> Looks good. I tested the change and tried to replicate the issue I had #818 with the uxarray.UxDataArray.Indexes disappearing but had no problems.
> 
> However, given this fix, I can think of two other places a similar change might be needed:
> 
> https://github.com/UXARRAY/uxarray/blob/ec5cc8f279ae7026e5f42ab0570105cd6861ced7/uxarray/remap/nearest_neighbor.py#L110
> 
> and
> 
> https://github.com/UXARRAY/uxarray/blob/ec5cc8f279ae7026e5f42ab0570105cd6861ced7/uxarray/remap/inverse_distance_weighted.py#L151

Good catches, I'll update these also.

philipc2:
@ahijevyc 

When remapping, should the coordinates from the original data array be dropped? It wouldn't make much sense to preserve them (unless I'm understanding it wrong)

ahijevyc (commented):
The _spatial_ coordinates should not be transferred, but if there are _non-spatial_ coordinates like `time`, `ensemble_member`, or `forecast_hour`, then those should be transferred to the destination. I think that's why I originally thought the `coords` argument was necessary.

philipc2:
> The _spatial_ coordinates should not be transferred, but if there are _non-spatial_ coordinates like `time`, `ensemble_member`, or `forecast_hour`, then those should be transferred to the destination. I think that's why I originally thought the `coords` argument was necessary.

That makes sense! I wasn't considering the non-spatial cases.

uxarray/remap/inverse_distance_weighted.py
```diff
@@ -148,7 +148,6 @@ def _inverse_distance_weighted_remap_uxda(
     uxda_remap = uxarray.core.dataarray.UxDataArray(
         data=destination_data,
         name=source_uxda.name,
-        coords=source_uxda.coords,
```
ahijevyc, line 151: @philipc2 Here's my solution to preserve any non-spatial coordinates (like time or ensemble_member) while dropping the spatial one:

Copy `source_uxda.coords` to a new `xarray.Coordinates` object. Then delete the key associated with the spatial dimension. Then use that new object as the coords argument to `uxarray.core.dataarray.UxDataArray`.

```
    destination_coords = source_uxda.coords
    del(destination_coords[destination_dim])

    # construct data array for remapping variable
    uxda_remap = uxarray.core.dataarray.UxDataArray(
        data=destination_data,
        name=source_uxda.name,
        coords=destination_coords,
```
Does this make sense?

ahijevyc (changes requested):
I provided a way to drop the remapped coordinate (which I agree, should not get transferred from the source) and still preserve any non-spatial coordinates.

uxarray/remap/inverse_distance_weighted.py
ahijevyc: @philipc2 Thanks .  Just to make sure I understand, does the _preserve_coordinates function slice along the spatial coordinate and return the sliced spatial coordinate (along with the non-spatial coordinates)? If so, that's great. I don't think my solution did that. My solution simply deleted the spatial coordinate. Retaining the properly-sliced spatial coordinate is good.  
Therefore, I don't think we really need a `dims_to_drop` option.  When I dropped it in my suggested code, I was just being lazy. I couldn't think of a way to provide the sliced spatial coordinates. We don't want to completely drop the spatial dimension, just slice it.  So could the `dims_to_drop` option be removed?

ahijevyc (commented):
I had a question about the new util function and whether it can be simplified.

uxarray/remap/inverse_distance_weighted.py
philipc2: I've updated the logic in the `UxDataArray.isel()` function, which should now correctly handle slicing the spatial dimensions.

As for the remapping, I've re-added the deletion approach you mentioned, with a few extra checks.

ahijevyc (changes requested):
Looks good. 
I couldn't think of a reason to find and delete the grid dimension found in the list ["n_node", "n_edge", "n_face"] when one could use simply delete `destination_dim`, which was defined earlier. So I removed a couple for-block and if-block levels.

ahijevyc (approved):
I approve

PR #1004, opened 2024-10-10
Non-Conservative Zonal Mean 

### LINKED ISSUE #93: Add zonal means (non-conservative)
Global and zonal means (conservative and non-conservative)

_Originally posted by @erogluorhan in https://github.com/UXARRAY/uxarray/discussions/46#discussioncomment-2746606_

philipc2:
Closes #93

## Overview

* Implements Non-Conservative Zonal Averaging (see #785 for a more detailed discussion)
* [User Guide on Zonal Means](https://uxarray--1004.org.readthedocs.build/en/1004/user-guide/zonal-average.html)


uxarray/core/zonal.py
```diff
@@ -0,0 +1,25 @@
...
+    shape = uxda.shape[:-1] + (len(latitudes),)
+    if isinstance(uxda.data, da.Array):
+        # Create a Dask array for storing results
+        averages = da.empty(shape, dtype=uxda.dtype)
+    else:
+        # Create a NumPy array for storing results
+        averages = np.empty(shape, dtype=uxda.dtype)
+
+    for i, lat in enumerate(latitudes):
+        indices = get_indices_func(lat, method)
```
philipc2, line 22: @hongyuchen1030 @amberchen122 

With the API design I've decided to go with in #989, the accurate implementation in #785 will be specified by having `method='accurate'` instead of `method='fast'`

I can handle translating the implementation in #785 to meet the API design.

philipc2:
@erogluorhan @hongyuchen1030 @paullric 

For zonal averaging, would the primary difference between Conservative and Non-Conservative be whether we apply a weighted average as opposed to performing a standard average?

hongyuchen1030:
> @erogluorhan @hongyuchen1030 @paullric
> 
> For zonal averaging, would the primary difference between Conservative and Non-Conservative be whether we apply a weighted average as opposed to performing a standard average?

As I explained before here https://github.com/UXARRAY/uxarray/pull/785#discussion_r1794446008

Both use weighted average

But non-conservative one use the arc length, conservative use the area as weights

uxarray/core/zonal.py
hongyuchen1030: Sounds good

philipc2:
> > @erogluorhan @hongyuchen1030 @paullric
> > For zonal averaging, would the primary difference between Conservative and Non-Conservative be whether we apply a weighted average as opposed to performing a standard average?
> 
> As I explained before here [#785 (comment)](https://github.com/UXARRAY/uxarray/pull/785#discussion_r1794446008)
> 
> Both use weighted average
> 
> But non-conservative one use the arc length, conservative use the area as weights

Sorry for the duplicate! Send this one without seeing your comment before. Appreciate the clarification

philipc2:
@aaronzedwick 

For reference, the pre-commit is failing in all the PR's and CI's too.

philipc2:
Hi @hongyuchen1030 

I've made some progress migrating the core functionality from #785 

Before requesting a comprehensive review, may you give this a look and suggest anything major that needs to still be added for it to be ready?

hongyuchen1030:
> Hi @hongyuchen1030
> 
> I've made some progress migrating the core functionality from #785
> 
> Before requesting a comprehensive review, may you give this a look and suggest anything major that needs to still be added for it to be ready?

So you mean we'are not going to merge #785, but instead we will merge this? The main reason #785 has conflicts is because there're too many redesign from the main and I have already lost track of these changes already. So it will be a good idea for just submit them here

As for the major, I will say just try to run the zonal-mean latlon bounds and see what's missing. Also for each helper function includes some testcases for them, you can directly use the one from #785.

philipc2:
> > Hi @hongyuchen1030
> > I've made some progress migrating the core functionality from #785
> > Before requesting a comprehensive review, may you give this a look and suggest anything major that needs to still be added for it to be ready?
> 
> So you mean we're not going to merge #785, but instead we will merge this? The main reason #785 has conflicts is because there're too many redesign from the main and I have already lost track of these changes already. So it will be a good idea for just submit them here
> 
> As for the major, I will say just try to run the zonal-mean with latlon bounds and see what's missing. Also for each helper function includes some testcases for them, you can directly use the one from #785.

That's correct. The goal was to have this PR address the redesigns while retaining the core content for #785

I am planning to document the advantages of using `spherical_bounding_box=True` in the [user guide notebook](https://uxarray--1004.org.readthedocs.build/en/1004/user-guide/zonal-average.html).

I'm also planning to create a "dummy" grid that contains a single large face where the spherical bounding box is needed to correctly compute the intersection and use it both in the notebooks and test case.

hongyuchen1030:
> > > Hi @hongyuchen1030
> > > I've made some progress migrating the core functionality from #785
> > > Before requesting a comprehensive review, may you give this a look and suggest anything major that needs to still be added for it to be ready?
> > 
> > 
> > So you mean we're not going to merge #785, but instead we will merge this? The main reason #785 has conflicts is because there're too many redesign from the main and I have already lost track of these changes already. So it will be a good idea for just submit them here
> > As for the major, I will say just try to run the zonal-mean with latlon bounds and see what's missing. Also for each helper function includes some testcases for them, you can directly use the one from #785.
> 
> That's correct. The goal was to have this PR address the redesigns while retaining the core content for #785
> 
> I am planning to document the advantages of using `spherical_bounding_box=True` in the [user guide notebook](https://uxarray--1004.org.readthedocs.build/en/1004/user-guide/zonal-average.html).
> 
> I'm also planning to create a "dummy" grid that contains a single large face where the spherical bounding box is needed to correctly compute the intersection and use it both in the notebooks and test case.

Sounds good. I have added @amberchen122 in this PR, please let me know if there's anything else needed from me

uxarray/core/dataarray.py
```diff
@@ -429,6 +434,76 @@ def integrate(
 
         return uxda
 
+    def zonal_mean(self, lat=(-90, 90, 10)):
+        """Compute the average along one or more lines of constant latitude.
+
+        The candidate faces at each line of constant latitude are determined using a robust algorithm that treats each edge
+        as a great circle arc, ensuring that extremes along those arcs are considered.
```
hongyuchen1030, line 441: We don't need to emphasize the "extreme" case specifically. We can just say this robust algorithm takes care of  the intersection point calculation accurate enough

uxarray/core/dataarray.py
```diff
@@ -429,6 +434,76 @@ def integrate(
 
         return uxda
 
+    def zonal_mean(self, lat=(-90, 90, 10)):
```
hongyuchen1030, line 437: We probably need to set the criteria for the step size: it must be integer or cannot be smaller than 1. Otherwise it will still hit some edge case that the current implementation can not take care of (We already have the more accurate one written out and it will be implemented in the future)

uxarray/plot/utils.py
```diff
@@ -33,3 +35,30 @@ def reset_mpl_backend(self):
...
+    """
+    Generate points along the great circle between two points.
+    p1 and p2 are arrays/lists of Cartesian coordinates.
+    """
+    # Normalize the input points
+    p1 = p1 / np.linalg.norm(p1)
+    p2 = p2 / np.linalg.norm(p2)
+
+    # Compute the angle between the points
+    omega = np.arccos(np.clip(np.dot(p1, p2), -1.0, 1.0))
```
hongyuchen1030, line 50: we already have the function that calculates the angle called `angle_between_vectors` right? That one uses a specific math formula to avoid some potential cancellation errors.

uxarray/plot/utils.py
```diff
@@ -33,3 +35,30 @@ def reset_mpl_backend(self):
...
+
+    if omega == 0:
```
hongyuchen1030, line 52: probably better to use `is_closed` with the error tolerance than using the equal sign

uxarray/grid/weights.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+from numba import njit
+from uxarray.constants import INT_FILL_VALUE
+
+
+@njit
+def _face_weights_from_edge_magnitudes(
```
hongyuchen1030, line 7: The edges here are not normal edges here. They are not great circle arcs, instead, they are just constant latitude lines. And they're constructed by ourselves. 
So probably the name `edge` is not appropriate

hongyuchen1030 (commented):
@philipc2 
I couldn't find the codes where we actually calculate the intersection points between great circle arcs and the constant latitude and utilize the sweeping line algorithm to calculate the weight. I wonder if you can kindly point me there? thanks

test/test_zonal.py
```diff
@@ -0,0 +1,67 @@
...
+    dsfile_var2_ne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_var2.nc"
+    test_file_2 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_test2.nc"
+    test_file_3 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_test3.nc"
+
+
+    @pytest.mark.parametrize("use_spherical_bounding_box", [True, False])
+    def test_non_conservative_zonal_mean_equator(self, use_spherical_bounding_box):
+        """Test _non_conservative_zonal_mean function with outCSne30 data.
+
+        Low error tolerance test at the equator.
```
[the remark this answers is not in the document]
philipc2: Good catch. I was using ChatGPT to re-format the test to use the pytest parameters and didn't realize it also modified the docstring.

uxarray/core/dataarray.py
philipc2: > We already have the more accurate one written out and it will be implemented in the future

Which other implementation are you referring to? Is it different than the one implement here that uses the bounds?

uxarray/plot/utils.py
philipc2: Good catch, didn't think to use that one here.

philipc2:
@hongyuchen1030 

I think I'll remove the edge weights for now from this PR, since this one focusses on the Conservative Zonal Average.

philipc2:
> I couldn't find the codes where we actually calculate the intersection points between great circle arcs and the constant latitude and utilize the sweeping line algorithm to calculate the weight. I wonder if you can kindly point me there? thanks

In this PR, I have implemented the Conservative Zonal Average, with each face being weighted by its area. 

From a discussion we had in https://github.com/UXARRAY/uxarray/pull/785#discussion_r1794446008
> Non-conservative and conservative both take weighted averages.
> 
> Non-conservative tasks the edge length as the weight factor while the conservative ones take the area.

I still need to look over `_get_zonal_faces_weight_at_constLat` and apply similar optimizations that I made in #1072

uxarray/core/dataarray.py
philipc2: Sounds good. Once you have published your work, we can reference the algorithm in your paper in the future for more details.

uxarray/core/dataarray.py
hongyuchen1030: So my point is we need to be careful of calling our package "robust", even my latest implementation (yes it's different than the one implemented here for calculating the intersection. But both of them use the bounds). We're using the bounds doesn't mean we're doing it robust. We are just doing it "more accurate" than not using the bounds.

We can say we tried a better method and it's accurate enough for most of the usages in the docstring. And we'll release an even more accurate version in the future. But none of them is guaranteeing
the robustness (we don't have to say this in the docstring ofc)

philipc2: I'll update the docstring to reflect this. Thank you for clarifying this.

test/test_cross_sections.py
```diff
@@ -174,3 +174,17 @@ def test_constant_lat_out_of_bounds(self):
...
+class TestExtremes:
+    def test_extreme_latitude(self):
+        node_lon = np.array([-40, -40, 40, 40])
+        node_lat = np.array([-20, 20, 20, -20])
+        face_node_connectivity = np.array([[0, 1, 2, 3]], dtype=np.int64)
+
+        uxgrid = ux.Grid.from_topology(node_lon, node_lat, face_node_connectivity)
+
+        out1 = uxgrid.get_faces_at_constant_latitude(lat=20)
+        out2 = uxgrid.get_faces_at_constant_latitude(lat=20)
```
rajeeja, line 188: How about ? 
```
-        out2 = uxgrid.get_faces_at_constant_latitude(lat=20)
+        out2 = uxgrid.get_faces_at_constant_latitude(lat=-20)
```

Also, what means extreme here? along the edges of the face? a small doc string would be good.

rajeeja (commented):
looks good, specially the dask lazy loading stuff and the framework that can be applied for non-conservative PR. Made a comment for a test.

Also, use `angle_between_vectors ` as pointed in the other review.

test/test_cross_sections.py
philipc2: Thanks for pointing this out, I made a typo in this tests.

Since we treat each edge as an arc, using the spherical bounds should yield the same result for 20 degrees and 21 degrees, which would otherwise fail if we didn't handle the spherical geometry.

rajeeja: Doing some trial and error, I found that +/-25.41 still intersects but not 25.42

philipc2: These are the bounds for the single face, computed using Hongyu's implementation. The maximum along the arc will be greater than 20 degrees since we treat it as a great circle arc. 

image

philipc2:
I am moving this PR back to a draft to make some corrections after our discussion during the UXarray meeting.

philipc2:
There is currently a large bottleneck in perfromance due to the weight computation. See below a profile of the 30km MPAS grid, with latitudes between -45 and 45 degrees using a 2 degree spacing.

image


Despite this, @rajeeja @aaronzedwick @amberchen122 , you may begin to review this PR.


@hongyuchen1030 and I will continue to investigate performance optimizations. If we do not finish them by the time reviews are done, we can continue to iterate after it is merged.

uxarray/core/dataarray.py
```diff
@@ -429,6 +434,78 @@ def integrate(
...
```
rajeeja, line 437: I think it'd be nice to get the API right, i.e. have some optional variables as such as `method` that represents `weights candidate faces by the length of intersection by a given line of constant latitude.` also the choice of conservative or non-conservative etc. This will make it easier for new additions to this work.

rajeeja (changes requested):

uxarray/core/dataarray.py
philipc2: Think we should call the method `non_conservative_zonal_mean` then, keeping the current parameters. Thoughts?

rajeeja: > Think we should call the method `non_conservative_zonal_mean` then, keeping the current parameters. Thoughts?

I like the current neat small function signature. How about keep it as is and have optional `conservative` `True/False`

philipc2: That works with me. However, I think adding the `conservative` parameter might make more sense once we implement the conservative method, not to confuse the user.

test/test_zonal.py
```diff
@@ -0,0 +1,69 @@
...
+        grid_path = self.gridfile_ne30
+        data_path = self.datafile_vortex_ne30
+        uxds = ux.open_dataset(grid_path, data_path)
+
+        # Test at the poles
+        res_n90 = uxds['psi'].zonal_mean(90)
+        res_p90 = uxds['psi'].zonal_mean(-90)
+
+        # Assert results are approximately 1 within a delta of 1
+        assert res_n90.values[0] == pytest.approx(1, abs=1)
```
hongyuchen1030, line 52: setting the delta as 1 is unnecessary here (That means a relative error of 100%), You can use the default one, whose relative error is approximately 10^-8

test/test_zonal.py
```diff
@@ -0,0 +1,69 @@
...
+        assert res_p90.values[0] == pytest.approx(1, abs=1)
```
hongyuchen1030, line 53: Same here, for the detla value

uxarray/grid/utils.py
```diff
@@ -80,6 +80,18 @@ def _angle_of_2_vectors(u, v):
...
+    jacobian = np.array(
+        [
+            [(y0 * z1 - z0 * y1), (x0 * z1 - z0 * x1)],
+            [2 * x_i_old, 2 * y_i_old],
+        ]
+    )
+    return jacobian
+
+
 def _inv_jacobian(x0, x1, y0, y1, z0, z1, x_i_old, y_i_old):
```
hongyuchen1030, line 95: Same here, we can remove all these jacobian related functions

uxarray/grid/utils.py
```diff
@@ -80,6 +80,18 @@ def _angle_of_2_vectors(u, v):
         return 2 * np.pi - angle_u_v_rad
 
 
+@njit(cache=True)
```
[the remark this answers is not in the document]
philipc2: Sounds good. I wasn't sure if we had other plans for these methods.

hongyuchen1030 (approved):
Looks good! Pinning @amberchen122 for a closer look of the implementation

amberchen122 (approved):
Great job! Thank you for transferring the changes over. Well done!

uxarray/core/dataarray.py
rajeeja: ok, fix this one and we can merge it then

rajeeja (changes requested):

philipc2:
Some timing results from my M1 Macbook for a 30km MPAS grid (~655,362 hexagons)

#### Single Timestep 

image

#### 1000 Timesteps

image

rajeeja:
Good work!!

erogluorhan:
Great work @hongyuchen1030 @philipc2 @amberchen122 !

Issue #991, opened 2024-10-07
Pre-Commit CI Failing 


philipc2:
https://github.com/PyCQA/docformatter/issues/289

PR #1007, opened 2024-10-10
JIT Cleanup

### LINKED ISSUE #969: standardize njit and jit-cache usage across the project, cleanup and use JIT related constants in constants.py

codecov doesn't work correctly and track @njit decorated functions, a way to enable disable numba across the project would be useful.

aaronzedwick:
Closes #969 

## Overview

Cleans up jit usage, standardizes `jit cache`, and removes redundant code. 

## PR Checklist

**General**
- [x] Add appropriate labels


philipc2:
This is great. I assume these global variables in Numba need to be set before importing UXarray right?

aaronzedwick:
> This is great. I assume these global variables in Numba need to be set before importing UXarray right?

Well, they aren't really global variables in Numba. There isn't much point in having the `disable_numba` function, as we have already shown once the function is loaded it won't change anything. That would be a global variable in Numba, but it doesn't help much once the function is loaded. And `ENABLE_JIT_CACHE` is just our own constant we have in the constants file. Honestly, I would be in favor of removing that as well and just rewriting all the functions with jit to have `@njit(cache=True)`. Since that is what we are doing anyway, I don't really see a reason to have an extra constant variable. We have never really even used it. It is basically the same as just saying `cache=True`.

rajeeja:
> > This is great. I assume these global variables in Numba need to be set before importing UXarray right?
> 
> Well, they aren't really global variables in Numba. There isn't much point in having the `disable_numba` function, as we have already shown once the function is loaded it won't change anything. That would be a global variable in Numba, but it doesn't help much once the function is loaded. And `ENABLE_JIT_CACHE` is just our own constant we have in the constants file. Honestly, I would be in favor of removing that as well and just rewriting all the functions with jit to have `@njit(cache=True)`. Since that is what we are doing anyway, I don't really see a reason to have an extra constant variable. We have never really even used it. It is basically the same as just saying `cache=True`.

+1 we should get rid of ENABLE_JIT_CACHE as you mention. For disable_numba what if we disable it at first - use case "codecov", in our current implementatation we don't have disable_numba.

philipc2:
> > This is great. I assume these global variables in Numba need to be set before importing UXarray right?
> 
> Well, they aren't really global variables in Numba. There isn't much point in having the `disable_numba` function, as we have already shown once the function is loaded it won't change anything. That would be a global variable in Numba, but it doesn't help much once the function is loaded. And `ENABLE_JIT_CACHE` is just our own constant we have in the constants file. Honestly, I would be in favor of removing that as well and just rewriting all the functions with jit to have `@njit(cache=True)`. Since that is what we are doing anyway, I don't really see a reason to have an extra constant variable. We have never really even used it. It is basically the same as just saying `cache=True`.

Yeah, I agee and don't really see much purpose in having a constant. Some functions can benefit from caching, while others may not. This should be something that we handle on a per-function basis.

aaronzedwick:
> > > This is great. I assume these global variables in Numba need to be set before importing UXarray right?
> > 
> > 
> > Well, they aren't really global variables in Numba. There isn't much point in having the `disable_numba` function, as we have already shown once the function is loaded it won't change anything. That would be a global variable in Numba, but it doesn't help much once the function is loaded. And `ENABLE_JIT_CACHE` is just our own constant we have in the constants file. Honestly, I would be in favor of removing that as well and just rewriting all the functions with jit to have `@njit(cache=True)`. Since that is what we are doing anyway, I don't really see a reason to have an extra constant variable. We have never really even used it. It is basically the same as just saying `cache=True`.
> 
> +1 we should get rid of ENABLE_JIT_CACHE as you mention. For disable_numba what if we disable it at first - use case "codecov", in our current implementatation we don't have disable_numba.

Disabling numba doesn't seem practical, or even really possible. We would have to somehow disable it before all the tests ran (no idea if that is possible, I don't know that it is but I am not well informed on the specifics of that), and it would slow down all the test cases. The CI would take forever to run since numba would be off for everything.

rajeeja:
> > > > This is great. I assume these global variables in Numba need to be set before importing UXarray right?
> > > 
> > > 
> > > Well, they aren't really global variables in Numba. There isn't much point in having the `disable_numba` function, as we have already shown once the function is loaded it won't change anything. That would be a global variable in Numba, but it doesn't help much once the function is loaded. And `ENABLE_JIT_CACHE` is just our own constant we have in the constants file. Honestly, I would be in favor of removing that as well and just rewriting all the functions with jit to have `@njit(cache=True)`. Since that is what we are doing anyway, I don't really see a reason to have an extra constant variable. We have never really even used it. It is basically the same as just saying `cache=True`.
> > 
> > 
> > +1 we should get rid of ENABLE_JIT_CACHE as you mention. For disable_numba what if we disable it at first - use case "codecov", in our current implementatation we don't have disable_numba.
> 
> Disabling numba doesn't seem practical, or even really possible. We would have to somehow disable it before all the tests ran (no idea if that is possible, I don't know that it is but I am not well informed on the specifics of that), and it would slow down all the test cases. The CI would take forever to run since numba would be off for everything.

I'll try some test to see, but disabling should be possible, even if it slows down one use case would be to get correct codecov and another to just check if sometime Numba has some issues going on with conversion to byte code etc.

aaronzedwick:
> > > > > This is great. I assume these global variables in Numba need to be set before importing UXarray right?
> > > > 
> > > > 
> > > > Well, they aren't really global variables in Numba. There isn't much point in having the `disable_numba` function, as we have already shown once the function is loaded it won't change anything. That would be a global variable in Numba, but it doesn't help much once the function is loaded. And `ENABLE_JIT_CACHE` is just our own constant we have in the constants file. Honestly, I would be in favor of removing that as well and just rewriting all the functions with jit to have `@njit(cache=True)`. Since that is what we are doing anyway, I don't really see a reason to have an extra constant variable. We have never really even used it. It is basically the same as just saying `cache=True`.
> > > 
> > > 
> > > +1 we should get rid of ENABLE_JIT_CACHE as you mention. For disable_numba what if we disable it at first - use case "codecov", in our current implementatation we don't have disable_numba.
> > 
> > 
> > Disabling numba doesn't seem practical, or even really possible. We would have to somehow disable it before all the tests ran (no idea if that is possible, I don't know that it is but I am not well informed on the specifics of that), and it would slow down all the test cases. The CI would take forever to run since numba would be off for everything.
> 
> I'll try some test to see, but disabling should be possible, even if it slows down one use case would be to get correct codecov and another to just check if sometime Numba has some issues going on with conversion to byte code etc.

We did try already. Once the function is loaded, it doesn't matter if numba is disabled after the function is loaded with the wrapper. Numba is either off or on it seems, for the whole package, and you'd have to disable it before the tests ran somehow. At least based on our testing. You would have to rewrap the functions somehow if you wanted to turn them on and off mid-function. Just seems like a lot of work, just to increase code coverage. As long as we make sure we are covering with tests the functions that numba is on, I don't see it being that useful to turn numba on/off.

philipc2 (approved):
This was much needed. Thanks @aaronzedwick

aaronzedwick:
> This was much needed. Thanks @aaronzedwick

Do you know what’s up with the precommit? It works locally.

philipc2:
> > This was much needed. Thanks @aaronzedwick
> 
> Do you know what’s up with the precommit? It works locally.

Nothing on your end. See #991

PR #1011, opened 2024-10-14
Constant Longitude Cross Sections 

### LINKED ISSUE #1012: Constant Longitude Cross Sections
Compute the cross section at a line of constant longitude

philipc2:
Closes #1012

## Overview

* Implements constant longitude cross sections and associated helpers
  *  The cross sections are between both pole points at a given longitude line. They will not form a closed arc 
* Updates constant lon and lat intersection algorithm to consider the case when an edge lies exactly on the line


aaronzedwick:
@philipc2 I am a bit uneducated on what these cross sections are used for, are there any resources you could point me towards so I can learn about what they are before giving a review?

philipc2:
> @philipc2 I am a bit uneducated on what these cross sections are used for, are there any resources you could point me towards so I can learn about what they are before giving a review?

Generally speaking, the cross sections are used for selecting slices of data along lines of latitude, longitude, or arbitrary great circle arcs.

For example, if I wanted to compute the average temperate at 20 degrees latitude, we would need to perform a cross section operation.

rajeeja (commented):
Is it possible to add an example with a box (lon, lat) bounding box, what would that mean, some real world case?

rajeeja:
> > Is it possible to add an example with a box (lon, lat) bounding box, what would that mean, some real world case?
> 
> Can you elaborate?
> 
> The implementation here is for constant-longitude cross sections, so it can extract faces along a single line of longitude, not a bounding box. 

I was thinking four lines - specifying a specific geographical region and showing some event
..

erogluorhan (approved):
Thanks for doing this; it looks good to me!

rajeeja (approved):

Issue #1051, opened 2024-10-30
Implement Gnomonic Projection


aaronzedwick:
### Proposed new feature or change:

Projects a point from a spherical surface to a plane using gnomonic projection.

Issue #1050, opened 2024-10-30
Implement Stereographic Projection


aaronzedwick:
### Proposed new feature or change:

Projects a point from a spherical surface to a plane using stereographic projection.

PR #1052, opened 2024-10-30
Regional Delaunay 

### LINKED ISSUE #1049: Implement Regional Delaunay

Create Regional Delaunay triangles by projecting the grid to a plane, generating Delaunay triangles, and reverting back to the grid.

aaronzedwick:
Closes #1049 #1050 #1051 

## Overview

Adds `Grid.to_planar` which takes a grid and projects it to a plane, using `_point_to_plane` which converts a point on the sphere to the plane. Also adds `_point_to_sphere` which converts a point on a plane back to the sphere.
## Expected Usage

```Python
import uxarray as ux

grid_points = [[10, -10, 10, -10], [-10, -10, 10, 10]]

grid_r = ux.Grid.from_points(grid_points, method="regional_delaunay")

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**

**Examples**
- [x] Any new notebook examples added to `docs/examples/` folder
- [x] Clear the output of all cells before committing
- [x] New notebook files added to `docs/examples.rst` toctree
- [x] New notebook files added to new entry in `docs/gallery.yml` with appropriate thumbnail photo in `docs/_static/thumbnails/`


aaronzedwick:
@philipc2 got any suggestions on how I might store this information? I am not completely certain about a solid way to do so.

philipc2:
> @philipc2 got any suggestions on how I might store this information? I am not completely certain about a solid way to do so.

Do these variables need to be cached, or can they be used once and then scrapped? Considering the computation of these variables will be pretty quick thanks to Numba, I lean towards not storing them directly as part of a `Grid`.

aaronzedwick:
> > @philipc2 got any suggestions on how I might store this information? I am not completely certain about a solid way to do so.
> 
> Do these variables need to be cached, or can they be used once and then scrapped? Considering the computation of these variables will be pretty quick thanks to Numba, I lean towards not storing them directly as part of a `Grid`.

Sorry, I think I miscommunication. I more meant, how should I return this information? Is using a dataset the best way, and constructing them all at once? Or should I have one function for each (node, face, edge) and return a np array?

philipc2:
> > > @philipc2 got any suggestions on how I might store this information? I am not completely certain about a solid way to do so.
> > 
> > 
> > Do these variables need to be cached, or can they be used once and then scrapped? Considering the computation of these variables will be pretty quick thanks to Numba, I lean towards not storing them directly as part of a `Grid`.
> 
> Sorry, I think I miscommunication. I more meant, how should I return this information? Is using a dataset the best way, and constructing them all at once? Or should I have one function for each (node, face, edge) and return a np array?

I see! I'd lean towards not having a `Grid.to_planar()` method and instead having the `_point_to_plane()` and `_point_to_sphere()` called as needed, returning the variables instead of storing them as a dataset.

uxarray/grid/geometry.py
```diff
@@ -1336,3 +1337,34 @@ def _construct_boundary_edge_indices(edge_face_connectivity):
     # If an edge only has one face saddling it than the mesh has holes in it
     edge_with_holes = np.where(edge_face_connectivity[:, 1] == INT_FILL_VALUE)[0]
     return edge_with_holes
+
+
+# @njit(cache=True)
```
philipc2, line 1342: Did you comment this out when debugging?

philipc2 (changes requested):
Can you test the behavior if the points are along the antemeridian? If you want to use data from the usage example, you could try the following. 

`uxgrid_regional.subset.bounding_circle((-180, 0), 10)`

philipc2 (commented):
Can you also expand upon the markdown in the notebook? I think the most important addition would be the "How It Works" section that describes the underlying algorithm. 

image

uxarray/grid/geometry.py
aaronzedwick: Oh, yeah I did.

aaronzedwick:
@philipc2 This projection technique (sterographic) can't handle south pole points, or rather any points along the z=1 axis. I was thinking maybe using a gnomonic projection would work better and be more accurate (as this one seems to not work very well for point in polygon either, the projection is distorted and gives incorrect results when the point is close to the edge). The gnomonic projection might have the same issue at the poles though, not certain. Thoughts? Do you think gnomonic projection would give more accurate results?

rajeeja (changes requested):
can you do something like `norm = np.linalg.norm(points, axis=1)` to calculate all norm at once?

also, is might faster to use set operation for filtering.

uxarray/grid/grid.py
```diff
@@ -387,6 +390,8 @@ def from_points(
             ds = _spherical_voronoi_from_points(_points, **kwargs)
         elif method == "spherical_delaunay":
             ds = _spherical_delaunay_from_points(_points, boundary_points)
+        elif method == "regional_delaunay":
+            ds = _regional_delaunay_from_points(_points, boundary_points)
```
philipc2, line 394: Please update the method parameter in the docstring. 

```
        method : str, optional
            The method to generate connectivity information. Supported methods are:

            - `'spherical_voronoi'`: Constructs a spherical Voronoi diagram.
            - `'spherical_delaunay'`: Constructs a spherical Delaunay triangulation.

            Default is `'spherical_delaunay'`.
```

philipc2 (changes requested):
Notebook changes are looking great!

One small comment below.

philipc2 (approved):
Great work @aaronzedwick

rajeeja (approved):

PR #1056, opened 2024-11-05
Point In Face & Get Faces Containing Point

### LINKED ISSUE #905: Point in Face

Add a function to check if a point is inside a polygon. Currently a pole point in polygon is implemented, possibly modify it and make a similar function? That function makes a reference edge using the point and the equator, which wouldn't work with all points.

aaronzedwick:
Adds several new features:
* `point_in_face` function to determine if a point is in a face.
*  `get_faces_containing_point` which given a point finds the face (or faces) the point resides in. 
* `max_face_radius` which finds the largest face radius and stores it within the grid. 
* `haversine_distance` function which finds the haversine distance between two given points.
* A few fixes to coordinate construction bugs

Closes #905, #1141

## Expected Usage

```Python
from uxarray.grid.geometry import point_in_polygon

# Defined polygon
polygon = [ [-10,  10, -10, 10], [10, 10, -10, -10]]

# Point to check
point = [10, 10]

point_in_polygon(polygon, point, inclusive=True)

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


philipc2 (changes requested):
Please include an ASV benchmark. I'd suggest doing a parameterized benchmark for the 120 and 480 km MPAS grids.

uxarray/grid/geometry.py
```diff
@@ -1574,3 +1583,88 @@ def inverse_stereographic_projection(x, y, central_lon, central_lat):
     )
 
     return lon, lat
+
+
+def point_in_polygon(polygon, point, ref_point, inclusive=True):
```
philipc2, line 1588: These functions will be called thousands, if not millions of times for many operations. Can you re-write this in Numba?

philipc2 (changes requested):
Please use the optimized functions from https://github.com/UXARRAY/uxarray/pull/1072 and try to write the function entirely in Numba. This may require us to pass in both the cartesian and spherical versions of point & polygon. Let me know if you have any questions!

uxarray/grid/geometry.py
aaronzedwick: When you say re-write in Numba, could you clarify? Do you mean to wrap the function in Numba, or do you mean something more?

philipc2: I'd start with wrapping it with `njit` and then seeing what modifications need to be made to be compatible. You can use the PR that I linked above with the re-writes in Numbs as a reference.

aaronzedwick:
> Please use the optimized functions from #1072 and try to write the function entirely in Numba. This may require us to pass in both the cartesian and spherical versions of point & polygon. Let me know if you have any questions!

Would it be okay to convert to spherical or cartesian inside the function based on what the user inputs, or should I require both?

philipc2:
> > Please use the optimized functions from #1072 and try to write the function entirely in Numba. This may require us to pass in both the cartesian and spherical versions of point & polygon. Let me know if you have any questions!
> 
> Would it be okay to convert to spherical or cartesian inside the function based on what the user inputs, or should I require both?

Requiring both up front would be better. In #1072, I observed that the conversion would take up about 30-40 percent of the total execution time. It would be much more practical to do a one-time conversion before calling `point_in_polygon`.

You can however introduce compatibility functions for testing (i.e. ones that have the option for conversion, see https://github.com/UXARRAY/uxarray/blob/41efd5c68d3d201e736875f1ac4e9c3fbb0a9ea0/uxarray/grid/intersections.py#L203-L227)

uxarray/grid/geometry.py
philipc2, line 1588: Wonder if `point_in_face` might be a more appropriate name?

philipc2:
The `point_in_polygon` functionality is going to be a great addition. 

Can we implement a `Grid.get_faces_containing_point()` function that returns the indices of the faces that contain a given point? 

```Python
uxgrid = ... 

# return the indicies of faces that contain the point (0, 90)
face_indices = uxgrid.get_faces_containing_point((0.0, 90.0))
```

I believe we have discussed this before, and a good approach would be to find the maximum radius of each face and use that to query our BallTree of the face centers. This should narrow down the possible candidates instead of needing to run `point_in_polygon` on the entire `Grid` each time.

hongyuchen1030:
> > Please use the optimized functions from #1072 and try to write the function entirely in Numba. This may require us to pass in both the cartesian and spherical versions of point & polygon. Let me know if you have any questions!
> 
> Would it be okay to convert to spherical or cartesian inside the function based on what the user inputs, or should I require both?

I recommend minimizing coordinate conversions as much as possible, as they can introduce unnecessary floating-point errors. Since we already construct and store both Cartesian and spherical coordinates internally in the data array, it’s better to stick to one coordinate system. If the other system is needed, we should first retrieve it from the stored conversions rather than recomputing it.

aaronzedwick:
> The `point_in_polygon` functionality is going to be a great addition.
> 
> Can we implement a `Grid.get_faces_containing_point()` function that returns the indices of the faces that contain a given point?
> 
> ```python
> uxgrid = ... 
> 
> # return the indicies of faces that contain the point (0, 90)
> face_indices = uxgrid.get_faces_containing_point((0.0, 90.0))
> ```
> 
> I believe we have discussed this before, and a good approach would be to find the maximum radius of each face and use that to query our BallTree of the face centers. This should narrow down the possible candidates instead of needing to run `point_in_polygon` on the entire `Grid` each time.

This would return a single face always correct? Since faces don't overlap.

philipc2:
> > The `point_in_polygon` functionality is going to be a great addition.
> > Can we implement a `Grid.get_faces_containing_point()` function that returns the indices of the faces that contain a given point?
> > ```python
> > uxgrid = ... 
> > 
> > # return the indicies of faces that contain the point (0, 90)
> > face_indices = uxgrid.get_faces_containing_point((0.0, 90.0))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I believe we have discussed this before, and a good approach would be to find the maximum radius of each face and use that to query our BallTree of the face centers. This should narrow down the possible candidates instead of needing to run `point_in_polygon` on the entire `Grid` each time.
> 
> This would return a single face always correct? Since faces don't overlap.

The majority of the time it would only return one, however for the case where the point is exactly on the nodes or edges (i.e. `inclusive=True`) there would be multiple indices returned, since those can be shared between multiple faces.

aaronzedwick:
> > > The `point_in_polygon` functionality is going to be a great addition.
> > > Can we implement a `Grid.get_faces_containing_point()` function that returns the indices of the faces that contain a given point?
> > > ```python
> > > uxgrid = ... 
> > > 
> > > # return the indicies of faces that contain the point (0, 90)
> > > face_indices = uxgrid.get_faces_containing_point((0.0, 90.0))
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I believe we have discussed this before, and a good approach would be to find the maximum radius of each face and use that to query our BallTree of the face centers. This should narrow down the possible candidates instead of needing to run `point_in_polygon` on the entire `Grid` each time.
> > 
> > 
> > This would return a single face always correct? Since faces don't overlap.
> 
> The majority of the time it would only return one, however for the case where the point is exactly on the nodes or edges (i.e. `inclusive=True`) there would be multiple indices returned, since those can be shared between multiple faces.

Ahh, that is true I hadn't thought of that. Thanks for the clarification, I will work on adding that to this PR

aaronzedwick:
@hongyuchen1030 Could you give the point in the polygon function a review? The first test case `test_point_inside` is the main test I have been using, which is failing. Philip and I tried debugging it yesterday, but couldn't find any issue with my code. We wanted to see if you saw any obvious misuse of the intersection code.

hongyuchen1030:
> @hongyuchen1030 Could you give the point in the polygon function a review? The first test case `test_point_inside` is the main test I have been using, which is failing. Philip and I tried debugging it yesterday, but couldn't find any issue with my code. We wanted to see if you saw any obvious misuse of the intersection code.

Looking into it right now

aaronzedwick:
> > @hongyuchen1030 Could you give the point in the polygon function a review? The first test case `test_point_inside` is the main test I have been using, which is failing. Philip and I tried debugging it yesterday, but couldn't find any issue with my code. We wanted to see if you saw any obvious misuse of the intersection code.
> 
> Looking into it right now
> 
> My error messages are
> 
> ```python
>         # Assert that the point is not in the polygon
>         self.assertFalse(
> >           point_in_polygon(polygon_xyz, polygon_lonlat, point_xyz, point_lonlat, ref_point_xyz, ref_point_lonlat))
> E       TypeError: point_in_polygon() takes from 4 to 5 positional arguments but 6 were given
> 
> test_geometry.py:1623: TypeError
> ```
> 
> And only the first one, `test_geometry.py::TestPointInPolygon::test_point_inside ` is passing, all others fail

The other test cases are not updated to handle the new conditions of passing in both lon and lat. The passing test case is only passing because I commented out the assertion statement. Sorry for the confusion.

hongyuchen1030:
> > > @hongyuchen1030 Could you give the point in the polygon function a review? The first test case `test_point_inside` is the main test I have been using, which is failing. Philip and I tried debugging it yesterday, but couldn't find any issue with my code. We wanted to see if you saw any obvious misuse of the intersection code.
> > 
> > 
> > Looking into it right now
> > My error messages are
> > ```python
> >         # Assert that the point is not in the polygon
> >         self.assertFalse(
> > >           point_in_polygon(polygon_xyz, polygon_lonlat, point_xyz, point_lonlat, ref_point_xyz, ref_point_lonlat))
> > E       TypeError: point_in_polygon() takes from 4 to 5 positional arguments but 6 were given
> > 
> > test_geometry.py:1623: TypeError
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And only the first one, `test_geometry.py::TestPointInPolygon::test_point_inside ` is passing, all others fail
> 
> The other test cases are not updated to handle the new conditions of passing in both lon and lat. The passing test case is only passing because I commented out the assertion statement. Sorry for the confusion.

By any chance you know which specific face is not passing?

aaronzedwick:
> > > > @hongyuchen1030 Could you give the point in the polygon function a review? The first test case `test_point_inside` is the main test I have been using, which is failing. Philip and I tried debugging it yesterday, but couldn't find any issue with my code. We wanted to see if you saw any obvious misuse of the intersection code.
> > > 
> > > 
> > > Looking into it right now
> > > My error messages are
> > > ```python
> > >         # Assert that the point is not in the polygon
> > >         self.assertFalse(
> > > >           point_in_polygon(polygon_xyz, polygon_lonlat, point_xyz, point_lonlat, ref_point_xyz, ref_point_lonlat))
> > > E       TypeError: point_in_polygon() takes from 4 to 5 positional arguments but 6 were given
> > > 
> > > test_geometry.py:1623: TypeError
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And only the first one, `test_geometry.py::TestPointInPolygon::test_point_inside ` is passing, all others fail
> > 
> > 
> > The other test cases are not updated to handle the new conditions of passing in both lon and lat. The passing test case is only passing because I commented out the assertion statement. Sorry for the confusion.
> 
> By any chance you know which specific face is not passing?

Half the faces don't pass, and the other half do. The first 3 at least fail, if I remember correctly. Interestingly, with your changes on a gca_gca intersection to make it cartesian only, it works better if the reference point is changed to be next to the pole but not the pole itself. However, there hasn't been any improvement when doing the same thing with the current implementation of the intersection code. So something seems to have changed from the changes you are making in that PR.

philipc2:
@hongyuchen1030 

https://github.com/UXARRAY/uxarray/blob/f783f612c0dc2d6a42f93d4621789e7376327fc5/test/test_intersections.py#L95-L114

I've added the test case here in #1112 

The correct intersection is computed up to `-89.9999999999999`, but after that it returns no intersection, such as when the point is set to exactly `-90.0`

aaronzedwick:
> @hongyuchen1030
> 
> https://github.com/UXARRAY/uxarray/blob/f783f612c0dc2d6a42f93d4621789e7376327fc5/test/test_intersections.py#L95-L114
> 
> I've added the test case here in #1112
> 
> The correct intersection is computed up to `-89.9999999999999`, but after that it returns no intersection, such as when the point is set to exactly `-90.0`

Thanks for this Philip. This is in the updated PR correct? Not the main branch?

aaronzedwick:
@hongyuchen1030 using your branch and the code from this branch, it works properly when the reference point is not on the pole, so whatever the issue seems to be fixed in that PR. Thanks for taking the time to look at this for me though!

hongyuchen1030:
> @hongyuchen1030 using your branch and the code from this branch, it works properly when the reference point is not on the pole, so whatever the issue seems to be fixed in that PR. Thanks for taking the time to look at this for me though!

I think me and @philipc2 had resolved this through the recently merged branch #1112 , Thanks again for @philipc2's help. With the updated algorithm, it should work for the the pole point as well now.

aaronzedwick:
@philipc2 do you know why this is failing? It doesn't have any issues locally when I run the tests, but for some reason, it fails on the CI, saying that it can't resize the array. Could you checkout the branch and try running the test cases?

Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found. I am pretty confused by that as well. So any insight into that would be appreciated if you have any ideas.

hongyuchen1030:
> @philipc2 do you know why this is failing? It doesn't have any issues locally when I run the tests, but for some reason, it fails on the CI, saying that it can't resize the array. Could you checkout the branch and try running the test cases?
> 
> Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found. I am pretty confused by that as well. So any insight into that would be appreciated if you have any ideas.

Can you point us to the specific testcase that's been failing? 

>Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found

Are you using the same setting? Did you clear the cache everytime you run it?

aaronzedwick:
> > @philipc2 do you know why this is failing? It doesn't have any issues locally when I run the tests, but for some reason, it fails on the CI, saying that it can't resize the array. Could you checkout the branch and try running the test cases?
> > Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found. I am pretty confused by that as well. So any insight into that would be appreciated if you have any ideas.
> 
> Can you point us to the specific testcase that's been failing?
> 
> > Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found
> 
> Are you using the same setting? Did you clear the cache everytime you run it?

The test cases that are failing are just the ones on the CI, which are listed under it if you click there, they are the two test cases I added for the point containing polygon grid function. 

Everything should be the same across environments, that I know of. That’s why I wanted to see if it passed or failed on someone else’s machine, to hopefully point me in some direction of what the error could be.

philipc2:
> > > @philipc2 do you know why this is failing? It doesn't have any issues locally when I run the tests, but for some reason, it fails on the CI, saying that it can't resize the array. Could you checkout the branch and try running the test cases?
> > > Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found. I am pretty confused by that as well. So any insight into that would be appreciated if you have any ideas.
> > 
> > 
> > Can you point us to the specific testcase that's been failing?
> > > Also, another odd problem is that half of the time the tests do fail locally, saying no faces were found in the subset, but the other half of the time faces are found
> > 
> > 
> > Are you using the same setting? Did you clear the cache everytime you run it?
> 
> The test cases that are failing are just the ones on the CI, which are listed under it if you click there, they are the two test cases I added for the point containing polygon grid function.
> 
> Everything should be the same across environments, that I know of. That’s why I wanted to see if it passed or failed on someone else’s machine, to hopefully point me in some direction of what the error could be.

Can you share your environment information? (Python Version & Installed packages)

philipc2:
The tests are failing on my machine as well.

hongyuchen1030:
> The test cases that are failing are just the ones on the CI, which are listed under it if you click there, they are the two test cases I added for the point containing polygon grid function.

>The test cases that are failing are just the ones on the CI, which are listed under it if you click there, they are the two test cases I added for the point containing polygon grid function.

I didn't see any `no faces found` on the CI. My point is: does your local fails at the same point as your CI?

uxarray/grid/grid.py
```diff
@@ -2414,3 +2418,52 @@ def get_faces_at_constant_longitude(self, lon: float):
...
+            r=max_face_radius,
+            element="face centers",
+            inverse_indices=True,
+        )
+
+        # Get the face's edges for the whole subset
+        face_edge_cartesian = _get_cartesian_face_edge_nodes(
+            subset.face_node_connectivity.values,
+            subset.n_face,
+            subset.n_max_face_edges,
```
philipc2, line 2439: This appears to be the issue here. 

If you pass in `subset.n_max_face_nodes` it works correctly (both values are equivalent).

When the connectivity construction for `self.face_edge_connectivity`, it leads to issues. I am looking deeper into it.

aaronzedwick:
> > The test cases that are failing are just the ones on the CI, which are listed under it if you click there, they are the two test cases I added for the point containing polygon grid function.
> 
> > The test cases that are failing are just the ones on the CI, which are listed under it if you click there, they are the two test cases I added for the point containing polygon grid function.
> 
> I didn't see any `no faces found` on the CI. My point is: does your local fails at the same point as your CI?

Ah, I see what you mean, no, my local is failing at a different point. But on the same test cases. I will try and reset my environment to make sure it's up to date.

uxarray/grid/grid.py
aaronzedwick: Ah, I see. So something with constructing the face_edge causes issues. Thanks for the fix for now, let me know if you figure out the issue. From what I saw, it seems to have something to do with the inverse indices, it uses the wrong ones, the ones gotten from the subset, instead of the ones stored inside `grid._ds["edge_node_connectivity"].attrs`. Maybe it is getting overwritten somehow?

philipc2: The inverse indices in the context of `edge_node_connectivity` are different than the `inverse_indices` from the subset. I don't believe that they are overwritten, but something is definitely off.

test/test_grid.py
```diff
@@ -726,9 +758,39 @@ def test_normalize_existing_coordinates_non_norm_initial():
...
+    point_xyz = np.array([grid.face_x[100].values, grid.face_y[100].values, grid.face_z[100].values])
+
+    assert len(grid.get_faces_containing_point(point_xyz=point_xyz)) == 1
+
+    # For an edge two faces should be found
+    point_xyz = np.array([grid.edge_x[100].values, grid.edge_y[100].values, grid.edge_z[100].values])
+
+    assert len(grid.get_faces_containing_point(point_xyz=point_xyz)) == 2
+
+    # For a node three faces should be found
```
rajeeja, line 783: This is not always true, a node can have 1, 2, 3 or higher number of faces. 
For 1, consider a node on the boundary, not on any other face.

For 2, consider a node on the boundary that is the actual node on the face, on both faces.

For 4/higher, consider a node that is the intersection of 4 faces or higher.

rajeeja (changes requested):

test/test_grid.py
aaronzedwick: That is true, in general. However this is a MPAS grid. Which is always going to have each node connected to 3 faces. So I think it’s an OK assumption here that it should result in 3 faces being found.

rajeeja: > That is true, in general. However this is a MPAS grid. Which is always going to have each node connected to 3 faces. So I think it’s an OK assumption here that it should result in 3 faces being found.

Ok, you can keep it, but if the grid has holes and the node is along the boundary. I think, it'd be good to test these edge cases.

uxarray/grid/grid.py
```diff
@@ -2415,3 +2426,33 @@ def get_faces_at_constant_longitude(self, lon: float):
...
+
+        # Get the face's edges for the whole subset
+        face_edge_cartesian = _get_cartesian_face_edge_nodes(
+            subset.face_node_connectivity.values,
+            subset.n_face,
+            subset.n_max_face_nodes,
+            subset.node_x.values,
+            subset.node_y.values,
+            subset.node_z.values,
+        )
```
philipc2, line 2451: This is very expensive to re-compute each time we call this function. Consider adding an internal property or storing this in `Grid._ds` like we do with our other methods.

uxarray/grid/geometry.py
```diff
@@ -1561,3 +1568,174 @@ def inverse_stereographic_projection(x, y, central_lon, central_lat):
...
+        )
+
+        # If the point is, add it to the list of faces
+        if contains_point:
+            index.append(inverse_indices[i])
+
+    return index
+
+
+def get_max_face_radius(self):
```
philipc2, line 1676: This should be `_populate_max_face_radius` similar to our other population methods.

philipc2 (changes requested):

uxarray/grid/grid.py
aaronzedwick: I can do that, but it is a new subset each time the operation is run. So won't it be recomputed anyway? It should only be a face or a few at max for each subset.

philipc2: A more efficient approach would be to store it in the original `Grid`, where it can be computed once, and then once we subset, it should return a `Grid` with the `face_edge_nodes_xyx` appropriately indexed.

Though, if it is only a few faces at most, the perfromance may be negligible, but we need to consider the cases where we may be running millions of point in polygon queries for higher-resolution grids.

benchmarks/mpas_ocean.py
```diff
@@ -183,3 +183,10 @@ def teardown(self, resolution, lat_step):
     def time_const_lat(self, resolution, lat_step):
         for lat in self.lats:
             self.uxgrid.cross_section.constant_latitude(lat)
+
+
+class PointInPolygon(GridBenchmark):
+    def time_whole_grid(self, resolution):
+        point_xyz = np.array([self.uxgrid.face_x[0].values, self.uxgrid.face_y[0].values, self.uxgrid.face_z[0].values])
+
+        self.uxgrid.get_faces_containing_point(point_xyz=point_xyz)
```
philipc2, line 192: From the GH Actions report, it appears that a single Point in Face query is taking approximately `27.0±0.1ms`

Ideally, we want to get this number significantly down.

philipc2: Did some digging:

Please include the following in a custom setup function:
```
_ = uxds.uxgrid.face_edge_connectivity
```

We don't want this to be included in the timing. I believe much of the 27ms is attributed to that. 

This is for a 30km Grid

image

philipc2: This is the performance I get for a 30km grid after doing the things mentioned above. pretty good for 30km grids

image

Only issue, is that it doesn't find any faces.

Here is my sample code I used.

```Python
import uxarray as ux
import numpy as np

import cProfile

profiler = cProfile.Profile()

grid_path = "/Users/philipc/PycharmProjects/uxarray/unstructured-grid-viz-cookbook/meshfiles/x1.655362.grid.nc"
data_path = "/Users/philipc/PycharmProjects/uxarray/unstructured-grid-viz-cookbook/meshfiles/x1.655362.data.nc"

uxds.uxgrid.normalize_cartesian_coordinates()

_ = uxds.uxgrid.face_edge_connectivity

point = np.array([0.0, 0.0, 1.0])
res = uxds.uxgrid.get_faces_containing_point(point)


profiler.enable()
res = uxds.uxgrid.get_faces_containing_point(point)
profiler.disable()

profiler.dump_stats('pface.prof')

print(res)
```

Can do `snakeviz pface.prof` to view the profiler.

uxarray/grid/grid.py
aaronzedwick: Oh, okay yeah that would work. I didn't realize it would keep it through the subset operation.

benchmarks/mpas_ocean.py
aaronzedwick: Thanks for looking into this for me! I am trying to implement these changes now. In terms of it not finding any faces, does the mesh have holes in it? There may not be a face at the pole possibly?

philipc2: The grid is a global MPAS atmosphere grid with no holes.

philipc2: I may have been on an older version of the code. Just ran it again after pulling and it works. A single face is detected. Going to run a few more tests.

aaronzedwick: Wait is it working locally for you? It does for me, but the CI is failing. Not sure why.

philipc2: It seems like it is working for the poles at least on my end. Haven't tested other points

rajeeja:
Try to print array shape and check why is it assesing [this](https://github.com/UXARRAY/uxarray/actions/runs/12789386769/job/35652701714?pr=1056#step:9:586)? 

works on my local also, it must be something with python 3.12/macos-14, we need to fix and can't ignore this, might uncover some bug.

uxarray/grid/grid.py
aaronzedwick: So the subset actually keeps the entire array for the whole grid. I can still make it work, that way, but that is why the tests where failing.

philipc2: > So the subset actually keeps the entire array for the whole grid. I can still make it work, that way, but that is why the tests where failing.

Yeah, we should always ensure that the subset never contains the entire array for any variable.

philipc2 (changes requested):
Please add the following to our API
`Grid.max_face_radius`
`Grid.get_faces_containing_point`

Lets not include `Grid.face_edge_nodes_xyz` in the API for now.

uxarray/grid/grid.py
```diff
@@ -2415,3 +2443,26 @@ def get_faces_at_constant_longitude(self, lon: float):
 
         faces = constant_lon_intersections_face_bounds(lon, self.face_bounds_lon.values)
         return faces
+
+    def get_faces_containing_point(self, point_xyz):
```
philipc2, line 2447: We should change `point_xyz` to `point`, which can be either cartesian or spherical. Many times the user will want to query a spherical point, which we can convert to x, y, z.

uxarray/grid/grid.py
```diff
@@ -1546,6 +1552,32 @@ def is_subset(self):
...
+                self.n_max_face_nodes,
+                self.node_x.values,
+                self.node_y.values,
+                self.node_z.values,
+            )
+            self._ds["face_edge_nodes_xyz"] = (
+                [self.n_face, self.n_max_face_edges, 2, 3],
+                face_edge,
+            )
+        return self._ds["face_edge_nodes_xyz"]
```
philipc2, line 1579: See the discussion below. 

https://github.com/UXARRAY/uxarray/pull/1130#discussion_r1924731908

aaronzedwick: Okay, I can use `face_edge_connectivity`. Should I remove this from the grid properties altogether?

uxarray/grid/grid.py
```diff
@@ -2449,3 +2481,56 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        elif len(point) == 3:
+            point_xyz = point
+            point_lonlat = np.array(_xyz_to_lonlat_deg(*point_xyz), dtype=np.float64)
+        else:
+            raise ValueError(
+                "Point must either be in spherical or cartesian coordinates."
+            )
+
+        # Get the maximum face radius of the grid
+        _ = self.face_edge_nodes_xyz
```
philipc2, line 2511: We shouldn't need this call, only to construct it for the subset below, right?

philipc2: Sounds good, it'll be much faster to just construct it for the subset.

aaronzedwick: You mentioned in a previous review comment I should pre-construct it for the whole grid, that way for larger grids it’s only constructed once. Since in bilinear remapping I call this function for every point. Do you think it would still be better for the more broader use case to construct it on the subset?

rajeeja (changes requested):
`point_in_face(faces_edge_cartesian[0], point, inclusive=True)`

this always requires calculation of edges, can we have another version of this - 

1. `point_in_face(face_id, point` or 
2. `point_in_face(face_conn, point)`

For the notebook example, we have 4 faces with id's say 0 to 4 and connectivity as shown below:
```

xarray.DataArray'face_node_connectivity'n_face: 4n_max_face_nodes: 6

array([[ 0,  1,  2,  3,  4,  5],
       [15,  7,  6, 12,  0,  5],
       [ 0, 12,  8,  9, 13,  1],
       [ 4, 14, 11, 10, 15,  5]])
```

getting edges adds another step that might not always be necessary and can be done internally for these other signatures, the two functions above would be more user friendly IMO.

aaronzedwick:
> `point_in_face(faces_edge_cartesian[0], point, inclusive=True)`
> 
> this always requires calculation of edges, can we have another version of this -
> 
> 1. `point_in_face(face_id, point` or
> 2. `point_in_face(face_conn, point)`
> 
> For the notebook example, we have 4 faces with id's say 0 to 4 and connectivity as shown below:
> 
> ```
> 
> xarray.DataArray'face_node_connectivity'n_face: 4n_max_face_nodes: 6
> 
> array([[ 0,  1,  2,  3,  4,  5],
>        [15,  7,  6, 12,  0,  5],
>        [ 0, 12,  8,  9, 13,  1],
>        [ 4, 14, 11, 10, 15,  5]])
> ```
> 
> getting edges adds another step that might not always be necessary and can be done internally for these other signatures, the two functions above would be more user friendly IMO.

This function is not a grid function, so passing just the face connectivity isn’t enough. You have to pass the node coordinates of the face. Either way, you have to parse the coordinates of the face you want to use. 

We can pass the node coordinates as nodes into it if you like, but I originally went with this implementation for two reasons. First, the conversion will in theory impact performance. And second, if the user has to construct the node coordinates and send them in anyway, it would require less effort to use the edge coordinates Cartesian function to create the array for the user. It is easier and ensures the correct array shape is being passed to the function. Also, that is the shape of array the ‘faces_containing_point’ use, as well as all of Hongyu’s code.

uxarray/grid/grid.py
aaronzedwick: @philipc2 could you clarify more here on what you want me to do about this?

rajeeja (approved):
bug fix for conversion _lonlat_rad_to_xyz in coordinates.py, max_face_radius and get_faces_containing_points are critical changes. good work

uxarray/grid/grid.py
philipc2: I think we should remove it.

philipc2 (changes requested):
Please fix the ASV benchmark, it is still failing on our CI.

benchmarks/mpas_ocean.py
```diff
@@ -194,6 +194,28 @@ def time_const_lat(self, resolution, lat_step):
...
+        point = np.array([0.0, 0.0, 1.0])
+        res = self.uxgrid.get_faces_containing_point(point)
+
+    def teardown(self, resolution):
+        del self.uxgrid
+
+    def time_whole_grid(self, resolution):
+        point_xyz = np.array([self.uxgrid.face_x[0].values, self.uxgrid.face_y[0].values, self.uxgrid.face_z[0].values])
+
+        self.uxgrid.get_faces_containing_point(point_xyz=point_xyz)
```
philipc2, line 216: File "/home/runner/work/uxarray/uxarray/benchmarks/mpas_ocean.py", line 216, in time_whole_grid
                 self.uxgrid.get_faces_containing_point(point_xyz=point_xyz)
             TypeError: Grid.get_faces_containing_point() got an unexpected keyword argument 'point_xyz'

aaronzedwick: Oh good catch, I changed the argument name to just `point`.

uxarray/grid/geometry.py
```diff
@@ -1561,3 +1568,234 @@ def inverse_stereographic_projection(x, y, central_lon, central_lat):
...
+        distances = haversine_distance(node_lon_rads, node_lat_rads, face_lon, face_lat)
+
+        # Store the max distance for this face
+        end_distances[ind] = np.max(distances)
+
+    # Return the maximum distance found across all faces
+    return np.max(end_distances)
+
+
+# @njit(cache=True)
```
philipc2, line 1769: Did you mean to comment out all the numba portions?

aaronzedwick: Yeah, they were causing caching issues locally; the changes I made were not being reflected when the functions were run with numba on. They should be uncommented out before the final merge. I can do that with the next commit.

philipc2:
Some perfromance issues, but they appear to be due to our `Grid.isel()`, which I am already working on optimizations for.

This is 10,000 queries on a 30km MPAS grid

image

rajeeja:
> Some perfromance issues, but they appear to be due to our `Grid.isel()`, which I am already working on optimizations for.
> 
> This is 10,000 queries on a 30km MPAS grid
> 
> image

do you generate this locally or asv generates it in some automated fashion?

philipc2:
> > Some perfromance issues, but they appear to be due to our `Grid.isel()`, which I am already working on optimizations for.
> > This is 10,000 queries on a 30km MPAS grid
> > image
> 
> do you generate this locally or asv generates it in some automated fashion?

```Python
import uxarray as ux
import numpy as np
import numba as nb

print(nb.__version__)
import cProfile

profiler = cProfile.Profile()


grid_path = "/Users/philipc/PycharmProjects/uxarray/unstructured-grid-viz-cookbook/meshfiles/x1.655362.grid.nc"

uxgrid=ux.open_grid(grid_path)

face_lon = uxgrid.face_lon.values.astype(np.float64)
face_lat = uxgrid.face_lat.values.astype(np.float64)



profiler.enable()
# Profile something here
for i in range(10000):
    point = np.array([face_lon[i], face_lat[i]])
    res = uxgrid.get_faces_containing_point(point)
profiler.disable()

# Save stats
profiler.dump_stats('point_in_polygon_10000.prof')
```

Then I ran `snakeviz point_in_polygon_10000.prof`

philipc2 (commented):
Can we include a test for points along a line of constant latitude?

```Python
    def test_point_along_arc():
        node_lon = np.array([-40, -40, 40, 40])
        node_lat = np.array([-20, 20, 20, -20])
        face_node_connectivity = np.array([[0, 1, 2, 3]], dtype=np.int64)

        uxgrid = ux.Grid.from_topology(node_lon, node_lat, face_node_connectivity)

        # point at exactly 20 degrees latitude
        out1 = uxgrid.get_faces_containing_point((0, 20))

        # point at 25.41 degrees latitude (max along the great circle arc)
        out2 = uxgrid.get_faces_containing_point((0, 25.41))

        nt.assert_array_equal(out1, out2)

File: uxarray/grid/grid.py [hunk 12]
```diff
@@ -2446,3 +2478,56 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        if len(point) == 2:
+            point_xyz = np.array(_lonlat_rad_to_xyz(*np.deg2rad(point)))
+            point_lonlat = point
+        elif len(point) == 3:
+            point_xyz = point
+            point_lonlat = np.array(_xyz_to_lonlat_deg(*point_xyz), dtype=np.float64)
+        else:
+            raise ValueError(
+                "Point must either be in spherical or cartesian coordinates."
+            )
```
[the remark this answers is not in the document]
aaronzedwick: In some of my tests I have found that using lonlat with a `dtype=np.float64` can cause issues on the nodes, it will not find all the faces nearby. I think this is possible due to rounding precission issues maybe?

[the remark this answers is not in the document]
philipc2: Casting to a higher precision shouldn't usually lead to issues.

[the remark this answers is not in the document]
aaronzedwick: It does, doing `dtype=np.float32` ensures that Hongyus algorithm recognizes that it is on the great circle arc. But when it is `float64` it doesn't seem to realize it is on the arc.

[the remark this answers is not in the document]
aaronzedwick: image
image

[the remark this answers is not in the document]
aaronzedwick: As you can see, in one case all three faces are found, in another only one is found. I do think this is only an issue on the nodes, I could put in a check where if the point is a node, I skip checking for point in polygon and just have it use the faces that the node is part of.

uxarray/grid/grid.py
```diff
@@ -2446,3 +2459,65 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
 
         """
         return faces_within_lat_bounds(lats, self.face_bounds_lat.values)
+
+    def get_faces_containing_point(self, point):
+        """Gets the indexes of the faces that contain a specific point
```
philipc2, line 2464: Identifies the indices of faces that contain a given point.

uxarray/grid/grid.py
```diff
@@ -2446,3 +2459,65 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+            subset.node_x.values,
+            subset.node_y.values,
+            subset.node_z.values,
+        )
+
+        # Get the original face indices from the subset
+        inverse_indices = subset.inverse_indices.face.values
+
+        # Check if any of the faces in the subset contain the point
+        index = _find_faces(face_edge_nodes_xyz, point_xyz, inverse_indices)
```
philipc2, line 2521:
```suggestion
        face_indices= _find_faces(face_edge_nodes_xyz, point_xyz, inverse_indices)
```

uxarray/grid/grid.py
```diff
@@ -2446,3 +2459,65 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+
+        return index
```
philipc2, line 2523:
```suggestion
        return face_indices
```

uxarray/grid/geometry.py
```diff
@@ -1561,3 +1568,234 @@ def inverse_stereographic_projection(x, y, central_lon, central_lat):
...
+    unique_intersections = set()
+
+    location = _classify_polygon_location(edges_xyz)
+
+    if location == 1:
+        ref_point_xyz = np.array([0.01745241, 0.0, -0.9998477], dtype=np.float64)
+    elif location == -1:
+        ref_point_xyz = np.array([0.01745241, 0.0, 0.9998477], dtype=np.float64)
+    else:
+        ref_point_xyz = np.array([0.01745241, 0.0, -0.9998477], dtype=np.float64)
```
philipc2, line 1616: Please define these as constants at the top of the file to avoid the use of magic numbers.

uxarray/grid/grid.py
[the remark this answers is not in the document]
philipc2: When I query with Cartesian coordinates in f64, I get the correct indices.

```Python
point = np.array([grid.node_x[0].values, grid.node_y[0].values, grid.node_z[0].values], dtype=np.float64)
```

Though, with f32 I get a Numba error.

```Python
point = np.array([grid.node_x[0].values, grid.node_y[0].values, grid.node_z[0].values], dtype=np.float32)
```

```
No implementation of function Function(<built-in function dot>) found for signature:
 
 >>> dot(array(float64, 1d, C), array(float32, 1d, C))
```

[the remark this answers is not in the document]
philipc2: This does appear to be a precision issue when using spherical coordinates, since we internally convert them to cartesian. We should try to avoid as many of these conversions as possible. @hongyuchen1030 and I have encountered this a lot in our previous implementations.

philipc2 (changes requested):

uxarray/grid/grid.py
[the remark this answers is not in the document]
aaronzedwick: The error you are getting happens within Hongyus code. I have had it error out sometimes when using cartesian coordinates that aren't in float64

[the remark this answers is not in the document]
aaronzedwick: This could be fixed by ensuring that the xyz coordinates are f64. But that still won't fix the precision issue.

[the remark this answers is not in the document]
philipc2: I wonder if adding a small `eps` to increase the radius of each bounding circle might help?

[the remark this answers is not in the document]
aaronzedwick: @philipc2 do you think it is okay just to ensure that the values are float64 and accept that if a node is sent in as a float32, it will only find one face? I think most of the time it will be f64, not f32 anyway. I see this as a very niche problem that shouldn't really cause many problems, but maybe I am wrong? That way this PR can get merged, as it just requires me to uncomment one line. Unless you view it as a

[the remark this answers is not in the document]
philipc2: I would still suggest digging a bit deeper into what is exactly causing this issue, since we should be careful about releasing something with an issue like this. 

If we can pin-point which helper function is leading to the issue, we should be able to fix that.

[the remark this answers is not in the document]
aaronzedwick: This should be solved now with the new error tolerance changes I pushed. Hongyu said the problem is trying to broadcast to a higher precision. We can go down in precision, but we can't expect accurate results when trying to go up in precision (which does make perfect sense).

uxarray/grid/grid.py
```diff
@@ -2446,3 +2459,81 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
 
         """
         return faces_within_lat_bounds(lats, self.face_bounds_lat.values)
+
+    def get_faces_containing_point(self, point, error_tolerance=ERROR_TOLERANCE):
```
philipc2, line 2463:
```suggestion
    def get_faces_containing_point(self, point, tolerance =ERROR_TOLERANCE):
```

Small suggestion to be consistent with how Xarray names their user-defined tolerance.

uxarray/grid/grid.py
```diff
@@ -1543,6 +1549,13 @@ def is_subset(self):
         """Returns `True` if the Grid is a subset, 'False' otherwise."""
         return self._is_subset
 
+    @property
+    def max_face_radius(self):
+        """Returns the maximum face radius of the grid"""
```
philipc2, line 1554: Please rephrase to match the style of our other properties. For reference:

```Python
    @property
    def edge_node_distances(self):
        """Distances between the two nodes that surround each edge in degrees.

        Dimensions ``(n_edge, )``
        """
```

uxarray/grid/grid.py
```diff
@@ -2446,3 +2459,81 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        Parameters
+        ----------
+        point : numpy.ndarray
+            A point in either cartesian coordinates or spherical coordinates
+        error_tolerance : numpy.ndarray
+            An optional error tolerance for points that lie on the nodes of a face
+
+        Returns
+        -------
+        index : array
```
philipc2, line 2474:
```suggestion
        index : numpy.ndarray
```

uxarray/grid/grid.py
```diff
@@ -2446,3 +2459,81 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+
+        # Try to find a subset in which the point resides
+        try:
+            subset = self.subset.bounding_circle(
+                r=np.rad2deg(max_face_radius),
+                center_coord=point_lonlat,
+                element="face centers",
+                inverse_indices=True,
+            )
+        # If no subset is found, it likely means the grid is a partial grid and the point is in an empty part
```
philipc2, line 2502: The comment above is good. How about checking the `self.partial_sphere_coverage` parameter and mentioning it in the warning?

philipc2 (changes requested):
Excellent work @aaronzedwick 

A few small comments below.

Also, may you please run a few small benchmarks and profile the code on the 30km MPAS grid, similar to what I shared above.

Something like the following would be a good example:
- Assume some fixed grid of center points, perhaps from a structured grid. 
- Compute the time it takes to determine which face each point lies within

Share the timing, with a screenshot of the profiling output like what I shared above.

uxarray/grid/grid.py
aaronzedwick: Good idea

aaronzedwick:
@philipc2 @rajeeja Here are the profiling results you asked for!

image
Most of the time comes from the subsetting method, the actual `_find_faces` function takes up very little time:
image

philipc2:
> @philipc2 @rajeeja Here are the profiling results you asked for!
> 
> image Most of the time comes from the subsetting method, the actual `_find_faces` function takes up very little time: image

Excellent! Once we optimize `isel` (and perhaps `node_x`) we will be in a great spot performance wise.

uxarray/grid/grid.py
```diff
@@ -1543,6 +1549,16 @@ def is_subset(self):
         """Returns `True` if the Grid is a subset, 'False' otherwise."""
         return self._is_subset
 
+    @property
+    def max_face_radius(self):
+        """Maximum face radius of the grid
+
+        Dimensions ``(1, )``
```
philipc2, line 1595: We can remove the dimension here since this is a scalar value.

uxarray/grid/grid.py
```diff
@@ -2485,3 +2501,87 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+            A point in either cartesian coordinates or spherical coordinates
+        tolerance : numpy.ndarray
+            An optional error tolerance for points that lie on the nodes of a face
+
+        Returns
+        -------
+        index : numpy.ndarray
+            Array of the face indices containing point. Empty if no face is found
+
+        """
```
philipc2, line 2519: Maybe mention something along the lines of:
- This function will typically return a single face, unless the point falls directly on a corner or edge, where there will instead be multiple values.

uxarray/grid/grid.py
philipc2: Also add units.

uxarray/grid/grid.py
```diff
@@ -2485,3 +2501,87 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        elif len(point) == 3:
+            point_xyz = point
+            point_lonlat = np.array(_xyz_to_lonlat_deg(*point_xyz), dtype=np.float64)
+        else:
+            raise ValueError(
+                "Point must either be in spherical or cartesian coordinates."
+            )
+
+        # Get the maximum face radius of the grid, plus a small adjustment for if the point is this exact radius away
+        max_face_radius = self.max_face_radius.values + 0.0001
```
philipc2, line 2534: Does using `ERROR_TOLLERANCE` work here instead, or is it too small?

philipc2 (changes requested):

uxarray/grid/grid.py
aaronzedwick: It is too small

uxarray/grid/grid.py
```diff
@@ -1582,6 +1588,13 @@ def is_subset(self):
         """Returns `True` if the Grid is a subset, 'False' otherwise."""
         return self._is_subset
 
+    @property
+    def max_face_radius(self):
+        """Maximum face radius of the grid in radians"""
+        if "max_face_radius" not in self._ds:
+            self._ds["max_face_radius"] = _populate_max_face_radius(self)
+        return self._ds["max_face_radius"]
```
philipc2, line 1596: Can we store this as degrees? It looks like we convert it radians anyways in the function call.

uxarray/grid/grid.py
```diff
@@ -2485,3 +2498,89 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
 
         """
         return faces_within_lat_bounds(lats, self.face_bounds_lat.values)
+
+    def get_faces_containing_point(self, point, tolerance=ERROR_TOLERANCE):
+        """Identifies the indices of faces that contain a given point. This function will typically return a
+        single face, unless the point falls directly on a corner or edge, where there will instead be multiple values.
```
philipc2, line 2504: "This function will typically return a single face, unless the point falls directly on a corner or edge, where there will instead be multiple values."

Can we move this inside of the description of the `index` parameter?

uxarray/grid/grid.py
```diff
@@ -2485,3 +2498,89 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        # If no subset is found, warn the user
+        except ValueError:
+            # If the grid is partial, let the user know the point likely lies outside the grid region
+            if self.partial_sphere_coverage:
+                warn(
+                    "No faces found. The grid has partial spherical coverage, and the point may be outside the defined region of the grid."
+                )
+            else:
+                warn("No faces found. Try adjusting the tolerance.")
+            return []
```
philipc2, line 2552: Can you return an empty NumPy array here of type `INT_DTYPE`

uxarray/grid/grid.py
```diff
@@ -2485,3 +2498,89 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
 
         """
         return faces_within_lat_bounds(lats, self.face_bounds_lat.values)
+
+    def get_faces_containing_point(self, point, tolerance=ERROR_TOLERANCE):
```
philipc2, line 2502: Can we slightly modify the structure here to support the following parameters
- point_xyz (Default None)
- point_lonlat (Default None)

The user **_needs_** to provide at least one of these, which would mean the logic below where we convert one of them to xyz or lonlat respectively can remain, however if they provide both we wouldn't need to do any extra conversion.

For many of our grids, we already have both the Spherical and Cartesian coordinate loaded, so we can leverage this. 

This reiterates @hongyuchen1030 's points about trying to avoid as many of these conversions, since they can lead to precision loss and trig functions are not the cheapest to run.

philipc2 (changes requested):

philipc2 (changes requested):
Please add a couple points in the "Overview" of the description of this PR.

rajeeja (approved):
Resolve comments.

philipc2 (changes requested):
For the notebook:
- Add an introduction below the title consistent with how other notebooks are foramtted.
- I don't think we should be showcasing the `point_in_face` function from the start. We should highlight the `Grid` functionality. I think we can restructure the notebook to remove the `point_in_face` part 
- Consider adding intermediate markdown blocks between each cell and try to avoid the inline comments, these should be in Markdown

image

philipc2:
> > For the notebook:
> > 
> > * Add an introduction below the title consistent with how other notebooks are formatted.
> > * I don't think we should be showcasing the `point_in_face` function from the start. We should highlight the `Grid` functionality. I think we can restructure the notebook to remove the `point_in_face` part
> > * Consider adding intermediate markdown blocks between each cell and try to avoid the inline comments, these should be in Markdown
> > 
> > image
> 
> Removing the `point_in_face` part doesn't leave much of the notebook left; do you think this function needs a notebook? It is a pretty self-explanatory function to use.

Maybe we can hold off on the notebook for right now, and keep the examples contained within the docstring?

philipc2:
Using this syntax/style would be good 
```Python
    Examples
    --------
    >>> import uxarray as ux
    Open a grid from a file path
    >>> uxgrid = ux.open_grid("grid_filename.nc")

    Lazily load grid variables using Dask
    >>> uxgrid = ux.open_grid("grid_filename.nc", chunks=-1)
````

**Comment by aaronzedwick** [comment]
> Using this syntax/style would be good
> 
> ```
>     Examples
>     --------
>     >>> import uxarray as ux
>     Open a grid from a file path
>     >>> uxgrid = ux.open_grid("grid_filename.nc")
> 
>     Lazily load grid variables using Dask
>     >>> uxgrid = ux.open_grid("grid_filename.nc", chunks=-1)
> ```

Ok! Yeah, I think this works better.

File: docs/userguide.rst [hunk 28]
```diff
@@ -82,6 +82,8 @@ These user guides provide detailed explanations of the core functionality in UXa
 `Dual Mesh Construction <user-guide/dual-mesh.ipynb>`_
  Construct the Dual Mesh of an unstructured grid
 
+`Point in Face <user-guide/point_in_face.ipynb>`_
+ Check if a point is within a face and find the faces containing a point
```
philipc2, line 86:
```suggestion
```

docs/userguide.rst
```diff
@@ -123,3 +125,4 @@ These user guides provide additional details about specific features in UXarray.
    user-guide/healpix.ipynb
    user-guide/holoviz.ipynb
    user-guide/from_file.ipynb
+   user-guide/point_in_face.ipynb
```
philipc2, line 128:
```suggestion
```

uxarray/grid/grid.py
```diff
@@ -2485,3 +2498,105 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+            a single face, unless the point falls directly on a corner or edge, where there will be multiple values.
+
+         Examples
+        --------
+        >>> import uxarray as ux
+        Open a grid from a file path
+        >>> uxgrid = ux.open_grid("grid_filename.nc")
+
+        Define the point's lonlat coordinates
+        >>> point_lonlat = np.array([45.2, 32.6], dtype=np.float64)
```
philipc2, line 2529: Can you add a cartesian example? Maybe the north pole point `[0.0, 0.0, 1.0]`

philipc2 (changes requested):

uxarray/grid/grid.py
```diff
@@ -2485,3 +2498,113 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        tolerance : numpy.ndarray
+            An optional error tolerance for points that lie on the nodes of a face.
+
+        Returns
+        -------
+        index : numpy.ndarray
+            Array of the face indices containing point. Empty if no face is found. This function will typically return
+            a single face, unless the point falls directly on a corner or edge, where there will be multiple values.
+
+         Examples
```
philipc2, line 2522:
```suggestion
        Examples
```

philipc2 (changes requested):

uxarray/grid/grid.py
```diff
@@ -2485,3 +2498,113 @@ def get_faces_between_latitudes(self, lats: Tuple[float, float]):
...
+        Define a cartesian point
+        >>> point_xyz = np.array([0.0, 0.0, 1.0], dtype=np.float64)
+
+        Find the indices of the faces that contain the given point
+        >>> lonlat_point_face_indices = uxgrid.get_faces_containing_point(
+        ...     point_lonlat=point_lonlat
+        ... )
+        >>> xyz_point_face_indices = uxgrid.get_faces_containing_point(
+        ...     point_xyz=point_xyz
+        ... )
```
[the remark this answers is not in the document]
aaronzedwick: Oh yeah, that is strange; let me try and fix it.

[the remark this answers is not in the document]
philipc2: I'm taking a look locally right now, may I push the changes directly @aaronzedwick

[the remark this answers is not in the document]
aaronzedwick: Sure, go ahead

philipc2 (approved):
Excellent work!

PR #1072, opened 2024-11-10
Optimized Geometry Functions using Numba

### LINKED ISSUE #936: Performance Issue: Unnecessary Re-conversion of Coordinates in `point_within_gca`

The current implementation of point_within_gca unnecessarily re-converts lat/lon coordinates from Cartesian coordinates, which negatively impacts performance.

The following lines re-compute lat/lon from Cartesian coordinates:

python
```
GCRv0_lonlat = np.array(
        _xyz_to_lonlat_rad_no_norm(gca_cart[0][0], gca_cart[0][1], gca_cart[0][2])
    )
GCRv1_lonlat = np.array(
    _xyz_to_lonlat_rad_no_norm(gca_cart[1][0], gca_cart[1][1], gca_cart[1][2])
)
```

However, Uxarray already stores lat/lon coordinates, which can be passed directly to avoid redundant computation. This approach reduces both time and space complexity while aligning better with the overall architecture.

Proposed Solution:

- Pass the existing lat/lon coordinates directly to eliminate the unnecessary conversion.
- Update upstream function calls to `point_within_gca`

philipc2:
Closes #936

## Overview

* Rewrites the following functions to avoid needing to convert between Cartesian and Spherical Coordinates
  *  `point_within_gca()`
  * `extreme_gca_latitude()`
  * `gca_gca_intersection()`
* Added compatibility functions with less-strict array and parameter requirements to use for testing.
  *  `_point_within_gca_cartesian()`
  * `_extreme_gca_latitude_cartesian()`
  * `_gca_gca_intersection_cartesian()`
* Removed the use of the `pyfma` package. 
* The bounds for each face is computed in parallel using Numba
image


philipc2:
@hongyuchen1030 

I still need to finish adjusting all the tests and ensure that everything is passing, but I wanted to quickly updated you on some initial timings. 
- 30km (~40,000 mostly hexagonal faces): 4 seconds

This was without any parallelization. Since the bounding box for each face can be computed interpedently, my next step will be to parallelize the loop that calls `_populate_face_latlon_bound`

uxarray/cross_sections/dataarray_accessor.py
```diff
@@ -17,60 +17,84 @@ def __repr__(self):
...
+        intersect with a specified line of constant latitude.
 
         Parameters
         ----------
         lat : float
             The latitude at which to extract the cross-section, in degrees.
+            Must be between -90.0 and 90.0
         use_spherical_bounding_box : bool, optional
-            If True, uses a spherical bounding box for intersection calculations.
+            If True, uses a spherical bounding box for obtaining candidate faces, which considers the extreme cases
```
hongyuchen1030, line 35: what does the spherical bounding box mean here?

uxarray/cross_sections/dataarray_accessor.py
```diff
@@ -17,60 +17,84 @@ def __repr__(self):
...
         faces = self.uxda.uxgrid.get_faces_at_constant_latitude(
             lat, use_spherical_bounding_box
         )
 
         return self.uxda.isel(n_face=faces)
 
-    def constant_longitude(self, lon: float, use_spherical_bounding_box=False):
-        """Extracts a cross-section of the data array at a specified constant
-        longitude.
+    def constant_longitude(self, lon: float, use_spherical_bounding_box: bool = True):
```
hongyuchen1030, line 65: since the longnitue will not suffer from the curvature, we don't actually need to build the bounding box here.

philipc2: Good point, although the alternative if we didn't use the bounding box would require us to check the intersection on every edge of every face, which is more expensive that checking the minimum and maximum latitude for each face when. Using the bounding box should be the preferred method if many cross section operations are being performed. 

I will correct the docstrings regarding the extremes though.

uxarray/cross_sections/dataarray_accessor.py
philipc2: `use_spherical_box` is used to select whether `Grid.bounds` should be used for performing the cross sections instead of checking every edge of every face using a naive approach.

philipc2:
@hongyuchen1030 

I might scrap the previous methods for computing the intersections and exclusively support the methods that use the bounding box now that the performance has significantly improved.

uxarray/cross_sections/dataarray_accessor.py
hongyuchen1030: If the lat-lon bounds aren’t pre-built, calling this function will end up looping through each edge of the faces and performing extra latitude checks, which can be pretty slow.

Ideally, if the lat-lon bounds are already available, we can use them directly. If not, we could do a quick longitude overlap check to keep things efficient.

For now, let’s just add a note about this in the docstring and leave it as a potential improvement for the future

hongyuchen1030:
> @hongyuchen1030
> 
> I might scrap the previous methods for computing the intersections and exclusively support the methods that use the bounding box now that the performance has significantly improved.

sounds great!

hongyuchen1030 (changes requested):
Did you remove all face_latlon_bounds usages in this PR?

Please prioritize merging the face_latlon_bounds here. You can set the `use_spherical_bounds` by default as `false`, but please try to at least keep the implementation in this PR

philipc2:
> Did you remove all face_latlon_bounds usages in this PR?
> 
> Please prioritize merging the face_latlon_bounds here. You can set the `use_spherical_bounds` by default as `false`, but please try to at least keep the implementation in this PR

I did the opposite. I removed the usages of the naive implementation. All constant longitude and constant latitude queries now use the `face_latlon_bounds`.

There is no longer a `use_spherical_bounds` parameter. I hope this clarifies.

hongyuchen1030:
> > Did you remove all face_latlon_bounds usages in this PR?
> > Please prioritize merging the face_latlon_bounds here. You can set the `use_spherical_bounds` by default as `false`, but please try to at least keep the implementation in this PR
> 
> I did the opposite. I removed the usages of the naive implementation. All constant longitude and constant latitude queries now use the `face_latlon_bounds`.
> 
> There is no longer a `use_spherical_bounds` parameter. I hope this clarifies.

Ohh I see. I guess I was misled by those “No implementation warning ” for `use_spherical_bounds` . Is there any reason we kept those?

philipc2:
> > > Did you remove all face_latlon_bounds usages in this PR?
> > > Please prioritize merging the face_latlon_bounds here. You can set the `use_spherical_bounds` by default as `false`, but please try to at least keep the implementation in this PR
> > 
> > 
> > I did the opposite. I removed the usages of the naive implementation. All constant longitude and constant latitude queries now use the `face_latlon_bounds`.
> > There is no longer a `use_spherical_bounds` parameter. I hope this clarifies.
> 
> Ohh I see. I guess I was misled by those “No implementation warning ” for `use_spherical_bounds` . Is there any reason we kept those?

Those are the `edge` queries. Right now I am only using the `face_bounds` to determine if a face intersects with a line of longitude or latitude.

uxarray/grid/grid.py
```diff
@@ -2254,32 +2283,26 @@ def get_faces_at_constant_latitude(self, lat, use_spherical_bounding_box=False):
...
         lon : float
-            The longitude at which to identify intersecting edges, in degrees.
+            The longitude at which to extract the cross-section, in degrees.
+            Must be between -90.0 and 90.0
         use_spherical_bounding_box : bool, optional
-            If `True`,
-            computes the bounding box for each face using great circle arcs for edges
-            and considers extreme minimums or maximums to increase accuracy.
-            Defaults to `False`.
+            If True, uses a spherical bounding box for obtaining candidate faces, which considers the extreme cases
```
hongyuchen1030, line 2304: Just for the information, we don't need to worry about the "curvature" of a great circle arc will affect the max/min longitude. This curvature thing will only affect the max/min latitude.

philipc2: Good catch. Forgot to update this section.

hongyuchen1030 (approved):
Good work!

Issue #954, opened 2024-09-13
Option to load minimal required coordinates and connectivity 


rajeeja:
### Proposed new feature or change:

This "minimal" reader would only reads corner nodes and face node connectivity, ignoring other info in grid file. #953

philipc2:
This could be generalized for all other grid formats also.

aaronzedwick:
I can tackle this, when you say corner nodes do you mean just read in the nodes of the grid? @rajeeja

rajeeja:
> I can tackle this, when you say corner nodes do you mean just read in the nodes of the grid? @rajeeja 

💯 Just be careful, sometimes there will be center nodes, edge nodes also. How to tackle those?

aaronzedwick:
It shouldn't be too hard, I figure we could even use our current readers, but pass in a `minimal` tag or something, which tells it to only read the `face_x/y/z/lat/lon` and `face_node_connectivity`

PR #1092, opened 2024-12-03
Consistently use Xarray for Grid Reader, Improved Dask Support

### LINKED ISSUE #1093: Better utilize Xarray when loading grids
Instead of always converting to NumPy, directly operate on Xarray arrays, which would preserve Dask arrays under the hood. This should improve the performance of loading in grids when converting them in the UGRID conventions.

philipc2:
Closes #1093 #1137

## Overview

- Eliminates the use of `.values` throughout the Grid readers, which preserves the original data (i.e. Dask)
- Removed a call to `Grid.n_nodes_per_face` that was triggering construction inside of the `repr`
- Fixed a bug where attributes were lost when standardizing coordinates 

## Related 

I'm not sure if we are able to use Xarray's Lazy Variables

https://docs.xarray.dev/en/latest/internals/internal-design.html

https://github.com/pydata/xarray/issues/5081


aaronzedwick:
@philipc2 why is #954 linked here? Are you planning to tackle this issue in this pull request? It doesn't look like you have any code in place that handles it yet, I have added an initial implementation for #954 in PR [1101](https://github.com/UXARRAY/uxarray/pull/1101). I didn't realize the issue was linked here, and it was one of my tasks for this iteration. I don't want to step on your toes or anything though if you have done work on that specific thing.

philipc2:
> @philipc2 why is #954 linked here? Are you planning to tackle this issue in this pull request? It doesn't look like you have any code in place that handles it yet, I have added an initial implementation for #954 in PR [1101](https://github.com/UXARRAY/uxarray/pull/1101). I didn't realize the issue was linked here, and it was one of my tasks for this iteration. I don't want to step on your toes or anything though if you have done work on that specific thing.

Hi @aaronzedwick 

This PR was started after a discussion with @florianziemen on better utilizing Xarray and Dask for loading our Grids. I linked #954 to this one because I was going to explore how to delay the loading of connectivity from a file until the user needs it, which would effectively work as a minimal reader.

Instead of having the user manually specify that they want a minimal grid to be read in, having it automatically delayed while still preserving the metadata (i.e. connectivity names that are available, dimensions, etc.) would better align with what Xarray does. 

The issue with a minimal reader is that the data would still be directly loaded into memory, including and computations that need to be done on it (i.e. converting to zero index, replacing fill values). For extremely large grids this leads to a very noticeable delay in the time it takes to print out information about a `Grid` or `UxDataset`

@florianziemen and @erogluorhan can share their experiences also. 

@aaronzedwick let's spend some time on Tuesday to look at this together, I might need some help anyways.

erogluorhan:
The only thing I'd like to share now is that Dask should be brought into the game only iif the user explicitly wants it (i.e. feeding UXarray with chunked data, Dask arrays). That said, I am glad to see the direction this PR is going towards, i.e. avoiding explicit Numpy/Dask array conversions from our Xarray data, and instead sticking to Xarray data structures as they already cater great IO experience.

uxarray/io/_icon.py
```diff
@@ -3,16 +3,21 @@
...
-    source_dims_dict = {"vertex": "n_node", "edge": "n_edge", "cell": "n_face"}
+    source_dims_dict = _icon_to_ugrid_dims(in_ds)
 
     # rename dimensions to match ugrid conventions
     in_ds = in_ds.rename_dims(source_dims_dict)
 
     # node coordinates
-    node_lon = np.rad2deg(in_ds["vlon"])
-    node_lat = np.rad2deg(in_ds["vlat"])
+    node_lon = 180.0 * in_ds["vlon"] / np.pi
```
aaronzedwick, line 19: Switching from `np.rad2deg` makes sense to avoid using numpy arrays, but it could present a performance bottleneck for larger datasets. Would it make sense to add a dedicated function within UXarray that does this conversion and wrap it with Numba for optimal performance? This could help speed up the conversion a lot for big datasets.

philipc2: This shouldn't introduce any performance issues. These operations are directly applied on the `xr.DataArray`, which can be done lazily if the data is stored as a `dask.Array`

florianziemen: What's wrong with this (apart from it possibly being more effective to first define `180/np.pi` as one constant and then multiply with it)? xarray / dask / ... should have an efficient implementation of multiplication.

uxarray/io/_esmf.py
```diff
@@ -73,22 +79,22 @@ def _read_esmf(in_ds):
...
+    if "start_index" in in_ds["elementConn"].attrs:
         start_index = in_ds["elementConn"].start_index
     else:
         # assume start index is 1 if one is not provided
         start_index = 1
 
     face_node_connectivity = in_ds["elementConn"].values.astype(INT_DTYPE)
 
-    for i, max_nodes in enumerate(n_nodes_per_face):
+    for i, max_nodes in enumerate(n_nodes_per_face.values):
```
erogluorhan, line 97: I think this is needed after all the above eliminations of `.values`, right? Could it still be handled without explicit conversion here? In case of Dask, isn't `n_nodes_per_face` be a Dask array, too?

I am wondering if `.data` instead of `.values` could effectively be used in such cases. In that case, `.data` would either return numpy arrays or Dask arrays, and maybe both cases would need to be handled separately.

uxarray/io/_icon.py
erogluorhan, line 19: @aaronzedwick has already resolved his comment, but this change is probably not even needed, per my [comment](https://github.com/UXARRAY/uxarray/pull/1092/files#r1982281882). Either way is fine with me though

uxarray/grid/grid.py
```diff
@@ -1875,14 +1875,14 @@ def compute_face_areas(
         # but is not the expected behavior behavior as we are in need to recompute if this function is called with different quadrature_rule or order
 
         if latlon:
-            x = self.node_lon.data
-            y = self.node_lat.data
+            x = self.node_lon.values
+            y = self.node_lat.values
```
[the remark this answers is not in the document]
philipc2: Rajeev's code for computing the face area is not written for Dask. This call is required here because the NumPy arrays are needed.

uxarray/grid/grid.py
```diff
@@ -1665,7 +1665,7 @@ def chunk(self, n_node="auto", n_edge="auto", n_face="auto"):
 
     def get_ball_tree(
         self,
-        coordinates: Optional[str] = "nodes",
+        coordinates: Optional[str] = "face centers",
```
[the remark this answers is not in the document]
philipc2: > Why are you changing default from nodes to face centers? 

Using face centers is leaner than the nodes because doing subsequent `isel()` calls will require the `node_face_connectivity` to be created if it isn't already present, which is very expensive for large grids.

I'll make sure the documentation is updated.

uxarray/io/_esmf.py
philipc2: Good catch! 

I didn't remove the `.values()` call above. Let me see if I can update it to not require it.

philipc2: I've replaced it with the following, which is what I used for MPAS

```Python
face_node_connectivity = in_ds["elementConn"].astype(INT_DTYPE)
face_node_connectivity = xr.where(face_node_connectivity != INT_FILL_VALUE, face_node_connectivity - start_index, face_node_connectivity)
```

erogluorhan: Thanks!

erogluorhan (commented):
Other than the documentation updates needed according to the changes of coordinates' default from "nodes" to "face centers", I am satisfied with this PR overall. Thanks! Once they are done, I'll approve.

aaronzedwick (approved):

rajeeja (approved):

PR #1095, opened 2024-12-03
o unittest.TestCase provides access to crucial assertion methods (lik…

rajeeja:
#1098 Running on command line `python -m unittest test.test_cross_sections.TestArcs.test_latitude_along_arc`

doesn't work without this change.


philipc2 (changes requested):
Should this PR merge into `main` instead of `philipc2/zonal-average`?

Issue #986, opened 2024-10-01
Refactor the convoluted test scripts


erogluorhan:
Our test directory and scripts were initially intended to have a test script in correspondence with each code scripts we have. For some while, we seem to have lost track of this (See the below example).  Also, the test directory has all of the test scripts as a plain single list, while our code dir has  modules. 

We can benefit from a test directory refactor to have things more readable and consistent.

For reference, see:

Regarding the existence & name of this test script: We had a tradition of creating test scripts in correspondence with the code scripts we have; this one is a bit off with that (I cannot remember if there are other test scripts in a similar case honestly). Should we be strict and simply put this test case into `test_grid.py` rather than creating its own file? Unsure what way personally.

_Originally posted by @erogluorhan in https://github.com/UXARRAY/uxarray/pull/922#discussion_r1775973579_

PR #1100, opened 2024-12-06
 Convert unittest to pytest

### LINKED ISSUE #1098: Convert unittest to pytest
Move from unittest to pytest for all tests.

rajeeja:
Fixes #1098


rajeeja:
Having seconds thoughts with this work, we might want to move to pytest and get rid of unittest all together. pytest has its expressive syntax, features like fixtures and parameterization, and extensive plugin ecosystem. thoughts @erogluorhan ?

erogluorhan:
Yes, that would make senss to me, @rajeeja , thanks for bringing that!

rajeeja:
> Yes, that would make senss to me, @rajeeja , thanks for bringing that!

We can completely go pytest and remove unittest/TestCase?

philipc2:
@rajeeja 

I'm all in favor of moving strictly to PyTest, although I believe we should all sit down sometime and draft out a design for how we want our testing suite to look like. We have moved along too long without any standards, and it has led to the issues that you are fixing here and also what is discussed in #986 

Do you have some availability this week to chat about this?

rajeeja:
Yes, let's talk today

erogluorhan:
> > Yes, that would make senss to me, @rajeeja , thanks for bringing that!
> 
> We can completely go pytest and remove unittest/TestCase?

Yes!

rajeeja:
3 files remain - `test_ helpers, geometry and integrate` rest converted to pytest. Please review @erogluorhan

test/test_api.py
```diff
@@ -14,126 +12,111 @@
...
-        data_paths = [
-            self.geoflow_data_v1, self.geoflow_data_v2, self.geoflow_data_v3
-        ]
+def test_open_geoflow_dataset():
+    """Loads a single dataset with its grid topology file using uxarray's
+    open_dataset call."""
 
-        uxds_v1 = ux.open_dataset(self.gridfile_geoflow, data_paths[0])
+    # Paths to Data Variable files
+    data_paths = [
```
erogluorhan, line 30: `test_copy_dataarray` below also defines such a `data_paths` var, but I am not sure either of these test cases actually needs to define it (since they only access one data file out of them). If still yes, then for reusing the code, maybe define it in the beginning of the test script instead, and then just access whatever needed from each test case, or don't define it at all and just access the actual data file needed

test/test_centroids.py
```diff
@@ -1,168 +1,145 @@
...
 
+def test_centroids_from_mean_verts_triangle():
+    """Test finding the centroid of a triangle."""
+    test_triangle = np.array([(0, 0, 1), (0, 0, -1), (1, 0, 0)])
+    expected_centroid = np.mean(test_triangle, axis=0)
+    norm_x, norm_y, norm_z = _normalize_xyz(
+        expected_centroid[0], expected_centroid[1], expected_centroid[2]
+    )
 
-class TestCentroids(TestCase):
```
erogluorhan, line 16: Test cases in this script were grouped into two different classes before. While I don't know the reason for that distinction, I am hesitant that this change here remove the distinction. You'd still be able to define the same classes with `pytest`, why not?

This question may apply to all the other scripts in the same situation.

erogluorhan (commented):
While this conversion is great overall, I have a couple inline comments below before moving forward.

test/test_centroids.py
rajeeja: We currently have:


```
class ABC(TestCase):
    def test1(self):
     def test2(self):
class XYZ(TestCase):
    def test1(self):

```
Better to move to this structure as a complete pytest way as it aligns better with pytest's focus on functions and fixtures. This is harder, but more long term and maintainable. 

```
def test_abc_1(setup_data):
    # .. your test logic from ABC.test1 ...

def test_abc_2(setup_data):
    # .. your test logic from ABC.test2 ...

def test_xyz_1(setup_data):
    # .. your test logic from XYZ.test1 ...

def test_xyz_2(setup_data):
    # .. your test logic from XYZ.test2 ...
```

Most of our classes are not complex, so converting all to functions is better IMO.

For the second pass, we can focus on  using pytest markers (@pytest.mark.slow, @pytest.mark.regression, etc.) to categorize your tests for selective running.

erogluorhan: Ok then it makes sense to me

erogluorhan (approved):
Thanks a lot for this conversion!

rajeeja:
> test_ helpers, geometry and integrate

Thanks for the review, will convert the remaining files - test_ helpers, geometry and integrate before merge.

rajeeja:
@philipc2 https://github.com/UXARRAY/uxarray/blob/main/test/test_integrate.py#L164 this test appears twice in main branch! also on line [189](https://github.com/UXARRAY/uxarray/blob/main/test/test_integrate.py#L189)

Fixed here. Please don't push another fix for the same.

philipc2 (approved):

PR #1109, opened 2024-12-10
Fix bug in `_map_dims_to_ugrid`, use Polars to improve SCRIP reader performance 

### LINKED ISSUE #1108: Checking `n_edge` in `_map_dims_to_ugrid` triggers construction of connectivity
When mapping dimensions to UGRID, if `n_edge` was not parsed in the source grid, an expensive computation to construct the `edge_node_connectivity` is executed, which is poorly optimized and leads to a large overhead when initially loading a `Grid`.

philipc2:
Closes #1108 #1110

### Bug Fix in `_map_dims_to_ugrid`

* @erogluorhan and @florianziemen have observed extremely long execution times when initially loading a `UxDataset`. I have isolated the issue to an expensive connectivity construction that happens internally if the `n_edge` dimension is not present in the parsed grid
* The logic in `_map_dims_to_ugrid` is updated to only check `n_edge` if the original grid provided an `n_edge `dimension. There is no need to check `n_edge` otherwise since no variable should be mapped to the edges in this case from the source grid.

### Use `polars` for unique calls
* Polars is used to replace `np.unique()` calls when constructing a `Grid`
   * For SCRIP grids on Derecho, the `ne1024pg2.g` grid went from about 2 mins 30 seconds to under 3 seconds to construct

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
 
       - name: Upload code coverage to Codecov
         if: github.repository == 'UXARRAY/uxarray'
-        uses: codecov/codecov-action@v5.0.7
+        uses: codecov/codecov-action@v5.1.1
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         with:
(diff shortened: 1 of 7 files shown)


erogluorhan (approved):
This is great to see; thanks for getting this fix up quick!

PR #1115, opened 2024-12-19
Latitude & Longitude Interval Cross Section, Updated Bounding Box Subset

### LINKED ISSUE #1104: Update `.subset.bounding_box()` to use spherical face bounds
Add an option to use the spherical face bounds to subset a grid as opposed to querying the BallTree coordinates.

philipc2:
Closes #1104 

## Overview

* Introduces bounded latitude & longitude cross sections, which returns a new grid with elements strictly between two lines of constant latitude or longitude
* Updates bounding box subset to use the spherical bounds instead of KDTree queries


aaronzedwick:
Glad to see this, definitely nice to have this updated. Quick clarification, does this update the bounding box and cross-section to be strictly between those bounds? So if I do a bounding box with a face center as the element, and a face center is inside the bounds but the whole face is not, will it include that face or not?

philipc2:
> Glad to see this, definitely nice to have this updated. Quick clarification, does this update the bounding box and cross-section to be strictly between those bounds? So if I do a bounding box with a face center as the element, and a face center is inside the bounds but the whole face is not, will it include that face or not?

The `subset.bounding_box()` methods now use the `Grid.bounds` to accurately determine the bounding box, instead of using an approximate query.

> So if I do a bounding box with a face center as the element, and a face center is inside the bounds but the whole face is not, will it include that face or not?

There is no longer an `element` needed, since we are not using the coordinates for the query, instead we use the actual geometry.

uxarray/subset/dataarray_accessor.py
```diff
@@ -19,10 +19,12 @@ def __repr__(self):
         prefix = "<uxarray.UxDataArray.subset>\n"
         methods_heading = "Supported Methods:\n"
 
-        methods_heading += "  * nearest_neighbor(center_coord, k, element, **kwargs)\n"
-        methods_heading += "  * bounding_circle(center_coord, r, element, **kwargs)\n"
```
rajeeja, line 23: are we getting rid of the `bounding_circle` method?

rajeeja (commented):
good addition, will review more today.

uxarray/subset/dataarray_accessor.py
philipc2: We are still keeping the `bounding_circle`, as retrieving all elements within some `r` of a point is still very useful for  geoscience applications.

aaronzedwick: Yeah, I am actually using this inside my ‘get_faces_containing_point’, so I would definitely want to keep it in place.

rajeeja (approved):
looks good, minor comment for tests

aaronzedwick (approved):

PR #978, opened 2024-09-29
DRAFT: API for accessing partitioned connectivity, reworked topological aggregations 

philipc2:
## Expected Usage

```Python
import uxarray as ux

grid_path = "/path/to/grid.nc"
data_path = "/path/to/data.nc"

uxds = ux.open_dataset(grid_path, data_path)

# this is how you use this function
some_output = uxds.some_function()

# this is another way to use this function
other_output = uxds.some_function(some_param = True)

```

PR #1117, opened 2024-12-21
Vectorize `coordinates._construct_face_centroids()`

### LINKED ISSUE #1116: Optimize Face Centroid Calculations
If `Grid.face_lon` does not exist, `_populate_face_centroids()`, actually `_construct_face_centroids()` in it, takes extremely long for large datasets. For instance, the benchmark/profiling below is for a ~4GB SCREAM dataset, around 5 mins:

@rajeeja FYI: I'm already working on this and have gotten optimized results, which will be good for "cartesian" parts of the face center calculations, but you may want to look into the `Welzl` parts as well, i.e. `_populate_face_centerpoints()`.

Image

erogluorhan:
Closes #1116.

## Overview
Optimizes `coordinates.py/_construct_face_centroids()` by replacing the for-loop with vectorization with the help of a masking method thanks to @hongyuchen1030 's [older suggestion](https://github.com/UXARRAY/uxarray/pull/811/files#r1670932138) 

Further optimization of the code would be possible, but maybe it is handled in future PRs.

FYI @philipc2 : The masking method this PR uses might be useful for cases we previously dealt with via partitioning, e.g. the nodal_averaging problem we look into last year

FYI @rajeeja : This is where the cartesian-based centroid calculations are being fixed. Welzl is a bit different but you may still want to look into it. 

## Results
#1116 shows the profiling results for the existing code for a 4GB SCREAM dataset where the execution time is around 5 mins. 

The optimized code gives around 6 seconds, see below:

Screenshot 2024-12-21 at 8 34 37 AM

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**


rajeeja:
Good fix, will test welzl and see appropriate fix.

philipc2:
Nice work @erogluorhan 

It appears that a few ASV benchmarks are failing (may or may not be due to the changes in this PR). I can look at it deeper into it on Monday.

erogluorhan:
All sound good, thanks both @rajeeja  @philipc2 !

erogluorhan:
I can't see any failing benchmarks though! @philipc2

philipc2:
> I can't see any failing benchmarks though! @philipc2 

If you take a look at the GitHub Bot above, it shows which ones failed. If they failed before and after, it means something was wrong before this PR. 

Looks like this PR didn't break anything, must have been something with the benchmark written before.

erogluorhan:
Ok, I see what you mean now, @philipc2 !

Also, re benchmarking, I'd expect `mpas_ocean.ConstructFaceLatLon.time_cartesian_averaging('120km')` to show the difference this PR creates with the optimization, but it does not. Dataset is pretty small I think for that, which makes me think big data benchmarking would be important in the future.

uxarray/grid/coordinates.py
```diff
@@ -327,15 +327,14 @@ def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_fa
...
-        centroid_y[face_idx] = np.mean(node_y[face_nodes[face_idx, 0:n_max_nodes]])
-        centroid_z[face_idx] = np.mean(node_z[face_nodes[face_idx, 0:n_max_nodes]])
+
+    # Mask to ignore invalid indices
+    mask = np.arange(face_nodes.shape[1])[None, :] < n_nodes_per_face[:, None]
+
+    # Calculate centroids
+    centroid_x = np.sum(node_x[face_nodes] * mask, axis=1) / n_nodes_per_face
+    centroid_y = np.sum(node_y[face_nodes] * mask, axis=1) / n_nodes_per_face
+    centroid_z = np.sum(node_z[face_nodes] * mask, axis=1) / n_nodes_per_face
```
philipc2, line 337: This here is problematic for grids that have variable element sizes. 

`face_node_connectivity` will contain fill values, which will lead to an index error here.

> IndexError: index -9223372036854775808 is out of bounds for axis 0 with size 83886080

This wasn't an issue for the SCREAM dataset because it is completely composed of quads.

philipc2 (changes requested):

philipc2:
Here is a minimal example that fails. Feel free to create a test case from this

```Python
node_lon = np.array([-20.0, 0.0, 20.0, -20, -40])
node_lat = np.array([-10.0, 10.0, -10.0, 10, -10])
face_node_connectivity = np.array([[0, 1, 2, -1], [0, 1, 3, 4]])

uxgrid = ux.Grid.from_topology(
    node_lon=node_lon,
    node_lat=node_lat,
    face_node_connectivity=face_node_connectivity,
    fill_value=-1,
)

uxgrid.construct_face_centers()

**Comment by philipc2** [comment]
Consider this implementation with Numba.

```Python
@njit(cache=True, parallel=True)
def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_face):
    centroid_x = np.zeros((face_nodes.shape[0]), dtype=np.float64)
    centroid_y = np.zeros((face_nodes.shape[0]), dtype=np.float64)
    centroid_z = np.zeros((face_nodes.shape[0]), dtype=np.float64)
    n_face = n_nodes_per_face.shape[0]

    for i_face in prange(n_face):
        n_max_nodes = n_nodes_per_face[i_face]

        x = np.mean(node_x[face_nodes[i_face, 0:n_max_nodes]])
        y = np.mean(node_y[face_nodes[i_face, 0:n_max_nodes]])
        z = np.mean(node_z[face_nodes[i_face, 0:n_max_nodes]])

        x, y, z = _normalize_xyz_scalar(x, y, z)

        centroid_x[i_face] = x
        centroid_y[i_face] = y
        centroid_z[i_face] = z
    return centroid_x, centroid_y, centroid_z
```

Without partitioning our arrays, it's really difficult to find an elegant vectorized solution directly using NumPy.

rajeeja:
> Consider this implementation with Numba.
> 
> ```python
> @njit(cache=True, parallel=True)
> def _construct_face_centroids(node_x, node_y, node_z, face_nodes, n_nodes_per_face):
>     centroid_x = np.zeros((face_nodes.shape[0]), dtype=np.float64)
>     centroid_y = np.zeros((face_nodes.shape[0]), dtype=np.float64)
>     centroid_z = np.zeros((face_nodes.shape[0]), dtype=np.float64)
>     n_face = n_nodes_per_face.shape[0]
> 
>     for i_face in prange(n_face):
>         n_max_nodes = n_nodes_per_face[i_face]
> 
>         x = np.mean(node_x[face_nodes[i_face, 0:n_max_nodes]])
>         y = np.mean(node_y[face_nodes[i_face, 0:n_max_nodes]])
>         z = np.mean(node_z[face_nodes[i_face, 0:n_max_nodes]])
> 
>         x, y, z = _normalize_xyz_scalar(x, y, z)
> 
>         centroid_x[i_face] = x
>         centroid_y[i_face] = y
>         centroid_z[i_face] = z
>     return centroid_x, centroid_y, centroid_z
> ```
> 
> Without partitioning our arrays, it's really difficult to find an elegant vectorized solution directly using NumPy.

Yes, it was a sticky one, I left it for a later PR to optimize this.

rajeeja:
this also might use a bit more memory due to the creation of the mask array - but that should be fine for larger grids and the speedup would be pretty good considering it uses vectorized operations and broadcasting.  Seems to work and I'm okay with approving this.

uxarray/grid/coordinates.py
rajeeja: How about fixing this and changing/standardizing the way we handle connectivity?

philipc2:
@erogluorhan 

Let me know if you have any thoughts about the Numba implementation that I shared.

uxarray/grid/coordinates.py
philipc2: > changing/standardizing the way we handle connectivity

Can you elaborate? I've explored storing connectivity as partitioned arrays in #978, although for something as simple as computing centroids, Numba seems to a significantly simpler implementation, performing closer to compiled language performance without needing to create an additional mask.

rajeeja: My understanding is that with -1 etc. the mask approach fails, isn't that correct? so, I was thinking adopting a convention that allows for masking to work would simply things.

Not sure about the performance as I think NumPy's vectorized operations might be highly optimized for numerical computations and hopefully outperform the explicit loop here, even when those loops are parallelized with Numba.

rajeeja (approved):

rajeeja:
> @erogluorhan
> 
> Let me know if you have any thoughts about the Numba implementation that I shared.

Fine with the Numba suggestion also, we have modify this PR to use that, we use Numba all over the project.

erogluorhan:
@philipc2 thanks very much for catching the original issue in this PR and suggesting the Numba solution! 

Also thank you and @rajeeja for the convo you kept running here!

Please see some considerations of mine below and let me know your thoughts about each of them, mostly for future, but at least for the scope of this PR, I am fine with the current Numba version, you all feel free to give it a review for this version:

1. Even ~5 secs with the method I tried, and also the current execution time of ~8secs (in Philip's local machine) are also sub-ideal for permanent long term performance of this function. We need to bring it to a better spot (say around a second or two) for this setup (4GB SCREAM). That's because when the user wants to use a time-series data with several time-steps, maybe for an even finer-res grid, the performance should still be okay.

2. Speaking of time-series, the user will likely use chunking schemes, e.g. each time step being a separate chunk. In such scenarios, we need to be sure Numba and Dask will play well together.

3. While this current Numba implementation seems to have a decent performance for now, we saw that vectorization would give us even better performance with my original attempt (if only we didn't have the masking issue with nonuniform connectivity array), and I thought even further parts of the code in that case would be optimized in the future. That said, partitioning the connectivity as Philip attempted in #978 still seems like the best bet to me to still achieve a vectorized performance. @philipc2 , was there any difficulty in that code that made  you stop and closing that PR?

Thoguhts?

philipc2:
Hi @erogluorhan 

> Even ~5 secs with the method I tried, and also the current execution time of ~8secs (in Philip's local machine) are also sub-ideal for permanent long term performance of this function. We need to bring it to a better spot (say around a second or two) for this setup (4GB SCREAM). That's because when the user wants to use a time-series data with several time-steps, maybe for an even finer-res grid, the performance should still be okay.

This is a one-time execution, so I personally think a couple of seconds is acceptable, especially for grids with millions of elements. The comparison is also not direct, as I ran my benchmark on a 3.75km MPAS grid (~84 million nodes). Do you recall the size of the SCREAM grid? 

> Speaking of time-series, the user will likely use chunking schemes, e.g. each time step being a separate chunk. In such scenarios, we need to be sure Numba and Dask will play well together.

The centroid computation here is independent of any time-series data, since this is tied to the `Grid` and not a `UxDataset`.

> While this current Numba implementation seems to have a decent performance for now, we saw that vectorization would give us even better performance with my original attempt (if only we didn't have the masking issue with nonuniform connectivity array), and I thought even further parts of the code in that case would be optimized in the future.

While I agree the speed is significantly in the masked approach better, it only appears that way due to the fixed-dimension. While partitioning as initially explored in #978 may improve our execution time, that PR has been stale for a while and I plan to re-visit it in the near future. However, for our needs, especially with the Face bounds optimization, Numba has proven to be an excellent choice for obtaining significant performance gains.

There was some discussion about Numba in #1124

Pinging @Huite if he has anything he'd like to comment on.

Huite:
Happy to comment!

You can easily circumvent the fill values for floating point data by utilizing NaNs and the relevant numpy methods. 
E.g. index with the -1 value, then overwrite those values with NaN, then use the appropriate NaN-aware function to filter.


## Vectorized

```python
face_nodes_x = node_x[face_node_connectivity]
face_nodes_x[face_node_connectivity == -1] = np.nan
centroid_x = np.nanmean(face_nodes_x, axis=1)
```

This allocates two intermediate arrays (the face_nodes_x and the boolean mask). That isn't too bad, although I'm not sure to what degree numpy automatically parallellizes these operations nowadays.

## Numba
With regards to numba, you're probably leaving quite some performance on the table:

```python
 x = np.mean(node_x[face_nodes[i_face, 0:n_max_nodes]])
```

Note that the `node_x[face_nodes[i_face, 0:n_max_nodes]]` continuously allocates two small (head-allocated) arrays. My guess is that this takes quite some memory, because it's right in the hottest part of the code. Ideally you pre-allocate and reuse the memory instead. The parallellization with `prange` makes this slightly trickier though.

You can always write your own non-allocating `mean`, e.g. something like:

```python
def mean(face, node_x):
     for index in face:
           accum = 0.0 
           if index == -1:
                break
           accum += node_x[face]
      return accum / index
```

This might not have optimal performance either, because memory is accessed all over the place with `node_x[face]`, which probably causes cache misses (depending on the size of the `node_x`).

You want to do indexing and computation of mean in parallel. The mean ideally operates over a contiguous piece of memory.

## Centroid versus center-of-mass

But something else to consider is whether the mean is really what you're after.

I compute a mean for triangles, but use a center of mass approach otherwise. The reason (which I should've commented!) is that there are quite a number of grids where there's stuff like hanging nodes (in quadtree meshes e.g.). 

See this sketch in paint. 

image

Node 3 is hanging. My expectation is that the user wants to get the spot marked by the x (which is the center). But if you compute the mean of the vertices, node 3 is going to "pull" the centroid towards the right, which is undesirable.

You can avoid this by using a center of mass approach instead, as I do:

https://github.com/Deltares/xugrid/blob/fc0bdcb0eaeff62887f1e62fadd61957629dff61/xugrid/ugrid/connectivity.py#L584

To deal with the fill values, I "close" the faces with `close_polygons` (as shapely calls it, I add another column to the face nodes, and I replace the fill values by the first node). The center of mass can then be computed easily, because the fill values will result in zero-area weights (thereby discounting their value).

My method generates relatively many temporary numpy arrays, but you could easily avoid those with fully scalar operations in numba.

My general philosophy is to rely on vectorized solution first, and primarily use numba when there's no other way, primarily to keep JIT latency down. Even though caching will make it only on first use, e.g. CI runners will not be able to utilize the caches. But it's debatable at which point numba becomes superior; I have 16 cores on my laptop so numba's `prange` will give huge speed ups compared to the vectorized approach...

erogluorhan:
@rajeeja you had already approved this before for the previous plan, and now that we've switched to Numba, please feel free to re-review or not.

philipc2:
Below are some profiling results for a 7.5km MPAS grid on my local M1 macbook. Excellent work! 

image

philipc2 (approved):
Nice work!

erogluorhan:
Thanks a lot for your contributions!

Issue #644, opened 2024-01-04
Exclusive Grid Indexing


philipc2:
### Proposed new feature or change:

Support exclusive grid indexing through `Grid.isel`

This method should attempt to return a UGRID-complient grid (i.e. at least one face) with no degenerate elements (a.k.a stray nodes or edges)

PR #1122, opened 2025-01-02
Add Inverse Face Indices to Subsetted Grids

### LINKED ISSUE #1114: Add an `inverse_indices` attribute to `isel`

Currently, when a subset is created, there is no way to know the original face indices of the subsetted grid. This would add an attribute to subset grids that can be called to retrieve the face indices of the original grid. 

This can be used in many ways, such as if a subset is taken of a large grid, we can use this attribute to only load the indices corresponding to that subset for data.

aaronzedwick:
Closes #1114  

## Expected Usage

```Python
import uxarray as ux

grid = ux.open_grid(grid_path)

coord = [0, 0]
subset = grid.subset.nearest_neighbor(coord, k=1, element="face centers", inverse_indices=True)

inverse_face_indices = subset.inverse_face_indices.values

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**

### DIFF
--- a/docs/user-guide/subset.ipynb
+++ b/docs/user-guide/subset.ipynb
@@ -131,7 +131,6 @@
    "cell_type": "markdown",
    "metadata": {
     "collapsed": false,
-    "jp-MarkdownHeadingCollapsed": true,
     "jupyter": {
      "outputs_hidden": false
     }
@@ -643,13 +642,6 @@
    "source": [
     "uxds.uxgrid.is_subset"
    ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": []
   }
  ],
  "metadata": {
--- a/docs/user-guide/subset.ipynb
+++ b/docs/user-guide/subset.ipynb
@@ -581,8 +581,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "subset_indices = uxds[\"relhum_200hPa\"][0].subset.bounding_circle(center_coord, r, element=\"face centers\", \n",
-    "                                                                 inverse_indices=(['face', 'node', 'edge'], True))"
+    "subset_indices = uxds[\"relhum_200hPa\"][0].subset.bounding_circle(\n",
+    "    center_coord,\n",
+    "    r,\n",
+    "    element=\"face centers\",\n",
+    "    inverse_indices=([\"face\", \"node\", \"edge\"], True),\n",
+    ")"
    ]
   },
   {
--- a/docs/user-guide/subset.ipynb
+++ b/docs/user-guide/subset.ipynb
@@ -131,6 +131,7 @@
    "cell_type": "markdown",
    "metadata": {
     "collapsed": false,
+    "jp-MarkdownHeadingCollapsed": true,
     "jupyter": {
      "outputs_hidden": false
     }
@@ -553,6 +554,98 @@
     "print(\"Bounding Box Mean:      \", bbox_subset_nodes.values.mean())\n",
     "print(\"Bounding Circle Mean:   \", bcircle_subset.values.mean())"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Retrieving Orignal Grid Indices"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Sometimes having the original grids' indices is useful. These indices can be stored within the subset with the `inverse_indices` variable. This can be used to store the indices of the original face centers, edge centers, and node indices. This variable can be used within the subset as follows:\n",
+    "\n",
+    "* Passing in `True`, which will store the face center indices\n",
+    "* Passing in a list of which indices to store, along with `True`, to indicate what kind of original grid indices to store.\n",
+    "    * Options for which indices to select include: `face`, `node`, and `edge`\n",
+    "\n",
+    "This currently only works when the element is `face centers`. Elements `nodes` and `edge centers` will be supported in the future."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "subset_indices = uxds[\"relhum_200hPa\"][0].subset.bounding_circle(center_coord, r, element=\"face centers\", \n",
+    "                                                                 inverse_indices=(['face', 'node', 'edge'], True))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "These indices can be retrieve through the grid:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "subset_indices.uxgrid.inverse_indices"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Determining if a Grid is a Subset"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To check if a Grid (or dataset using `.uxgrid`) is a subset, we can use `Grid.is_subset`, which will return either `True` or `False`, depending on whether the `Grid` is a subset. Since `subset_indices` is a subset, using this feature we will return `True`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "subset_indices.uxgrid.is_subset"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The file we have been using to create these subsets, `uxds`, is not a subset, so using the same call we will return `False:`"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "uxds.uxgrid.is_subset"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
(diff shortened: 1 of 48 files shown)


uxarray/grid/grid.py
```diff
@@ -1506,6 +1506,18 @@ def global_sphere_coverage(self):
         (i.e. contains no holes)"""
         return not self.partial_sphere_coverage
 
+    @property
+    def inverse_face_indices(self):
+        if self._is_subset:
+            return self._ds["inverse_face_indices"]
```
philipc2, line 1512: Please add a docstring. I would suggest throwing an exception if the grid is not a subset.

uxarray/grid/grid.py
```diff
@@ -1506,6 +1506,18 @@ def global_sphere_coverage(self):
...
+
+    @property
+    def _is_subset(self):
```
philipc2, line 1515: If we want to make this a property, can we make it public? 
```suggestion
    def is_subset(self):
```

uxarray/grid/grid.py
```diff
@@ -1506,6 +1506,18 @@ def global_sphere_coverage(self):
...
+        """Boolean indicator for whether the Grid is from a subset or not."""
+        if "_is_subset" not in self._ds:
+            self._ds["_is_subset"] = False
+        return self._ds["_is_subset"]
```
philipc2, line 1519: A simple `self._is_subset` should suffice instead of storing it inside of the internal data.

I would suggest adding a `self._is_subset` attribute inside of the constructor, which is set to false by default. 

https://github.com/UXARRAY/uxarray/blob/7c4f3e10019229816525f7796545d191f65c42d4/uxarray/grid/grid.py#L158-L163

uxarray/grid/grid.py
```diff
@@ -2201,7 +2213,7 @@ def get_dual(self):
 
         return dual
 
-    def isel(self, **dim_kwargs):
+    def isel(self, inverse_indices=False, **dim_kwargs):
```
philipc2, line 2216: We need to be careful when storing the inverse indices here.

Depending on which dimension we are slicing on, the inverse indices will either point to node, edge, or face indices.

For the majority of cases, we are interested in the `inverse_face_indices`, however having the option for `inverse_edge_indices` and `inverse_node_indices` may be needed for certain operations.

With this, I'd consider having the following properties under a Grid:
- `inverse_face_indices`: The original face indices if a grid was sliced along the face dimension 
- `inverse_edge_indices`: The original edge indices if a grid was sliced along the edge dimension, 
- `inverse_node_indices`: The original node indices if a grid was sliced along the node dimension 

Regarding the parameter name, I'd also suggest something a bit more descriptive
- `store_inverse_face_indices`
- `store_inverse_node_indices`
- `store_inverse_edge_indices`

uxarray/grid/slice.py
```diff
@@ -132,4 +131,9 @@ def _slice_face_indices(grid, indices, inclusive=True):
             # drop any conn that would require re-computation
             ds = ds.drop_vars(conn_name)
 
+    if inverse_indices:
+        ds["inverse_face_indices"] = indices
+
+    ds["_is_subset"] = True
```
philipc2, line 137: Once you make the changes as described above, pass in `is_subset=True` to the constructor below.

philipc2 (changes requested):
Initial implementation looks good; a few design considerations.

uxarray/grid/grid.py
aaronzedwick: So from my understanding of the code, the way I did it always returns face_indices, since the other two functions (nodes and edges) end up calling the face subset method. If we have the face indices, there isn't much need to get the other ones, as they can be abstracted from the face indices, right?

aaronzedwick: Or does the code just call it face_indices, when in reality it is nodes and edges?

philipc2: I would still suggest naming the variable `inverse_face_indicies`, but that is a fair point.

philipc2: > Or does the code just call it face_indices, when in reality it is nodes and edges?

Doing `isel(n_edge=[2, 4, 6])` will return a grid with faces that contain the edges corresponding to those indices. The indexed here is not exclusive, meaning that if a face contains at least one of the edges, it will be included. 


In this case, the inverse indices should remain `[2, 4, 6]`.

aaronzedwick: So does anything need to be changed with regards to this?

aaronzedwick:
> May you include this in the `cross_section` implementations also?

Sure.

philipc2:
> > May you include this in the `cross_section` implementations also?
> 
> Sure. Would that be a different attribute or stored in the same inverse indices?

Should be the same parameter for consistency, stored within the `inverse_indices`

uxarray/grid/grid.py
philipc2: I believe with your current implementation, doing `isel(n_edge=[2,4,6], inverse_indices=True)` would only store the face-indices corresponding to the faces that contain those edges.

Consider the case where we perform a subset of a node-centered or edge-centered data variable. In this scenario, knowing the inverse face indices is not immediately beneficial, since we would still need the node indices.

Can you implement the following signature for `isel()`

```Python
def isel(self, inverse_indices: : Union[List[str], Set[str], bool] = False, **dim_kwargs):
```

It would work in the following way:
- `False`: Current behavior in main, no indices are stored
- `True`: Current implementation you put together, the inverse face indices are stored
- A list or set (i.e. `['face', 'edge'])` the inverse indices of the provided elements are stored, some combination of `face`, `edge`, and `node` when applicable 

What do you think? This way we can keep the `inverse_indices` with the same name.

aaronzedwick: So let me see if I understand, if a face is the target of the subset, we store the face indices. If an edge is we store both the face and the edge indices? And same with the node?

philipc2: > if a face is the target of the subset, we store the face indices. If an edge is we store both the face and the edge indices? And same with the node?

That's how I would expect it to behave for the `True` case 

However, if the user explicitly declares which inverse indices to store (i.e. `["node", "edge", "face"]`) by passing in a list or set into `inverse indices`, we should only store the ones the user requests.

As for the node and edge cases, I need to re-visit how we are doing the selection, since currently I don't believe it is exclusive as discussed in #644 

Let me take a look at this and I can put together a quick PR that would add inclusive/exclusive indexing to `isel`.

philipc2: However, for the scope of this PR, let's move forward with handling the `inverse_indices` only for the face-centered case. We can add the more advanced functionality at a later point.

uxarray/grid/grid.py
```diff
@@ -242,7 +244,9 @@ def __init__(
     cross_section = UncachedAccessor(GridCrossSectionAccessor)
 
     @classmethod
-    def from_dataset(cls, dataset, use_dual: Optional[bool] = False, **kwargs):
+    def from_dataset(
+        cls, dataset, use_dual: Optional[bool] = False, is_subset=False, **kwargs
```
philipc2, line 248: Can we exclusive `is_subset` from the parameters and access it through the `kwargs`?

Can you pass in `is_subset=kwargs.get('is_subset', False)`  into the `cls()` constructor below?

philipc2 (approved):

philipc2 (changes requested):
Accidently approved in the previous review, please disregard.

philipc2 (changes requested):
Please include a test that directly uses `Grid.isel()`. Docstrings need to be updated to describe the `inverse_indices`parameter.

philipc2 (changes requested):
Nice work.

There are a few instances where you refer to "face indices" as "face centers".

Can you make sure that we refer to the indices as "node", "edge" and "face" instead of reffering to them as "node center", "edge center", or "face center"

rajeeja:
Seems to work, still parsing the code, in the meantime - it'd be good to add or extend one of the examples in subset.ipynb

aaronzedwick:
> Seems to work, still parsing the code, in the meantime - it'd be good to add or extend one of the examples in subset.ipynb

Yeah, this had crossed my mind as well.

docs/user-guide/subset.ipynb
```diff
@@ -131,6 +131,7 @@
    "cell_type": "markdown",
```
philipc2, line 131: <p>Let's add a user-facing property for this. Accessing the hidden attributes is not a best practice.</p><p> </p><p><code>Grid.is_subset</code> which will access <code>Grid._is_subset</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1122/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='41'/>

uxarray/grid/grid.py
```diff
@@ -1506,6 +1528,20 @@ def global_sphere_coverage(self):
...
+        """Indices for a subset that map each face in the subset back to the original grid"""
+        if self.is_subset:
+            return self._inverse_indices
+        else:
+            raise Exception(
+                "Grid is not a subset, therefore no inverse face indices exist"
+            )
+
+    @property
+    def is_subset(self):
```
philipc2, line 1542:
```suggestion
    def is_subset(self) -> bool:
```

uxarray/grid/grid.py
```diff
@@ -1506,6 +1528,20 @@ def global_sphere_coverage(self):
...
+        return self._is_subset
```
philipc2, line 1543: Please include a very brief docstring to `Grid.is_subset`

test/test_subset.py
```diff
@@ -104,19 +104,49 @@ def test_grid_bounding_box_subset():
...
 def test_uxda_isel_with_coords():
     uxds = ux.open_dataset(GRID_PATHS[0], DATA_PATHS[0])
     uxds = uxds.assign_coords({"lon_face": uxds.uxgrid.face_lon})
     sub = uxds['bottomDepth'].isel(n_face=[1, 2, 3])
 
     assert "lon_face" in sub.coords
     assert len(sub.coords['lon_face']) == 3
+
+
+def test_inverse_indices():
```
philipc2, line 124: One small suggestion
- When testing the `isel` functionality, please check that the stored inverse indices for the faces is the same one that is expecte.d

philipc2 (changes requested):
Great work @aaronzedwick !

Three small comments, otherwise all looks fantastic.

philipc2 (approved):
Nice work!

PR #1130, opened 2025-01-21
Optimize Zonal Weights, Cartesian Extreme Lat

### LINKED ISSUE #1131: Implement `extreme_gca_lat` using strictly cartesian coordinates

philipc2:
Closes #1131 

## Overview

* Implements a Cartesian-Only version of `extreme_gca_lat`, which returns the maximum `z` instead of the maximum `lat`, not requring any spherical coordinates to be passed in or converted.
* Implements `get_non_conservative_zonal_face_weights_at_const_lat`, an optimized zonal weight calculation which falls back to the original implementation for values near the equator. 

## Comparison

image


uxarray/grid/integrate.py
```diff
@@ -1,184 +1,356 @@
...
-        The list of latitude and longitude bounds of candidate faces.
-        It should have the same shape as the face_edges_cart_candidate.
-        Shape: (n_faces(candidate),,2, 2),
-        [...,[lat_min, lat_max], [lon_min, lon_max],...]
 
-    is_latlonface : bool, optional, default=False
-        A global flag to indicate if faces are latlon face. If True, then treat all faces as latlon faces. Latlon face means
-        That all edge is either a longitude or constant latitude line. If False, then all edges are GCA.
-         Default is False.
+def get_non_conservative_zonal_face_weights_at_const_lat_original(
```
erogluorhan, line 54: A few comments on this:
- Why is this in the user API?
- The naming is way too long, especially for an user API. It feels like it should have been something like `get_zonal_face_weights` with arguments `method` (with options "conservative" and "non-conservative") and `cross-section` with options "constant_lat" and "constant_lon"
- It is pretty confusing to the user that there is an "original" function and there is one with a normal name (without "original") in the API
- If it will have to be kept in the user API, I am not sure about such a function returning `polars.dataframe` would be  a useful scenario

uxarray/grid/integrate.py
```diff
@@ -285,212 +456,227 @@ def _get_faces_constLat_intersection_info(
...
 
-            raise  # Re-raise the exception if it's not the expected ValueError
+    total_arc = np.sum(arc_lengths)
+    return arc_lengths / total_arc
 
 
-def _process_overlapped_intervals(intervals_df):
-    """Process the overlapped intervals using the sweep line algorithm,
-    considering multiple intervals per face.
+def get_non_conservative_zonal_face_weights_at_const_lat(
```
erogluorhan, line 640: Similar naming and API-related comments as the `get_non_conservative_zonal_face_weights_at_const_lat_original ` function.

uxarray/grid/grid.py
```diff
@@ -1340,6 +1342,27 @@ def node_face_connectivity(self, value):
         assert isinstance(value, xr.DataArray)
         self._ds["node_face_connectivity"] = value
 
+    @property
+    def face_edge_nodes_xyz(self):
```
erogluorhan, line 1346: Not sure about the motive, but adding this to grid properties would mean additional memory usage of `3 * 2 * max_edges * num_faces`, which seems huge for high-res grids. It'd be a better memory utilization to use the `face_edge_connectivity`, which should be six times smaller in size, to compute such data, i.e. face_edge_nodes, on the run and then probably get it discarded (likely with the function's scope ends).

erogluorhan (commented):
Made a quick review and  a few comments.

uxarray/grid/integrate.py
philipc2: > Why is this in the user API?

I will convert it back to internal; it wasn't the intention to expose it to the user.

> The naming is way too long, especially for an user API. It feels like it should have been something like get_zonal_face_weights with arguments method (with options "conservative" and "non-conservative") and cross-section with options "constant_lat" and "constant_lon"

This is fair, I can try to trim down the names of the function.

uxarray/grid/grid.py
philipc2: Yeah that sounds like a better implementation. For the 3.75km grid, it would over 10GB to store this array, which is not practical at all.

uxarray/grid/integrate.py
erogluorhan: As you will convert it back to internal, I am less concerned about the naming unless you have an immediate idea to implement

hongyuchen1030 (approved):
Looks good to me. Great work!

PR #1151, opened 2025-02-04
Performance Quickfix for Subsetting & Slicing 

philipc2:
Closes #XXX

### DIFF
--- a/uxarray/grid/slice.py
+++ b/uxarray/grid/slice.py
@@ -107,22 +107,29 @@ def _slice_face_indices(
 
     face_indices = indices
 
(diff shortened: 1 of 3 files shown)


aaronzedwick:
How much does this increase the performance of subsetting?

philipc2:
> How much does this increase the performance of subsetting?

This is a quick fix that removes unnecessary computations. It should only be noticeable for cases where extra connectivity variables are constructed. So it will vary.

erogluorhan (approved):
Thanks a lot for this improvement!

aaronzedwick (approved):

PR #1120, opened 2024-12-23
Area correction for when an edge is the line of constant latitude

rajeeja:
Closes #813

PR #1162, opened 2025-02-18
Update `conf.py` to fix equation rendering. 

philipc2:
## Overview

- Cleans up `conf.py` to fix equation rendering.
- Small corrections as well to `conf.py`


rajeeja:
I fixed that in #1120 it was using `mathjax3`

rajeeja:
> > I fixed that in #1120 it was using `mathjax3`
> 
> I see! Though, this implementation avoids the deprecation.
> 
> https://www.mathjax.org/cdn-shutting-down/

oh! this fix is timely if cdn is shutting. Thanks! Once merged I can try this on my bigger formula in area correction PR.

PR #1179, opened 2025-03-10
Update units for parsed distances from MPAS 

philipc2:
Closes #XXX

### DIFF
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -1385,7 +1385,7 @@ class Grid:
 
     @property
     def edge_node_distances(self):
-        """Distances between the two nodes that surround each edge in degrees.
+        """Distances between the two nodes that surround each edge in radians.
 
         Dimensions ``(n_edge, )``
         """
@@ -1402,7 +1402,7 @@ class Grid:
     @property
     def edge_face_distances(self):
         """Distances between the centers of the faces that saddle each edge in
-        degrees.
+        radians.
 
         Dimensions ``(n_edge, )``
         """
--- a/uxarray/io/_mpas.py
+++ b/uxarray/io/_mpas.py
@@ -403,7 +403,7 @@ def _parse_edge_faces(in_ds, out_ds, mesh_type):
 
 def _parse_edge_node_distances(in_ds, out_ds):
     """Parses ``edge_node_distances``"""
-    edge_node_distances = in_ds["dvEdge"]
+    edge_node_distances = in_ds["dvEdge"] / in_ds.attrs["sphere_radius"]
 
     out_ds["edge_node_distances"] = edge_node_distances.assign_attrs(
         descriptors.EDGE_NODE_DISTANCES_ATTRS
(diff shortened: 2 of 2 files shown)


philipc2:
@fluidnumerics-joe 

After some further digging into the `30km` MPAS grid, there appear to be two issues:
- The parsed in `edge_node_distances`, even after the unit conversion, seem to be slightly off. I need to do `uxgrid._ds['edge_node_distances'] *= 1.02` to get the `SpatialHash` to construct. Interestingly, even computing it on our own after deleting the parsed version doing `del uxgrid._ds['edge_node_distances']` still requires the `1.02` factor.
- There are exactly `214` out of `40962` faces that seem to incorrectly return `-1` as the index. These seem to be entirely contained along the antimeridian. This appears to occur for the other grids too. I would assume this is probably a limitation of the current implementation. 

```Python
sh = uxgrid.get_spatial_hash(reconstruct=True)

face_lon = uxgrid.face_lon.values
face_lat = uxgrid.face_lat.values

invalid = []
for i, (lon, lat) in enumerate(zip(face_lon, face_lat)):
    a, b = sh.query((lon, lat))
    if i != a[0]:
        invalid.append(i)

uxgrid.isel(n_face=invalid).plot(periodic_elements="split", projection=ccrs.Orthographic(central_longitude=-180))
```

image

philipc2:
We've considered adding a feature to split periodic elements along the `antimeridian` and return a new `Grid`, which I would assume should fix the issue with the `SpatialHash`

We already do this in our visualization functionality internally, but haven't implemented it to return a new `Grid`.

fluidnumerics-joe:
I think it's fair to say we need to work in logic for periodicity. There's two ways to do this I think

1. When checking if an element is inside a hash cells, we can check the x-value as is. If no overlap is found, we can use the periodic image of some of the points. 

2. Convert the points to Cartesian point and use a 3-d hash grid in Cartesian space. We'd have to think about selecting a proper hash grid cells size in this case. Last, the hash table would contain a lot of empty rows, but this may not really be an issue.


Is there a way to get the lat, lon of the vertices for a face that already takes into account reflection over the antimeridian?

philipc2:
> I think it's fair to say we need to work in logic for periodicity. There's two ways to do this I think
> 
> 1. When checking if an element is inside a hash cells, we can check the x-value as is. If no overlap is found, we can use the periodic image of some of the points.
> 2. Convert the points to Cartesian point and use a 3-d hash grid in Cartesian space. We'd have to think about selecting a proper hash grid cells size in this case. Last, the hash table would contain a lot of empty rows, but this may not really be an issue.
> 
> Is there a way to get the lat, lon of the vertices for a face that already takes into account reflection over the antimeridian?

We have a `Grid.antimeridian_face_indicies` variable that stores the indices of faces that cross the antimeridian. We also have a `Grid.to_geodataframe(periodic_elements='split')` which consists of MultiPolygons. This is how we handle the "stiching" for our visualization functions. 

image

image

Issue #880, opened 2024-08-09
Rewrite Geometry Functions With Latest Refractors


hongyuchen1030:
### Summary

As discussed thoroughly in [#785](#785), many of the geometry functions were implemented before multiple refactors and changes to data structures, and they have not been updated to align with the current implementation. A rewrite is needed to significantly improve performance.

### Affected Geometry Functions

The following geometry functions are in need of a rewrite:

- `_pole_point_inside_polygon`
- `extreme_gca_latitude`
- `_insert_pt_in_latlonbox`
- `_get_cartesian_face_edge_nodes`
- `_get_lonlat_rad_face_edge_nodes`

### Key Issues

To improve performance, we need to eliminate any unnecessary coordinate conversion calls, such as `_xyz_to_lonlat_rad`. 

One proposed solution is to provide both the `latlon_rad` and `cartesian` points in the function call to avoid needing conversions. For example:

```python
def _point_within_gca(pt_latlon,
                      pt_xyz,
                      gca_a_latlon,
                      gca_a_xyz,
                      gca_b_latlon,
                      gca_b_xyz,
                      is_directed=False):
 ```
 
 ### Request
 A new API design is required for these geometry functions to address the issues mentioned and enhance performance.

philipc2:
@hongyuchen1030 

Should we link this with #1180

hongyuchen1030:
> [@hongyuchen1030](https://github.com/hongyuchen1030)
> 
> Should we link this with [#1180](https://github.com/UXARRAY/uxarray/pull/1180)

I think for the agile development purpose, it's best to separate these two. And we should also merged [#1180](https://github.com/UXARRAY/uxarray/pull/1180) before we started this considering the fact that 1180 is done now and we still need to final confirmation for settling down the design

PR #1180, opened 2025-03-12
DRAFT: Optimized Non-Conservative Zonal Average

### LINKED ISSUE #1181: Optimize Non-Conservative Zonal Average in `zonal.py`
## Overview

The goal of this issue is to optimize the non-conservative zonal average routine by:
- Avoiding redundant recalculations
- Reusing intermediary computational steps to minimize overhead

**Note:** The current implementation uses data copying in its general helper function to support broad user scenarios. This behavior will remain for user-facing functions.

## Proposed Approach

- **Dedicated Helper Functions:**  
  Build specialized helper functions tailored for the non-conservative zonal average routine.

- **Specialized Routine for Internal Use:**  
  Implement an optimized version that avoids unnecessary data copying internally, while preserving the general helper function that uses copies for general user use.

- **Intermediary Step Reuse:**  
  Identify and reuse common intermediary steps in the computation to avoid redundant calculations.

- **Module Organization:**  
  Create a separate `zonal.py` file that bundles all related functions for clarity and maintainability.

## Tasks

- [ ] Review the current implementation in `zonal.py` for inefficiencies.
- [ ] Design and implement specialized helper functions optimized for the zonal average routine.
- [ ] Maintain the current general helper function (with data copies) for users who need its generality.
- [ ] Implement caching or reuse of intermediary steps to reduce recalculation.
- [ ] Test the new implementation to ensure performance improvements and correctness.
- [ ] Update documentation and add unit tests for the new code.

## Example Reference

*In a similar non-conservative zonal average implementation, dedicated helper functions are built to work specifically for that routine. While general helper functions may use data copying for broad usability, an optimized internal version can focus on performance by reusing intermediary steps and minimizing overhead.*

hongyuchen1030:
Closes #1181 #880 

## Overview
- Introduce specialized helper functions for the optimized non-conservative zonal average routine in `zonal.py`
- Implement caching and reuse of intermediary computational steps to reduce redundant recalculations
- Retain the general-purpose helper functions (which use data copying) for broader usability
- Improve overall performance of the zonal averaging computation

## Expected Usage

```Python
import uxarray as ux

```

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**

**Examples**

### DIFF
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -3,7 +3,10 @@ import dask.array as da
 
 
 from uxarray.grid.integrate import _zonal_face_weights, _zonal_face_weights_robust
-from uxarray.grid.utils import _get_cartesian_face_edge_nodes
+from uxarray.grid.utils import (
+    _get_cartesian_face_edge_nodes,
+    _get_lonlat_rad_face_edge_nodes,
+)
 
 
 
@@ -19,23 +22,35 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
         # Create a NumPy array for storing results
         result = np.zeros(shape, dtype=uxda.dtype)
 
-    faces_edge_nodes_xyz = _get_cartesian_face_edge_nodes(
+
+    faces_edges_cartesian = _get_cartesian_face_edge_nodes(
         uxgrid.face_node_connectivity.values,
         uxgrid.n_face,
-        uxgrid.n_max_face_nodes,
+        uxgrid.n_max_face_edges,
         uxgrid.node_x.values,
         uxgrid.node_y.values,
         uxgrid.node_z.values,
     )
 
-    bounds = uxgrid.bounds.values
+
+    faces_edges_lonlat_rad = _get_lonlat_rad_face_edge_nodes(
+        uxgrid.face_node_connectivity.values,
+        uxgrid.n_face,
+        uxgrid.n_max_face_edges,
+        uxgrid.node_lon.values,
+        uxgrid.node_lat.values,
+    )
+
+    # Check if the bounds are available
+    if "bounds" not in uxgrid._ds:
+        uxgrid._populate_bounds()
 
     for i, lat in enumerate(latitudes):
         face_indices = uxda.uxgrid.get_faces_at_constant_latitude(lat)
 
         z = np.sin(np.deg2rad(lat))
 
-        faces_edge_nodes_xyz_candidate = faces_edge_nodes_xyz[face_indices, :, :, :]
+        faces_edge_nodes_xyz_candidate = faces_edges_cartesian[face_indices, :, :, :]
 
         n_nodes_per_face_candidate = n_nodes_per_face[face_indices]
 
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -6,6 +6,8 @@ from uxarray.grid.integrate import _zonal_face_weights, _zonal_face_weights_robu
 from uxarray.grid.utils import _get_cartesian_face_edge_nodes
 
 
+#TODO: Optimize the entire zonal average function to avoid unnecessary computations
+
 def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=False):
     """Computes the non-conservative zonal mean across one or more latitudes."""
     uxgrid = uxda.uxgrid
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -3,10 +3,6 @@ import dask.array as da
 
 
 from uxarray.grid.integrate import _zonal_face_weights, _zonal_face_weights_robust
-from uxarray.grid.utils import (
-    _get_cartesian_face_edge_nodes,
-    _get_lonlat_rad_face_edge_nodes,
-)
 
 
 
@@ -22,28 +18,8 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
         # Create a NumPy array for storing results
         result = np.zeros(shape, dtype=uxda.dtype)
 
-
-    faces_edges_cartesian = _get_cartesian_face_edge_nodes(
-        uxgrid.face_node_connectivity.values,
-        uxgrid.n_face,
-        uxgrid.n_max_face_edges,
-        uxgrid.node_x.values,
-        uxgrid.node_y.values,
-        uxgrid.node_z.values,
-    )
-
-
-    faces_edges_lonlat_rad = _get_lonlat_rad_face_edge_nodes(
-        uxgrid.face_node_connectivity.values,
-        uxgrid.n_face,
-        uxgrid.n_max_face_edges,
-        uxgrid.node_lon.values,
-        uxgrid.node_lat.values,
-    )
-
-    # Check if the bounds are available
-    if "bounds" not in uxgrid._ds:
-        uxgrid._populate_bounds()
+    bounds = uxgrid.bounds.values
+    faces_edges_cartesian = uxgrid.face_edges_cartesian.values
 
     for i, lat in enumerate(latitudes):
         face_indices = uxda.uxgrid.get_faces_at_constant_latitude(lat)
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -19,7 +19,7 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
         result = np.zeros(shape, dtype=uxda.dtype)
 
     bounds = uxgrid.bounds.values
-    faces_edges_cartesian = uxgrid.face_edges_cartesian.values
+    faces_edges_cartesian = uxgrid.faces_edges_cartesian.values
 
     for i, lat in enumerate(latitudes):
         face_indices = uxda.uxgrid.get_faces_at_constant_latitude(lat)
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -6,7 +6,6 @@ from uxarray.grid.integrate import _zonal_face_weights, _zonal_face_weights_robu
 from uxarray.grid.utils import _get_cartesian_face_edge_nodes
 
 
-#TODO: Optimize the entire zonal average function to avoid unnecessary computations
 
 def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=False):
     """Computes the non-conservative zonal mean across one or more latitudes."""
@@ -60,3 +59,4 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
         )
 
     return result
+
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -5,7 +5,6 @@ import dask.array as da
 from uxarray.grid.integrate import _zonal_face_weights, _zonal_face_weights_robust
 
 
-
 def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=False):
     """Computes the non-conservative zonal mean across one or more latitudes."""
     uxgrid = uxda.uxgrid
@@ -50,4 +49,3 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
         )
 
     return result
-
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -29,8 +29,8 @@ from uxarray.grid.intersections import (
     gca_gca_intersection,
 )
 from uxarray.grid.utils import (
-    _get_cartesian_face_edge_nodes,
-    _get_lonlat_rad_face_edge_nodes,
+    _get_cartesian_faces_edge_nodes,
+    _get_lonlat_rad_faces_edge_nodes,
 )
 from uxarray.utils.computing import allclose, isclose
 
@@ -1346,42 +1346,10 @@ def compute_temp_latlon_array(
     return temp_latlon_array
 
 
-def _populate_face_edges_cartesian(grid, return_array=False):
-    faces_edges_cartesian = _get_cartesian_face_edge_nodes(
-        grid.face_node_connectivity.values,
-        grid.n_face,
-        grid.n_max_face_edges,
-        grid.node_x.values,
-        grid.node_y.values,
-        grid.node_z.values,
-    )
-
-    faces_edges_cartesian_xarray = xr.DataArray(
-        faces_edges_cartesian,
-        dims=["n_face", "n_max_face_edges", "two", "three"],
-        attrs={
-            "cf_role": "face_edges_cartesian",
-            "_FillValue": INT_FILL_VALUE,
-            "long_name": "Provide the Cartesian coordinates of the edge for each face",
-            "start_index": INT_DTYPE(0)
-        },
-    )
-
-    if return_array:
-        return faces_edges_cartesian_xarray
-    else:
-        grid._ds["face_edges_cartesian"] = faces_edges_cartesian_xarray
-
-
 def _populate_faces_edges_cartesian(grid, return_array=False):
-    faces_edges_cartesian = _get_cartesian_face_edge_nodes(
-        grid.face_node_connectivity.values,
-        grid.n_face,
-        grid.n_max_face_edges,
-        grid.node_x.values,
-        grid.node_y.values,
-        grid.node_z.values,
-    )
+    faces_edges_cartesian = _get_cartesian_faces_edge_nodes(grid.face_node_connectivity.values, grid.n_face,
+                                                            grid.n_max_face_edges, grid.node_x.values,
+                                                            grid.node_y.values, grid.node_z.values)
 
     faces_edges_cartesian_xarray = xr.DataArray(
         faces_edges_cartesian,
@@ -1401,13 +1369,9 @@ def _populate_faces_edges_cartesian(grid, return_array=False):
 
 
 def _populate_faces_edges_spherical(grid, return_array=False):
-    faces_edges_lonlat_rad = _get_lonlat_rad_face_edge_nodes(
-        grid.face_node_connectivity.values,
-        grid.n_face,
-        grid.n_max_face_edges,
-        grid.node_lon.values,
-        grid.node_lat.values,
-    )
+    faces_edges_lonlat_rad = _get_lonlat_rad_faces_edge_nodes(grid.face_node_connectivity.values, grid.n_face,
+                                                              grid.n_max_face_edges, grid.node_lon.values,
+                                                              grid.node_lat.values)
 
     faces_edges_lonlat_rad_xarray = xr.DataArray(
         faces_edges_lonlat_rad,
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -1346,13 +1346,91 @@ def compute_temp_latlon_array(
     return temp_latlon_array
 
 
+def _populate_face_edges_cartesian(grid, return_array=False):
+    faces_edges_cartesian = _get_cartesian_face_edge_nodes(
+        grid.face_node_connectivity.values,
+        grid.n_face,
+        grid.n_max_face_edges,
+        grid.node_x.values,
+        grid.node_y.values,
+        grid.node_z.values,
+    )
+
+    faces_edges_cartesian_xarray = xr.DataArray(
+        faces_edges_cartesian,
+        dims=["n_face", "n_max_face_edges", "two", "three"],
+        attrs={
+            "cf_role": "face_edges_cartesian",
+            "_FillValue": INT_FILL_VALUE,
+            "long_name": "Provide the Cartesian coordinates of the edge for each face",
+            "start_index": INT_DTYPE(0)
+        },
+    )
+
+    if return_array:
+        return faces_edges_cartesian_xarray
+    else:
+        grid._ds["face_edges_cartesian"] = faces_edges_cartesian_xarray
+
+
+def _populate_faces_edges_cartesian(grid, return_array=False):
+    faces_edges_cartesian = _get_cartesian_face_edge_nodes(
+        grid.face_node_connectivity.values,
+        grid.n_face,
+        grid.n_max_face_edges,
+        grid.node_x.values,
+        grid.node_y.values,
+        grid.node_z.values,
+    )
+
+    faces_edges_cartesian_xarray = xr.DataArray(
+        faces_edges_cartesian,
+        dims=["n_face", "n_max_face_edges", "two", "three"],
+        attrs={
+            "cf_role": "faces_edges_cartesian",
+            "_FillValue": INT_FILL_VALUE,
+            "long_name": "Provide the Cartesian coordinates of the edge for each face",
+            "start_index": INT_DTYPE(0)
+        },
+    )
+
+    if return_array:
+        return faces_edges_cartesian_xarray
+    else:
+        grid._ds["faces_edges_cartesian"] = faces_edges_cartesian_xarray
+
+
+def _populate_faces_edges_spherical(grid, return_array=False):
+    faces_edges_lonlat_rad = _get_lonlat_rad_face_edge_nodes(
+        grid.face_node_connectivity.values,
+        grid.n_face,
+        grid.n_max_face_edges,
+        grid.node_lon.values,
+        grid.node_lat.values,
+    )
+
+    faces_edges_lonlat_rad_xarray = xr.DataArray(
+        faces_edges_lonlat_rad,
+        dims=["n_face", "n_max_face_edges", "two", "two"],
+        attrs={
+            "cf_role": "faces_edges_spherical",
+            "_FillValue": INT_FILL_VALUE,
+            "long_name": "Provide the Spherical coordinates (lon lat) in radians of the edge for each face",
+            "start_index": INT_DTYPE(0)
+        },
+    )
+
+    if return_array:
+        return faces_edges_lonlat_rad_xarray
+    else:
+        grid._ds["faces_edges_spherical"] = faces_edges_lonlat_rad_xarray
+
+
 def _populate_bounds(
-    grid,
-    faces_edges_cartesian: np.ndarray = None,
-    faces_edges_lonlat_rad: np.ndarray = None,
-    is_latlonface: bool = False,
-    is_face_GCA_list=None,
-    return_array=False
+        grid,
+        is_latlonface: bool = False,
+        is_face_GCA_list=None,
+        return_array=False
 ):
     """Populates the bounds of the grid based on the geometry of its faces,
     taking into account special conditions such as faces crossing the
@@ -1430,26 +1508,11 @@ def _populate_bounds(
     # TODO: Is it possible to have a more flexible normalization? (Avoid duplicated normalizations)
     grid.normalize_cartesian_coordinates()
 
-    # Prepare data for Numba functions
-    # TODO: remove the duplicate call for this function in both zonal average and bounds
-    if faces_edges_cartesian is None:
-        faces_edges_cartesian = _get_cartesian_face_edge_nodes(
-            grid.face_node_connectivity.values,
-            grid.n_face,
-            grid.n_max_face_edges,
-            grid.node_x.values,
-            grid.node_y.values,
-            grid.node_z.values,
-        )
+    # If the face_edges_cartesian and face_edges_lonlat_rad didn't exist,
+    # The following call will automatically populate them
+    faces_edges_cartesian = grid.faces_edges_cartesian.values
 
-    if faces_edges_lonlat_rad is None:
-        faces_edges_lonlat_rad = _get_lonlat_rad_face_edge_nodes(
-            grid.face_node_connectivity.values,
-            grid.n_face,
-            grid.n_max_face_edges,
-            grid.node_lon.values,
-            grid.node_lat.values,
-        )
+    faces_edges_lonlat_rad = grid.faces_edges_spherical.values
 
     n_nodes_per_face = grid.n_nodes_per_face.values
 
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -103,7 +103,7 @@ def _unique_points(points, tolerance=ERROR_TOLERANCE):
 
 @njit(cache=True)
 def _pad_closed_face_nodes(
-    face_node_connectivity, n_face, n_max_face_nodes, n_nodes_per_face
+        face_node_connectivity, n_face, n_max_face_nodes, n_nodes_per_face
 ):
     """Pads a closed array of face nodes by inserting the first element at any
     point a fill value is encountered.
@@ -123,14 +123,14 @@ def _pad_closed_face_nodes(
 
 
 def _build_polygon_shells(
-    node_lon,
-    node_lat,
-    face_node_connectivity,
-    n_face,
-    n_max_face_nodes,
-    n_nodes_per_face,
-    projection=None,
-    central_longitude=0.0,
+        node_lon,
+        node_lat,
+        face_node_connectivity,
+        n_face,
+        n_max_face_nodes,
+        n_nodes_per_face,
+        projection=None,
+        central_longitude=0.0,
 ):
     """Builds an array of polygon shells, which can be used with Shapely to
     construct polygons."""
@@ -266,7 +266,7 @@ def _grid_to_polygon_geodataframe(grid, periodic_elements, projection, project, 
 
 
 def _build_geodataframe_without_antimeridian(
-    polygon_shells, projected_polygon_shells, antimeridian_face_indices, engine
+        polygon_shells, projected_polygon_shells, antimeridian_face_indices, engine
 ):
     """Builds a ``spatialpandas.GeoDataFrame`` or
     ``geopandas.GeoDataFrame``excluding any faces that cross the
@@ -295,10 +295,10 @@ def _build_geodataframe_without_antimeridian(
 
 
 def _build_geodataframe_with_antimeridian(
-    polygon_shells,
-    projected_polygon_shells,
-    antimeridian_face_indices,
-    engine,
+        polygon_shells,
+        projected_polygon_shells,
+        antimeridian_face_indices,
+        engine,
 ):
     """Builds a ``spatialpandas.GeoDataFrame`` or ``geopandas.GeoDataFrame``
     including any faces that cross the antimeridian."""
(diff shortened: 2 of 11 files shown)


uxarray/grid/geometry.py
```diff
@@ -1407,25 +1466,14 @@ def _populate_bounds(
     """
 
     # Ensure grid's cartesian coordinates are normalized
+    # TODO: Is it possible to have a more flexible normalization? (Avoid duplicated normalizations)
```
hongyuchen1030, line 1469: @ philipc2 Can we avoid duplicate normalization here? For example if the nodes are already normalized, we will not normalize them.

For example, set an attributes in the grid `node_x`,`node_y`,`node_z` to indicate if they're already normalized or not.

rajeeja:
ASV Benchmark failing?

philipc2:
pre-commit.ci autofix

hongyuchen1030:
@philipc2 

The benchmark still doesn't work out for this PR, can you take a look? Thanks

rajeeja (approved):
Approving, we can fix the benchmarking for forked repo's later.

uxarray/core/zonal.py
```diff
@@ -18,23 +17,15 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
...
     bounds = uxgrid.bounds.values
+    faces_edges_cartesian = uxgrid.faces_edges_cartesian.values
 
     for i, lat in enumerate(latitudes):
         face_indices = uxda.uxgrid.get_faces_at_constant_latitude(lat)
 
         z = np.sin(np.deg2rad(lat))
 
-        faces_edge_nodes_xyz_candidate = faces_edge_nodes_xyz[face_indices, :, :, :]
+        faces_edge_nodes_xyz_candidate = faces_edges_cartesian[face_indices, :, :, :]
```
philipc2, line 28: Wouldn't a better implementation here be to only construct the `faces_edges_cartesian` for the indices that we require, instead of needing to construct the entire array and then index?

For each line of constant latitude, typically only a small percentage of the total faces will be candidates.

uxarray/grid/grid.py
```diff
@@ -1443,6 +1445,26 @@ def face_areas(self, value):
...
+    @property
+    def faces_edges_spherical(self):
+        """Latitude Longitude Coordinates for each Face in radians.
+
+        Dimensions ``(n_face, n_max_face_edges, 2, 2)``
+        """
+        if "faces_edges_spherical" not in self._ds:
+            _populate_faces_edges_spherical(self)
+        return self._ds["faces_edges_spherical"]
+
```
philipc2, line 1467: I'm not sure if caching these throughout the entire lifetime of the `Grid` is something we want to do. 

For the 42 million element MPAS grid, each of these would be multiple GB in size. 

From a previous discussion with @erogluorhan 

https://github.com/UXARRAY/uxarray/pull/1130#discussion_r1924731908

In my other comment, we should instead construct only the desired faces for these variables in the zonal average calls, instead of storing the pre-computed array.

Since this variable simply manipulates existing connectivity and coordinate variables, we won't be losing any information or doing any expensive re-computations.

philipc2 (changes requested):

uxarray/grid/grid.py
hongyuchen1030: Hi Philip, thanks for your input. Just to clarify a few points:

1. This property is constructed **only** when needed to populate the face_latlon bounds, **similar to how it was handled in the old design.**
2. Previously, **we populated these** for the whole grid when constructing the latlon bounds.
3. Also, we never permanently saved this property. It was created more than once—once for the face_latlon bounds and then partially re-populated for the zonal average.
4. With the new version, we **only** cache the populated face_edge_nodes property when required for the latlon bounds, which helps avoid duplicate calls later in the process.

So, if you take a closer look at the previous implementation, you'll notice **we didn't actually populate any extra data for the entire grid**. Hope this clears things up!

uxarray/core/zonal.py
hongyuchen1030: > Something like the following should work if we put it inside of the loop?
> 
> ```python
> _get_cartesian_face_edge_nodes(
>        uxgrid.face_node_connectivity.values[face_indices], # only candidate faces
>        len(face_indices), # number of candidate faces
>        uxgrid.n_max_face_nodes,
>        uxgrid.node_x.values,
>        uxgrid.node_y.values,
>        uxgrid.node_z.values,
>     )
> ```

Please refer to my comment below: the lalon bounds building will required this connectivity for the entire grid anyway

uxarray/grid/grid.py
philipc2: Thanks for the detailed response!

> This property is constructed only when needed to populate the face_latlon bounds, similar to how it was handled in the old design.

That's consistent with how we've handled `Grid` variables. My concern is that both of these new variables are extremely large, and inherently are derived from existing variables that are already stored in our `Grid` (`face_node_connectivity` and either set of coordinates).

Caching something like the face bounds is practical because the time needed to generate it is the main bottle-neck in the initial zonal average calls. Additionally, it only contains `(n_face x 4)` total entries, compared to the `(n_face x n_max_face_nodes x 2 x (2/3))` for the `edge_nodes` here 


> This property is constructed only when needed to populate the face_latlon bounds, similar to how it was handled in the old design.

This makes sense, but my issue lies in whether we need to store the `face_edges` after constructing the bounds. As I mentioned, for our high-resolution grids these arrays require a significant amount of memory. 


> With the new version, we only cache the populated face_edge_nodes property when required for the latlon bounds, which helps avoid duplicate calls later in the process.

I do see the intention with removing the duplicate calls, but right now the bottle-neck with the zonal average is the initial bounds construction. 

For a `30km` grid (~600k hexagons), the initial execution time for the default `.zonal_average()` call is ~33 seconds on my machine. Subsequent calls are about 400ms.

The initial call is dominated by the bounds construction.

philipc2: One area of optimization could be to parallelize this Python loop with Numba.

https://github.com/hongyuchen1030/uxarray/blob/29862c2823abe49b33942ba36961d9891a21bd38/uxarray/core/zonal.py#L23-L49

The zonal average for each line of constant latitude can be computed independently.

hongyuchen1030: Thanks for the detailed response! I'm glad we're on the same page regarding bounds construction—it’s an inevitable step, and we need the full property for it.

If we look into the current implementation closely, as we loop from -90° to 90° for candidate faces, we're essentially rebuilding the entire face_edge_nodes connectivity. 

If the concern is mainly about the memory footprint for storing this grid property, **we can consider removing it after the zonal average call.**

philipc2: > I'm glad we're on the same page regarding bounds construction—it’s an inevitable step, and we need the full property for it.

Yep! Having the bounds cached is extremely important.

> If we look into the current implementation closely, as we loop from -90° to 90° for candidate faces, we're essentially rebuilding the entire face_edge_nodes connectivity.

I consider the `face_edge_nodes` to be one step past a `connectivity` variable. Our definition of a connectivity variable is an array of indices that point to some geometry (nodes, edges, or faces).

When we enter `_compute_non_conservative_zonal_mean()`, we compute the `face_edge_nodes` before the `latitude` loop, so we have access to the entire array for each line of constant latitude. Once we finish the zonal average call, we no longer need to store the `face_edge_nodes`

Here's a breakdown of the zonal average call for the example I provided above.

image

I think there is definitely some further optimization that we could do with this helper. It's not currently written with Numba. 

I'd like to see how much we could optimize this helper before we consider caching computed value.

hongyuchen1030: Yes, exactly. This is why I believe we should store the connectivity information throughout the entire zonal average call—it doesn't contradict anything you've mentioned.

First, let's think thoroughly about the routine. Here's the detailed breakdown for the zonal average call:

1. **Before the zonal average call:** We build the latlon bounds, which requires the complete sets of faces_edges_cartesian and faces_edges_spherical for the entire grid.
2. **During the zonal weight calculation:** As we loop from -90° to 90° for candidate faces, we effectively need the full grid's connectivity information again. Caching the connectivity saves us from rebuilding it for each slice.
3. **After the zonal average call:** Once we're done, we remove the face_edge_nodes property since it's no longer needed.

Additionally, regarding parallelization: rather than focusing solely on paralleling each latitude slice, we should tackle the bottleneck—the latlon bound construction. Caching the connectivity also aids in parallelization, as each thread can read its slice of face_edge_nodes data directly without needing to rebuild it.

Also, I think it's important to consider the big picture. Instead of honing in on individual helper functions, we should think about the remapping operation as a whole. This integrated pipeline approach will give us a better framework for redesigning the process, rather than treating each function in isolation.

hongyuchen1030:
Hi @philipc2 ,

Thanks again for the discussion—I'm glad we reached an agreement. I've added the latest commit, which removes the "faces_edges_cartesian" and "faces_edges_spherical" properties at the end of the zonal_mean call. This change should help save duplicate call time and prevent unnecessary memory expansion.

Please take a look when you have a moment. Once this is merged, I'll open another PR where we can integrate and carefully parallelize the latlonbound and zonal_avg calls.

Thanks!

hongyuchen1030:
And I also thought carefully about optimizing the loop you mentioned:

> Something like the following should work if we put it inside of the loop?
> 
> ```python
> _get_cartesian_face_edge_nodes(
>        uxgrid.face_node_connectivity.values[face_indices], # only candidate faces
>        len(face_indices), # number of candidate faces
>        uxgrid.n_max_face_nodes,
>        uxgrid.node_x.values,
>        uxgrid.node_y.values,
>        uxgrid.node_z.values,
>     )
> ```


I know it sounds good to build the connectivity for each face slice inside the loop, and it would be very helpful if the user is querying just a few latitudes. However, consider the case of a batch of latitudes: for every latitude call, we would be querying the associated `uxgrid.node_x.values`, `uxgrid.node_y.values`, and `uxgrid.node_z.values` for the candidate faces, converting them into Cartesian face edge nodes, and then discarding them. Meanwhile, we already have this information available from the `latlon_bound` build.

This brings up another general design idea (which also applies to S2 geometry): we should separate the single call from the batch call. Our current version works best for a batch call for candidate faces—which is a common use case. In a future optimization, we can also consider optimizing the function for a single latitude call, perhaps even creating a dedicated function tailored for that scenario.

Let me know your thoughts.

philipc2:
> @philipc2 And I also thought carefully about optimizing the loop you mentioned:
> 
> > Something like the following should work if we put it inside of the loop?
> > ```python
> > _get_cartesian_face_edge_nodes(
> >        uxgrid.face_node_connectivity.values[face_indices], # only candidate faces
> >        len(face_indices), # number of candidate faces
> >        uxgrid.n_max_face_nodes,
> >        uxgrid.node_x.values,
> >        uxgrid.node_y.values,
> >        uxgrid.node_z.values,
> >     )
> > ```
> 
> I know it sounds good to build the connectivity for each face slice inside the loop, and it would be very helpful if the user is querying just a few latitudes. However, consider the case of a batch of latitudes: for every latitude call, we would be querying the associated `uxgrid.node_x.values`, `uxgrid.node_y.values`, and `uxgrid.node_z.values` for the candidate faces, converting them into Cartesian face edge nodes, and then discarding them. Meanwhile, we already have this information available from the `latlon_bound` build.
> 

This is a good point. Keeping the call outside of the loop should be good for our current application. 

> This brings up another general design idea (which also applies to S2 geometry): we should separate the single call from the batch call. Our current version works best for a batch call for candidate faces—which is a common use case. In a future optimization, we can also consider optimizing the function for a single latitude call, perhaps even creating a dedicated function tailored for that scenario.

I like that idea.

uxarray/grid/grid.py
```diff
@@ -1443,6 +1445,44 @@ def face_areas(self, value):
...
+    @faces_edges_spherical.setter
+    def faces_edges_spherical(self, value):
+        self._ds["faces_edges_spherical"] = value
+
+    @faces_edges_spherical.deleter
+    def faces_edges_spherical(self):
+        if "faces_edges_spherical" in self._ds:
+            del self._ds["faces_edges_spherical"]
+
```
philipc2, line 1485: This is a step in the right direction, although I don't think we should store the `face_edges` in `self._ds` due to the memory concerns.

I'm happy to keep the two `faces_edges_spherical` and `faces_edges_cartesian` as part of the API, but instead they construct the variables by calling `_get_cartesian_faces_edge_nodes`.

Inside of `zonal.py`, calling this attribute within the scope of the function will automatically manage the memory associated with it once we exit the function.

philipc2:
@hongyuchen1030 

Here's an optimized version of `get_cartesian_face_edge_nodes` using Numba. Here we re-use the `face_edge_connectivity` and `edge_node_connectivity` that can be computed internally and re-used.

This runs in about `90ms` for me, down from the `~500ms` we currently have. 



```Python
@njit(cache=True, parallel=True)
def get_cartesian_face_edge_nodes(face_edge_connectivity, edge_node_connectivity,
                                  node_x, node_y, node_z,
                                  n_edges_per_face, n_face, n_max_face_edges):

    # Allocate result array
    # I used np.nan here instead of INT_FILL_VALUE to be consistent with the double type (INT_FILL_VALUE is for our int64 arrays)
    face_edges = np.full((n_face, n_max_face_edges, 2, 3), np.nan, dtype=np.float64)

    # Construct edge_nodes for each face in parallel
    for face_i in prange(n_face):
        n_edges = n_edges_per_face[face_i]
        edge_indices = face_edge_connectivity[face_i, :n_edges]
        node_indices = edge_node_connectivity[edge_indices]  

        first_nodes = node_indices[:, 0]
        second_nodes = node_indices[:, 1]

        # Assign x coordinates
        face_edges[face_i, :n_edges, 0, 0] = node_x[first_nodes]
        face_edges[face_i, :n_edges, 1, 0] = node_x[second_nodes]

        # Assign y coordinates
        face_edges[face_i, :n_edges, 0, 1] = node_y[first_nodes]
        face_edges[face_i, :n_edges, 1, 1] = node_y[second_nodes]

        # Assign z coordinates
        face_edges[face_i, :n_edges, 0, 2] = node_z[first_nodes]
        face_edges[face_i, :n_edges, 1, 2] = node_z[second_nodes]

    return face_edges

```

The same can be applied to the spherical one, with changes only needed to the assignment of coordinates.

uxarray/grid/grid.py
hongyuchen1030: Keeping `face_edges` in zonal was my first thought as well. However, I then realized that l`atlonbound `also needs to use this property. That means we would have to modify the `latlonbound` API to accept incoming face_edges, which involves more API changes than I initially expected—and I’m wary of making such significant modifications to the existing API and functionality.

Do you have any suggestions on where we might store this information to address the memory concerns without causing a large API overhaul?

Thanks!

hongyuchen1030:
> ```python
> face_node_connectivity
> ```

Thanks for the information! This is actually what I had in mind for `get_cartesian_face_edge_nodes` using `face_edge_connectivity` and `edge_node_connectivity`. However, I recall that we had a lengthy discussion about why we should separate the steps—first populating `face_node_connectivity` (using  `face_edge_connectivity` and `edge_node_connectivity`) and storing it in `_ds`, and then populating face_edge_connectivity when needed.

Do you still recall the rationale behind that decision? I'm trying to avoid duplicate iterations as much as possible.

Thanks!

uxarray/grid/grid.py
philipc2: We can keep the two `@property` implementations, and have them return the output of `_get_cartesian_face_edge_nodes()` and `_get_lonlat_rad_face_edge_nodes()` and replace the following two lines with calls to `Grid.face_edges_cartesian` and `Grid.face_edges_spherical`

https://github.com/UXARRAY/uxarray/blob/a6aa6294975f189e8c36dd3d14da745526e53e06/uxarray/grid/geometry.py#L1412-L1428

philipc2:
> > ```python
> > face_node_connectivity
> > ```
> 
> Thanks for the information! This is actually what I had in mind for `get_cartesian_face_edge_nodes` using `face_edge_connectivity` and `edge_node_connectivity`. However, I recall that we had a lengthy discussion about why we should separate the steps—first populating `face_node_connectivity` (using `face_edge_connectivity` and `edge_node_connectivity`) and storing it in `_ds`, and then populating face_edge_connectivity when needed.
> 
> Do you still recall the rationale behind that decision? I'm trying to avoid duplicate iterations as much as possible.
> 
> Thanks!

Most of the time, the `face_node_connectivity` is already present and we do not need to construct it. The other two connectivity variables are automatically populated when we call their respective attribute.

It makes sense to me to re-use those connectivity variables instead of doing the extra operations within the function call, which each time essentially "derives" the same information already present in the connectivity.

hongyuchen1030:
Hi Kelly,

Thank you for the detailed explanation and your help. I appreciate the
guidance regarding the pairing requirements.

I will reach out again next week about both semi-private and private lesson
options. If a suitable pairing for the semi-private lesson isn’t available,
I’m happy to continue with a private lesson.

Thanks again for your assistance.

Best regards,

Victoria

hongyuchen1030:
> @hongyuchen1030
> 
> Here's an optimized version of `get_cartesian_face_edge_nodes` using Numba. Here we re-use the `face_edge_connectivity` and `edge_node_connectivity` that can be computed internally and re-used.
> 
> This runs in about `90ms` for me, down from the `~500ms` we currently have.
> 
> ```python
> @njit(cache=True, parallel=True)
> def get_cartesian_face_edge_nodes(face_edge_connectivity, edge_node_connectivity,
>                                   node_x, node_y, node_z,
>                                   n_edges_per_face, n_face, n_max_face_edges):
> 
>     # Allocate result array
>     # I used np.nan here instead of INT_FILL_VALUE to be consistent with the double type (INT_FILL_VALUE is for our int64 arrays)
>     face_edges = np.full((n_face, n_max_face_edges, 2, 3), np.nan, dtype=np.float64)
> 
>     # Construct edge_nodes for each face in parallel
>     for face_i in prange(n_face):
>         n_edges = n_edges_per_face[face_i]
>         edge_indices = face_edge_connectivity[face_i, :n_edges]
>         node_indices = edge_node_connectivity[edge_indices]  
> 
>         first_nodes = node_indices[:, 0]
>         second_nodes = node_indices[:, 1]
> 
>         # Assign x coordinates
>         face_edges[face_i, :n_edges, 0, 0] = node_x[first_nodes]
>         face_edges[face_i, :n_edges, 1, 0] = node_x[second_nodes]
> 
>         # Assign y coordinates
>         face_edges[face_i, :n_edges, 0, 1] = node_y[first_nodes]
>         face_edges[face_i, :n_edges, 1, 1] = node_y[second_nodes]
> 
>         # Assign z coordinates
>         face_edges[face_i, :n_edges, 0, 2] = node_z[first_nodes]
>         face_edges[face_i, :n_edges, 1, 2] = node_z[second_nodes]
> 
>     return face_edges
> ```
> 
> The same can be applied to the spherical one, with changes only needed to the assignment of coordinates.

Hi @philipc2 

Have you compared the new version's results with the old one? I set up the following test and found that the results differ:
```python
def test_cartesian_test():
    grid = ux.open_grid(gridfile_CSne8)
    new_faces_edges_cartesian = get_cartesian_face_edge_nodes(
        grid.face_edge_connectivity.values,
        grid.edge_node_connectivity.values,
        grid.node_x.values,
        grid.node_y.values,
        grid.node_z.values,
        grid.n_edges_per_face,  # or grid.n_face, depending on what you intended
        grid.n_max_face_edges
    )

    old_faces_edges_cartesian = _get_cartesian_faces_edge_nodes(
        grid.face_node_connectivity.values, grid.n_face,
        grid.n_max_face_edges, grid.node_x.values,
        grid.node_y.values, grid.node_z.values
    )

    nt.assert_array_equal(new_faces_edges_cartesian, old_faces_edges_cartesian)
```
The error messages are:

```python
args = (<built-in function eq>, array([[[[ 0.63933827, -0.42719217, -0.63933827],
         [ 0.57735027, -0.57735027, -0.5773... 0.63933827]],

        [[-0.63933827,  0.42719217,  0.63933827],
         [-0.48565274,  0.48565274,  0.72683068]]]]))
kwds = {'err_msg': '', 'header': 'Arrays are not equal', 'strict': False, 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 4516 / 9216 (49%)
E           Max absolute difference among violations: 0.19509032
E           Max relative difference among violations: 1.60290531e+16
E            ACTUAL: array([[[[ 0.639338, -0.427192, -0.639338],
E                    [ 0.57735 , -0.57735 , -0.57735 ]],
E           ...
E            DESIRED: array([[[[ 0.57735 , -0.57735 , -0.57735 ],
E                    [ 0.639338, -0.427192, -0.639338]],
E           ...

```

philipc2:
> > @hongyuchen1030
> > Here's an optimized version of `get_cartesian_face_edge_nodes` using Numba. Here we re-use the `face_edge_connectivity` and `edge_node_connectivity` that can be computed internally and re-used.
> > This runs in about `90ms` for me, down from the `~500ms` we currently have.
> > ```python
> > @njit(cache=True, parallel=True)
> > def get_cartesian_face_edge_nodes(face_edge_connectivity, edge_node_connectivity,
> >                                   node_x, node_y, node_z,
> >                                   n_edges_per_face, n_face, n_max_face_edges):
> > 
> >     # Allocate result array
> >     # I used np.nan here instead of INT_FILL_VALUE to be consistent with the double type (INT_FILL_VALUE is for our int64 arrays)
> >     face_edges = np.full((n_face, n_max_face_edges, 2, 3), np.nan, dtype=np.float64)
> > 
> >     # Construct edge_nodes for each face in parallel
> >     for face_i in prange(n_face):
> >         n_edges = n_edges_per_face[face_i]
> >         edge_indices = face_edge_connectivity[face_i, :n_edges]
> >         node_indices = edge_node_connectivity[edge_indices]  
> > 
> >         first_nodes = node_indices[:, 0]
> >         second_nodes = node_indices[:, 1]
> > 
> >         # Assign x coordinates
> >         face_edges[face_i, :n_edges, 0, 0] = node_x[first_nodes]
> >         face_edges[face_i, :n_edges, 1, 0] = node_x[second_nodes]
> > 
> >         # Assign y coordinates
> >         face_edges[face_i, :n_edges, 0, 1] = node_y[first_nodes]
> >         face_edges[face_i, :n_edges, 1, 1] = node_y[second_nodes]
> > 
> >         # Assign z coordinates
> >         face_edges[face_i, :n_edges, 0, 2] = node_z[first_nodes]
> >         face_edges[face_i, :n_edges, 1, 2] = node_z[second_nodes]
> > 
> >     return face_edges
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The same can be applied to the spherical one, with changes only needed to the assignment of coordinates.
> 
> Hi @philipc2
> 
> Have you compared the new version's results with the old one? I set up the following test and found that the results differ:
> 
> ```python
> def test_cartesian_test():
>     grid = ux.open_grid(gridfile_CSne8)
>     new_faces_edges_cartesian = get_cartesian_face_edge_nodes(
>         grid.face_edge_connectivity.values,
>         grid.edge_node_connectivity.values,
>         grid.node_x.values,
>         grid.node_y.values,
>         grid.node_z.values,
>         grid.n_edges_per_face,  # or grid.n_face, depending on what you intended
>         grid.n_max_face_edges
>     )
> 
>     old_faces_edges_cartesian = _get_cartesian_faces_edge_nodes(
>         grid.face_node_connectivity.values, grid.n_face,
>         grid.n_max_face_edges, grid.node_x.values,
>         grid.node_y.values, grid.node_z.values
>     )
> 
>     nt.assert_array_equal(new_faces_edges_cartesian, old_faces_edges_cartesian)
> ```
> 
> The error messages are:
> 
> ```python
> args = (<built-in function eq>, array([[[[ 0.63933827, -0.42719217, -0.63933827],
>          [ 0.57735027, -0.57735027, -0.5773... 0.63933827]],
> 
>         [[-0.63933827,  0.42719217,  0.63933827],
>          [-0.48565274,  0.48565274,  0.72683068]]]]))
> kwds = {'err_msg': '', 'header': 'Arrays are not equal', 'strict': False, 'verbose': True}
> 
>     @wraps(func)
>     def inner(*args, **kwds):
>         with self._recreate_cm():
> >           return func(*args, **kwds)
> E           AssertionError: 
> E           Arrays are not equal
> E           
> E           Mismatched elements: 4516 / 9216 (49%)
> E           Max absolute difference among violations: 0.19509032
> E           Max relative difference among violations: 1.60290531e+16
> E            ACTUAL: array([[[[ 0.639338, -0.427192, -0.639338],
> E                    [ 0.57735 , -0.57735 , -0.57735 ]],
> E           ...
> E            DESIRED: array([[[[ 0.57735 , -0.57735 , -0.57735 ],
> E                    [ 0.639338, -0.427192, -0.639338]],
> E           ...
> ```

Looks like there is some small differences. We can move this optimizing into the work I'm putting together in #1195 and keep this PR smaller in scope.

hongyuchen1030:
> Looks like there is some small differences. We can move this optimizing into the work I'm putting together in #1195 and keep this PR smaller in scope.

@philipc2 
Sounds good, Can you approve this PR so it can be merged? Thanks

philipc2:
> > Looks like there is some small differences. We can move this optimizing into the work I'm putting together in #1195 and keep this PR smaller in scope.
> 
> @philipc2 Sounds good, Can you approve this PR so it can be merged? Thanks

Yep! Wrapping up a final review.

uxarray/grid/grid.py
philipc2: @hongyuchen1030 

Were you able to make this change?

uxarray/grid/geometry.py
```diff
@@ -1407,25 +1466,14 @@ def _populate_bounds(
...
-        grid.face_node_connectivity.values,
-        grid.n_face,
-        grid.n_max_face_edges,
-        grid.node_x.values,
-        grid.node_y.values,
-        grid.node_z.values,
-    )
+    # If the face_edges_cartesian and face_edges_lonlat_rad didn't exist,
+    # The following call will automatically populate them
+    faces_edges_cartesian = grid.faces_edges_cartesian.values
```
hongyuchen1030, line 1474: @philipc2 

So in this PR, it's already replaced with the call to `grid.faces_edges_cartesian.values` and `grid.faces_edges_spherical.values`

uxarray/grid/grid.py
hongyuchen1030: I am unable to find such a leftover design in this PR. I have also pinned you to the corresponding `populate_bounds` section because the whole point of the PR was to reduce such duplicate calls. I wonder which commit this file belongs to?

philipc2 (changes requested):
Just to clarify a couple things.

I would prefer not to handle the deletion of the allocated arrays through the `Grid` attributes. 

If we want to keep the attributes (i.e. `Grid.faces_edges_cartesian`), these should not store anything as part of the `Grid._ds`. We would instead use them for convenience within our calls (i.e. what you've already done inside of zonal average).

This loops back to the discussion I had previously with @erogluorhan . Pinging him here if he has anything to add.

I do see the benefit of caching and don't mean to invalidate the optimizations in this PR, however the current issue that is still making the usage of some of our functionality impractical for the highest resolution grids is still the execution time of `Grid.bounds`

I hope this makes sense.

uxarray/grid/geometry.py
philipc2: I believe that we handle this here

https://github.com/UXARRAY/uxarray/blob/a6aa6294975f189e8c36dd3d14da745526e53e06/uxarray/grid/grid.py#L1983

uxarray/grid/grid.py
erogluorhan: Sorry I didn't notice before that I was pinged here. My thoughts are:

I see the motive behind avoiding duplicate calls as much as we can. The memory consumption s also another limiting factor, so we should be looking for the sweet spot between performance vs. memory optimizations to provide the best user experience. 

That said, the discussion in https://github.com/UXARRAY/uxarray/pull/1130#discussion_r1924731908 is important I believe. You two, Hongyu and Philip, may know the details of this PR much better than me, but my first impressions is that `faces_edges_spherical ` is being generated from an existing (or once-generated) connectivity info, and its implementation seems vectorized (despite presence of some potentially expensive Numpy functions like unraveling in it), so I'd be more worried on the added memory consumption here than avoiding duplicate calls.

I am happy to look into more details of the code if needed though.

uxarray/core/zonal.py
```diff
@@ -18,23 +17,15 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
...
-        uxgrid.face_node_connectivity.values,
-        uxgrid.n_face,
-        uxgrid.n_max_face_nodes,
-        uxgrid.node_x.values,
-        uxgrid.node_y.values,
-        uxgrid.node_z.values,
-    )
-
     bounds = uxgrid.bounds.values
+    faces_edges_cartesian = uxgrid.faces_edges_cartesian.values
```
erogluorhan, line 21: I've seen a lot of these `.values` calls in this PR. I see that conversion to Numpy arrays is required for Numba executions, but I want to emphasize a couple things here:

- Are all of the `.values` conversions in this code required for the Numba injections here. Could any of them be avoided?
   - Asking this because Xarray-to-Numpy conversions, i.e. `.values` calls, are dangerous when we have chunked Xarray inputs, which may otherwise not fit into the memory. These `.values` may break the executions in such cases since they will attempt to convert the whole data into Numpy arrays, which couldn't fit into memory.
- Even for those that would be required for Numba calls, we will still need to find a robust longer-term solution, which will probably be a combination of `map_blocks()` or `apply_ufunc()` with `.data` calls where chunked Dask arrays would be efficiently processed through one of these methods.

hongyuchen1030: Hi @erogluorhan ,

These `.values` calls aren’t something new introduced in this PR— If you read the current implementation, you will notice that they follow the existing implementation in the `bounds` function. We’re simply maintaining the same approach here. If you have suggestions for a better long-term solution to handle chunked inputs more robustly (e.g., through `map_blocks()` or `apply_ufunc()` with `.data`), We can open a new PR to address those overall design improvements.

Thanks for bringing this up!

philipc2:
I apologize if I've caused any confusion.


> I understand that you’d prefer not to keep these variables in Grid._ds, but you previously mentioned that it’s okay to have them as properties there. So I'm not entirely clear on where you’d like to store them instead.

Our properties don't need to return cached variables only. We could simply wrap the building functions for `face_edges_cartesian` and `face_edges_spherical` and use those in our implementations. This is operationally the same, but reduces the overall length of our code.

I do see the benefit in caching the variables, especially considering the breakoutof the zonal average call above. However, before we chose to cache and re-use this values, I believe that optimizing the helper function should be the first step.

Since constructing the `face_edges` only involves the manipulation of connectivity variables and indexing existing coordinates, I'd prefer not to use up extra memory throughout the life-span of the Grid.

The current issue with how you've implemented things is that it assumes the user is going to immediately call `.zonal_average()` after constructing the bounds, which isn't always the case. 

Also, the performance of our Zonal Average functionality is at a good spot right now. The current bottle neck still lies in the initial bounds construction, which has come a long way since our initial implementation, but could still use extra optimizations. The change here would not lead to a noticeable change in perfromance.


| Resolution | Nodes      | Faces      | Edges       | Grid Load Time (s) | Bounds Construction Time (s) | Total Time (s) |
|------------|------------|------------|-------------|-------------------:|----------------------------:|---------------:|
| 30km       | 1,310,720  | 655,362    | 1,966,080   | 1.923              | 2.541                      | 4.464          |
| 15km       | 5,242,880  | 2,621,442  | 7,864,320   | 7.489              | 10.191                     | 17.68          |
| 7.5km      | 20,971,520 | 10,485,762 | 31,457,280  | 29.228             | 37.866                     | 67.095         |
| 3.75km     | 83,886,080 | 41,943,042 | 125,829,120 | 113.028            | 155.486                    | 268.514        |

To summarize:
- We should investigate further optimizations into constructing the `face_edge_` variables before considering caching
  - If we reach a wall here, we could consider adding a configuration parameter that the user can select if they would like to consider caching. 
- We should continue to optimize the Bounds construction, since this is the most noticeable slowdown (even though it is a one-time execution) 

I am happy to have a discussion about this during this weeks UXarray meeting.

uxarray/core/zonal.py
erogluorhan: Thanks @hongyuchen1030 ! Just to clarify, yes, I can see those `.values` in the older code, but I just wanted to remind how keeping those `.values` calls in the code would be dangerous for larger-than-memory data. 

So, even though we've had them in the code so far, we will need to avoid them (or get rid of the existing ones) as much as we can in our future code to fully optimize UXarray for memory usage and parallelization.  

If you want to keep this PR's scope limited to what you've done already though, I am just fine with it. Things I am pointing out can be handled in separate work.

See the [Consistently use Xarray for Grid Reader, Improved Dask Support(https://github.com/UXARRAY/uxarray/pull/1092) PR, which was one of the most recent optimization work we had, you haven't seen yet. This work didn't involve any Numba though, so it was relatively easier to handle. 

To handle Numba and this together, we'll need to look into mechanisms I mentioned in the first comment.

hongyuchen1030: Thanks for the information, @erogluorhan .

I agree that addressing the extensive use of `.values` throughout UXarray is a significant task that goes beyond the scope of this PR—it essentially calls for an overall redesign of the codebase to optimize memory usage and parallelization.

 I believe the points you've raised should be tracked in a new PR or issue dedicated to this redesign. That way, we can ensure these considerations are not lost and are properly addressed in a future effort.

For now, I'll keep the changes in this PR as-is.

Thanks again for your insights!

hongyuchen1030:
> That is understandable, however having the bounds is a valuable addition to the user API, as it has applications to visualization and subsetting/cross-section functionality that we have implemented.

I can proceed to revoke the changes. 

Just some comments on this:
>That is understandable, however having the bounds is a valuable addition to the user API, as it has applications to visualization and subsetting/cross-section functionality that we have implemented.

This still means the latlonbound is served as an intermediate process. And the right way to do is to optimize the subsetting/cross-section functionality as a whole, and even provide a tailored latlonbounds building for these functionalities in the future

uxarray/core/zonal.py
erogluorhan: Sounds good to me; thanks @hongyuchen1030 !

philipc2:
@hongyuchen1030 

I am putting together optimizations to the constructing of `Grid.bounds` in #1205 to compute the face edge nodes for each face as they are needed, instead of pre-computing the entire array.

Once that is merged, we should considering doing something similar for the zonal face weights. We could tackle that as part of this PR.

erogluorhan:
@hongyuchen1030 given that @philipc2 's improvements have gone into release, should we close this one and its corresponding issue #1181?

erogluorhan:
Confirmed with @hongyuchen1030 that this can be closed.

Issue #1176, opened 2025-03-07
Add flag for area correction


rajeeja:
### Proposed new feature or change:

As a followup to PR #1120 change function signature, work with only cartesian coordinates, remove tests that use this function directly with coords_type. 

Possible work with edges defining the face. 
Also, need the flag to specify line of constant lattitude for each edge

```
def calculate_face_area(
    x,
    y,
    z,
    quadrature_rule="gaussian",
    order=4,
    coords_type="spherical",
    latitude_adjusted_area=False,
):
```

PR #1188, opened 2025-03-19
Update Face Area API to only use Cartesian Coordinates

rajeeja:
algorithm works with cartesian coords only, towards #1176


philipc2 (approved):
Nice work on this!

PR #1203, opened 2025-04-08
Use `n_max_face_nodes` instead of `n_max_face_edges` in Bounds construction 

### LINKED ISSUE #1202: Use `n_max_face_nodes` instead of `n_max_face_edges` in Bounds construction
When constructing the face_edges_nodes, we pass in `n_max_face_edges`, which triggers the computation of 
the `face_edge_connectivity`. We can use `n_max_face_nodes` instead, since they are equivalent. 

This significantly improves the perfromance if the `face_edge_connectivity` is not already present in the Grid.

## Before
Image

## After

Image

philipc2:
Closes #1202 

## Overview

- Avoids triggering the computation of `Grid.face_edge_connectivity` when constructing the face edge coordinates.
- Improves the execution time (see #1202) when the connectivity is not present

### DIFF
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -1413,7 +1413,7 @@ def _populate_bounds(
     faces_edges_cartesian = _get_cartesian_face_edge_nodes(
         grid.face_node_connectivity.values,
         grid.n_face,
-        grid.n_max_face_edges,
+        grid.n_max_face_nodes,
         grid.node_x.values,
         grid.node_y.values,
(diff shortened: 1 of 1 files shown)


philipc2:
> Great catch! Approving this but curious: There are a number of places where `n_max_face_edges` is used, would there be similar opportunities in those places, too?

Yes. I could do a sweep and update those accordingly too.

philipc2:
Closing this, as it will not be needed after #1205

PR #1204, opened 2025-04-09
Remove min/max lat/lon print statement in spatialhash initialization

fluidnumerics-joe:
## Overview
This PR is a simple patch that removes an annoying print statement in the spatial hash initialization that prints the min/max lat/lon values.


philipc2 (approved):
Good catch! We must have missed this.

fluidnumerics-joe:
> Good catch! We must have missed this.

Lots of moving parts :)

PR #1207, opened 2025-04-10
Changed np.atan2() to np.arctan2() arcs.py

### LINKED ISSUE #1206: np.atan2() vs. np.arctan2() in grid/arcs.py
I was curious whether there is a reason to use np.atan2() as opposed to np.arctan2() here:
https://github.com/UXARRAY/uxarray/blob/ca64f4eb472c5a30b23e734b37878601b3db7f1c/uxarray/grid/arcs.py#L369
Perhaps it would be a very simple way to maintain backwards-compatibility by using the latter?

I was running the non-conservative UxDataArray.zonal_mean() on v2025.3.0 with numpy 1.26.4 (trying to figure out how to solve numpy update dependency issues), which gave the error
`AttributeError: module 'numpy' has no attribute 'atan2'. Did you mean: 'arctan2'?`

jpan84:
Closes #1206

This calls the equivalent underlying function but maintains backwards-compatibility with older numpy versions that didn't have atan2().

### DIFF
--- a/uxarray/grid/arcs.py
+++ b/uxarray/grid/arcs.py
@@ -366,6 +366,6 @@ def compute_arc_length(pt_a, pt_b):
     rho = np.sqrt(1.0 - z1 * z2)
     cross_2d = x1 * y2 - y1 * x2
     dot_2d = x1 * x2 + y1 * y2
-    delta_theta = np.atan2(cross_2d, dot_2d)
+    delta_theta = np.arctan2(cross_2d, dot_2d)
 
     return rho * abs(delta_theta)
--- a/uxarray/grid/coordinates.py
+++ b/uxarray/grid/coordinates.py
@@ -53,7 +53,7 @@ def _xyz_to_lonlat_rad_no_norm(
         Latitude in radians
     """
 
-    lon = np.atan2(y, x)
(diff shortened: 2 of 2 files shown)


philipc2:
We should change these too:

https://github.com/UXARRAY/uxarray/blob/ca64f4eb472c5a30b23e734b37878601b3db7f1c/uxarray/grid/coordinates.py#L56

https://github.com/UXARRAY/uxarray/blob/ca64f4eb472c5a30b23e734b37878601b3db7f1c/uxarray/grid/coordinates.py#L79

https://github.com/UXARRAY/uxarray/blob/ca64f4eb472c5a30b23e734b37878601b3db7f1c/uxarray/grid/coordinates.py#L781

jpan84:
I've changed the 3 instances in coordinates.py as well

philipc2 (approved):
Thanks for fixing this & bringing this to our attention!

rajeeja:
> `np.asin()` an `np.acos()` to preserve backward combability?

I think this is not needed unless we use very old ~10 yr old version of numpy.

rajeeja (approved):
Thanks fot the fixes. @jpan84

PR #1229, opened 2025-04-29
o Add healpix bilinear remapping

rajeeja:


aaronzedwick:
This should be helpful for the upcoming hackathon, thanks @rajeeja

docs/user-guide/remapping.ipynb
```diff
@@ -692,10 +692,40 @@
     ")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "9ae38e84",
+   "metadata": {},
+   "source": [
+    "#### Also do a coarse HealPix grid and downsampling with it"
```
aaronzedwick, line 700: Can this be changed to something more like this:
"Here we also demonstrate downsampling to a coarse HealPix grid."

aaronzedwick (commented):
Thanks @rajeeja!

PR #1231, opened 2025-05-02
DRAFT: Support `ccrs.Geodetic` source projection, add Polygon plot to Grid 

### LINKED ISSUE #1230: HEALPix (maybe even others) connectivities and split visualizations have issues around the pole
### Version

v2025.04.0

Conda

### What happened?

Using UXarray to visualize some coarse resolution HEALPix grids, I notice some issues as follows:

1. Face edges do not seem like great circle arc in ours but the do in the easy.gems visualizations.

UXarray    
Image 

easy.gems
Image

3. There should be some issues around the Poles, which might be due to connectivity generation and/or periodic elements handling (Western faces being split but the Eastern ones, no splitting seems needed from easy.gems), but there does not seem to be issues with the actual corner nodes and face centers:

Image

Image

Image

Image

Image

Already clarified above I believe

```python
# Let us open a pretty coarse HEALPix grid because we will visualize it later
uxgrid = ux.Grid.from_healpix(zoom=0, pixels_only=False)

uxgrid.plot(
    periodic_elements="split",         
    projection=ccrs.Mollweide(),
    width=800,
    title="HEALPix, zoom=0")

# Other plotting code in the figures above can also be inspected here
```

philipc2:
Closes #1230

## Overview

- Adds support for using Cartopy's `Geodetic` projection as a source projection, allowing for better visualization of grids on the surface of a sphere and for handling splitting along the dateline: https://github.com/holoviz/geoviews/issues/107
- Using the Geodetic projection is especially useful for coarser grids where the distortion is more noticeable, and over the poles 
- Implements a `Grid.plot.polygons()` method that plots shaded polygons using the index of each face. 

### No Source Projection

image

### Geodetic Source Projection

image

### DIFF
--- a/test/test_plot.py
+++ b/test/test_plot.py
@@ -3,6 +3,7 @@ import uxarray as ux
 import holoviews as hv
 import pytest
 from pathlib import Path
+import cartopy.crs as ccrs
 
 current_path = Path(os.path.dirname(os.path.realpath(__file__)))
 
@@ -24,6 +25,7 @@ def test_topology():
         uxgrid.plot.mesh(backend=backend)
         uxgrid.plot.edges(backend=backend)
         uxgrid.plot.nodes(backend=backend)
+        uxgrid.plot.polygons(backend=backend)
         uxgrid.plot.node_coords(backend=backend)
         uxgrid.plot.corner_nodes(backend=backend)
         uxgrid.plot.face_centers(backend=backend)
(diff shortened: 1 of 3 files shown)


erogluorhan (changes requested):
This is great! Some thoughts on a few things though:

- The name `grid.plot.polygons()` can potentially be confused with our `polygons()` visualizations that plot data values based polygons instead. 
   - How about renaming it from `polygons` to something else? 
      - Just an idea: This is only for topology diagnostics purposes, right? Maybe something to emphasize this aspect better, which'd also help avoid confusions with the other polygons? e.g. `grid.plot.shade_faces()` , etc?
- In addition to renaming, if still needed, maybe also consider either adding a bigger emphasis on the purpose of this plotting function into the docstrings or displaying a warning message when this is called
- Values in the color bars in both cases of `fill` argument for this function do not make much sense, e.g. face indices being float, or those float values around 1 when it is polygons.

philipc2:
Closing this, as the `ccrs.GeoDetic()` projection is already supported. Will reopen an issue and PR with changes to documentation and/or usage if needed in the future.

PR #1232, opened 2025-05-02
Derive HEALPix area and add missing parameter to `UxDataset.from_healpix()`

philipc2:
Closes #XXX

## Overview

- Derives the HEALPix area for each pixel instead of computing it internally
- Adds the `pixels_only` parameter to `UxDataset.from_healpix()`

### DIFF
--- a/test/test_healpix.py
+++ b/test/test_healpix.py
@@ -3,6 +3,7 @@ import healpix as hp
 import pytest
 import os
 from pathlib import Path
+import numpy as np
 
 
 current_path = Path(os.path.dirname(os.path.realpath(__file__)))
@@ -18,6 +19,10 @@ def test_to_ugrid(resolution_level):
     expected_n_face = hp.nside2npix(hp.order2nside(resolution_level))
 
     assert uxgrid.n_face == expected_n_face
+
+    n_side = uxgrid._ds.attrs["n_side"]
+    expected_area = np.ones(uxgrid.n_face) * np.pi / 3 * n_side**2
+    assert np.array_equal(uxgrid.face_areas.values, expected_area)
 
 @pytest.mark.parametrize("resolution_level", [0, 1, 2, 3])
 def test_boundaries(resolution_level):
@@ -41,6 +46,9 @@ def test_dataset():
 
     assert uxds.uxgrid.source_grid_spec == "HEALPix"
     assert "n_face" in uxds.dims
+
+    uxds = ux.UxDataset.from_healpix(ds_path, pixels_only=False)
+    assert "face_node_connectivity" in uxds.uxgrid._ds
 
 
 
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -1902,8 +1902,11 @@ class Grid:
         array([0.00211174, 0.00211221, 0.00210723, ..., 0.00210723, 0.00211221,
             0.00211174])
         """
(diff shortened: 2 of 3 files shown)


uxarray/grid/grid.py
```diff
@@ -1902,8 +1902,13 @@ def compute_face_areas(
...
         """
-        # if self._face_areas is None: # this allows for using the cached result,
-        # but is not the expected behavior behavior as we are in need to recompute if this function is called with different quadrature_rule or order
+        if self.source_grid_spec == "HEALPix":
+            # Derive HEALPix Area
+            # Ω_pix = 4π / (12 * Nside²) = π / (3 * Nside²)
+            # Górski et al. 2005, “HEALPix: A Framework for High‐Resolution Discretization and Fast Analysis of Data Distributed on the Sphere”
+            n_side = self._ds.attrs["n_side"]
+            face_area = np.pi / (3 * n_side**2)
+            return np.ones(self.n_face) * face_area, None
```
erogluorhan, line 1911: While I get the reasoning about this for achieving actual equal-area faces for HEALPix, I am a bit concerned about deriving HEALPix area as in here:

- Actually we are not generating our edges in the topology in the same way as HEALPix does (I mean the actual topology, not visualization), right? So, each pixel's edges being generated in our way shouldn't be resulting in equal areas as it is being attempted to derived here . 
- We are kind of trying to hard-code the area values now in here.
- Where would this be a problem? 
   - Probably where an operator uses both edge information and area information.

test/test_healpix.py
```diff
@@ -19,6 +20,10 @@ def test_to_ugrid(resolution_level):
 
     assert uxgrid.n_face == expected_n_face
 
+    n_side = uxgrid._ds.attrs["n_side"]
+    expected_area = np.ones(uxgrid.n_face) * np.pi / (3 * n_side**2)
+    assert np.array_equal(uxgrid.face_areas.values, expected_area)
```
erogluorhan, line 25: See my comments on the area derivation part in `grid.py`

erogluorhan (changes requested):
Please see my comments this PR trying to derive areas according to HEALPix without actually changing the edge topology. Maybe for now:

- Hold off on deriving the areas for HEALPix like it is done in this PR
- Create an issue about the face areas currently not being equal
   - Maybe also discuss things like whether the edge topology generation should be changed for HEALPix? This issue would be something good to discuss during the upcoming hackathon.

rajeeja:
Didn't go through the change, but for a HEALPix mesh the area of each pixel is predetermined and equal across the entire sphere. We don't need to compute it using our routines, it can be a simple formula and we can bypass our routines.

`4*Pi/(12 * NSIDE**2)`

philipc2:
> Didn't go through the change, but for a HEALPix mesh the area of each pixel is predetermined and equal across the entire sphere. We don't need to compute it using our routines, it can be a simple formula and we can bypass our routines.
> 
> `4*Pi/(12 * NSIDE**2)`

That's correct. This is what I'm looking to implement here.

philipc2:
@erogluorhan 

I'm happy to close this for now and keep #1236 open, unless you envision us making any of these changes in the near future.

erogluorhan:
> @erogluorhan
> 
> I'm happy to close this for now and keep #1236 open, unless you envision us making any of these changes in the near future.

Yeah, that works with me!

PR #1255, opened 2025-05-14
Add Azimuthal Averaging 

### LINKED ISSUE #1254: Compute azimuthal mean in radial rings around a central point.

Bin data points by their radius from a central point and compute the weighted mean for each bin to obtain a radial profile of the input variable

jpan84:
Closes #1254 

Compute azimuthal mean in radial rings around a central point.

## Overview
Like `uxarray.UxDataArray.zonal_mean()`, this aggregates over faces to output the data with a new horizontal dimension `radius`. Following `uxarray.UxDataArray.subset.bounding_circle`, this function uses `ball_tree.query_radius` to select grid elements in a circular neighborhood from one center coordinate. For radii r=[r1, r2, r3, ...], face centers at distance d (given in great-circle degrees), where r2<d<=r3 will be used for the azimuthal mean at r3.

This is one solution for issue #1254 

This may be tangentially relevant to issue #930.

## Expected Usage
```Python
import uxarray as ux

grid_path = '/glade/p/cesmdata/inputdata/share/scripgrids/ne120np4_pentagons_100310.nc'
data_path = '/glade/work/jpan/azim_test_data.nc'

uxda = uxds['OMEGA500']
print(uxda)

azim_mean, hit_count = uxda.azimuthal_mean((183., -18.933), 5, 0.3)

print(azim_mean)
print(hit_count)
plt.plot(azim_mean.radius, azim_mean.isel(time=0))
plt.show()
```

## PR Checklist
**General**
- [✓] An issue is linked created and linked
- [ ] Add appropriate labels
- [✓] Filled out Overview and Expected Usage (if applicable) sections

**Testing**

**Documentation**
- [✓ ] Docstrings have been added to all new functions
- [N/A ] Docstrings have updated with any function changes

**Examples**


jpan84:
sorry, this is my 1st time creating a new feature. I'm not sure how to properly link this to the issue or attach a "new label" feature.

philipc2:
This seems similar to #941

philipc2:
> > > The functionality seems great!
> > > I think I am hesitant though about returning `UxDataset` out of such an aggregation function (it likewise applies to `zonal_mean()` as well; I think I overlooked it before): Shouldn't the logic be to create x number of scalars that are corresponding to the mean of each circular region, and not to worry about providing an USG topology? If so, shouldn't return types simply be np.ndarray, or something similar?
> > > cc: @rajeeja
> > 
> > 
> > Good suggestion. Since there's no longer a valid grid associate with the output, the result should be an `xarray.DataArray/Dataset` (like with our cross-sections).
> 
> Absolutely! Probably multiple `xarray.DataArray`'s as needed rather than a `xarray.Dataset`, but not fully against the latter.

Yep! Added a `return_hit_counts` parameter to return the hits as an optional DataArray as opposed to being stored in the Dataset

erogluorhan (commented):
Code looks great to me; thanks very much! No plans with an immediate notebook into the user guide with this PR?  If the preference is to add that later, could you please create an issue regarding that, if none exists. Also, who is going to write that guide,  @jpan84 maybe you?

philipc2:
> No plans with an immediate notebook into the user guide with this PR? If the preference is to add that later, could you please create an issue regarding that, if none exists. Also, who is going to write that guide, @jpan84 maybe you?

Any thoughts @jpan84? I'm happy to move forward with this PR with the simple docstring example, and create an issue to develop a notebook down the line.

erogluorhan (approved):
I am approving it to help make its way forward. Thanks a lot for this addition!

rajeeja:
Good PR. Thanks this work and I agree that we should not return Uxdataset here as already handled. Couple of minor points: 
Handle case where no faces are found at all `np.all(hit_count == 0)`  also some parameter validations outer_radius <= 0 or radius_step <= 0, len(center_coord) != 2 , something for lon/lat deg..

rajeeja (changes requested):
Minor safeguards as pointed in comments

Issue #669, opened 2024-01-19
Support reading GeoDataFrame 


philipc2:
### Proposed new feature or change:

Support reading GeoDataFrames with geometries and data variables into UXarray data structures.

kmsampson:
Thanks @philipc2 for adding this as an issue!

philipc2:
@rajeeja 

Can you check if we can close this?

rajeeja:
We have uxarray and ugrid compliant support already. Closing this now.

PR #1263, opened 2025-05-15
more geometry for geopandas

rajeeja:
#669


uxarray/plot/accessor.py
```diff
@@ -31,7 +31,16 @@ def __init__(self, uxgrid: Grid) -> None:
         self._uxgrid = uxgrid
 
     def __call__(self, **kwargs) -> Any:
-        return self.edges(**kwargs)
+        # Decide what to plot based on what's in the Grid
+        if self._uxgrid.n_face > 0:
+            # If we have faces, use the existing edges plotting
+            return self.edges(**kwargs)
+        elif self._uxgrid.n_edge > 0:
```
erogluorhan, line 38: While I get these checks here, is it practically possible to have grids where we wouldn't have faces but edges, or not even edges but nodes?

uxarray/plot/accessor.py
```diff
@@ -77,11 +87,32 @@ def points(self, element="nodes", backend=None, **kwargs):
...
+        # Set default kwargs for better visualization
+        if "color" not in kwargs:
+            kwargs["color"] = "blue"
+        if "size" not in kwargs:
+            kwargs["size"] = 50
+        if "title" not in kwargs:
+            if self._uxgrid.n_face == 0 and self._uxgrid.n_edge == 0:
+                kwargs["title"] = "Points"
+            else:
+                kwargs["title"] = f"{element.capitalize()}"
```
erogluorhan: A few things: 
- With these new defaults here, are we bypassing the defaults of `points_df.hvplot.points()`? 
- I don't know if we need to bypass them, but if you think that it contributes to a better plot, I'd at least suggest that we read these from constants.py or something similar to that maybe for plotting constants rather than hard-coding into the function. 
- Not sure providing a default title would help or confuse though

uxarray/plot/accessor.py
```diff
@@ -77,11 +87,32 @@ def points(self, element="nodes", backend=None, **kwargs):
...
+            else:
+                central_longitude = 0.0
+            kwargs["crs"] = ccrs.PlateCarree(central_longitude=central_longitude)
+
         verts = {"lon": lon, "lat": lat}
 
         points_df = pd.DataFrame.from_dict(verts)
 
-        return points_df.hvplot.points("lon", "lat", **kwargs)
+        return points_df.hvplot.points("lon", "lat", geo=True, **kwargs)
```
erogluorhan: I've seen cases where we needed to use `geo=True`, but what is the expectation with defaulting to it for the function? Also what is the extent, I mean I'd be curious if it'd impact how the plots behave overall?

uxarray/plot/accessor.py
```diff
@@ -309,6 +340,95 @@ def face_area_distribution(
 
         return histogram
 
+    def linestrings(
+        self,
+        periodic_elements="exclude",
+        backend=None,
+        engine="spatialpandas",
+        **kwargs,
+    ):
```
erogluorhan: Maybe I am missing a detail here, but this function seems pretty much like same/similar results with `.edges()` plot. Could you please clarify?

uxarray/plot/accessor.py
```diff
@@ -309,6 +340,95 @@ def face_area_distribution(
...
+        # Set default style parameters
+        if "color" not in kwargs:
+            kwargs["color"] = "blue"
+        if "line_width" not in kwargs and "linewidth" not in kwargs:
+            if backend == "matplotlib":
+                kwargs["linewidth"] = 2
+            else:
+                kwargs["line_width"] = 2
+        if "title" not in kwargs:
+            kwargs["title"] = "LineString Geometries"
```
erogluorhan: Similar thoughts to the above comments about these new defaults above

uxarray/plot/accessor.py
```diff
@@ -309,6 +340,95 @@ def face_area_distribution(
...
+            for edge in edge_conn:
+                start_node, end_node = edge
+                if start_node >= 0 and end_node >= 0:  # Ensure valid indices
+                    line = LineString(
+                        [
+                            (node_lon[start_node], node_lat[start_node]),
+                            (node_lon[end_node], node_lat[end_node]),
+                        ]
+                    )
+                    geometries.append(line)
```
erogluorhan: The construction of `geometries` in the for-loop looks more like a helper function rather than plotting, and since it is iterative, it could benefit from Numba I believe. What is the execution performance of this for km-scale grids?

erogluorhan (commented):
Thanks a lot for this! I'll need to make a second pass to review `uxarray/io/_geopandas.py` and its tests, but please find my initial review about others

uxarray/plot/accessor.py
rajeeja: Yes, sometimes there are point clouds or simply railway lines etc. that geopandas support reading. It is not common, but may be useful in some cases.

uxarray/plot/accessor.py
rajeeja: Excellent observation.`.linestring()` manually constructs LineString geometries from edge_node_connectivity, while, `.edges() ` use the existing to_geodataframe() method with edge data. I am thinking harder to see if there is any reason to maintain both; might modify the existing` .edges()` to work for both.

uxarray/plot/accessor.py
rajeeja: When plotting lon/lat coordinates, we almost always want proper geographic projection handling, without geo=True, hvplot treats coordinates as simple x/y values, which can lead to distorted visualizations (especially at higher latitudes)

rajeeja: Have you encountered specific use cases where geo=False would be preferred for lon/lat data?

uxarray/plot/accessor.py
rajeeja: Right observation, I got carried away in the notebook that I was curating for this PR. The idea was to have some defaults that would improve the out-of-box experience. Another approach: we strip back to just the essential geographic defaults and let hvplot handle the rest? - this would be cleaner.

uxarray/plot/accessor.py
erogluorhan: I see. Did you test the capability of whole UXarray to handle those scenarios, not only this vis. func.? Asking because we made some strong assumptions about having faces in the topology, e.g. face-node-connectivity is a minimum requirement.

uxarray/plot/accessor.py
erogluorhan: As long as the code is clean and readable, and duplicate code is avoided, either would be fine.

uxarray/plot/accessor.py
rajeeja: Ofcourse a linestring or point cloud would fail for integrate area calculation or other functions, I think there is still value in supporting these and the user would know that these aren't meant for major 2D grid type operations.

erogluorhan: OK. Looks like (for the text cases you put together) UXarray is able to at least read without any issues grids with linestrings, multilinestrings, points, or multipoints, right? Then we should be fine

uxarray/plot/accessor.py
erogluorhan: I see often that the users' workflows include plots of the data without any projection to quick inspect it in addition to many they generate with projections. 

I think I really like the boiler-plate of `polygons()` in this same file, and you may want to use the same aprroach here and in edges:

```
if projection is not None:
    kwargs["projection"] = projection
    kwargs["geo"] = True
```

I think this really addresses your point well: If projection is desired, then make `geo=True`. The only caveat here would be that if the user also specified `geo=False` in `kwargs`, we are forcing it to `True`, but for a good reason, and that shouldn't be an issue I think. 

Thoughts?

rajeeja: > 
> Thoughts?

Agreed. I'll pushes changes soon. Testing all notebook plots, I broke some, fixing them now.

rajeeja: done

uxarray/plot/accessor.py
```diff
@@ -31,24 +31,33 @@ def __init__(self, uxgrid: Grid) -> None:
...
+        # Decide what to plot based on what's in the Grid
+        if self._uxgrid.n_face > 0:
+            # If we have faces, use the existing edges plotting
+            return self.edges(**kwargs)
+        elif self._uxgrid.n_edge > 0:
+            # If we have edges but no faces, plot edges
+            return self.edges(**kwargs)
+        else:
+            # If we only have nodes, plot points
+            return self.points(**kwargs)
```
philipc2, line 43: This is a pretty strict requirement that we have enforced with the UGRID conventions, where we need to have valid faces. 

Otherwise, much of our functionality breaks. The only case where we have loosened up this requirement was for HEALPix, where we simply delay the creation of the `face_node_connectivity`.

rajeeja: We do support loading of point clouds etc., agree that it won't be UGRID compliant, but it serves other viz. analysis purposes 

Other functinality won't be used with this loaded.

philipc2: > We do support loading of point clouds etc., agree that it won't be UGRID compliant, but it serves other viz. analysis purposes

While we do support point clouds, the connectivity is generated by at the loading step, so there are always faces represented.

> Other functinality won't be used with this loaded.

Is there anyway we could safeguard this?

rajeeja: > > We do support loading of point clouds etc., agree that it won't be UGRID compliant, but it serves other viz. analysis purposes
> 
> While we do support point clouds, the connectivity is generated by at the loading step, so there are always faces represented.
> 
> > Other functinality won't be used with this loaded.
> 
> Is there anyway we could safeguard this?

Calling a validate before all our functions.

philipc2: > Calling a validate before all our functions.

This could work, but would create an undesired performance impact.

philipc2:
The added functionality here is great, however I have some concerns as I've mentioned about with regards to compliance with the UGRID conventions and what would make the most sense with our package.

Supporting Polygons and MultiPolygons is great, as we can comfortably represent these in the UGRID conventions and there are many cases where users may provide data in these formats. However, loading in other geometries like Points or LineStrings leads to the compliance issues.

I'll ping @erogluorhan if he has any feedback. But to summarize, I think we should only support Polygon and MultiPolygon geometries, both for the reading and output.

rajeeja:
Based on our discussion during the Tuesday meeting, I will close this PR.

Issue #1260, opened 2025-05-15
slow import


brianpm:
I've noticed that recent versions of uxarray seem to take a while to import. On casper I am usually seeing more than 10seconds (and less than a minute, with quite a bit of variation). Using version '2025.5.1' with python 3.12.3.

rajeeja:
I noticed this also, indeed this is an issue, I wrote a script to test this out on my local osx machine, v2024.07.1 is an outlier at 11.26 s, recent one's are around 5.5, v2024.08.2 and v2024.04.0 are in 4 s range. 

Will run the profiler to check what's taking long..

```
v2025.05.1: 5.3290 s

v2025.05.0: 5.4735 s

v2025.04.0: 5.5810 s

v2025.03.0: 5.5754 s

v2025.02.1: 5.4720 s

v2025.02.0: 5.4160 s

v2025.01.0: 5.5482 s

v2024.12.0: 5.3332 s

v2024.11.1: 5.5428 s

v2024.11.0: 6.3427 s

v2024.10.1: 6.2609 s

v2024.10.0: 5.3786 s

v2024.08.2: 4.2361 s

v2024.08.1: 4.2115 s

v2024.08.0: 4.3526 s

v2024.07.1: 11.2629 s

v2024.07.0: 4.2546 s

v2024.06.0: 4.1834 s

v2024.05.0: 4.5880 s

v2024.04.0: 4.2596 s
```

rajeeja:
There are several, but this on: 

```
from holoviews.operation.datashader import rasterize as hds_rasterize
```

takes forever.. from file: uxarray/plot/dataarray_plot.py

philipc2:
#1265

PR #1265, opened 2025-05-16
Reduce Import TIme 

rajeeja:
Need to do more profiling and lazy loading etc. lots of room for improvement #1260


rajeeja:
On my local the time has gone down from 5.4 to 3.1 now!

philipc2:
Nice work @rajeeja 

Going to push a few more changes that I had in mind also.

philipc2:
## Before
image


## After

image

aaronzedwick:
Why are we removing our color maps? @philipc2

philipc2:
> Why are we removing our color maps? @philipc2

These were being computed eagerly when loading the package and leading to degraded import times, and the blue/green ones are not great in terms of accessibility.

PR #1271, opened 2025-05-20
DRAFT: Native Matplotlib Plotting Routines 

philipc2:
Closes #XXX

### DIFF
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -424,7 +424,7 @@ class Grid:
             ds = _regional_delaunay_from_points(_points, boundary_points)
         else:
             raise ValueError(
-                f"Unsupported method '{method}'. Expected one of ['spherical_voronoi', 'spherical_delaunay']."
+                f"Unsupported method '{method}'. Expected one of ['spherical_voronoi', 'spherical_delaunay', 'regional_delaunay']."
             )
 
         return cls.from_dataset(dataset=ds, source_grid_spec=method)
--- a/uxarray/plot/matplotlib.py
+++ b/uxarray/plot/matplotlib.py
@@ -1,0 +1,95 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from uxarray import UxDataArray
+
+
+def uxplot(data: UxDataArray, ax=None, **kwargs):
+    fig = ax.get_figure()
+    fig.canvas.draw()
+    res = _sample_grid(data, ax)
+
+    im = ax.imshow(
+        res,
+        origin="lower",
+        extent=ax.get_xlim() + ax.get_ylim(),
+        **kwargs,
+    )
+
+    return im
+
+
+def uxcontour(data: UxDataArray, ax=None, **kwargs):
+    fig = ax.get_figure()
+    fig.canvas.draw()
+    res = _sample_grid(data, ax)
+
+    im = ax.contour(
+        res,
+        origin="lower",
+        extent=ax.get_xlim() + ax.get_ylim(),
+        **kwargs,
+    )
+
+    return im
+
+
+def _get_points_from_axis(ax):
+    import cartopy.crs as ccrs
+
+    x0, x1 = ax.get_xlim()
+    y0, y1 = ax.get_ylim()
+
+    _, _, nx, ny = np.array(ax.bbox.bounds, dtype=int)
+
+    dx = (x1 - x0) / nx
+    dy = (y1 - y0) / ny
+    x = np.linspace(x0 + dx / 2, x1 - dx / 2, nx)
+    y = np.linspace(y0 + dy / 2, y1 - dy / 2, ny)
+    x2d, y2d = np.meshgrid(x, y)
+
+    lonlat = ccrs.PlateCarree().transform_points(ax.projection, x2d, y2d)
+
+    valid = np.isfinite(lonlat[..., 0]) & np.isfinite(lonlat[..., 1])
+
+    lons = np.deg2rad(lonlat[..., 0][valid])
+    lats = np.deg2rad(lonlat[..., 1][valid])
+
+    xs = np.cos(lats) * np.cos(lons)
+    ys = np.cos(lats) * np.sin(lons)
+    zs = np.sin(lats)
+    pts = np.vstack((xs, ys, zs)).T
+
+    return pts, valid, nx, ny
+
+
+def _sample_grid(
+    data: UxDataArray,
+    ax=None,
+):
+    pts, valid, nx, ny = _get_points_from_axis(ax)
+    face_indices, counts = data.uxgrid.get_faces_containing_point(
+        point_xyz=pts,
+        return_counts=True,
+    )
+
+    # TODO: instead of picking the first face, consider an average
+
+    # pick the first face
+    first_face = face_indices[:, 0]
+    first_face[counts == 0] = -1
+
+    # build an array of values for each valid point
+    flat_vals = np.full(first_face.shape, np.nan, dtype=float)
+    mask_has_face = first_face >= 0
+    flat_vals[mask_has_face] = data.values[first_face[mask_has_face]]
+
+    # scatter back into a full raster via the valid mask
+    res = np.full((ny, nx), np.nan, dtype=float)
+    res.flat[np.flatnonzero(valid.ravel())] = flat_vals
+
+    return res


philipc2:
@brianpm 

Let me know if you have any feedback on this. I'm curious whether you think `imshow()` is an appropriate name here or if you have any other ideas.

philipc2:
Rendering some data from the hackathon (HEALPix zoom=10) only takes about 5s on Derecho.

image

kafitzgerald (commented):
I'm a little worried that the API here might be confusing and not particularly intuitive for users familiar with conventions in the broader scientific Python ecosystem for a couple of reasons.

The first being that it deviates from other high level plotting APIs in the scientific Python ecosystem that use `.plot()` and `.hvplot()`.  This is definitely beyond the scope of this PR, but adds some potential for confusion and complexity here. 

And the second that `imshow()` here is pretty different from how it's used in Xarray and Matplotlib (e.g. there's a good bit going on with the resampling, Cartopy is used by default, etc.).  Maybe a new name would be appropriate?  

That said, I don't know that I have a great suggestion for what to do here.  I suspect that something chained onto the `.plot()` accessor would be more intuitive, but that seems a bit complicated and potentially confusing given what's already present. Maybe implementing something under `uxarray.plot.mpl.plot()` for a high level Matplotlib plotting interface could work?

erogluorhan:
> I'm a little worried that the API here might be confusing and not particularly intuitive for users familiar with conventions in the broader scientific Python ecosystem for a couple of reasons.
> 
> The first being that it deviates from other high level plotting APIs in the scientific Python ecosystem that use `.plot()` and `.hvplot()`. This is definitely beyond the scope of this PR, but adds some potential for confusion and complexity here.
> 
> And the second that `imshow()` here is pretty different from how it's used in Xarray and Matplotlib (e.g. there's a good bit going on with the resampling, Cartopy is used by default, etc.). Maybe a new name would be appropriate?
> 
> That said, I don't know that I have a great suggestion for what to do here. I suspect that something chained onto the `.plot()` accessor would be more intuitive, but that seems a bit complicated and potentially confusing given what's already present. Maybe implementing something under `uxarray.plot.mpl.plot()` for a high level Matplotlib plotting interface could work?
> 
> **Edited to add:** I should also say it's really exciting to see this and I think folks will very much appreciate having better support for Matplotlib and the performance improvements.

I agree with the concerns here. Also, the `uxarray.plot.mpl.<vis.func.>`suggestion makes sense.

philipc2:
Thanks for the great feedback @kafitzgerald !

> That said, I don't know that I have a great suggestion for what to do here. I suspect that something chained onto the .plot() accessor would be more intuitive, but that seems a bit complicated and potentially confusing given what's already present. Maybe implementing something under uxarray.plot.mpl.plot() for a high level Matplotlib plotting interface could work?

I like the idea of having it named `.plot()`.

> The first being that it deviates from other high level plotting APIs in the scientific Python ecosystem that use .plot() and .hvplot(). This is definitely beyond the scope of this PR, but adds some potential for confusion and complexity here.

Agreed. It would be nice to have it better integrated under our `.plot()`, but that is something we can include down the line if we chose to refactor our existing plotting API. Keeping it as a standalone function seems like a good middle ground for now.

kafitzgerald:
The more I think about it, I'm less sure about my suggestion of `plot` over `imshow`.

I do still like the idea of laying groundwork to eventually imitate the approach xarray has taken though with `plot` and `plot.wrapped_plotting_function` with some default plot types based upon dimensionality and additional wrapped plotting functions (i.e. some of the ones Brian mentioned).  I think I'm just less confident which plotting function should serve as the default(s) for UXarray and that seems like a conversation for later.  

I suppose `imshow` isn't necessarily something that could be used directly in UXarray and therefore may not be quite as confusing as I was initially thinking, but I could still see wanting to use an `imshow` visualization function w/out Cartopy and/or having more control over the resampling for something called `imshow`.  Maybe a new name (`geoimshow` or one of Brian's suggestions) makes sense for this until a more general `imshow` function is implemented?

Sorry for the back and for on this.

philipc2:
@brianpm @kafitzgerald 

Actually, what about something like the following? Instead of having seperate plotting functions at this stage, what if we instead expose the sampling that we do, which can be passed into whichever function the user desires? 

```Python
import cartopy.crs as ccrs
import uxarray as ux 
from uxarray.plot.matplotlib import sample_grid

fig, ax = plt.subplots(
    subplot_kw={'projection': ccrs.Orthographic(central_longitude=-90, central_latitude=45)},
    figsize=(10, 5),
)

uxda = ux.UxDataArray(...)

# Sample the unstructured grid onto a GeoAxes
sampled_data = sample_grid(uxda, ax=ax, method='nearest')

# Can add other methods eventually 
# sampled_data = sample_grid(uxda, ax=ax, method='bilinear') 

# Show the raster data
ax.imshow(sampled_data, ...)

# Draw Contours on the sampled data 
ax.contour(sampled_data, ...)
ax.contourf(sampled_data, ...)
```

mkavulich:
> Instead of having seperate plotting functions at this stage, what if we instead expose the sampling that we do, which can be passed into whichever function the user desires?

To offer a user perspective, this would be desirable for our command-line plotting applications, where we are creating PNG graphics files for later use rather than trying to visualize them immediately.

kafitzgerald:
> Exposing the sampling as opposed to writing separate functions at this point might be better, especially since we do plan to have a more direct `.plot()` accessor eventually.

I like this for the time being and the demo above looks good to me!

philipc2:
Failing tests due to a Numpy compatibility issue with Numba, see #1292

philipc2:
I've made the requested changes, adding a `UxDataArray.rasterize_to_geoaxes()` method.

I like this design a lot more, and it directly enables more workflows than were possible with the initial implementation.

philipc2:
I have also updated our user guide section on [Plotting with Matplotlib and Cartopy](https://uxarray--1271.org.readthedocs.build/en/1271/user-guide/mpl.html)

docs/api.rst
```diff
@@ -275,6 +277,7 @@ UxDataArray
    UxDataArray.plot.points
    UxDataArray.plot.line
    UxDataArray.plot.scatter
+   UxDataArray.rasterize_to_geoaxes
```
erogluorhan:
```suggestion
   UxDataArray.to_raster
```

docs/api.rst
```diff
@@ -285,6 +288,15 @@ UxDataset
...
    UxDataset.plot
 
+Standalone Functions
+~~~~~~~~~~~~~~~~~~~~
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor_method.rst
+
+   uxarray.plot.imshow
```
erogluorhan: Need to remove this now

test/test_plot.py
```diff
@@ -91,3 +94,19 @@ def test_scatter():
     uxds = ux.open_dataset(gridfile_mpas, gridfile_mpas)
     _plot_line = uxds['bottomDepth'].zonal_average().plot.scatter()
     assert isinstance(_plot_line, hv.Scatter)
+
+
+
+def test_rasterize_to_geoaxes():
```
erogluorhan:
```suggestion
def test_to_raster():
```

uxarray/grid/grid.py
```diff
@@ -2182,149 +2169,62 @@ def to_geodataframe(
...
-            if not override:
-                return self._line_collection_cached_parameters["line_collection"]
+        if self._cached_line_collection:
+            return copy.deepcopy(self._cached_line_collection)
 
         line_collection = _grid_to_matplotlib_linecollection(
             grid=self,
-            periodic_elements=periodic_elements,
-            projection=projection,
+            periodic_elements="ingore",
```
erogluorhan, line 2218:
```suggestion
            periodic_elements="ignore",
```

erogluorhan (commented):
Thanks for putting this together! I like it overall; just see a few comments below please.

docs/user-guide/mpl.ipynb
```diff
@@ -5,379 +5,321 @@
    "id": "2cea9b83a226284a",
```
erogluorhan, line 5: <p>You meant to say "... rasterization is extremely fast" I believe.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1271/discussion/#comment-2161996591'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

erogluorhan, line 5: <p>Typo: geometiries</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1271/discussion/#comment-2161996594'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='9'/>

philipc2: <p>Indeed. Good catch </p>

docs/api.rst
```diff
@@ -223,25 +207,59 @@ Grid Accessor
 
    UxDataset.uxgrid
 
-I/O & Conversion
-~~~~~~~~~~~~~~~~
+
+Conversion Methods
+------------------
+
+UXarray provides functionality to convert to other data formats to enable workflows with existing tools, such as
```
erogluorhan:
```suggestion
UXarray provides functionality to convert its unstructured grids representation to other data structures that can be ingested by existing, widely used tools, such as Matplotlib and Cartopy. This allows users to keep using their workflows with such tools.
```

How about something like this?

docs/api.rst
```diff
@@ -223,25 +207,59 @@ Grid Accessor
...
+Matplotlib and Cartopy.
```
erogluorhan:
```suggestion
```

This removal if the above suggestion is accepted.

docs/api.rst
philipc2: This sounds great.

kafitzgerald (approved):
Looks good to me!

Issue #1280, opened 2025-05-29
encode exodus from healpix


rajeeja:
### Version

latest

### How did you install UXarray?

Source

### What happened?

KeyError: 'node_lon'

### What did you expect to happen?

encoding to throw no error

### Can you provide a MCVE to repoduce the bug?

```python
import uxarray as ux
h = ux.Grid.from_healpix(zoom=3)
exa = h.encode_as("Exodus")
```

erogluorhan:
Double check to see if it should have been fixed by now

Issue #1279, opened 2025-05-29
encode/save to UGRID after healpix creation


rajeeja:
### Version

latest

### How did you install UXarray?

Source

### What happened?

created healpix mesh and tried to write ugrid, resulted in error.

We might need iterate through the attributes of the input dataset and convert any boolean values to integers (0 for `False`, 1 for `True`) before assigning them to the output dataset's attributes. This should resolve the issue when writing to netCDF.

### What did you expect to happen?

file written correctly.
`
TypeError: illegal data type for attribute b'nest', must be one of dict_keys(['S1', 'i1', 'u1', 'i2', 'u2', 'i4', 'u4', 'i8', 'u8', 'f4', 'f8']), got b1`

### Can you provide a MCVE to repoduce the bug?

```python
import uxarray as ux
h = ux.Grid.from_healpix(zoom=3)
uxa = h.encode_as("UGRID")
uxa.to_netcdf("uxa.nc")
```

philipc2:
Fixed in #1282

PR #1282, opened 2025-05-29
Deprecate Grid.encode_as, Add esmf writer, Add tests, fix to_netcdf bug after healpix mesh creation 

### LINKED ISSUE #1242: Exodus writer does not write `num_elem` dimension
https://github.com/UXARRAY/uxarray/blob/ff253ede22aa804c1192f87694b5dde275f90852/uxarray/io/_exodus.py#L109-L293

rajeeja:
Closes #1242 #1279 and #1280

### DIFF
--- a/uxarray/grid/dual.py
+++ b/uxarray/grid/dual.py
@@ -1,5 +1,5 @@
 import numpy as np
-from numba import njit
+from numba import njit, prange
 
 from uxarray.constants import INT_DTYPE, INT_FILL_VALUE
 
@@ -26,7 +26,6 @@ def construct_dual(grid):
     dual_node_z = grid.face_z.values
 
     # Get other information from the grid needed
-    n_node = grid.n_node
     node_x = grid.node_x.values
     node_y = grid.node_y.values
     node_z = grid.node_z.values
@@ -35,10 +34,18 @@ def construct_dual(grid):
     # Get an array with the number of edges for each face
     n_edges_mask = node_face_connectivity != INT_FILL_VALUE
     n_edges = np.sum(n_edges_mask, axis=1)
+    max_edges = node_face_connectivity.shape[1]
+
+    # Only nodes with 3+ edges can form valid dual faces
+    valid_node_indices = np.where(n_edges >= 3)[0]
+
+    construct_node_face_connectivity = np.full(
+        (len(valid_node_indices), max_edges), INT_FILL_VALUE, dtype=INT_DTYPE
+    )
 
     # Construct and return the faces
     new_node_face_connectivity = construct_faces(
-        n_node,
+        valid_node_indices,
         n_edges,
         dual_node_x,
         dual_node_y,
@@ -47,14 +54,16 @@ def construct_dual(grid):
         node_x,
         node_y,
         node_z,
+        construct_node_face_connectivity,
+        max_edges,
     )
 
     return new_node_face_connectivity
 
 
-@njit(cache=True)
+@njit(cache=True, parallel=True)
 def construct_faces(
-    n_node,
+    valid_node_indices,
     n_edges,
     dual_node_x,
     dual_node_y,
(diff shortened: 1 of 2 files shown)


uxarray/grid/grid.py
```diff
@@ -1864,14 +1864,21 @@ def encode_as(self, grid_type: str) -> xr.Dataset:
         warn(
             "Grid.encode_as will be deprecated in a future release. Please use Grid.to_xarray instead."
         )
+        grid_type = grid_type.lower()
+        # Node boundaries/coords are not populated by default. Accessing node_lon will trigger their population.
+        # Do this before encoding to ensure that the node boundaries/coords are populated.
+        if "node_lon" not in self._ds.variables:
+            _ = self.node_lon
```
philipc2: We don't need the if statement here, since that will trigger if we simply call the `node_lon` property.

uxarray/grid/grid.py
```diff
@@ -1864,14 +1864,21 @@ def encode_as(self, grid_type: str) -> xr.Dataset:
         warn(
             "Grid.encode_as will be deprecated in a future release. Please use Grid.to_xarray instead."
         )
+        grid_type = grid_type.lower()
```
philipc2: Is this necessary? I would like to keep them uppercase.

uxarray/grid/grid.py
```diff
@@ -1864,14 +1864,21 @@ def encode_as(self, grid_type: str) -> xr.Dataset:
...
+        # Node boundaries/coords are not populated by default. Accessing node_lon will trigger their population.
+        # Do this before encoding to ensure that the node boundaries/coords are populated.
+        if "node_lon" not in self._ds.variables:
+            _ = self.node_lon
 
-        if grid_type == "UGRID":
+        grid_type = grid_type.lower()
```
philipc2: See above

philipc2 (changes requested):

uxarray/grid/grid.py
rajeeja: yes, give flexibility and doesn't impose too much. I used EXODUS, it specifically wanted Exodus. Rather just check the spelling and not the case. We can stick to all caps in all examples etc. to keep things uniform on our end - enforcing as a requirement via code isn't a good idea - i think.

uxarray/grid/grid.py
rajeeja: Good observations. Fixed.

rajeeja:
@erogluorhan please review.

test/test_healpix.py
```diff
@@ -99,3 +99,31 @@ def test_invalid_cells():
...
+    uxa_ugrid = h.encode_as("UGRID")
+    uxa_exodus = h.encode_as("Exodus")
+
+
+    tmp_filename_ugrid = tmp_path / "healpix_test_ugrid.nc"
+    tmp_filename_exodus = tmp_path / "healpix_test_exodus.exo"
+
+    # Save to netCDF
+    uxa_ugrid.to_netcdf(tmp_filename_ugrid)
+    uxa_exodus.to_netcdf(tmp_filename_exodus)
```
erogluorhan: I know the two issues were about ugrid and Exodus only, but should you have added encode to the encode formats as will into this test case?

uxarray/io/_ugrid.py
```diff
@@ -93,30 +93,42 @@ def _encode_ugrid(ds):
     """Encodes an unstructured grid represented under a ``Grid`` object as a
     ``xr.Dataset`` with an updated grid topology variable."""
 
-    if "grid_topology" in ds:
-        ds = ds.drop_vars(["grid_topology"])
+    # Create a copy of the input dataset to avoid modifying the original Grid._ds
+    out_ds = ds.copy()
```
erogluorhan: Good catch, but no need for use of `.copy()` and duplicate variables (i.e. `out_ds`) if you can keep the original code and replace only:

`ds["grid_topology"] = grid_topology_da` 

on Line 117 with the following: 

`ds = ds.assign(grid_topology=grid_topology_da)`

This way, since there is an assignment back to `ds`, it will not change the original `_ds` (will not behave like pass-by-reference)

uxarray/io/_ugrid.py
```diff
@@ -93,30 +93,42 @@ def _encode_ugrid(ds):
...
 
-    grid_topology = ugrid.BASE_GRID_TOPOLOGY_ATTRS
+    if "grid_topology" in out_ds:
+        out_ds = out_ds.drop_vars(["grid_topology"])
 
-    if "n_edge" in ds.dims:
-        grid_topology["edge_dimension"] = "n_edge"
+    grid_topology_attrs = ugrid.BASE_GRID_TOPOLOGY_ATTRS.copy()  # Use a copy
```
erogluorhan: No need for `copy()` since none of the keys below are present in `ugrid.BASE_GRID_TOPOLOGY_ATTRS`, but nice to have to stay 100% safe in the future (even the the logic here seems like get the ugrid base attributes from dictionary first and add more not modify the base).

erogluorhan (commented):
Thanks very much for this fix! I like it overall; just a few minor suggestions below.

uxarray/grid/grid.py
philipc2: That's fair, as long as we are consistent throughout the entire codebase (include other areas that check for the grid type, I'm okay with either.

uxarray/io/_ugrid.py
rajeeja: Thanks for this, totally makes sense. I have made the change.

test/test_healpix.py
rajeeja: Yes, will add all here. This is where I wanted to check and see if we want to deprecate encode_as and use to_xarray as we mention is our deprecation warning.

erogluorhan: Oh, I see! Sounds good to me! @philipc2 any thoughts re `encode_as`?

philipc2: I'm good with deprecating it in this release. We've had that warning or a while. 

`to_xarray()` is the preferred method.

test/test_weighted_mean.py
```diff
@@ -119,7 +119,7 @@ def test_csne30_equal_area():
     unweighted_mean = uxds['psi'].mean()
 
     # with equal area, both should be equal
-    nt.assert_equal(weighted_mean, unweighted_mean)
+    nt.assert_almost_equal(weighted_mean, unweighted_mean)
```
philipc2, line 122: I'm curious to what changed here?

rajeeja: Not sure what changed here, this has appeared on my local machine for several weeks now.

test/test_healpix.py
rajeeja: handle it here or in another PR?

uxarray/grid/grid.py
```diff
@@ -1830,11 +1830,14 @@ def encode_as(self, grid_type: str) -> xr.Dataset:
         """Encodes the grid as a new `xarray.Dataset` per grid format supplied
         in the `grid_type` argument.
```
philipc2: Let's move all the logic into `to_xarray()` and fully deprecate `encode_as()`

test/test_grid.py
```diff
@@ -96,15 +96,15 @@ def test_grid_with_holes():
...
+    grid_CSne30.to_xarray("UGRID")
+    grid_RLL1deg.to_xarray("UGRID")
+    grid_RLL10deg_CSne4.to_xarray("UGRID")
 
-    grid_CSne30.encode_as("Exodus")
-    grid_RLL1deg.encode_as("Exodus")
-    grid_RLL10deg_CSne4.encode_as("Exodus")
+    grid_CSne30.to_xarray("Exodus")
+    grid_RLL1deg.to_xarray("Exodus")
+    grid_RLL10deg_CSne4.to_xarray("Exodus")
```
philipc2: For all `to_xarray()` tests, we need to also add a step that reads in the saved output and checks to make sure the output is valid.

Right now, these tests only check if the function can be executed, which may not be sufficient.

uxarray/core/dataset.py
```diff
@@ -445,24 +445,38 @@ def to_array(self) -> UxDataArray:
...
-        Parameters
-        ----------
-        grid_format : str, default="UGRID"
-            The format in which to convert the grid. Supported values are "UGRID" and "HEALPix". The dimensions will
-            match the selected grid format.
-
-        Returns
-        -------
-        xr.Dataset
-            The ``ux.UXDataset`` represented as a ``xr.Dataset``
```
philipc2: These changes should not be made here. The purpose to `to_xarray()` for `UxDataset` to provide a method to fall back down to Xarray, which in this case doesn't have anything to do with the Grid information.

philipc2 (changes requested):

philipc2:
Looks like this PR also some changes from #1303

Is that intended?

test/test_healpix.py
erogluorhan: Let's just do it here and mention it in the description

uxarray/grid/grid.py
erogluorhan: I agree; let's fully deprecate `encode_as()`

erogluorhan (commented):
Looks like all of my initial concerns have been addressed, so I will be fine once `encode_as` is deprecated and all @philipc2 's points are addressed.

EDIT: Also once code coverage is fixed

test/test_scrip.py
```diff
@@ -45,10 +45,10 @@ def test_read_ugrid():
 #         assert isinstance(getattr(uxgrid_url, "node_lat"), xr.DataArray)
 #         assert isinstance(getattr(uxgrid_url, "face_node_connectivity"), xr.DataArray)
 
-def test_encode_ugrid():
-    """Read an Exodus dataset and encode that as a UGRID format."""
+def test_to_xarray_ugrid():
```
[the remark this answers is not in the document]
rajeeja: we might want to combine all io tests and make more manageable.

test/test_grid.py
rajeeja: thanks for this comment, found a few bugs in the readers. Fixes pushed.

rajeeja:
> Looks like all of my initial concerns have been addressed, so I will be fine once `encode_as` is deprecated and all @philipc2 's points are addressed.
> 
> EDIT: Also once code coverage is fixed

coverage is 0.5% higher now ;)

erogluorhan (approved):
This now looks good to me; thanks for doing it! Just a single small thing: There is one left reference to "encode_as" in a test script comment; please rephrase it.

philipc2:
@rajeeja 

As discussed in https://github.com/UXARRAY/uxarray/discussions/1324, do you think you could put together a User Guide notebook showcasing this functionality?

rajeeja:
On it

Issue #1220, opened 2025-04-22
xarray features not working on datasets


wwieder:
### Version

v2025.3.0

### How did you install UXarray?

Conda

### What happened?

A few examples using groupby and resample don't seem to be working as I'd expect from experience with xarray. Briefly, I'm trying to create a climatology of monthly means from a decade of monthly CESM history files (`h0`)

**First the win**:  After reading in a dataset with three variables dimensioned _time (120) x n_face_, this works, returning a _month (12) x n_face_ array:
- `ds0['TSA'].groupby("time.month").mean('time')`

The following three examples all fail as a:
- data array using resample `ds0['TSA'].resample(time="1ME").mean()` the time dimension isn't reduced (still _120 x n_faces_; or as a
- dataset using both resample and groupby (e.g., `ds0.groupby("time.month").mean('time')`) throws a bunch of errors.  As what the data array, groupby seems to reduce time dimensions as expected, but resample does not.  The end of the error message is below
```
/glade/work/wwieder/conda-envs/ldf_v0.0/lib/python3.11/site-packages/uxarray/formatting_html.py:24, in <listcomp>(.0)
     22 def _grid_sections(grid, max_items_collapse=15):
     23     cartesian_coordinates = list(
---> 24         [coord for coord in ugrid.CARTESIAN_COORDS if coord in grid._ds]
     25     )
     26     spherical_coordinates = list(
     27         [coord for coord in ugrid.SPHERICAL_COORDS if coord in grid._ds]
     28     )
     29     descriptor = list(
     30         [desc for desc in descriptors.DESCRIPTOR_NAMES if desc in grid._ds]
     31     )

AttributeError: 'NoneType' object has no attribute '_ds'
```

I'm not sure this is in scope for efforts to enable xarray functionality in uxarray, but I thought I'd create an issue here

### What did you expect to happen?

I though resample and groupy by would both produce climatological means of monthly history output that would work for a data array or a dataset.d

### Can you provide a MCVE to repoduce the bug?

```python
import os,sys
from glob import glob
from os.path import join
import numpy as np
import pandas as pd
from netCDF4 import num2date
import xarray as xr
import uxarray as ux 
import matplotlib
import matplotlib.pyplot as plt

path = '/glade/derecho/scratch/hannay/archive/'
suff = '/lnd/hist/'
cases = ['b.e30_alpha06e.B1850C_LTso.ne30_t232_wgx3.156']
[...]

PR #1308, opened 2025-06-24
Add `Grid` preservation to groupby, groupby_bin & resample xarray operations

rajeeja:
UxArray crashes with an AttributeError: 'NoneType' object has no attribute '_ds' when attempting to display the HTML representation of a UxDataset that has lost its grid reference during certain operations, particularly groupby operations.

added null checking in `_obj_repr_with_grid()` and overloaded groupby and resample methods to preserve uxgrid after operations. 

test cases added to test these chages.

cftime, I thought would be handle by xarray, but was causing issues:
We now detecti cftime objects, extract year/month information directly from them and perform manual binning without requiring conversion to pandas datetime


rajeeja:
> The issue in #1220 is due to the `Grid` being lost after certain Xarray operations. The changes here only impact the printing, which is not handling the underlying issue. 

Right, I put what you mentioned in the PR description. The checks here should still be fine as a safety mechanism.

uxarray/formatting_html.py
```diff
@@ -19,6 +19,9 @@ def _grid_header(grid, header_name=None):
 
 
 def _grid_sections(grid, max_items_collapse=15):
+    if grid._ds is None:
+        print(f"Warning: grid._ds is None for {type(grid)}")
+        return []
```
philipc2: Can we throw a warning here using the `warnings` module and provide the user with some extra information? Something like

```Python
import warnings

warnings.warn(
    "Grid repr requested but the grid is None. "
    "This often means it was dropped by an unsupported Xarray operation.",
    UserWarning
)
```

rajeeja: done. How about custom __groupby_ classmethod in `core/dataset.py` which
         handles the `uxgrid` propagation?

rajeeja: > done. How about custom __GroupBy_ classmethod in `core/dataset.py` which handles the `uxgrid` propagation?

rajeeja: @philipc2 I have added this change and now the problem in the issue is fixed.

uxarray/core/dataset.py
```diff
@@ -327,6 +327,63 @@ def from_healpix(
...
+            return cls(
+                result, uxgrid=original_uxgrid, source_datasets=original_source_datasets
+            )
+        elif isinstance(result, xr.DataArray):
+            from uxarray.core.dataarray import UxDataArray
+
+            return UxDataArray(result, uxgrid=original_uxgrid)
+        return result
+
+    def groupby(self, group, squeeze: bool = False, restore_coord_dims: bool = None):
```
philipc2: Please include tests for these.

uxarray/core/dataset.py
```diff
@@ -327,6 +327,63 @@ def from_healpix(
...
+                        result = orig_method(*args, **kwargs)
+                        return self._wrap_groupby_result(
+                            result, original_uxgrid, original_source_datasets
+                        )
+
+                    return wrapped_method
+
+                setattr(
+                    groupby_obj, method_name, create_wrapped_method(original_method)
+                )
```
philipc2: Can you explain the need for this part?

philipc2 (changes requested):

uxarray/core/dataset.py
rajeeja: this is to ensure that when users perform groupby-aggregation operations (such as `.mean()`, `.sum()`, etc.) on a `UxDataset`, the resulting object continues to be a `UxDataset` and retains its `uxgrid` attribute.

philipc2 (commented):
This looks good to me. I'm going to take a deeper look on Monday when I'm back from PTO this week.

philipc2:
Actually, looks like the `groupby` works as expected, it's only the `resample` now.

uxarray/core/dataset.py
```diff
@@ -327,6 +327,61 @@ def from_healpix(
...
+                result, uxgrid=original_uxgrid, source_datasets=original_source_datasets
+            )
+        elif isinstance(result, xr.DataArray):
+            from uxarray.core.dataarray import UxDataArray
+
+            return UxDataArray(result, uxgrid=original_uxgrid)
+        return result
+
+    def groupby(self, group, squeeze: bool = False, restore_coord_dims: bool = None):
+        """Override to return a groupby object that preserves uxgrid on aggregation."""
```
philipc2: Instead of creating a new docstring, let's use the original docstring from the Xarray method. See an example of how we did it here:

https://github.com/UXARRAY/uxarray/blob/61ff40a67b9d36e1a2c045afbfee275f84819150/uxarray/core/dataset.py#L522-L530

rajeeja: Good idea, will do

uxarray/core/dataset.py
```diff
@@ -327,6 +327,349 @@ def from_healpix(
...
+            closed=closed,
+            label=label,
+            base=base,
+            offset=offset,
+            origin=origin,
+            keep_attrs=keep_attrs,
+        )
+
+        # Return the wrapper
+        return UxResampleWrapper(self, indexer, indexer, kwargs)
```
philipc2: I'd prefer to not define a class within the method body.

Is there a way we can achieve the `resample` functionality without needing a custom wrapper, similar to what you did for `groupby`?

philipc2 (changes requested):

uxarray/core/dataset.py
rajeeja: > I'd prefer to not define a class within the method body.
> 
> Is there a way we can achieve the `resample` functionality without needing a custom wrapper, similar to what you did for `groupby`?

I tried that, but resample required a complete reimplementation because xarray's version has limitations, I was getting failures with "Resampling only supported along single dimensions" when used with UxDatasets. Also, resample needed special handling for climate-specific time types -`cftime` objects

philipc2: Could you provide an example of the limitations? I'd like for us to stay as close as possible to Xarray's implementation, a doing a ground-up reimplementation is not very sustainable and leaves us susceptibleto upstream changes on their end. 

> Also, resample needed special handling for climate-specific time types -cftime objects

I beleive Xarray's implementation already supports this, unless that is something that `cf_xarray` does.

rajeeja: I agree with the concerns, tried most of yesterday to not have to do this, but for getting the code pointed in #1220 to work, these changes were needed, will try to think more if I can find another solutions to this. 

```
import xarray as xr
import pandas as pd
import numpy as np

# Create a time index and a second dimension
times = pd.date_range("2000-01-01", periods=4, freq="D")
other_dim = ["a", "b"]

# Create a DataArray with two dimensions: time and other_dim
data = xr.DataArray(
    np.random.rand(4, 2),
    coords={"time": times, "other_dim": other_dim},
    dims=["time", "other_dim"]
)

# Attempt to resample along both dimensions (incorrect usage)
try:
    # This will raise the error
    resampled = data.resample(time="2D", other_dim="1D").mean()
except Exception as e:
    print("Error:", e)

```

Error: Resampling only supported along single dimensions.


--
While Xarray does have support for `cftime` objects, I found that certain operations (especially when combined with our custom indexing or grid structures) didn’t behave as expected. In some cases, the resampling failed or produced incorrect results unless additional handling was added.

rajeeja: done

rajeeja (commented):
now both groupby and resample work

philipc2:
https://docs.xarray.dev/en/stable/generated/xarray.Dataset.groupby_bins.html#xarray.Dataset.groupby_bins

uxarray/core/dataset.py
```diff
@@ -327,6 +327,189 @@ def from_healpix(
...
+        for method_name in [
+            "mean",
+            "sum",
+            "min",
+            "max",
+            "std",
+            "var",
+            "median",
+            "count",
+        ]:
```
philipc2: I wonder if there is any way to do this without harding code the aggregation methods?

rajeeja:
> @rajeeja
> 
> Have you had a chance to confirm if #1220 works correctly now?

Yes, tested on glade. The problem shown in the original issue works.

uxarray/core/dataset.py
rajeeja: I am trying to do a lightweight proxy that forwards attribute access.

philipc2: That would be great!

rajeeja: pushed changes

uxarray/core/dataset.py
```diff
@@ -314,7 +340,7 @@ def from_healpix(
 
         if face_dim not in ds.dims:
             raise ValueError(
-                f"The provided face dimension '{face_dim}' is present in the provided healpix dataset."
+                f"The provided face dimension '{face_dim}' is not present in the provided healpix dataset."
```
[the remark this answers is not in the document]
philipc2: @rajeeja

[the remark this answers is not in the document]
rajeeja: it is set to not present, so we are good, not sure what this is trying to say with 343 and 342 

the code in this branch seems fine: 

```
        if face_dim not in ds.dims:
            raise ValueError(
                f"The provided face dimension '{face_dim}' is not present in the provided healpix dataset."
                f"Please set 'face_dim' to the dimension corresponding to the healpix face dimension."
            )
```

uxarray/core/dataset.py
```diff
@@ -23,6 +23,33 @@
...
+            return wrap_result(result)
+        except Exception as exc:  # Only catch Exception, not BaseException
+            # WARNING: Catching Exception is risky. Only use on_error for known, recoverable errors.
+            if on_error is not None:
+                return on_error(exc, *args, **kwargs)
+            raise
+
+    _wrapped.__doc__ = getattr(attr, "__doc__", _def_proxy_wrap_callable_doc)
+    _wrapped.__name__ = getattr(attr, "__name__", "wrapped")
+    return _wrapped
```
philipc2: Can we move this to `core/utils.py`

uxarray/core/dataset.py
```diff
@@ -327,6 +354,200 @@ def from_healpix(
...
+                return iter(self._gb)
+
+            def __repr__(self):
+                return repr(self._gb)
+
+            def __dir__(self):
+                # Improve IDE/tab-completion discoverability
+                return sorted(set(dir(self.__class__)) | set(dir(self._gb)))
+
+        return _UxGroupByProxy(groupby_obj, original_uxgrid, original_source_datasets)
```
philipc2: I'm not sure if having separate Proxy classes is the best approach here. There is a lot of duplicate code between each of  the three methods you've written.

uxarray/core/dataset.py
```diff
@@ -327,6 +354,200 @@ def from_healpix(
...
+            return cls(
+                result, uxgrid=original_uxgrid, source_datasets=original_source_datasets
+            )
+        elif isinstance(result, xr.DataArray):
+            from uxarray.core.dataarray import UxDataArray
+
+            return UxDataArray(result, uxgrid=original_uxgrid)
+        return result
+
+    def groupby(self, group, squeeze: bool = False, restore_coord_dims: bool = None):
```
philipc2: These parameters do not match what Xarray has.

uxarray/core/dataset.py
rajeeja: I understand the concerns. I will  try to address this in a followup PR `_UxGroupByProxy, _UxResampleProxy, _UxGroupByBinsProxy`

uxarray/core/dataset.py
rajeeja: I can address this.

uxarray/core/dataset.py
rajeeja: Good idea, I want to do a followup for this also.

rajeeja:
> While this does seem to work, I'm skeptical if this is the best solution in terms of maintainability moving forward. Pinging @erogluorhan so he can take a look.

Good observation and I have been thinking about an alternate approach that might be better - How about wrapping the xarray accessor objects (GroupBy/Bin, Resample, `Rolling, Coarsen,Weighted`(found these are in need also:)) instead of using proxies? When methods like groupby(), resample(), or rolling() are called, we return lightweight wrappers that delegate to xarray but preserve uxgrid in results.

uxarray/core/dataset.py
```diff
@@ -327,6 +343,46 @@ def from_healpix(
...
+        ACCESSOR_METHODS = {
+            "groupby": "UxDatasetGroupByAccessor",
+            "groupby_bins": "UxDatasetGroupByAccessor",  # Uses same accessor as groupby
+            "resample": "UxDatasetResampleAccessor",
+            "rolling": "UxDatasetRollingAccessor",
+            "coarsen": "UxDatasetCoarsenAccessor",
+            "weighted": "UxDatasetWeightedAccessor",
+            "rolling_exp": "UxDatasetRollingExpAccessor",
+            "cumulative": "UxDatasetRollingAccessor",  # Uses same accessor as rolling
+        }
```
philipc2: Could we move these into `accessors.py`

uxarray/core/dataarray.py
```diff
@@ -1424,6 +1440,46 @@ def get_dual(self):
...
+        ACCESSOR_METHODS = {
+            "groupby": "UxDataArrayGroupByAccessor",
+            "groupby_bins": "UxDataArrayGroupByAccessor",  # Uses same accessor as groupby
+            "resample": "UxDataArrayResampleAccessor",
+            "rolling": "UxDataArrayRollingAccessor",
+            "coarsen": "UxDataArrayCoarsenAccessor",
+            "weighted": "UxDataArrayWeightedAccessor",
+            "rolling_exp": "UxDataArrayRollingExpAccessor",
+            "cumulative": "UxDataArrayRollingAccessor",  # Uses same accessor as rolling
+        }
```
philipc2: Could we move these into `accessors.py`

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
     assert isinstance(dual, UxDataset)
     assert len(uxds.data_vars) == len(dual.data_vars)
 
+
+def test_groupby_preserves_uxgrid():
```
philipc2: Since all of these test the `DataArray` and `Dataset` implementations, should we considering moving them into their own module?

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+    temp_data = np.random.rand(12)
+
+    # Create a simple xarray Dataset
+    xr_ds = xr.Dataset(
+        {"temperature": ("time", temp_data)},
+        coords={"time": times}
+    )
+
+    # Open the minimal dataset with a real grid
+    try:
```
philipc2: Please don't use `try/except` and and print statements in these tests.

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+    ds_result = uxds_time.resample(time="1W").mean()
+    assert hasattr(ds_result, "uxgrid"), "uxgrid not preserved on Dataset resample"
+    assert ds_result.uxgrid is original_uxgrid, "uxgrid not identical after Dataset resample"
+
+
+def test_resample_reduces_time_dimension():
+    """Test that resample properly reduces the time dimension."""
+    import numpy as np
+    import pandas as pd
+    import pytest
```
philipc2: Imports should be at the top of the module.

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+    except Exception as e:
+        import traceback
+        traceback.print_exc()
+        pytest.fail(f"Error in resample test: {e}")
+
+def test_resample_preserves_uxgrid():
+    """Test that resample preserves the uxgrid attribute."""
+    import numpy as np
+    import pandas as pd
+    import pytest
```
philipc2: Imports

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+    assert ds_result.uxgrid == original_grid
+
+
+
+def test_resample_preserves_uxgrid_and_reduces_time():
+    """Test that resample operations preserve uxgrid and reduce time dimension."""
+    import numpy as np
+    import pandas as pd
+    import pytest
+    import xarray as xr
```
philipc2: Imports

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+    quarterly = uxds.resample(time="Q").mean()
+    assert hasattr(quarterly, "uxgrid"), "uxgrid not preserved with cftime resampling"
+    assert "time" in quarterly.dims, "time dimension missing after cftime resample"
+    assert quarterly.dims["time"] < uxds.dims["time"], "time dimension not reduced with cftime"
+
+
+def test_rolling_preserves_uxgrid():
+    """Test that rolling operations preserve the uxgrid attribute."""
+    import numpy as np
+    import pandas as pd
```
philipc2: Imports

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+
+    # Test that rolling window operations work correctly
+    assert len(da_result.time) == len(uxds.time), "rolling should preserve time dimension length"
+    assert not np.isnan(da_result.values[6:]).any(), "rolling mean should have valid values after window size"
+
+
+def test_coarsen_preserves_uxgrid():
+    """Test that coarsen operations preserve the uxgrid attribute."""
+    import numpy as np
+    import pandas as pd
```
philipc2: Imports

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+
+    # Test that coarsen reduces dimension correctly
+    assert len(da_result.time) == 8, "coarsen by 3 should reduce 24 points to 8"
+    assert ds_result.dims["time"] == 8, "coarsen should reduce time dimension"
+
+
+def test_weighted_preserves_uxgrid():
+    """Test that weighted operations preserve the uxgrid attribute."""
+    import numpy as np
+    import pandas as pd
```
philipc2: Imports

test/test_dataset.py
```diff
@@ -60,6 +60,341 @@ def test_get_dual():
...
+    # Test that weighted operations reduce dimensions correctly
+    assert "time" not in da_result.dims, "weighted mean over time should remove time dimension"
+    assert "n_face" in da_result.dims, "face dimension should be preserved"
+    assert da_result.shape == (n_face,), "result should only have face dimension"
+
+
+def test_cumulative_preserves_uxgrid():
+    """Test that cumulative operations preserve the uxgrid attribute."""
+    import numpy as np
+    import pandas as pd
```
philipc2: Imports

erogluorhan:
> While this does seem to work, I'm skeptical if this is the best solution in terms of maintainability moving forward. Pinging @erogluorhan so he can take a look.

Sorry for the delayed response! Seems like @rajeeja made some changes since this ping. Should I still have a look, @philipc2 ?

philipc2:
> > While this does seem to work, I'm skeptical if this is the best solution in terms of maintainability moving forward. Pinging @erogluorhan so he can take a look.
> 
> Sorry for the delayed response! Seems like @rajeeja made some changes since this ping. Should I still have a look, @philipc2 ?

Yes please! I think two reviews are needed for the scope of this PR.

test/test_accessors.py
```diff
@@ -0,0 +1,329 @@
...
+import pandas as pd
+import xarray as xr
+import uxarray as ux
+from uxarray import UxDataset
+import pytest
+
+try:
+    import constants
+except ImportError:
+    from . import constants
```
philipc2:
```suggestion
```

philipc2 (commented):
This can be removed.

test/test_accessors.py
```diff
@@ -0,0 +1,329 @@
...
+    assert monthly.dims["time"] <= 12, "monthly resampling should give 12 or fewer time points"
+
+
+def test_resample_with_cftime():
+    """Test that resample works with cftime objects."""
+
+    try:
+        import cftime
+    except ImportError:
+        pytest.skip("cftime package not available")
```
philipc2, line 160: Imports should be at the top without and try/except blocks.

uxarray/core/accessors.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def __len__(self):
+        return len(self._xr_obj)
+
+    def __repr__(self):
+        # Subclasses should override this for specific names
+        return f"{self.__class__.__name__}({repr(self._xr_obj)})"
+
+    def __dir__(self):
+        # Combine our methods with xarray's for better IDE support
```
philipc2, line 136: Can you elaborate on the better IDE support?

uxarray/core/accessors.py
```diff
@@ -0,0 +1,285 @@
...
+        "bfill",
+        "fillna",
+        "where",
+        "interpolate_na",
+        "reduce",
+        "map",
+        "apply",
+        "assign",
+        "assign_coords",
+    }
```
philipc2, line 67: This seems beyond the scope of the PR, do we need to do this for every method?

philipc2 (changes requested):
A few comments.

uxarray/core/accessors.py
rajeeja: You're right - this list is actually redundant since xarray automatically preserves UxDataset types. I can 
  remove it to keep the PR focused on core delegation functionality

erogluorhan:
> > > While this does seem to work, I'm skeptical if this is the best solution in terms of maintainability moving forward. Pinging @erogluorhan so he can take a look.
> > 
> > 
> > Sorry for the delayed response! Seems like @rajeeja made some changes since this ping. Should I still have a look, @philipc2 ?
> 
> Yes please! I think two reviews are needed for the scope of this PR.

Sure! Give me some time, maybe until tomorrow, I will give this a review before the weekend.

erogluorhan (commented):
Please find my thoughts about the Grid preservation and cftime handling below:

1. I might really be lost in the code, but shouldn't we just override the xarray functions, wherever there is a loss of the Grid object, and simply add back the `uxgrid` to the data array/set object to be returned? I think that might be a much simpler and adhoc solution, and that way I don't think we should be concerned about `_repr_`. 

I'd imagine this like something as follows (for UxDataset):

```
def resample(self, ...):
  xrds = self.resample(...)
  uxds = uxarray.core.dataset.UxDataset(xrds, uxgrid=self.uxgrid)
  
  return uxds
```

2. Not sure about the cftime handling since even [`xarray` does not work](https://github.com/UXARRAY/uxarray/issues/1220#issuecomment-2821825579) on the dataset pointed out in the #1220 . Didn't look into those aprts of this PR yet either. Nevertheless, if we can get it to work with uxarray easily, maybe it is worth handling here.

rajeeja:
> Please find my thoughts about the Grid preservation and cftime handling below:
> 
> 1. I might really be lost in the code, but shouldn't we just override the xarray functions, wherever there is a loss of the Grid object, and simply add back the `uxgrid` to the data array/set object to be returned? I think that might be a much simpler and adhoc solution, and that way I don't think we should be concerned about `_repr_`.
> 
> I'd imagine this'd look something like (for UxDataset):
> 
> ```
> def resample(self, ...):
>   xrds = self.resample(...)
>   uxds = uxarray.core.dataset.UxDataset(xrds, uxgrid=self.uxgrid)
>   
>   return uxds
> ```

 The accessor pattern is necessary because:

    # Without accessors, this breaks:
`   ds.resample('D').mean()  # resample returns xr.Resample, mean() loses grid`
   
   # With accessors:
```
   ds.resample('D')  # returns UxDatasetResampleAccessor
      .mean()         # accessor wraps mean(), preserves grid
```

  Direct override only handles the first call, not the chained operations. The accessor wraps the entire
  operation chain.

PR #1313, opened 2025-07-07
Gradient Operator 

### LINKED ISSUE #1299: Add Gradients
As mentioned in #1298, our current "gradient" implementation is essentially a directional derivative. 

We should support the “full” 3D gradient for a scalar field defined on an arbitrary horizontally unstructured surface mesh on the sphere.

philipc2:
Closes #1299

## Overview

* Implements a `UxDataArray.gradient()` operator, which computes the zonal and meridional gradients
* Develops a User Guide notebook showcasing the gradient  operator.


erogluorhan:
> @rajeeja @erogluorhan
> 
> This should be good for a review now.

Thanks, will do it today

rajeeja:
> > @rajeeja @erogluorhan
> > This should be good for a review now.
> 
> Thanks, will do it today

+1

uxarray/core/gradient.py
```diff
@@ -33,27 +33,404 @@ def _calculate_edge_node_difference(d_var, edge_nodes):
...
+                        num_node_neighbors += 1
+
+    return bool_bdy, node_neighbors[0:num_node_neighbors]
+
+
+def _compute_gradient(data):
+    from uxarray import UxDataArray
+
+    uxgrid = data.uxgrid
+
```
rajeeja, line 95: Maybe we should check if input data contains NaN values?

rajeeja (approved):
Looks good. Thanks for the good work @egallmeier and @philipc2 
minor comment inline for data inspection before operation

erogluorhan (commented):
This looks great to me; thanks a lot for all the great efforts @egallmeier and @philipc2 !

Only a couple things:

- `calculus.ipynb` looks like a leftover that should probably be removed
- Re new MPAS data and grid files for the gradients test (i.e. "subset" ones): are thopse subsets of existing files we have? Is the subsetting procedure complex that should not be in the testing script? I am trying to understand if we should really add them or generate on the fly.

erogluorhan:
Depending on whether `calculus.ipynb` will be removed from this PR or not, I can approve it now or then

philipc2:
> Depending on whether `calculus.ipynb` will be removed from this PR or not, I can approve it now or then

I will be removing that notebook.

erogluorhan (approved):
Looks great!

uxarray/core/gradient.py
philipc2: The current implementation does handle `nan` values. The gradient is only computed at valid data values. 

## Entire Grid

image


## Half `nan` values 
image

Issue #1329, opened 2025-08-02
IO Test Overhaul


rajeeja:
#1282 PR has new ESMF reader/writer and other test fixes including some roundtrip (loading/reloading and checking reader/writer) they all need to avoid duplication and be optimized.

rajeeja:
[CI / Python (3.13, windows-latest) (pull_request)](https://github.com/UXARRAY/uxarray/actions/runs/16759116143/job/47449464330?pr=1282)Successful in 43m, want to reduce this time.

PR #1331, opened 2025-08-05
Base class for IO test files

rajeeja:
#1329 better performance and more maintainable, improved coverage.


rajeeja:
> This looks great. I like the idea of centralized tests for all grid formats.
> 
> Though, I still would like to have separate modules (i.e. `test_mpas`) instead of putting all in one module.
> 
> Maybe something like the following structure?
> 
> 1. Create a base set of tests we can run on all grid formats (similar to what you've done), which can be imported and used within each of the individual modules. This would still allow us to add extra tests via inheritance if needed.

I have made this change, but I like the one-module design is better. It is typically the case that one reader-writer breaks the others, and we don't test it. If needed, we can always run a part of the tests from that one module.

rajeeja:
adding only test_io_common.py for shared IO functionality while keeping all format-specific tests unchanged. We should create a follow-up issue to reorganize the entire test suite into a hierarchical structure (mirroring uxarray's module organization), breaking down large files like test_geometry.py (60KB) into smaller, more manageable components.

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+                grid = ux.open_grid(grid_path)
+
+        # Check UGRID compliance
+        # 1. Connectivity should use proper fill values
+        assert grid.face_node_connectivity._FillValue == INT_FILL_VALUE
+
+        # 2. Coordinates should be in valid ranges
+        if hasattr(grid.node_lon, 'units') and 'degree' in str(grid.node_lon.units):
+            assert grid.node_lon.min() >= -180
+            assert grid.node_lon.max() <= 360
```
philipc2:
```suggestion
            assert grid.node_lon.max() <= 180
```

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+            assert grid.node_lat.min() >= -90
+            assert grid.node_lat.max() <= 90
+
+        # 3. Check that grid has been properly standardized by uxarray
+        # (Not all input files have Conventions attribute, but uxarray should handle them)
```
philipc2: Is this unfinished?

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+    def test_grid_properties_consistency(self):
+        """Test that all grids have consistent basic properties after loading."""
+        grids = []
+
+        # Load a few different format grids
+        test_files = [
+            ("ugrid", "ugrid/quad-hexagon", "grid.nc"),
+            ("mpas", "mpas/QU/480", "grid.nc"),
+            ("exodus", "exodus/outCSne8", "outCSne8.g"),
+        ]
```
philipc2: Is there a reason for using this instead of `IO_READ_TEST_FORMATS`

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+
+    @pytest.mark.parametrize("write_format", WRITABLE_FORMATS)
+    def test_write_invalid_path(self, write_format):
+        """Test error handling for invalid write paths."""
+        # Create a simple test grid
+        grid = ux.open_grid(current_path / "meshfiles" / "ugrid" / "quad-hexagon" / "grid.nc")
+
+        # Try to write to invalid path
+        with pytest.raises((OSError, IOError, PermissionError)):
+            grid._ds.to_netcdf("/invalid/path/file.nc")
```
philipc2: I don't think we need this test.

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def test_read_nonexistent_file(self):
+        """Test error handling for non-existent files."""
+        with pytest.raises((FileNotFoundError, OSError)):
+            ux.open_grid("nonexistent_file.nc")
```
philipc2: I don't think we need this test.

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+
+class TestIODatasetCommon:
+    """Common tests for dataset (grid + data) operations."""
+
+    def test_dataset_basic_operations(self):
+        """Test basic dataset operations across formats."""
+        # Test that we can open datasets with different grid formats
+        test_cases = [
+            ("ugrid", "ugrid/quad-hexagon", "grid.nc", "data.nc"),
+        ]
```
philipc2: Should we use `IO_READ_TEST_FORMATS` instead of just this one?

test/test_io_common.py
```diff
@@ -0,0 +1,212 @@
...
+    ("ugrid", "ugrid/outCSne30", "outCSne30.ug"),
+    ("ugrid", "ugrid/outRLL1deg", "outRLL1deg.ug"),
+    ("mpas", "mpas/QU/480", "grid.nc"),
+    ("esmf", "esmf/ne30", "ne30pg3.grid.nc"),
+    ("exodus", "exodus/outCSne8", "outCSne8.g"),
+    ("exodus", "exodus/mixed", "mixed.exo"),
+    ("scrip", "scrip/outCSne8", "outCSne8.nc"),
+    ("icon", "icon/R02B04", "icon_grid_0010_R02B04_G.nc"),
+    ("fesom", "fesom/pi", None),  # Special case - multiple files
+]
```
philipc2: If we plan to only cover grid functionality here, this is fine, otherwise we need to math the corresponding data file too for more comprehensive dataset tests.

test/test_io_common.py
rajeeja: fixed.

test/test_io_common.py
rajeeja: good catch. fixed.

test/test_io_common.py
rajeeja: grid only is what I planned.

test/test_io_common.py
```diff
@@ -0,0 +1,203 @@
...
+            if not grid_path.exists() or not data_path.exists():
+                continue
+
+            # Open dataset
+            dataset = ux.open_dataset(grid_path, data_path)
+
+            # Basic checks
+            assert dataset is not None
+            assert hasattr(dataset, 'uxgrid')
+            assert len(dataset.data_vars) > 0
```
philipc2: If we plan to only cover grid functionality for now, let's hold off on this component.

test/test_io_common.py
```diff
@@ -0,0 +1,203 @@
...
+    ("esmf", "esmf/ne30", "ne30pg3.grid.nc"),
+    ("exodus", "exodus/outCSne8", "outCSne8.g"),
+    ("exodus", "exodus/mixed", "mixed.exo"),
+    ("scrip", "scrip/outCSne8", "outCSne8.nc"),
+    ("icon", "icon/R02B04", "icon_grid_0010_R02B04_G.nc"),
+    ("fesom", "fesom/pi", None),  # Special case - multiple files
+]
+
+# Formats that support writing
+WRITABLE_FORMATS = ["ugrid", "exodus"]
```
philipc2: Should we add Scrip and ESMF now after your recent additions?

philipc2: Though, we don't use this right now so I'm fine with either removing it or keeping it as a placeholder.

test/test_io_common.py
```diff
@@ -0,0 +1,203 @@
...
+        assert hasattr(grid, 'node_lat')
+
+        # Check required dimensions
+        assert 'n_node' in grid._ds.dims
+        assert 'n_face' in grid._ds.dims
+
+        # Validate grid
+        grid.validate()
+
+    @pytest.mark.parametrize("format_name,subpath,filename", IO_READ_TEST_FORMATS)
```
philipc2: Please use this consistently across all tests. There are a couple below that don't use this.

test/test_io_common.py
```diff
@@ -0,0 +1,203 @@
...
+            pytest.skip(f"No test file found for {format_name}")
+
+        # Open grid
+        if format_name == "mpas":
+            grid = ux.open_grid(test_file, use_dual=False)
+        else:
+            grid = ux.open_grid(test_file)
+
+        # Check that data arrays are lazy (dask arrays) where applicable
+        # This is format-dependent, so we just check it doesn't error
```
philipc2:
```suggestion
```

test/test_io_common.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            grid_path = current_path / "meshfiles" / subpath / filename
+            if not grid_path.exists():
+                pytest.skip(f"Test file not found: {grid_path}")
+
+            # Handle special cases
+            if format_name == "mpas":
+                grid = ux.open_grid(grid_path, use_dual=False)
+            else:
+                grid = ux.open_grid(grid_path)
```
philipc2: Could you look into creating a [`pytest.fixture` ](https://docs.pytest.org/en/6.2.x/fixture.html) which could handle the entire grid-loading step and be used to significantly reduce the amount of code needed to write many of these tests?

I've used these before for some of the tests, and they are very helpful. 

https://github.com/UXARRAY/uxarray/blob/c54d01fe9b63b2dbd20e8b841edca82ee2657f6a/test/test_point_in_face.py#L15-L22

What do you think?

philipc2: @rajeeja 

A single fixture that would return a grid across all formats defined in `IO_READ_TEST_FORMATS` could be very elegant.

test/test_io_common.py
```diff
@@ -0,0 +1,203 @@
...
+
+# Format conversion test pairs - removed for now as format conversion
+# requires more sophisticated handling than simple to_netcdf
+
+
+class TestIOCommon:
+    """Common IO tests across all formats."""
+
+    @pytest.mark.parametrize("format_name,subpath,filename", IO_READ_TEST_FORMATS)
+    def test_basic_read(self, format_name, subpath, filename):
```
philipc2: The fixture would act as a "basic read" test, since if any of the fixtures fail, then there is an issue in simply reading in a grid.

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+    ("ugrid", "ugrid/outCSne30", "outCSne30.ug"),
+    ("ugrid", "ugrid/outRLL1deg", "outRLL1deg.ug"),
+    ("mpas", "mpas/QU/480", "grid.nc"),
+    ("esmf", "esmf/ne30", "ne30pg3.grid.nc"),
+    ("exodus", "exodus/outCSne8", "outCSne8.g"),
+    ("exodus", "exodus/mixed", "mixed.exo"),
+    ("scrip", "scrip/outCSne8", "outCSne8.nc"),
+    ("icon", "icon/R02B04", "icon_grid_0010_R02B04_G.nc"),
+    ("fesom", "fesom/pi", None),  # Special case - multiple files
+]
```
philipc2: Could we include HEALPix in here?

There's no grid file associated with it, but within our fixture below we can construct a basic one by doing:

```Python
grid = ux.Grid.from_healpix(zoom=1, pixels_only=False)
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+        if hasattr(grid.node_lon, 'units') and 'degree' in str(grid.node_lon.units):
+            assert grid.node_lon.min() >= -180
+            assert grid.node_lon.max() <= 180
+            assert grid.node_lat.min() >= -90
+            assert grid.node_lat.max() <= 90
+
+        # 3. Check that grid has been properly standardized by uxarray
+        # (Not all input files have Conventions attribute, but uxarray should handle them)
+
+    def test_grid_properties_consistency(self):
```
philipc2: Please update this one to use the fixture.

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+        grid = grid_from_format
+
+        # Basic validation
+        assert grid is not None
+        assert hasattr(grid, 'face_node_connectivity')
+        assert hasattr(grid, 'node_lon')
+        assert hasattr(grid, 'node_lat')
+
+        # Check required dimensions
+        assert 'n_node' in grid._ds.dims
```
philipc2:
```suggestion
        assert 'n_node' in grid.dims
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+        assert 'n_face' in grid._ds.dims
```
philipc2:
```suggestion
        assert 'n_face' in grid.dims
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def test_basic_read(self, grid_from_format):
+        """Test that all formats can be read successfully."""
+        grid = grid_from_format
+
+        # Basic validation
+        assert grid is not None
+        assert hasattr(grid, 'face_node_connectivity')
+        assert hasattr(grid, 'node_lon')
+        assert hasattr(grid, 'node_lat')
```
philipc2: Since we've defined our coordinates, connectivity, and other descriptor variables as properties, a Grid will always have these attached as attributes, meaning this check is always True.

You could do:

```Python
assert `face_node_connectivity` in grid.connectivity
assert 'node_lon' in grid.coordinates
assert 'node_lat' in grid.coordinates

File: test/test_io_common.py [hunk 19]
```diff
@@ -0,0 +1,151 @@
...
+                grid = ux.open_grid(grid_path, use_dual=False)
+            else:
+                grid = ux.open_grid(grid_path)
+
+            # Check that all grids have the essential properties
+            assert hasattr(grid, 'n_node')
+            assert hasattr(grid, 'n_face')
+            assert hasattr(grid, 'face_node_connectivity')
+            assert hasattr(grid, 'node_lon')
+            assert hasattr(grid, 'node_lat')
```
philipc2: Same as my comment above.

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+            assert grid.node_lon.dtype in [np.float32, np.float64]
+            assert grid.node_lat.dtype in [np.float32, np.float64]
+
+
+
+    def test_lazy_loading(self, grid_from_format):
+        """Test that grids support lazy loading where applicable."""
+        grid = grid_from_format
+
+        assert grid._ds is not None
```
philipc2:
```suggestion
    def test_lazy_loading(self, grid_from_format):
        """Test that grids support lazy loading where applicable."""
        grid = grid_from_format

        assert grid._ds is not None
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+from pathlib import Path
+import tempfile
+import os
+from numpy.testing import assert_array_equal, assert_allclose
+from uxarray.constants import ERROR_TOLERANCE, INT_DTYPE, INT_FILL_VALUE
+
+try:
+    import constants
+except ImportError:
+    from . import constants
```
philipc2:
```suggestion
try:
    import constants
except ImportError:
    from . import constants
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+
+            # Check that all grids have the essential properties
+            assert hasattr(grid, 'n_node')
+            assert hasattr(grid, 'n_face')
+            assert hasattr(grid, 'face_node_connectivity')
+            assert hasattr(grid, 'node_lon')
+            assert hasattr(grid, 'node_lat')
+
+            # Check data types are consistent
+            assert grid.face_node_connectivity.dtype in [np.int32, np.int64, INT_DTYPE]
```
philipc2:
```Python
np.issubdtype(grid.face_node_connectivity.dtype, np.integer)
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+            assert grid.node_lon.dtype in [np.float32, np.float64]
```
philipc2:
```Python
np.issubdtype(grid.node_lon.dtype, np.floating)

File: test/test_io_common.py [hunk 24]
```diff
@@ -0,0 +1,151 @@
...
+        assert hasattr(grid, 'node_lat')
+
+        # Check required dimensions
+        assert 'n_node' in grid._ds.dims
+        assert 'n_face' in grid._ds.dims
+
+        # Validate grid
+        grid.validate()
+
+    def test_ugrid_compliance_after_read(self, grid_from_format):
```
philipc2:
```suggestion
    def test_ugrid_compliance(self, grid_from_format):
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+class TestIOCommon:
+    """Common IO tests across all formats."""
+
+    def test_basic_read(self, grid_from_format):
+        """Test that all formats can be read successfully."""
+        grid = grid_from_format
+
+        # Basic validation
+        assert grid is not None
+        assert hasattr(grid, 'face_node_connectivity')
```
philipc2: Everything below here should be in `test_ugrid_compliance`

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TestIOCommon:
+    """Common IO tests across all formats."""
+
+    def test_basic_read(self, grid_from_format):
+        """Test that all formats can be read successfully."""
+        grid = grid_from_format
+
+        # Basic validation
+        assert grid is not None
```
philipc2:
```suggestion
        assert isinstance(grid, ux.Grid)
```

test/test_io_common.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Add format info to the grid for test identification
+    grid._test_format = format_name
+    return grid
+
+
+class TestIOCommon:
+    """Common IO tests across all formats."""
+
+    def test_basic_read(self, grid_from_format):
```
philipc2:
```suggestion
    def test_return_type(self, grid_from_format):
```

test/test_io_common.py
```diff
@@ -0,0 +1,140 @@
...
+        # 3. Check that grid has been properly standardized by uxarray
+        # (Not all input files have Conventions attribute, but uxarray should handle them)
+
+    def test_grid_properties_consistency(self, grid_from_format):
+        """Test that all grids have consistent basic properties after loading."""
+        grid = grid_from_format
+
+        # Check that all grids have the essential properties
+        assert hasattr(grid, 'n_node')
+        assert hasattr(grid, 'n_face')
```
philipc2: Check `grid.dims`

test/test_io_common.py
```diff
@@ -0,0 +1,140 @@
...
+        assert 'face_node_connectivity' in grid.connectivity
+        assert 'node_lon' in grid.coordinates
+        assert 'node_lat' in grid.coordinates
+
+        # Check required dimensions
+        assert 'n_node' in grid.dims
+        assert 'n_face' in grid.dims
+
+        # Validate grid
+        grid.validate()
```
philipc2: Can we move these into `test_ugrid_compliance`

test/test_io_common.py
philipc2: @rajeeja 

Can we remove this? I don't think is doing anything right now.

test/test_io_common.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Validate grid structure
+        grid.validate()
+
+        # Check UGRID compliance
+        # 1. Connectivity should use proper fill values
+        assert grid.face_node_connectivity._FillValue == INT_FILL_VALUE
+
+        # 2. Coordinates should be in valid ranges
+        if hasattr(grid.node_lon, 'units') and 'degree' in str(grid.node_lon.units):
```
philipc2: I don't think we need this check here.

test/test_io_common.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Check that all grids have the essential properties
+        assert 'n_node' in grid.dims
+        assert 'n_face' in grid.dims
+        assert 'face_node_connectivity' in grid.connectivity
+        assert 'node_lon' in grid.coordinates
+        assert 'node_lat' in grid.coordinates
+
+        # Check data types are consistent
+        assert grid.face_node_connectivity.dtype in [np.int32, np.int64, INT_DTYPE]
```
philipc2: Use `np.issubdtype` with `np.integer` here

test/test_io_common.py
philipc2: @rajeeja 

Can we remove this test?

test/test_io_common.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def test_return_type(self, grid_from_format):
+        """Test that all formats can be read successfully."""
+        grid = grid_from_format
+
+        # Basic validation
+        assert isinstance(grid, ux.Grid)
+
+    def test_ugrid_compliance(self, grid_from_format):
+        """Test that grids from all formats meet basic UGRID standards."""
```
philipc2: With this new module, lets try and keep the docstrings for each test as detailed as possible.

Can you elaborate on what we are testing and the motivation for it in each of them?

rajeeja: Sure, I can add more to the docstrings. This checks for required UGRID topology/coordinates, core dimensions, correct fill values in connectivity, reasonable degree-based coordinate ranges,and that grid.validate() passes.

test/test_io_common.py
```diff
@@ -0,0 +1,150 @@
...
+        assert np.issubdtype(grid.node_lon.dtype, np.floating)
+        assert np.issubdtype(grid.node_lat.dtype, np.floating)
+
+    def test_lazy_loading(self, grid_from_format):
+        """Confirm a backing xarray Dataset exists (grid._ds). Keeps IO and
+        compute lazy where supported.
+        """
+        grid = grid_from_format
+
+        assert grid._ds is not None
```
philipc2:
```suggestion
```

test/test_io_common.py
```diff
@@ -0,0 +1,132 @@
...
+        assert 'face_node_connectivity' in grid.connectivity
+        assert 'node_lon' in grid.coordinates
+        assert 'node_lat' in grid.coordinates
+
+        # Required dimensions
+        assert 'n_node' in grid.dims
+        assert 'n_face' in grid.dims
+
+        # Validate grid structure
+        grid.validate()
```
philipc2: Does validate throw Exceptions or Warnings in it's current state?

philipc2: If it throws warnings, we might want to check for that.

rajeeja: Grid.validate() returns True on success; if any check fails it raises RuntimeError("Mesh validation failed."). I'll add another guard for warning about this.

Issue #1332, opened 2025-08-06
open_dataset incorrectly renames dimensions of size 2, causing duplicate dimension names


ahijevyc:
### Version

2025.06.0

### How did you install UXarray?

Conda

### What happened?

When opening a UGRID-compliant NetCDF file using uxarray.open_dataset, any dimension with a size of 2 (e.g., lon_lat = 2, min_max = 2) is automatically and silently renamed to the generic dimension name 'two'.

If a variable in the file uses more than one of these dimensions, it results in an xarray.DataArray with duplicate dimension names (e.g., dims=('n_face', 'two', 'two')). This triggers an xarray UserWarning and makes the resulting UxDataArray or UxDataset difficult to use, as subsequent attempts to fix the dimensions (as the warning suggests) are blocked by xarray's safety features (AttributeError: you cannot assign dims... and TypeError: 'Grid' object does not support item assignment).

For example, first create a 1-degree SCRIP convention lat lon grid with NCO tool ncremap on derecho:
```csh
module load nco
ncremap -g grd.nc -G latlon=180,360#lon_typ=grn_wst
```

See remaining code below.

### What did you expect to happen?

No warnings like this:
/glade/work/ahijevyc/conda-envs/uxarray/lib/python3.12/site-packages/xarray/namedarray/core.py:503: UserWarning: Duplicate dimension names present: dimensions {'two'} appear more than once in dims=('n_face', 'two', 'two'). We do not yet support duplicate dimension names, but we do allow initial construction of the object. We recommend you rename the dims immediately to become distinct, as most xarray functionality is likely to fail silently if you do not. To rename the dimensions you will need to set the ``.dims`` attribute of each variable, ``e.g. var.dims=('x0', 'x1')``.

### Can you provide a MCVE to repoduce the bug?

```python
import os
import uxarray

data_path = "/glade/campaign/cisl/vast/uxarray/data/e3sm_keeling/ENSO_ctl_1std/unstructured/20231220.F2010.ENSO_ctl.lagreg.ne30pg2_EC30to60E2r2.keeling.eam.h0.0006-12.nc"
grid_file = (
    "/glade/campaign/cisl/vast/uxarray/data/e3sm_keeling/E3SM_grid/ne30pg2_grd.nc"
)
uxda = uxarray.open_dataset(grid_file, data_path).TS.squeeze(dim="time")

lon_bounds = (-74, -64)
lat_bounds = (18, 24)
[...]

PR #1334, opened 2025-08-06
dimension name should be lat_lon because lat comes before lon

ahijevyc:
This came up when dealing with #1332.

The dimension `lon_lat` should be named `lat_lon` because latitude is in the zero-index and longitude is in the one-index.
This is analogous to how the `min_max` dimension contains the minimum value in the zero-index and maximum value in the one-index. You wouldn't want to have a dimension name like `max_min` because min comes first.

## Overview
rename dimension `lon_lat` -> `lat_lon` because the latitude index comes before longitude.

After tracing the way `bounds_array` is constructed in `_populate_face_bounds` it is likely that latitude comes first and longitude comes second. That's why it should be renamed.

All the other variables with lat and lon in their names seems to be consistent with the order of lat and lon in their data.

### DIFF
--- a/uxarray/grid/bounds.py
+++ b/uxarray/grid/bounds.py
@@ -97,7 +97,7 @@ def _populate_face_bounds(
 
     bounds_da = xr.DataArray(
         bounds_array,
-        dims=["n_face", "lon_lat", "min_max"],
+        dims=["n_face", "lat_lon", "min_max"],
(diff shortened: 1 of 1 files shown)


rajeeja (approved):

philipc2 (approved):
Nice catch!

PR #1336, opened 2025-08-11
Update CI workflow to improve execution time 

philipc2:
# Overview
- Removed `macos-14` workflow, which is no longer needed as `macos-latest` now defaults to ARM architecture 
- Removed CodeCov from this workflow, which was taking a significant amount of time and should not run each time.


rajeeja:
This is a good PR. Was worried about the same last week

PR #1341, opened 2025-08-13
R Tree 

rajeeja:
- ASV QueryTreeStructures (branch-restricted):
       - time_ball_tree: ~462 μs
       - time_kd_tree: ~737 μs
       - time_r_tree: ~1.48 ms
       
It might be that for single-point query, KD/BallTree stay faster. R-tree (spatialpandas’ HilbertRtree) should be better for broad-phase overlap pruning and many box queries, especially on large grids 

EDIT: After replacing spatialpandas.spatialindex.HilbertRtree with rtree.index.Index

   1 R-tree query (avg):    0.016ms  ⚡ FASTEST
   2 KD-tree query (avg):   0.060ms
   3 Ball-tree query (avg): 0.073ms


uxarray/grid/grid.py
```diff
@@ -1890,6 +1896,17 @@ def get_spatial_hash(
...
 
+    def get_r_tree(self, reconstruct: bool = False):
+        """Get an R-tree over per-face 3D AABBs for fast face-box intersection queries.
+
+        Returns a lightweight adapter exposing intersects(box), where box is either
+        (xmin, ymin, zmin, xmax, ymax, zmax) for 3D or (xmin, ymin, xmax, ymax) for XY fallback.
+        """
+        if reconstruct or not hasattr(self, "_rtree") or self._rtree is None:
+            rtree, boxes, dim = _rtree_build(self.bounds)
+            self._rtree = _RtreeAdapter(rtree, boxes, dim)
```
philipc2: If possible, we should use the original `spatialpandas.HilbertRtree` here and not return a wrapped class since we would not be able to use this tree in Numba functions.

The spatialpandas implementation is written using Numba.

rajeeja: sounds good, I was debating if that would be any useful to do adapter.., removed it now.

philipc2: No worries!

uxarray/grid/grid.py
```diff
@@ -1890,6 +1893,21 @@ def get_spatial_hash(
 
         return self._spatialhash
 
+    def get_r_tree(self, reconstruct: bool = False):
+        """Build or retrieve a spatialpandas.spatialindex.HilbertRtree over per-face AABBs.
```
philipc2: Instead of using "per-face AABBs" in the docstring, can you mention that it is built on the bounding boxes?

uxarray/grid/grid.py
```diff
@@ -1890,6 +1893,21 @@ def get_spatial_hash(
...
+
+        The returned object is the original Numba-backed HilbertRtree instance so it can be
+        used directly in Numba via its `.numba_rtree` attribute. Boxes are built in 3D
+        (xmin, ymin, zmin, xmax, ymax, zmax) when supported.
+        """
+        if reconstruct or not hasattr(self, "_rtree") or self._rtree is None:
+            rtree, boxes, dim = _rtree_build(self.bounds)
+            self._rtree = rtree
+            # cache for advanced users; may be useful for debugging or custom queries
```
philipc2, line 1933:
```suggestion
            # cache for advanced users; may be useful for debugging or custom queries
```

uxarray/grid/r_tree.py
```diff
@@ -0,0 +1,150 @@
+import numpy as np
+from numba import njit, prange
+from spatialpandas.spatialindex import HilbertRtree
+
+
+@njit(cache=True)
+def _face_aabb_xyz_kernel(
```
philipc2, line 14: Docstring

uxarray/grid/r_tree.py
```diff
@@ -0,0 +1,150 @@
...
+            if z < zmin:
+                zmin = z
+            if z > zmax:
+                zmax = z
+
+    return (xmin - eps, ymin - eps, zmin - eps, xmax + eps, ymax + eps, zmax + eps)
+
+
+@njit(cache=True, parallel=True)
+def face_aabb_xyz(
```
philipc2, line 94: Docstring

uxarray/grid/r_tree.py
```diff
@@ -0,0 +1,150 @@
...
+    for i in prange(n):
+        lat0 = lat_bounds[i, 0]
+        lat1 = lat_bounds[i, 1]
+        lon0 = lon_bounds[i, 0]
+        lon1 = lon_bounds[i, 1]
+        boxes[i, :] = _face_aabb_xyz_kernel(lat0, lat1, lon0, lon1, eps)
+    return boxes
+
+
+def construct_face_rtree_from_bounds(bounds_da, p: int = 10, page_size: int = 512):
```
philipc2: Docstring

uxarray/grid/r_tree.py
```diff
@@ -0,0 +1,150 @@
...
+        rtree = HilbertRtree(boxes, p, page_size)
+        dim = 3
+    except Exception:
+        xy_boxes = np.column_stack([boxes[:, 0], boxes[:, 1], boxes[:, 3], boxes[:, 4]])
+        rtree = HilbertRtree(xy_boxes, p, page_size)
+        dim = 2
+    return rtree, boxes, dim
+
+
+def aabb_overlap3(b1: np.ndarray, b2: np.ndarray) -> bool:
```
philipc2, line 185: Docstring

uxarray/grid/r_tree.py
```diff
@@ -0,0 +1,150 @@
...
+        (b1[3] < b2[0])
+        or (b2[3] < b1[0])
+        or (b1[4] < b2[1])
+        or (b2[4] < b1[1])
+        or (b1[5] < b2[2])
+        or (b2[5] < b1[2])
+    )
+
+
+def faces_aabb_overlap_from_bounds(bounds_da, i: int, j: int) -> bool:
```
philipc2, line 208: Docstring

uxarray/grid/r_tree.py
```diff
@@ -0,0 +1,150 @@
...
+    arr = bounds_da.values
+    lat_bounds = arr[[i, j], 0, :]
+    lon_bounds = arr[[i, j], 1, :]
+    boxes = face_aabb_xyz(lat_bounds, lon_bounds)
+    return aabb_overlap3(boxes[0], boxes[1])
+
+
+def find_intersecting_face_pairs(bounds_da):
```
philipc2, line 230: Docstring

philipc2 (commented):
I'm not sure if the implementation is working correctly. I'm looking to query each face center against the r-tree, which should return a single element, since the coordinate lies in the center of the bounding box

```Python
x = uxgrid.face_x.values[0]
y = uxgrid.face_y.values[0]
z = uxgrid.face_z.values[0]

rt.intersects((x, y, z, x, y, z))
```

But right now, it is returning four elements.

uxarray/grid/r_tree.py
philipc2: Perhaps the issue is here. This is where I left off because I was having similar issues. Could you look into it? Feel free to re-write anything.

philipc2:
For most cases, I would assume the following should hold true.
- Doing a Face Center $(x, y, z)$ query should return a single face, since the bounding boxes should not overlap
- Doing a Corner Node $(x, y, z)$ should return the faces that share that node

rajeeja:
> For most cases, I would assume the following should hold true.
> 
> * Doing a Face Center 
>     (
>     x
>     ,
>     y
>     ,
>     z
>     )
>    query should return a single face, since the bounding boxes should not overlap
> * Doing a Corner Node 
>     (
>     x
>     ,
>     y
>     ,
>     z
>     )
>    query should return the faces that share that node



> For most cases, I would assume the following should hold true.
> 
> * Doing a Face Center 
>     (
>     x
>     ,
>     y
>     ,
>     z
>     )
>    query should return a single face, since the bounding boxes should not overlap
> * Doing a Corner Node 
>     (
>     x
>     ,
>     y
>     ,
>     z
>     )
>    query should return the faces that share that node

I think R-tree is working correctly as a spatial index - it's meant to quickly filter candidates for exact geometric tests. To get the behavior you describe, we'd need to performs exact point-in-polygon on the R-tree results. Didn't we do this @aaronzedwick ?

philipc2:
The R-Tree is built on the bounding boxes, so we should still be able to do a Point in Box check (not to be confused with Point in Polygon which uses our K-D Tree to determine candidates and then run a most robust algorithm like @aaronzedwick mentioned)

The K-D Trees are built on coordinates, so they simply return the nearest neighbor. While doing Point checks isn't the primary use of the R-Tree, it's still concerning that the results are not consistent with what we would expect.

philipc2:
Another option we could consider is to build a 2D R-Tree, with elements along the antimeridian split. However, If we can stick to Cartesian that would be preferred.

philipc2:
> > We might want to switch over to the `[rtree](https://pypi.org/project/rtree/)` package.
> 
> Good catch, I like this option. It should be easy to migrate SpatialPandas to rtree, should I start doing that?

I think so, `rtree` is very fast too.

PR #1344, opened 2025-08-14
Fix MPAS reader bug: Normalize after loading

rajeeja:


rajeeja:
All other formats normalize.

philipc2 (approved):
Should the face area be updated too? 

I think so. Good catch!

uxarray/io/_mpas.py
```diff
@@ -190,6 +190,13 @@ def _parse_node_xyz_coords(in_ds, out_ds, mesh_type):
         node_y = node_y.rename({"nCells": ugrid.NODE_DIM})
         node_z = node_z.rename({"nCells": ugrid.NODE_DIM})
 
+    # Normalize coordinates to unit sphere if needed
+    if "sphere_radius" in in_ds.attrs and in_ds.attrs["sphere_radius"] != 1.0:
+        radius = in_ds.attrs["sphere_radius"]
```
philipc2: I'd add a single line before the normalization calls like this

```python
radius = in_ds.attrs.get("sphere_radius", 1.0)
```

And then to normalize, the if statement can simply before

```Python
if sphere_radius != 1.0:
   ... 
```

I'd then apply this to the rest of the checks.

test/test_mpas.py
```diff
@@ -133,3 +133,100 @@ def test_distance_units():
...
+    # For a unit sphere, face areas should be much smaller than total area
+    max_face_area = np.max(face_areas)
+    min_face_area = np.min(face_areas)
+
+    assert max_face_area < 0.1 * expected_total_area, "Maximum face area seems too large for unit sphere"
+    assert min_face_area > 1e-10, "Minimum face area seems too small"
+
+    # Check edge lengths are normalized for unit sphere
+    # Edge lengths should be reasonable for a unit sphere (max possible is pi for antipodal points)
+    if "edge_node_distances" in uxgrid._ds:
```
philipc2, line 222: Should we normalize `edge_node_distances` and `edge_face_distances` too in our MPAS reader too?

rajeeja: it is already done` _parse_edge_node/face_distances()` divides dvEdge by sphere_radius

PR #1345, opened 2025-08-14
DRAFT: conservative zonal means

rajeeja:
#527

### DIFF
--- a/uxarray/core/dataarray.py
+++ b/uxarray/core/dataarray.py
@@ -20,7 +20,10 @@ from uxarray.core.gradient import (
     _calculate_grad_on_edge_from_faces,
 )
 from uxarray.core.utils import _map_dims_to_ugrid
-from uxarray.core.zonal import _compute_non_conservative_zonal_mean
+from uxarray.core.zonal import (
+    _compute_conservative_zonal_mean_bands,
+    _compute_non_conservative_zonal_mean,
+)
 from uxarray.cross_sections import UxDataArrayCrossSectionAccessor
 from uxarray.formatting_html import array_repr
 from uxarray.grid import Grid
@@ -425,16 +428,24 @@ class UxDataArray(xr.DataArray):
 
         return uxda
 
-    def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
+    def zonal_mean(
+        self, lat=(-90, 90, 10), mode: str = "non-conservative", bands=None, **kwargs
+    ):
         """Compute averages along lines of constant latitude.
 
         Parameters
         ----------
         lat : tuple, float, or array-like, default=(-90, 90, 10)
-            Latitude values in degrees. Can be specified as:
+            For mode='non-conservative':
             - tuple (start, end, step): Computes means at intervals of `step` in range [start, end]
             - float: Computes mean for a single latitude
             - array-like: Computes means for each specified latitude
+            For mode='conservative':
+            - tuple (start, end, step): Interpreted as band edges via np.arange(start, end+step, step)
+        mode : {"non-conservative", "conservative"}, default="non-conservative"
+            Select the zonal averaging scheme.
+        bands : array-like, optional
+            Only used when mode='conservative'. Explicit band edges in degrees (size >= 2).
 
         Returns
         -------
@@ -444,59 +455,98 @@ class UxDataArray(xr.DataArray):
 
         Examples
         --------
-        # All latitudes from -90° to 90° at 10° intervals
+        # All latitudes from -90° to 90° at 10° intervals (non-conservative)
         >>> uxds["var"].zonal_mean()
 
-        # Single latitude at 30°
+        # Single latitude at 30° (non-conservative)
         >>> uxds["var"].zonal_mean(lat=30.0)
 
-        # Range from -60° to 60° at 10° intervals
-        >>> uxds["var"].zonal_mean(lat=(-60, 60, 10))
+        # Conservative bands from -60° to 60° at 10° band width
+        >>> uxds["var"].zonal_mean(lat=(-60, 60, 10), mode="conservative")
 
         Notes
         -----
-        Only supported for face-centered data variables. Candidate faces are determined
-        using spherical bounding boxes - faces whose bounds contain the target latitude
-        are included in calculations.
+        Only supported for face-centered data variables.
         """
         if not self._face_centered():
             raise ValueError(
                 "Zonal mean computations are currently only supported for face-centered data variables."
             )
 
-        if isinstance(lat, tuple):
-            # zonal mean over a range of latitudes
-            latitudes = np.arange(lat[0], lat[1] + lat[2], lat[2])
-            latitudes = np.clip(latitudes, -90, 90)
-        elif isinstance(lat, (float, int)):
-            # zonal mean over a single latitude
-            latitudes = [lat]
-        elif isinstance(lat, (list, np.ndarray)):
-            # zonal mean over an array of arbitrary latitudes
-            latitudes = np.asarray(lat)
-        else:
-            raise ValueError(
-                "Invalid value for 'lat' provided. Must either be a single scalar value, tuple (min_lat, max_lat, step), or array-like."
+        face_axis = self.dims.index("n_face")
+
+        if mode == "non-conservative":
+            if isinstance(lat, tuple):
+                latitudes = np.arange(lat[0], lat[1] + lat[2], lat[2])
+                latitudes = np.clip(latitudes, -90, 90)
+            elif isinstance(lat, (float, int)):
+                latitudes = [lat]
+            elif isinstance(lat, (list, np.ndarray)):
+                latitudes = np.asarray(lat)
+            else:
+                raise ValueError(
+                    "Invalid value for 'lat' provided. Must either be a single scalar value, tuple (min_lat, max_lat, step), or array-like."
+                )
+
+            res = _compute_non_conservative_zonal_mean(
+                uxda=self, latitudes=latitudes, **kwargs
             )
 
-        res = _compute_non_conservative_zonal_mean(
-            uxda=self, latitudes=latitudes, **kwargs
-        )
+            dims = list(self.dims)
+            dims[face_axis] = "latitudes"
 
-        face_axis = self.dims.index("n_face")
-        dims = list(self.dims)
-        dims[face_axis] = "latitudes"
+            uxda = UxDataArray(
+                res,
+                uxgrid=self.uxgrid,
+                dims=dims,
+                coords={"latitudes": latitudes},
+                name=self.name + "_zonal_mean"
+                if self.name is not None
+                else "zonal_mean",
+                attrs={"zonal_mean": True, "scheme": "non-conservative"},
+            )
+            return uxda
 
-        uxda = UxDataArray(
-            res,
-            uxgrid=self.uxgrid,
-            dims=dims,
-            coords={"latitudes": latitudes},
-            name=self.name + "_zonal_mean" if self.name is not None else "zonal_mean",
-            attrs={"zonal_mean": True},
-        )
+        elif mode == "conservative":
+            if bands is None:
+                if isinstance(lat, tuple):
+                    edges = np.arange(lat[0], lat[1] + lat[2], lat[2])
+                    edges = np.clip(edges, -90, 90)
+                else:
+                    raise ValueError(
+                        "For mode='conservative', provide 'bands' (edges) or a tuple (start, end, step) via 'lat'."
+                    )
+            else:
+                edges = np.asarray(bands, dtype=float)
 
-        return uxda
+            if edges.ndim != 1 or edges.size < 2:
+                raise ValueError("'bands' must be 1D with at least two edges (degrees)")
+
+            res = _compute_conservative_zonal_mean_bands(self, edges)
+
+            # Use band centers as coordinate values
+            centers = 0.5 * (edges[:-1] + edges[1:])
+
+            dims = list(self.dims)
+            dims[face_axis] = "latitudes"
+
+            uxda = UxDataArray(
+                res,
+                uxgrid=self.uxgrid,
+                dims=dims,
+                coords={"latitudes": centers},
+                name=self.name + "_zonal_mean"
+                if self.name is not None
+                else "zonal_mean",
+                attrs={
+                    "zonal_mean": True,
+                    "scheme": "conservative",
+                    "lat_band_edges": edges,
+                },
+            )
+            return uxda
+        else:
+            raise ValueError("mode must be one of {'non-conservative','conservative'}")
 
     # Alias for 'zonal_mean', since this name is also commonly used.
     zonal_average = zonal_mean
--- a/uxarray/core/zonal.py
+++ b/uxarray/core/zonal.py
@@ -1,6 +1,12 @@
 import numpy as np
 
+from uxarray.grid.arcs import compute_arc_length
+from uxarray.grid.area import get_gauss_quadrature_dg
 from uxarray.grid.integrate import _zonal_face_weights, _zonal_face_weights_robust
+from uxarray.grid.intersections import (
+    gca_const_lat_intersection,
+    get_number_of_intersections,
+)
 from uxarray.grid.utils import _get_cartesian_face_edge_nodes_array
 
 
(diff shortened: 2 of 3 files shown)


uxarray/core/zonal.py
```diff
@@ -60,3 +66,127 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
         result[tuple(idx)] = weighted
 
     return result
+
+
+def _band_overlap_area_for_face(face_edges_xyz, z_min, z_max, order=3):
```
philipc2: We should be using Numba here (if possible)

philipc2:
Is this using @hongyuchen1030 and @paullric 

> Check out this pull request on  [ReviewNB](https://app.reviewnb.com/UXARRAY/uxarray/pull/1345)
> 
> See visual diffs & provide feedback on Jupyter Notebooks.
> 
> Powered by [ReviewNB](https://www.reviewnb.com/?utm_source=gh)

We should not create a new notebook, instead extend the existing zonal average notebook.

rajeeja:
> Is this using @hongyuchen1030 and @paullric
> 
> > Check out this pull request on  [ReviewNB](https://app.reviewnb.com/UXARRAY/uxarray/pull/1345)
> > See visual diffs & provide feedback on Jupyter Notebooks.
> > Powered by [ReviewNB](https://www.reviewnb.com/?utm_source=gh)
> 
> We should not create a new notebook, instead extend the existing zonal average notebook.

will do, once we agree on several things here, for now - it is a playground for trying out things.

philipc2:
This could be helpful and reduce some of the code that you would need to write:

https://github.com/UXARRAY/uxarray/blob/c54d01fe9b63b2dbd20e8b841edca82ee2657f6a/uxarray/grid/grid.py#L2487-L2501

Once we obtain all faces between latitudes, we can then perform the intersection checks using the lower and upper bounds to partition the results like you are doing.

uxarray/core/dataarray.py
```diff
@@ -425,16 +428,24 @@ def integrate(
 
         return uxda
 
-    def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
+    def zonal_mean(
+        self, lat=(-90, 90, 10), mode: str = "non-conservative", bands=None, **kwargs
```
philipc2: I'd suggest a boolean `conservative` parameter. 

Also, the bands here are derived from the `lat` parameter , unless I'm understanding incorrectly.

rajeeja: our current non-conservative is for a line, for this we need a band (area)

rajeeja:
> Some notes from @hongyuchen1030's presentation to keep in mind:
> 
> > Given a set of latitude intervals, determine the fractional area from each of the faces that are crossed by the areal band.  These weights can be used to compute conservative zonal averages.

  The implementation:
   1. Identifies which faces overlap with each latitude band
   2. Computes the fractional area of overlap using numerical integration (Gauss-Legendre quadrature)
   3. Uses these fractional areas as weights when averaging the data values

hongyuchen1030:
> > Some notes from @hongyuchen1030's presentation to keep in mind:
> > > Given a set of latitude intervals, determine the fractional area from each of the faces that are crossed by the areal band.  These weights can be used to compute conservative zonal averages.
> 
> The implementation:
> 
> 1. Identifies which faces overlap with each latitude band
> 2. Computes the fractional area of overlap using numerical integration (Gauss-Legendre quadrature)
> 3. Uses these fractional areas as weights when averaging the data values

Did you include the area correction for constant latitude line?

rajeeja:
TODO: Add tests, coverage dropping

uxarray/core/dataarray.py
philipc2: The current one creates lines defined by the tuple `lat=(min, max, step)`, so in the default example, we'd have lines at `-90`, `-80` ... to `80`, `90`. The non-conservative one samples directly at these lines, while conservative should form bands between each of these lines.

uxarray/core/zonal.py
rajeeja: Done

uxarray/core/dataarray.py
rajeeja: it works this way

philipc2 (changes requested):
Please update the existing user guide notebook instead of creating a new one.

uxarray/core/dataarray.py
```diff
@@ -444,59 +451,107 @@ def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
 
         Examples
         --------
-        # All latitudes from -90° to 90° at 10° intervals
+        # Non-conservative averaging at 10° intervals
```
erogluorhan:
```suggestion
        # Non-conservative averaging from -90° to 90° at 10° intervals by default
```

uxarray/core/dataarray.py
```diff
@@ -444,59 +451,107 @@ def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
...
         >>> uxds["var"].zonal_mean()
 
-        # Single latitude at 30°
+        # Single latitude (non-conservative)
```
erogluorhan:
```suggestion
        # Single latitude (non-conservative) over 30° latitude
```

uxarray/core/dataarray.py
```diff
@@ -425,16 +428,20 @@ def integrate(
 
         return uxda
 
-    def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
+    def zonal_mean(self, lat=(-90, 90, 10), conservative: bool = False, **kwargs):
         """Compute averages along lines of constant latitude.
```
erogluorhan:
```suggestion
        """Compute non-conservative or conservative averages along lines of constant latitude or latitude bands.
```

uxarray/core/dataarray.py
```diff
@@ -444,59 +451,107 @@ def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
...
+                attrs={
+                    "zonal_mean": True,
+                    "scheme": "conservative",
+                    "lat_band_edges": edges,
+                },
+            )
+            return uxda
 
     # Alias for 'zonal_mean', since this name is also commonly used.
     zonal_average = zonal_mean
```
erogluorhan: I find aliases by assignment to be potentially confusing  for future, and it may easily be overlooked in the code. Also this way is probably inconsistent with how we did similar things before (see `GridPlotAccessor`'s `nodes()`, `node_coords()` etc. for instance). 

What do you think?

erogluorhan (commented):
Thanks a lot for doing this! I like the code and the docs, except only I put a few inline comments. 

The actual logic should ideally be reviewed and approved by @hongyuchen1030 though, if at all possible.

docs/user-guide/zonal-average.ipynb
```diff
@@ -7,7 +7,7 @@
    "source": [
```
erogluorhan, line 7: <p>This first plot of the global data is taking a valuable space that can better be used for the section below. Myayb e this plot should be moved to just before the first zonal avg plot, also providing some brief intro to it</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1345/discussion/#comment-2317090010'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

erogluorhan, line 7: <p>"A zonal average (or zonal mean) is a statistical measure that represents the average of a variable along one or more lines of constant latitude or over a latitudinal band."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1345/discussion/#comment-2317090015'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

erogluorhan, line 7: <p>Displaying one of those 6 lat bands on the global map would be a great visual help for the reader, if easily possible</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1345/discussion/#comment-2317090017'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='15'/>

erogluorhan, line 7: <p>All bold. Probably only the first parts are to be bold instead. Also, looks like some text editing is needed.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1345/discussion/#comment-2317090022'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='16'/>

erogluorhan, line 7: <pre class="ql-syntax" spellcheck="false">Line #6.    ax1.set_title("Global Field")
</pre><p>This global plot is not showing up in the figure currently.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/UXARRAY/uxarray/pull/1345/discussion/#comment-2317090024'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='21'/>

rajeeja:
@erogluorhan Thank you for the thorough review. All points are valid and I will address them today.

uxarray/core/zonal.py
```diff
@@ -60,3 +67,243 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
...
+        Points sorted by longitude angle
+    """
+    n_points = points.shape[0]
+    if n_points <= 1:
+        return points.copy()
+
+    # Calculate angles (longitude)
+    angles = np.empty(n_points, dtype=np.float64)
+    for i in range(n_points):
+        angles[i] = np.arctan2(points[i, 1], points[i, 0])
```
hongyuchen1030: When trying to calculate angles, always try to use the following robust way that incoporate the Kahan's angle formulation to avoid potential floating point precision issues.

https://github.com/UXARRAY/uxarray/blob/484bc2e82813c61c22c4f287d1b257ee7ecf8863/uxarray/grid/utils.py#L9

uxarray/core/zonal.py
```diff
@@ -60,3 +67,243 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
...
+
+@njit(cache=True)
+def _compute_band_overlap_area(
+    face_edges_xyz, z_min, z_max, quadrature_rule="gaussian", order=4
+):
+    """Compute overlap area between a face and latitude band using area.py functions.
+
+    This function finds the intersection polygon between a face and latitude band,
+    then uses calculate_face_area from area.py with latitude_adjusted_area=True
+    for accurate area computation when edges lie on constant latitude lines.
```
hongyuchen1030, line 137: So it always use the mathematically accurate area and intersection calculation?

uxarray/core/zonal.py
```diff
@@ -60,3 +67,243 @@ def _compute_non_conservative_zonal_mean(uxda, latitudes, use_robust_weights=Fal
...
+    result : array
+        Zonal means for each band
+    """
+    import dask.array as da
+
+    uxgrid = uxda.uxgrid
+    face_axis = uxda.get_axis_num("n_face")
+
+    # Pre-compute face properties
+    faces_edge_nodes_xyz = _get_cartesian_face_edge_nodes_array(
```
hongyuchen1030, line 236: @philipc2 Per my memory, we have stored this property inside the data array to avoid such re-population over and over again during the zonal average/remapping pipeline?

philipc2: We chose to not pre-compute and store these because of the memory requirement. The overall cost to compute these is negligible in the context of the broader computation.

hongyuchen1030: Sounds good.

uxarray/core/dataarray.py
```diff
@@ -458,59 +465,107 @@ def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
...
+                    "Invalid value for 'lat' provided. Must be a scalar, tuple (min_lat, max_lat, step), or array-like."
+                )
 
-        face_axis = self.dims.index("n_face")
-        dims = list(self.dims)
-        dims[face_axis] = "latitudes"
+            # Filter out conservative-specific kwargs
+            filtered_kwargs = {
+                k: v for k, v in kwargs.items() if k not in ["mode", "bands"]
+            }
```
philipc2: Is this necessary after the changes to use the `lat` parameter?

uxarray/core/dataarray.py
```diff
@@ -458,59 +465,107 @@ def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
...
+                res,
+                uxgrid=self.uxgrid,
+                dims=dims,
+                coords={"latitudes": centers},
+                name=self.name + "_zonal_mean"
+                if self.name is not None
+                else "zonal_mean",
+                attrs={
+                    "zonal_mean": True,
+                    "scheme": "conservative",
```
philipc2: Can we have a boolean conservative attribute instead?

uxarray/core/dataarray.py
philipc2: > I find aliases by assignment to be potentially confusing for future, and it may easily be overlooked in the code.

May you elaborate on the confusion? This is mostly added as a fall-back for users who may expect a `zonal_average` signature, as opposed to `zonal_mean`. 

We only document the use of `zonal_mean()`, however if a user accidentally/instinctively uses the wrong signature, they will still have the expected functionality.

rajeeja:
> Nice work @rajeeja
> 
> Have you had a chance to take a look at the execution time for higher resolution grids?

 HEALPix zoom 11 end-to-end time: 298.28 s.

uxarray/core/dataarray.py
erogluorhan: For instance, online docs will be about `zonal_mean` only, so those users whose `zonal_average` attempt is seamlessly covered by this alias might have issues when it comes to making simple searches online for UXarray resources.

I am not too worried though. So, if you both think this should be the way, I am okay with it.

erogluorhan:
> The intent here is a user-facing convenience for a widely used term (“zonal_average”) while keeping a single implementation and a single documented API (“zonal_mean”). A thin wraper would be good, or we can just remove it you think it will be cleaner - no confusion.

I think a convenience one way or other is necessary, you're right. I thought a thin wrapper would be a better idea than an alias, but I am also not too worried. So, don't remove it unless you will add a thin wrapper

uxarray/core/dataarray.py
```diff
@@ -510,16 +513,20 @@ def integrate(
...
+        """Compute non-conservative or conservative averages along lines of constant latitude or latitude bands.
 
         Parameters
         ----------
         lat : tuple, float, or array-like, default=(-90, 90, 10)
-            Latitude values in degrees. Can be specified as:
-            - tuple (start, end, step): Computes means at intervals of `step` in range [start, end]
-            - float: Computes mean for a single latitude
-            - array-like: Computes means for each specified latitude
+            Latitude specification:
```
hongyuchen1030, line 522: Should we constrain the smallest step size? Per my memory, the current implementation robustness level only supports an integer level of the band.

hongyuchen1030: And if we are supporting a random float, at least reflect it in the test case, especially when closed to the pole

rajeeja: added this support.

uxarray/core/dataarray.py
```diff
@@ -529,62 +536,135 @@ def zonal_mean(self, lat=(-90, 90, 10), **kwargs):
...
+            centers = 0.5 * (edges[:-1] + edges[1:])
+
+            dims = list(self.dims)
+            dims[face_axis] = "latitudes"
+
+            warnings.warn(
+                "zonal_mean returns an xarray.DataArray (no grid topology). Returning UxDataArray is deprecated and will be removed.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
philipc2: Do we need this warning if the feature is already deprecated?

PR #1349, opened 2025-08-20
Moving tests from flat to hierarchical structure

rajeeja:
This branch reorganizes the test suite by consolidating scattered test file paths and constants into centralized pytest fixtures (gridpath, datasetpath, and mesh_constants) via `conftest.py`, replacing the previous approach where paths were defined individually in each test file. The migration standardizes how all 40+ test files access test data, making the test suite more maintainable. Running Tests:
Tests must now be run from the repository root directory using python -m pytest instead of just pytest . in test/ directory as the fixtures need to resolve paths relative to the project root. For example:
   - Run all tests: python -m pytest test/
   - Run specific test file: python -m pytest test/test_grid.py
   - Run specific test function: python -m pytest test/test_grid.py::test_grid_validate
   - Run tests in parallel: python -m pytest -n auto test/

### DIFF
--- a/test/io/test_ugrid.py
+++ b/test/io/test_ugrid.py
@@ -8,10 +8,6 @@ import uxarray as ux
 from uxarray.constants import INT_DTYPE, INT_FILL_VALUE
 
 
-# Constants now come from fixtures
-
-
-
 def test_read_ugrid(gridpath, mesh_constants):
     """Reads a ugrid file."""
     uxgrid_ne30 = ux.open_grid(str(gridpath("ugrid", "outCSne30", "outCSne30.ug")))
--- a/test/grid/test_arcs.py
+++ b/test/grid/test_arcs.py
@@ -5,8 +5,6 @@ import pytest
 from uxarray.grid.coordinates import _lonlat_rad_to_xyz
 from uxarray.grid.arcs import point_within_gca
 
-# Data files - using centralized paths directly
-
 def test_pt_within_gcr():
     # The GCR that's exactly 180 degrees will raise a ValueError
     gcr_180degree_cart = np.asarray([
--- a/test/grid/test_grid.py
+++ b/test/grid/test_grid.py
@@ -20,20 +20,6 @@ from uxarray.grid.arcs import extreme_gca_latitude
 
 from uxarray.grid.validation import _find_duplicate_nodes
 
-
-# quad_hex_grid_path is now available from centralized paths as QUAD_HEXAGON_GRID
-
-# These will be set in test methods that need them
-shp_filename = None
-grid_CSne30 = None
-
-
-
-# These will be set in test methods that need them
-grid_mpas = None
-grid_exodus = None
-grid_ugrid = None
-
 f0_deg = [[120, -20], [130, -10], [120, 0], [105, 0], [95, -10], [105, -20]]
 f1_deg = [[120, 0], [120, 10], [115, 0],
           [ux.INT_FILL_VALUE, ux.INT_FILL_VALUE],
@@ -306,8 +292,9 @@ def test_grid_properties(gridpath):
     assert n_faces == grid_geoflow.n_face
     assert n_face_nodes == grid_geoflow.n_max_face_nodes
 
-def test_read_shpfile():
+def test_read_shpfile(test_data_dir):
     """Reads a shape file and write ugrid file."""
+    shp_filename = test_data_dir / "shp" / "grid_fire.shp"
     with pytest.raises(ValueError):
         grid_shp = ux.open_grid(shp_filename)
 
--- a/test/io/test_scrip.py
+++ b/test/io/test_scrip.py
@@ -8,10 +8,6 @@ import uxarray as ux
 from uxarray.constants import INT_DTYPE, INT_FILL_VALUE
 
 
-# Constants now come from fixtures
-
-
-
 def test_read_ugrid(gridpath, mesh_constants):
     """Reads a ugrid file."""
     uxgrid_ne30 = ux.open_grid(str(gridpath("ugrid", "outCSne30", "outCSne30.ug")))
--- a/test/test_cross_sections.py
+++ b/test/test_cross_sections.py
@@ -2,20 +2,8 @@ import uxarray as ux
 import pytest
 import numpy as np
 import xarray as xr
-from pathlib import Path
 
 import numpy.testing as nt
-
-# Define the current path and file paths for grid and data
-current_path = Path(__file__).resolve().parent
-quad_hex_grid_path = current_path / 'meshfiles' / "ugrid" / "quad-hexagon" / 'grid.nc'
-quad_hex_data_path = current_path / 'meshfiles' / "ugrid" / "quad-hexagon" / 'data.nc'
-quad_hex_node_data = current_path / 'meshfiles' / "ugrid" / "quad-hexagon" / 'random-node-data.nc'
-cube_sphere_grid = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30.ug"
-
-
-csne8_grid = current_path / "meshfiles" / "scrip" / "ne30pg2" / "grid.nc"
-csne8_data = current_path / "meshfiles" / "scrip" / "ne30pg2" / "data.nc"
 
 
 
@@ -26,8 +14,8 @@ from uxarray.grid.intersections import constant_lat_intersections_face_bounds
 
 
 
-def test_constant_lat_subset_grid():
-    uxgrid = ux.open_grid(quad_hex_grid_path)
+def test_constant_lat_subset_grid(gridpath):
+    uxgrid = ux.open_grid(gridpath("ugrid", "quad-hexagon", "grid.nc"))
 
     grid_top_two = uxgrid.subset.constant_latitude(lat=0.1)
     assert grid_top_two.n_face == 2
@@ -41,8 +29,8 @@ def test_constant_lat_subset_grid():
     with pytest.raises(ValueError):
         uxgrid.subset.constant_latitude(lat=10.0)
 
-def test_constant_lon_subset_grid():
-    uxgrid = ux.open_grid(quad_hex_grid_path)
+def test_constant_lon_subset_grid(gridpath):
+    uxgrid = ux.open_grid(gridpath("ugrid", "quad-hexagon", "grid.nc"))
 
     grid_left_two = uxgrid.subset.constant_longitude(lon=-0.1)
     assert grid_left_two.n_face == 2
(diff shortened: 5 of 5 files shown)


rajeeja:
Add a new file paths.py, if we agree I can update all test files to use centralized paths  ? - @erogluorhan @philipc2 

there are some folders with one file, but that is fine IMO, also some big files that can be broken into smaller, it'd be help if we agree on things before I start to implement things.

philipc2:
I'm personally not a huge fan of having sub directories and separate testing modules for each python module. Looking at what other packages do (Xarray, NumPy, etc), each of which cover a larger scope than we do, they don't have a design like this.

Moving forward, I think the overall plan should still be to have a single `tests` directory with individual testing modules (i.e. `test_grid.py`), but start to clean up and remove any duplicate tests. 

I'm curious to what @erogluorhan has to say too.

philipc2:
I like @rajeeja's suggestion for a hybrid design. I don't think a 1:1 match for each directory and module is the best, however having some sub-directories (i.e. `grid`, `core`) could work well.

test/paths.py
```diff
@@ -0,0 +1,89 @@
+"""
+Centralized test file paths for all mesh files.
+This module provides consistent paths to test data files for all test modules.
+"""
```
philipc2: I like this. Much better than re-defining the path each time.

test/core/test_api.py
```diff
@@ -8,9 +8,11 @@
 try:
     import constants
 except ImportError:
-    from . import constants
+    import sys
+    sys.path.append(str(Path(__file__).parent.parent))
+    import constants
 
-current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+current_path = Path(os.path.dirname(os.path.realpath(__file__))).parent
```
erogluorhan: The `paths.py` will help get rid of these, and you will incorporate it to all of these scripts, right?

erogluorhan (commented):
I like the way this PR brings our test suite to. See my single inline comment plus the following:

* If there is a single file in any subfolder, maybe put that file into the main test dir until multiple files are present
* > Sounds good, what do you think about the paths.py file ? that would change all of the tests and how it gets the files, please take a look. It be good to be on the same page before I make all those changes. @erogluorhan
   *  I like the paths.py a lot!

test/core/test_api.py
rajeeja: Yes, that is absolutely correct

test/io/test_ugrid.py
```diff
@@ -8,17 +8,16 @@
...
-current_path = Path(os.path.dirname(os.path.realpath(__file__)))
-
-gridfile_ne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30.ug"
-gridfile_RLL1deg = current_path / "meshfiles" / "ugrid" / "outRLL1deg" / "outRLL1deg.ug"
-gridfile_RLL10deg_ne4 = current_path / "meshfiles" / "ugrid" / "ov_RLL10deg_CSne4" / "ov_RLL10deg_CSne4.ug"
-gridfile_exo_ne8 = current_path / "meshfiles" / "exodus" / "outCSne8" / "outCSne8.g"
+gridfile_ne30 = OUTCSNE30_GRID
+gridfile_RLL1deg = OUTRLL1DEG_GRID
+gridfile_RLL10deg_ne4 = OV_RLL10DEG_CSNE4_GRID
+gridfile_exo_ne8 = EXODUS_OUTCSNE8
```
philipc2: Minor nitpick, but I've noticed that each testing module is redefining these variables. Shouldn't we instead use the imported variables from the `paths` module?

So for example, we'd use `OUTCSNE30_GRID` directly instead of creating a variable `gridfile_ne30`

test/grid/test_centroids.py
```diff
@@ -13,10 +13,13 @@
     _normalize_xyz,
 )
 
-current_path = Path(os.path.dirname(os.path.realpath(__file__)))
+# Import centralized paths
+import sys
+sys.path.append(str(Path(__file__).parent.parent))
```
philipc2: Are these two lines necessary?

rajeeja: Yes, these sys.path.append lines are currently necessary because test files in subdirectories need to import from paths.py in the parent test/ directory.  Do you think a cleaner solution would be to add __init__.py files to make test/ a proper Python package and then use relative imports? or something else?

rajeeja: Another option would be to move the test paths into test/conftest.py as pytest fixtures, which is how many packages like pandas and scikit-learn handle test data paths. This would eliminate the imports entirely.

philipc2: I like that idea. We should be adding `__init__.py` to each of the sub directories anyways I believe. Ideally, in each of the testing modules, having the following would be the cleanest. 

```Python
# Paths is in the parent directory, can use the . syntax 
from .paths import GRID_FILE # or whichever are used
```

rajeeja: `_test/conftest.py_`  (automatically loaded by pytest) is better I think, it is cleaner and more pytest-idiomatic; We can move all path constants to _conftest.py_ as fixtures - it will be little bit of a rewrite, but more maintainable @erogluorhan what do you tihnk?

rajeeja: only downside is we'd be locked to pytest ecosystem

philipc2: We should only be using `pytest` anyways. I'm okay with that.

erogluorhan: Yeah, I like that idea to have a cleaner path access with imports eliminated.

erogluorhan (commented):
As soon as [this](https://github.com/UXARRAY/uxarray/pull/1349#discussion_r2330553046) is implemented, this PR should be good to go I believe.

test/core/test_accessors.py
```diff
@@ -15,19 +15,11 @@
...
-dsfile_v1_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "v1.nc"
-mpas_ds_path = current_path / 'meshfiles' / "mpas" / "QU" / 'mesh.QU.1920km.151026.nc'
-
-
-def test_groupby_preserves_uxgrid():
+def test_groupby_preserves_uxgrid(datapath):
     """Test that groupby operations preserve the uxgrid attribute."""
     # Create a dataset from a file
-    uxds = ux.open_dataset(mpas_ds_path, mpas_ds_path)
+    mesh_path = datapath("mpas", "QU", "mesh.QU.1920km.151026.nc")
```
erogluorhan: This is much cleaner!

test/conftest.py
```diff
@@ -0,0 +1,49 @@
+"""Pytest configuration and fixtures for uxarray tests."""
+
+import pytest
+import numpy as np
+from pathlib import Path
+
+
+@pytest.fixture
+def datapath():
```
erogluorhan: Is this name a pytest convention, or can we do something else (e.g. filepath)? 

Asking because having a distinction between "data" vs "grid" would be really helpful in the UXarray context, and seeing "datapath" for grid files feels a bit awkward.

philipc2: Agree on the `datapath` name. Perhaps `datasetpath` would be more appropriate if it's returning a pair of grid and data paths. 

EDIT:

After looking at a few of the examples below, I would lean towards a seperate `grid` and `data fixture if possible.

erogluorhan: Not sure if having separate fixtures would be needed here as the main role of both would be the exact same thing: file access. My point was more about the "data" phrase in the `datapath` fixture's name as it makes it hard to get the immediate distinction between grid files and data files. 

So, if we can find a better name, maybe we should be good?

EDIT: However, I also get that having `gridpath` and `datapath` fixtures called next to each other in the test scripts would be very helpful for the readability of the tests (even though they basically do the same job)

philipc2: > EDIT: However, I also get that having gridpath and datapath fixtures called next to each other in the test scripts would be very helpful for the readability of the tests (even though they basically do the same job)

That's what I was thinking, from a readability perspective it may be more clear.

rajeeja: datapath came from my research on pandas. Just made the changes. Good  feedback. Thanks!
Both are the same, but if we ever need different behavior (e.g., different validation, different base paths), we can modify them independently

test/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+        path = base_path.joinpath(*args)
+        if not path.exists():
+            pytest.skip(f"Test grid file not found: {path}")
+        return path
+
+    return _get_path
+
+
+@pytest.fixture
+def datasetpath():
```
philipc2, line 48: Could we extend this to support returning a list of data paths if the user provides multiple, such as for the geoflow case?

test/core/test_api.py
```diff
@@ -5,22 +5,21 @@
...
-current_path = Path(os.path.dirname(os.path.realpath(__file__)))
 
-geoflow_data_path = current_path / "meshfiles" / "ugrid" / "geoflow-small"
-gridfile_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "grid.nc"
-geoflow_data_v1 = geoflow_data_path / "v1.nc"
-geoflow_data_v2 = geoflow_data_path / "v2.nc"
-geoflow_data_v3 = geoflow_data_path / "v3.nc"
+geoflow_data_v1 = GEOFLOW_V1
+geoflow_data_v2 = GEOFLOW_V2
+geoflow_data_v3 = GEOFLOW_V3
```
philipc2: See my previous comment. Anyway to use the fixture for this?

test/core/test_dataarray.py
```diff
@@ -6,34 +6,33 @@
...
-dsfile_var2_ne30 = current_path / "meshfiles" / "ugrid" / "outCSne30" / "outCSne30_vortex.nc"
 
-gridfile_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "grid.nc"
-dsfile_v1_geoflow = current_path / "meshfiles" / "ugrid" / "geoflow-small" / "v1.nc"
 
 def test_to_dataset():
     """Tests the conversion of UxDataArrays to a UXDataset."""
-    uxds = ux.open_dataset(gridfile_ne30, dsfile_var2_ne30)
-    uxds_converted = uxds['psi'].to_dataset()
+    uxds = ux.open_dataset(OUTCSNE30_GRID, OUTCSNE30_VAR2)
```
philipc2: Use fixture

test/core/test_dataarray.py
```diff
@@ -6,34 +6,33 @@
...
+    uxds_converted = uxds['var2'].to_dataset()
 
     assert isinstance(uxds_converted, UxDataset)
     assert uxds_converted.uxgrid == uxds.uxgrid
 
 def test_get_dual():
     """Tests the creation of the dual mesh on a data array."""
-    uxds = ux.open_dataset(gridfile_ne30, dsfile_var2_ne30)
-    dual = uxds['psi'].get_dual()
+    uxds = ux.open_dataset(OUTCSNE30_GRID, OUTCSNE30_VAR2)
```
philipc2: Use fixture

philipc2 (changes requested):
Can you update the tests to use the fixtures? There are many spots that still use the raw paths.

test/core/test_dataarray.py
rajeeja: I only changed on file for now test/core/test_accessors.py - confirming things before changes to all files

philipc2: Got it!

test/conftest.py
rajeeja: Yes, that can be done datasetpath("ugrid", "geoflow-small", ["v1.nc", "v2.nc", "v3.nc"]) would return a list of paths

PR #1351, opened 2025-08-27
Add warning when constructing a PolyCollection/LineCollection with a projection

philipc2:
# Overview
- Adds a warning when a user provides a projection to `to_polycollection()` and `to_linecollection()` calls. 
- Updates logic to prevent duplicate projection arguments from being passed in, which caused errors

### DIFF
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -2217,6 +2217,14 @@ class Grid:
         """
         import cartopy.crs as ccrs
 
+        if kwargs.get("projection", False):
+            warn(
+                "Explicit providing a projection when construction a PolyCollection has been deprecated. Please provide a projection when"
+                "constructing a GeoAxes instead.",
+                stacklevel=2,
+            )
+            kwargs.pop("projection")
+
         if self._cached_poly_collection:
             return copy.deepcopy(self._cached_poly_collection)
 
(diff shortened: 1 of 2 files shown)


uxarray/grid/grid.py
```diff
@@ -2237,15 +2245,23 @@ def to_linecollection(
...
         ----------
         **kwargs: dict
-            Key word arguments to pass into the construction of a PolyCollection
+            Key word arguments to pass into the construction of a LineCollection
         """
         import cartopy.crs as ccrs
 
+        if kwargs.get("projection", False):
+            warn(
+                "Explicitly providing a projection when construction a LolyCollection has been deprecated. Please provide a projection when setting up"
```
erogluorhan:
```suggestion
                "Explicitly providing a projection when constructing a LineCollection has been deprecated. Please provide a projection when setting up"
```

uxarray/grid/grid.py
```diff
@@ -2217,6 +2217,14 @@ def to_polycollection(
         """
         import cartopy.crs as ccrs
 
+        if kwargs.get("projection", False):
+            warn(
+                "Explicitly providing a projection when construction a PolyCollection has been deprecated. Please provide a projection when setting up"
```
erogluorhan:
```suggestion
                "Explicitly providing a projection when constructing a PolyCollection has been deprecated. Please provide a projection when setting up"
```

erogluorhan: In addition to above, the last sentence reads like incomplete, how about something like "You provide a projection later, e.g. when setting up your plot with the PolyCollection."?

uxarray/grid/grid.py
erogluorhan: In addition to above, the last sentence reads like incomplete, how about something like "You provide a projection later, e.g. when setting up your plot with the LineCollection."?

erogluorhan (changes requested):
Thanks for this fix! See my simple change requests.

uxarray/grid/grid.py
philipc2: Great suggestion.

philipc2:
@erogluorhan 

I think the updated warning should be much more informative now. Let me know what you think.

PR #1352, opened 2025-08-28
Add `UxDataset.isel()` 

philipc2:
# Overview
- Introduces a grid-informed implementation of `UxDataset.isel()`


uxarray/core/utils.py
```diff
@@ -80,3 +80,19 @@ def match_chunks_to_ugrid(grid_filename_or_obj, chunks):
             chunks[original_grid_dim] = chunks[ugrid_grid_dim]
 
     return chunks
+
+
+def _validate_indexers(indexers, indexers_kwargs, func_name, ignore_grid):
```
erogluorhan, line 85: This is good addition!

uxarray/core/dataarray.py
```diff
@@ -1297,56 +1297,37 @@ def isel(
...
+            indexers, indexers_kwargs, "isel", ignore_grid
+        )
 
         # Grid Branch
         if not ignore_grid and len(grid_dims) == 1:
             # pop off the one grid‐dim indexer
-            grid_dim = grid_dims[0]
-            grid_indexer = idx_map.pop(grid_dim)
+            grid_dim = grid_dims.pop()
+            grid_indexer = indexers.pop(grid_dim)
```
erogluorhan: `pop`ing from `indexers` wouldn't lead to any issues later in this code and after the return, right (similarly for `grid_dims`)? Asking this as it will remove last element in-place

philipc2: Correct. We pop the grid dimension form the `indexes` because we are performing the slicing ourselves within `_slice_from_grid`

erogluorhan: Oh, right! Thanks for clarifying!

PR #1354, opened 2025-08-30
Add `pixel_ratio` option to `.to_raster()`

### LINKED ISSUE #1356: Matplotlib PolyCollection example failing for main branch
https://uxarray.readthedocs.io/en/latest/user-guide/mpl.html#visualize-data-with-polycollection

`GEOSException: UnsupportedOperationException: getX called on empty Point`

<details>
<summary>Full traceback</summary>

```
---------------------------------------------------------------------------
GEOSException                             Traceback (most recent call last)
Cell In[8], line 17
     14 ax.add_feature(cfeature.COASTLINE)
     15 ax.add_feature(cfeature.BORDERS)
---> 17 ax.add_collection(poly_collection)
     18 ax.set_global()
     19 plt.title("PolyCollection")

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/matplotlib/axes/_base.py:2367, in _AxesBase.add_collection(self, collection, autolim)
   2363 if autolim:
   2364     # Make sure viewLim is not stale (mostly to match
   2365     # pre-lazy-autoscale behavior, which is not really better).
   2366     self._unstale_viewLim()
-> 2367     datalim = collection.get_datalim(self.transData)
   2368     points = datalim.get_points()
   2369     if not np.isinf(datalim.minpos).all():
   2370         # By definition, if minpos (minimum positive value) is set
   2371         # (i.e., non-inf), then min(points) <= minpos <= max(points),
   2372         # and minpos would be superfluous. However, we add minpos to
   2373         # the call so that self.dataLim will update its own minpos.
   2374         # This ensures that log scales see the correct minimum.

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/matplotlib/collections.py:273, in Collection.get_datalim(self, transData)
    270     return transforms.Bbox.null()
    272 if not transform.is_affine:
--> 273     paths = [transform.transform_path_non_affine(p) for p in paths]
    274     # Don't convert transform to transform.get_affine() here because
    275     # we may have transform.contains_branch(transData) but not
    276     # transforms.get_affine().contains_branch(transData).  But later,
    277     # be careful to only apply the affine part that remains.
    279 offsets = self.get_offsets()

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/matplotlib/transforms.py:2427, in CompositeGenericTransform.transform_path_non_affine(self, path)
   2425     return path
   2426 elif not self._a.is_affine and self._b.is_affine:
-> 2427     return self._a.transform_path_non_affine(path)
   2428 else:
   2429     return self._b.transform_path_non_affine(
   2430                             self._a.transform_path(path))

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/cartopy/mpl/geoaxes.py:174, in InterProjectionTransform.transform_path_non_affine(self, src_path)
    171     return mpath.Path(self.transform(src_path.vertices))
    173 geom = cpath.path_to_shapely(src_path)
--> 174 transformed_geom = self.target_projection.project_geometry(
    175     geom, self.source_projection)
    177 result = cpath.shapely_to_path(transformed_geom)
    179 # store the result in the cache for future performance boosts

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/cartopy/crs.py:833, in Projection.project_geometry(self, geometry, src_crs)
    831 if not method_name:
    832     raise ValueError(f'Unsupported geometry type {geom_type!r}')
--> 833 return getattr(self, method_name)(geometry, src_crs)

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/cartopy/crs.py:836, in Projection._project_point(self, point, src_crs)
    835 def _project_point(self, point, src_crs):
--> 836     return sgeom.Point(*self.transform_point(point.x, point.y, src_crs))

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/shapely/geometry/point.py:91, in Point.x(self)
     88 @property
     89 def x(self):
     90     """Return x coordinate."""
---> 91     return float(shapely.get_x(self))

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/shapely/decorators.py:88, in multithreading_enabled.<locals>.wrapped(*args, **kwargs)
     86     for arr in array_args:
     87         arr.flags.writeable = False
---> 88     return func(*args, **kwargs)
     89 finally:
     90     for arr, old_flag in zip(array_args, old_flags):

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.13/site-packages/shapely/_geometry.py:298, in get_x(point, **kwargs)
    273 @multithreading_enabled
    274 def get_x(point, **kwargs):
    275     """Return the x-coordinate of a point.
    276 
    277     Parameters
   (...)    296 
    297     """
--> 298     return lib.get_x(point, **kwargs)

GEOSException: UnsupportedOperationException: getX called on empty Point
```
</details>

### LINKED ISSUE #1363: Add options to `.to_raster()`
Since `.to_raster()`[^a] doesn't expose any options, it seems that currently the only way to make it try harder and produce a sharper image (assuming you have high-resolution data) is to set the figure DPI to something higher than the default (100) _before_ calling `.to_raster()`. This is a bit annoying if you're using Jupyter though, because unless you reset the figure DPI to the default afterwards, it will display the figure blown up.

Anyway, I feel like this method should at least be mentioned in <https://uxarray.readthedocs.io/en/latest/user-guide/mpl.html#rasterization>. And/or option(s) could be added to `.to_raster()`.

[^a]: thanks for adding this, have been enjoying it

_Originally posted by @zmoon in https://github.com/UXARRAY/uxarray/discussions/1353_

zmoon:
## Overview

As discussed in #1353, this PR introduces a `pixel_ratio` setting to the `UxDataArray.to_raster()` method, allowing you to easily control the resolution of the rasterization when using Matplotlib+Cartopy without HoloViz.

One issue is that the get-faces-containing-points gets pretty slow as you increase pixel ratio. So I also added an option to pass in a pre-computed mapping (`pixel_mapping`). Example use-cases could be if you are making a lot of plots of the same scene, or overlaying multiple fields in one plot, or tweaking things and want a faster iterate cycle. But I wasn't sure what would be the best way to expose _creating the mapping_ to the user.

Also added an example and a few other docs tweaks.

## Expected Usage

```python
import cartopy.crs as ccrs
import uxarray as ux

fig, ax = plt.subplots(
    subplot_kw={"projection": ccrs.Robinson()}, figsize=(8, 4), constrained_layout=True,
)

ax.set_extent([-125, -67, 24.5, 50], crs=ccrs.PlateCarree())
raster = uxds["dataVar"].to_raster(ax=ax, pixel_ratio=2.0)

ax.imshow(raster, origin="lower", extent=ax.get_xlim() + ax.get_ylim())
```

## PR Checklist

**General**
- Add appropriate labels

**Testing**

**Documentation**
- [x] Internal functions have a preceding underscore (`_`)

**Examples**
- Any new notebook examples added to `docs/examples/` folder
- [x] Clear the output of all cells before committing
- New notebook files added to `docs/examples.rst` toctree
- New notebook files added to new entry in `docs/gallery.yml` with appropriate thumbnail photo in `docs/_static/thumbnails/`


zmoon:
> One issue is that the get-faces-containing-points gets pretty slow as you increase pixel ratio.

bench

A little non-rigorous benchmark with the global QU 480 example to demonstrate.

<details>

```python
import time

import cartopy.crs as ccrs
import pandas as pd

import uxarray as ux
from uxarray.plot.matplotlib import _get_raster_pixel_to_face_mapping as get_pixel_mapping

base_path = "test/meshfiles/mpas/QU/480/"
grid_path = base_path + "grid.nc"
data_path = base_path + "data.nc"

fig, ax = plt.subplots(
    subplot_kw={"projection": ccrs.Robinson()}, figsize=(8, 4), constrained_layout=True,
)

ax.set_global()

results = []
for pixel_ratio in [0.01, 0.05, 0.1, 0.25, 0.5, 0.75, 1, 2, 3, 4, 6, 8]:
    for _ in range(5):
        tic = time.perf_counter_ns()
        _ = get_pixel_mapping(uxds, ax=ax, pixel_ratio=pixel_ratio)
        toc = time.perf_counter_ns()
        results.append((pixel_ratio, toc - tic))

df = pd.DataFrame(results, columns=["pixel_ratio", "time_ns"])
df.to_csv("bench.csv", index=False)
</details>
```

philipc2:
> > One issue is that the get-faces-containing-points gets pretty slow as you increase pixel ratio.
> 
> bench
> A little non-rigorous benchmark with the global QU 480 example to demonstrate.

Do you have the system specs for the benchmark run here?

uxarray/grid/geometry.py
```diff
@@ -138,7 +138,15 @@ def _build_polygon_shells(
         .swapaxes(1, 2)
     )
 
-    return polygon_shells
+    # XXX: fixing the PolyCollection example for now
```
zmoon: Sorry the hacky way I did this seems to have messed with gdf creation a bit, failing two tests. Maybe better to create a filter function and apply it within `_grid_to_matplotlib_polycollection`?

philipc2:
> I was also curious about this warning that surfaces to the user:
> 
> ```
> uxarray/grid/point_in_face.py:168: NumbaPerformanceWarning: np.dot() is faster on contiguous arrays, called on (Array(float64, 1, 'C', False, aligned=True), Array(float64, 1, 'A', False, aligned=True))
>   hits = _get_faces_containing_point(
> ```
> 
> Did you look at trying to address that? If you decide(d) it's not important or whatever, maybe we can catch and filter it so the user doesn't see it.

I've seen this error before, but didn't notice it in our workflows. 

I'll raise an issue about this and we can tackle it, should improve our performance.

uxarray/grid/geometry.py
zmoon: In f8a013a I moved this inside `_grid_to_matplotlib_polycollection`.

uxarray/plot/accessor.py
```diff
@@ -19,9 +19,17 @@
...
-    Accessed through `Grid.plot()` or `Grid.plot.specific_routine()`
+    You can call :func:`Grid.plot`
+    which plots the edges, or you can specify a routine.
+    For example:
+
+    - :func:`Grid.plot.edges`
+    - :func:`Grid.plot.nodes`
+    - :func:`Grid.plot.face_centers`
+
+    See examples in :doc:`/user-guide/plotting`.
```
philipc2, line 32: Nice additions here! The rendered API docs look much better now:

image

philipc2 (approved):
Thank you so much for your contribution @zmoon 

This looks great!

erogluorhan (commented):
EDIT: Oh, I missed your LRU caching discussion before putting this following review in. Anyways, let's still discuss.

Overall this looks great; thanks very much for the contribution, @zmoon ! 

I like the whole code and docs, but I'd like to discuss the `get_raster_pixel_mapping()`:

I think that would only be useful for the `to_raster()` cases, right? And, actually for cases where there are multiple plot calls with a high `pixel_ratio` value where calling `get_faces_containing_point` after the first one would be redundant. If that's correct, how about: 

Introduce a caching mechanism (i.e. the user makes the first call to `plot` with a `cache=True`, which would cache the pixel mapping to `Grid` or wherever relevant, then other plot calls would use that cached one)?  `to_raster()` would always check if there is cached pixel_mapping, and use that one if so.

We did similar things so far.

For this though, we'd need to make sure the memory consumption for the pixel mapping is not huge.

Thoughts?

zmoon:
I think the @erogluorhan plan sounds good. I am happy to remove `pixel_mapping` stuff and work on that here, or in a separate PR.

> For this though, we'd need to make sure the memory consumption for the pixel mapping is not huge.

Regarding this aspect, assuming we are going to cache more than one `ax`+`pixel_ratio` setting, we could expose a limit to how many as a Grid class attribute. And/or other potential settings like a max array size. So that user could have some control. But if you think that's not necessary that's cool too.

erogluorhan:
@zmoon @philipc2 , re-thinking this, caching something that is `Axis`-dependent (i.e. for visualization purposes only) into the core structures, `Grid` etc., does not make as much sense to me now. 

I am now leaning towards either @zmoon 's first suggestion of exposing a function for mapping creation, or maybe something like:

`raster, pixel_mapping = uxds["dataVar"].to_raster(ax=ax, pixel_ratio=2.0)` when user wants to re-use `pixel_mapping` or 

`raster, ~ = uxds["dataVar"].to_raster(ax=ax, pixel_ratio=2.0)` in most cases.

zmoon:
Yeah, I think I'd slightly prefer the kwarg, keep `to_raster()` just returning `raster` by default. I will give it a go in this PR then unless you object.

philipc2:
I like the idea of returning the `pixel_mapping` through a kwarg.

erogluorhan:
Sounds like we have a plan; looking forward to your edits, @zmoon !

uxarray/core/dataarray.py
```diff
@@ -334,9 +360,48 @@ def to_raster(self, ax: GeoAxes):
...
+            if isinstance(pixel_mapping, xr.DataArray):
+                pixel_ratio_input = pixel_ratio
+                pixel_ratio = pixel_mapping.attrs["pixel_ratio"]
+                if pixel_ratio_input != 1 and pixel_ratio_input != pixel_ratio:
+                    warn(
+                        "Pixel ratio mismatch: "
+                        f"{pixel_ratio_input} passed but {pixel_ratio} in pixel_mapping. "
+                        "Using the pixel_mapping attribute.",
+                        stacklevel=2,
+                    )
```
philipc2: Should we validate whether the other attributes match with the provied `ax`? 
- `ax_projection`
- `ax_xlim`
- `ax_ylim`
- `ax_width`
- `ax_height`

We could include this logic within `_nearest_neighbor_resample` or as a small utility function (as we might support other sampling methods in the future.

philipc2 (changes requested):

.gitignore
```diff
@@ -150,9 +151,13 @@ test/write_*
...
 docs/user_api/_autosummary/
-
 docs/notebook-examples.txt
-
+docs/*-examples-gallery.txt
+docs/user-guide/oQU120.data.nc
+docs/user-guide/oQU120.grid.nc
+docs/user-guide/oQU480.data.nc
+docs/user-guide/oQU480.grid.nc
+docs/user-guide/psi_healpix.nc
```
erogluorhan, line 160: I don't think we have netCDF files in the docs/user-guide dir, and probably these shouldn't be here. Are these from your local?

.gitignore
```diff
@@ -141,6 +141,7 @@ test/dask-worker-space/
 /dask-worker-space/
 .vscode/
 generated/
+.ipynb_checkpoints/
```
erogluorhan, line 144: Normally an `ipynb_checkpoints` folder shouldn't be auto-generated in the root dir, since we don't have any notebook files there. Probably no need for this.

docs/conf.py
```diff
@@ -78,13 +78,14 @@ def __getattr__(cls, name):
     "iris": ("https://scitools-iris.readthedocs.io/en/latest", None),
     "numpy": ("https://numpy.org/doc/stable", None),
     "scipy": ("https://docs.scipy.org/doc/scipy", None),
-    "numba": ("https://numba.pydata.org/numba-doc/latest", None),
+    "numba": ("https://numba.readthedocs.io/en/latest", None),
```
erogluorhan: Great catch! However, it looks like Numba uses https://numba.readthedocs.io/en/stable, not /latest. Since packages might switch from stable to latest or vice versa over time, why don't we just link to https://numba.readthedocs.io?

docs/conf.py
```diff
@@ -136,7 +137,7 @@ def __getattr__(cls, name):
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This pattern also affects html_static_path and html_extra_path.
-exclude_patterns = []
+exclude_patterns = ["_build", "**.ipynb_checkpoints"]
```
erogluorhan, line 140: Are these necessary?

erogluorhan (commented):
This looks great! Docs looks really good! I only had a few inline comments.

docs/conf.py
zmoon: Well the `_build` one not if you start from clean every time. But yes otherwise myst-nb will find what it stored after executing the last time and essentially execute every notebook twice, I can tell you from experience (and the docs build already takes a while 😅).

<details>
<summary>Demonstrating</summary>

```
...
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/examples/visualization/mpas_topology.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/examples/working-with-mpas-grids.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/getting-started/quick-overview.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/user-guide/advanced-plotting.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/user-guide/advanced-plotting.ipynb: Executed notebook in 0.96 seconds [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/user-guide/area_calc.ipynb: Executing notebook using local CWD [mystnb]
...
/mnt/c/Users/zmoon/git/uxarray/docs/examples/visualization/mpas_topology.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/examples/visualization/mpas_topology.ipynb: Executed notebook in 12.84 seconds [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/examples/working-with-mpas-grids.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/examples/working-with-mpas-grids.ipynb: Executed notebook in 8.26 seconds [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/getting-started/quick-overview.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/getting-started/quick-overview.ipynb: Executed notebook in 11.98 seconds [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/user-guide/advanced-plotting.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/user-guide/advanced-plotting.ipynb: Executed notebook in 1.01 seconds [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/user-guide/area_calc.ipynb: Executing notebook using local CWD [mystnb]
/mnt/c/Users/zmoon/git/uxarray/docs/user-guide/area_calc.ipynb: Executed notebook in 13.59 seconds [mystnb]
...
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/examples/visualization/mpas_topology.ipynb: WARNING: document isn't included in any toctree
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/examples/working-with-mpas-grids.ipynb: WARNING: document isn't included in any toctree
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/getting-started/quick-overview.ipynb: WARNING: document isn't included in any toctree
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/user-guide/advanced-plotting.ipynb: WARNING: document isn't included in any toctree
/mnt/c/Users/zmoon/git/uxarray/docs/_build/jupyter_execute/user-guide/area_calc.ipynb: WARNING: document isn't included in any toctree
...
```
</details>

docs/conf.py
zmoon: Yeah especially with Numba, I think going with stable is better. Or like you suggest, what they choose the default to be is fine too. I can update the rest of them too if you want.

erogluorhan: Sounds good!

docs/conf.py
erogluorhan: Hmm interesting! AFAIK we should be starting from clean every time, but @philipc2 might know that better

.gitignore
zmoon: This syntax (without leading `/`) is to ignore any checkpoints folder, not just at root, just helpful for editing the notebooks.

.gitignore
zmoon: These are generated in the docs build. The culprits:

* https://uxarray.readthedocs.io/en/latest/user-guide/remapping.html#data
* https://uxarray.readthedocs.io/en/latest/user-guide/healpix.html#remapping

.gitignore
erogluorhan: Oops, missed that; sounds good then!

.gitignore
erogluorhan: Oh okay, I wasn't aware as it's been quite a while that I've not done a local docs build. Good then

uxarray/core/dataarray.py
zmoon: I think this a good idea. I can add this. I was trying to leave `_nearest_neighbor_resample` more lower-level-y like it was, so maybe the utility function route.

docs/conf.py
zmoon: IMHO when you are developing it's a little inconvenient to have to remove `_build` every time, this just alleviates the need for that.

zmoon: I also usually use [`nb_execution_mode = "cache"`](https://myst-nb.readthedocs.io/en/stable/computation/execute.html#notebook-execution-modes), which is also helpful, if you want to consider that.

erogluorhan: Yeah, for local docs builds, this makes complete sense.

docs/conf.py
zmoon: I'm reading that it by design won't follow redirects, so you have to choose (but I'll try just in case). If that's the case, stable better?

zmoon: Went to stable in f7708869. I think that is the most useful, since likely most people will be using stable releases of all of these.

uxarray/core/dataarray.py
zmoon: @philipc2 were you okay with that approach I pushed? it's a bit much but could help with expanding to more methods

philipc2: @zmoon 

I think it's great; I can give a detailed review in the morning, but from quickly glancing at it I'm happy with how it looks.

docs/conf.py
erogluorhan: Sorry for the delay! Yes, stable makes sense. Thanks!

philipc2 (approved):
Looks great! Thank you for this contribution.

erogluorhan (approved):
This looks great; thanks a lot for the contribution!

Issue #1216, opened 2025-04-21
Support Face area calculations for sphere radius > 1


philipc2:
### Discussed in https://github.com/UXARRAY/uxarray/discussions/1214

<div type='discussions-op-text'>

<sup>Originally posted by **st-ncar** April 18, 2025</sup>
I work with MPAS data on a sphere of radius $r_E > 1$ and look for ways to compute integrals of quantities defined at the cell centers of the primal mesh. This requires computing the face areas (either explicitly or implicitly as part of `UxDataArray.integrate()`) and I am not sure I fully understand the behavior of `uxarray`. There seem to be at least two ways to obtain the face areas:

- `Grid.compute_face_areas()`: computes the face area from the node coordinates (spherical or Cartesian).
- `Grid.face_areas`: returns `face_areas` if contained in `Grid`, otherwise calls the function above

Intuitively, I would expect that both methods (with either coordinate system) give the same answer. When I tested this, I got a different answer in each case:

1. `Grid.compute_face_areas(latlon=True)` seems to give the face areas on the unit sphere ($r = 1$)
2. `Grid.compute_face_areas(latlon=False)` gives very small face areas (`1e-16` or smaller). I wasn't sure how to interpret this.
3. `Grid.face_areas` seems to give the face area on the sphere with radius $r = r_E$ (perhaps these areas are read from the grid file directly?)

I can rescale the result of method 1 to recover method 3. It also seems that `UxDataArray.integrate()` uses method 1 internally and I have to rescale the result to get the expected integral value on a sphere with radius $r_E$.

I was wondering if someone can help me interpret the result of the 3 methods above (what I wrote are just my guesses). Also, do you have a recommendation for which method I should use?  And in case the above interpretation is correct, would it be possible to make `Grid.compute_face_areas()` aware of `Grid.attrs['sphere_radius']` so that there is no need for rescaling between method 1 and 3?</div>

rajeeja:
#TODO: validation function enhancements

rajeeja:
> @rajeeja 
> 
> Maybe we could consider a `Grid.set_radius()` method that the user can use to set their radius? Though, we'd need to be careful to make sure this doesn't break any of our existing functionality. 

This would need a thorough review of everything and usage of this new 'r'.

Easier option is to just normalize it and have this as a variable for any use feature record

PR #1357, opened 2025-09-04
Indroduce Grid property sphere_radius

rajeeja:
closes #1214 #1216  Add sphere_radius property to Grid class to preserve and access the physical sphere radius from source grids (e.g., Earth's radius in MPAS ocean meshes), enabling proper scaling of results from unit sphere calculations to physical units.


philipc2 (approved):
Nice work on this!

PR #1375, opened 2025-09-16
Preserve Longitude Attributes when Normalizing 

### LINKED ISSUE #1137: Attributes when calling `_set_desired_longitude_range`
Longitude attributes are lost when calling this function.

philipc2:
Closes #1137


rajeeja:
Good catch. Are there other places in the codebase where similar attribute loss might occur (some math op on xarray..)?

philipc2:
> Good catch. Are there other places in the codebase where similar attribute loss might occur (some math op on xarray..)?

I don't believe so, at least not any areas I've noticed.

erogluorhan (approved):
Thanks very much for this fix!

Issue #1236, opened 2025-05-03
Incorrect HEALPix Area Calculation


philipc2:
The area of each HEALPix pixel can be derived using `Ω_pix = 4π / (12 * Nside²) = π / (3 * Nside²)` 

Our internal area calculation does not produce the correct area. From my understanding, I believe this is due to boundaries between pixels not being defined as a typical edge/gca, which throws off our internal calculations. Approximating the boundary as a GCA leads to the area be

```Python
>>> uxgrid = ux.Grid.from_healpix(zoom=0, pixels_only=True)
>>> uxgrid.face_areas

array([0.94349613, 0.94349613, 0.94349613, 0.94349613, 1.2620424 ,
       1.2620424 , 1.2620424 , 1.2620424 , 0.94349613, 0.94349613,
       0.94349613, 0.94349613])
```

The expected area should be `~1.047`

The faces near the equator have a larger area, with those near the poles having a smaller area.

PR #1379, opened 2025-09-23
Fix HEALPix face area calculation to enforce equal area property

rajeeja:
Address https://github.com/UXARRAY/uxarray/issues/1236 and update face area calculation API


uxarray/grid/grid.py
```diff
@@ -1483,10 +1483,30 @@ def face_areas(self) -> xr.DataArray:
         from uxarray.conventions.descriptors import FACE_AREAS_ATTRS, FACE_AREAS_DIMS
 
         if "face_areas" not in self._ds:
-            face_areas, self._face_jacobian = self.compute_face_areas()
-            self._ds["face_areas"] = xr.DataArray(
-                data=face_areas, dims=FACE_AREAS_DIMS, attrs=FACE_AREAS_ATTRS
-            )
+            # Check if this is a HEALPix grid
+            if self._ds.attrs.get("source_grid_spec") == "HEALPix":
```
erogluorhan, line 1492: How about moving the HEALPIx-specific calculations under this case into `io/_healpix.py`?

test/io/test_healpix.py
```diff
@@ -95,6 +95,106 @@ def test_invalid_cells():
     with pytest.raises(ValueError):
         uxda = ux.UxDataset.from_healpix(xrda)
 
+@pytest.mark.parametrize("resolution_level", [0, 2])  # Test lowest and a mid-level
```
erogluorhan, line 98: Maybe this is enough, but what about adding a third one that is a bit higher, level 7 or 9?

rajeeja:
> Implementation looks good except a few inline comments and the following:
> 
> Per today's discussion:
> 
> * Did you have a chance to look into how the current UXarray area calculation behaves between coarse and fine resolutions, compared to the HEALPIx areas at those levels?
> * Could you please have a look into where in the documentation we could make some clear statements about how our geometric representation of HEALPix pixels (i.e. GCAs) differ from their own algorithmic geometry, and how we did this area adjustments because of that?
Screenshot 2025-09-23 at 20 32 31
Resolution 0 (coarse): 20.5% max error, 100% faces affected                                                                                                                                                                                            
Resolution 7 (fine):   10.0% max error, 0.01% faces affected 
Overall we can say that mean errors decrease dramatically, but max errors persist

Based on my analysis of where these max errors occur, I can see 
   1. HEALPix's 12 base pixels (0 resolution)  have fundamentally different shapes
   2. Equatorial pixels: More "square-like" in HEALPix's projection
   3. Mid-latitude pixels: More "diamond-like" or distorted
   4. UXarray's GCA representation doesn't perfectly capture these shape differences

These errors are systematic are related to how different HEALPix pixel types (equatorial vs mid-latitude) are represented using Great Circle Arcs. This is a fundamental geometric representation issue, not a numerical precision problem.

**So, the area property in this PR preserves HEALPix's equal-area property!**

test/io/test_healpix.py
rajeeja: That would be too computationally expensive for testing. How about level 4, or maybe we can skip it.

erogluorhan: Yeah, let's not expand testing. 0 and 2 should suffice then, maybe replace 2 with 4, up to you, but probably not a big difference in anything.

rajeeja: leaving it at 2.

erogluorhan (changes requested):
Reviewing this PR's latest revisions on `healpix.ipynb` made me notice something and further think about how to deliver this area adjustment in the best possible (least confusing?) way:

- With the added parts, the "Working with HEALPix" notebook now discusses Best Practices, i.e. about using the `face_areas` property and avoiding the `compute_face_areas` function when it is a HEALPix.
   - This might easily be overlooked or forgotten though. Then I started to think why we should provide a `compute_face_areas` API even though we have a `face_areas` API already. Why don't we have it only as an internal API?
      - Then I dug into finding where `compute_face_areas` has been used so far. It is used in: 
         - `grid.calculate_total_face_area`, 
         - `dataarray.integrate`, and 
         - `dataset.integrate`. 
      - Also, none of them seems to have been modified in this PR to use `face_areas` to leverage the HEALPix area adjustment.

I'd suggest that:
- We convert `compute_face_areas` into internal API
   - Relatedly, remove its references from the docs, tests, etc.
   - Also relatedly, revise `healpix.ipynb` again to only mention how `face_areas` is adjusted when it is a HEALPix grid (after discussing the systematic difference between UXarray and HEALPix as you already did)
- Turn `compute_face_areas` usages mentioned above into `face_areas` instead
- Likewise, create `compute_healpix_face_areas` as an internal API, not user API

Do all of these make sense to you?

rajeeja:
> Reviewing this PR's latest revisions on `healpix.ipynb` made me notice something and further think about how to deliver this area adjustment in the best possible (least confusing?) way:
> 
> * With the added parts, the "Working with HEALPix" notebook now discusses Best Practices, i.e. about using the `face_areas` property and avoiding the `compute_face_areas` function when it is a HEALPix.
>   
>   * This might easily be overlooked or forgotten by the user though; hence it could lead confusions. Then I started to think why we should provide a `compute_face_areas` API even though we have a `face_areas` API already. Why don't we have it only as an internal API?
>     
>     * Then I dug into finding where `compute_face_areas` has been used so far. It is used in:
>       
>       * `grid.calculate_total_face_area`,
>       * `dataarray.integrate`, and
>       * `dataset.integrate`.
>     * Also, none of them seems to have been modified in this PR to use `face_areas` to leverage the HEALPix area adjustment.
> 
> I'd suggest that:
> 
> * We convert `compute_face_areas` into internal API
>   
>   * Relatedly, remove its references from the docs, tests, etc.
>   * Also relatedly, revise `healpix.ipynb` again to only mention how `face_areas` is adjusted when it is a HEALPix grid (after discussing the systematic difference between UXarray and HEALPix as you already did)
> * Turn `compute_face_areas` usages mentioned above into `face_areas` instead
> * Likewise, create `compute_healpix_face_areas` as an internal API, not user API
> 
> Do these make sense?

Good observation/plan and I agree with most of proposed changes. It did look cumbersome while doing this notebook.
A couple of minor concerns - what if users need to force recomputation of area for some workflow? and maybe some backward compatibiluty issues.

erogluorhan:
> Good observation/plan and I agree with most of proposed changes. It did look cumbersome while doing this notebook. A couple of minor concerns - what if users need to force recomputation of area for some workflow? and maybe some backward compatibiluty issues.

Yeah, that's a good point! 

Thinking out loud: How about making all the suggested changes above in my comment,but adding an admonition into the HEALPix notebook about its workflow possibility and guiding the reader about an internal function (`_compute_face_areas()`) that could be used for that purpose?

rajeeja:
Address #1236

uxarray/core/dataarray.py
```diff
@@ -484,9 +484,14 @@ def integrate(
         >>> integral = uxds["psi"].integrate()
         """
         if self.values.shape[-1] == self.uxgrid.n_face:
-            face_areas, face_jacobian = self.uxgrid.compute_face_areas(
-                quadrature_rule, order
-            )
+            # For default parameters, use cached face_areas property for better performance
```
erogluorhan: Is this related to HEALPix area adjustment (It doesn't look like so)? If not, I don't think we need this if-check here because calling just the `face_areas` will either bring the result of previously called  `_compute_face_areas()` or call it for the first time. 

So, I believe the change here should only be replacement of `compute_face_areas` with `face_areas`

uxarray/core/dataset.py
```diff
@@ -586,10 +586,14 @@ def integrate(self, quadrature_rule="triangular", order=4):
 
         integral = 0.0
 
-        # call function to get area of all the faces as a np array
-        face_areas, face_jacobian = self.uxgrid.compute_face_areas(
-            quadrature_rule, order
-        )
+        # For default parameters, use cached face_areas property for better performance
```
erogluorhan: The same as the above `dataarray` comment

uxarray/grid/grid.py
```diff
@@ -1926,12 +1939,19 @@ def calculate_total_face_area(
         Sum of area of all the faces in the mesh : float
         """
 
-        # call function to get area of all the faces as a np array
-        face_areas, face_jacobian = self.compute_face_areas(
-            quadrature_rule, order, latitude_adjusted_area=latitude_adjusted_area
-        )
-
-        return np.sum(face_areas)
+        # For default parameters, use cached face_areas property for better performance
```
erogluorhan: Similar to the comments in `dataarray` and `dataset`, is this related to the HEALPix area adjustment? If not, just replace `compute_face_areas` use with `grid.face_areas` without any if-check.

uxarray/grid/grid.py
```diff
@@ -1942,6 +1962,10 @@ def compute_face_areas(
         """Face areas calculation function for grid class, calculates area of
         all faces in the grid.
 
+        .. deprecated:: 2025.01.0
```
erogluorhan:
```suggestion
        .. deprecated:: 2025.10.0
```

erogluorhan (approved):
Looks great; thanks a lot for this fix!

PR #1381, opened 2025-09-26
ESMF netCDF4 engine issue found with upstream CI failure

rajeeja:
Fix ESMF file loading by forcing netCDF4 engine [xarray was using scipy netcdf](https://github.com/UXARRAY/uxarray/actions/runs/18024951189/job/51349778834#step:8:967)

 - Modified uxarray/core/api.py to set engine='netcdf4' for .nc files

### DIFF
--- a/uxarray/core/utils.py
+++ b/uxarray/core/utils.py
@@ -28,21 +28,14 @@ def _open_dataset_with_fallback(filename_or_obj, chunks=None, **kwargs):
     xr.Dataset
         The opened dataset
     """
-    # First attempt: use xarray's default read engine
-    original_error = None
     try:
+        # Try opening with xarray's default read engine
         return xr.open_dataset(filename_or_obj, chunks=chunks, **kwargs)
-    except Exception as e:
-        original_error = e
-
-    # Second attempt: fallback to netcdf4 engine
-    try:
+    except Exception:
+        # If it fails, use the "netcdf4" engine as backup
         # Extract engine from kwargs to prevent duplicate parameter error
         engine = kwargs.pop("engine", "netcdf4")
         return xr.open_dataset(filename_or_obj, engine=engine, chunks=chunks, **kwargs)
-    except Exception:
-        # If both attempts fail, raise the original error to preserve behavior
-        raise original_error
 
 
 def _map_dims_to_ugrid(
--- a/test/grid/grid/test_initialization.py
+++ b/test/grid/grid/test_initialization.py
@@ -49,7 +49,7 @@ def test_grid_init_verts_fill_values():
 def test_read_shpfile(test_data_dir):
     """Reads a shape file and write ugrid file."""
     shp_filename = test_data_dir / "shp" / "grid_fire.shp"
-    with pytest.raises(ValueError):
+    with pytest.raises((ValueError, FileNotFoundError, OSError)):
         grid_shp = ux.open_grid(shp_filename)
 
 


erogluorhan (commented):
The problem seems to be related to some files only (which is probably not h5 compatible). Instead of the proposed solution here, doing something like the following makes more sense to me:

```
# Attempt to use Xarray directly for remaining input types
try:
   # Try opening with xarray's default read engine, which may be "h5netcdf" or "scipy" after v2025.09.0
   grid_ds = xr.open_dataset(grid_filename_or_obj, chunks=grid_chunks, **kwargs)
except:
   # If it fails (in most cases, for some files only), use the "netcdf4" engine as backup
   grid_ds = xr.open_dataset(grid_filename_or_obj, engine="netcdf4", chunks=grid_chunks, **kwargs)

grid = Grid.from_dataset(grid_ds, use_dual=use_dual)
```

This way, we can make use of faster engines for many files but only back up to  "netcdf4" whenever needed. 

If this makes sense to you, please use the same mechanisms in the following code locations as well (Should we consider writing a utility function for all of them?):
- `api.py`: `open_dataset()` - L193
- `grid.py`: `from_file()` - L351
- `dataset.py`: `from_healpix()` - L329
- `utils.py`: `match_chunks_to_ugrid()` - L72

rajeeja:
> The problem seems to be related to some files only (which is probably not h5 compatible). Instead of the proposed solution here, doing something like the following makes more sense to me:
> 
> ```
> # Attempt to use Xarray directly for remaining input types
> try:
>    # Try opening with xarray's default read engine, which may be "h5netcdf" or "scipy" after v2025.09.0
>    grid_ds = xr.open_dataset(grid_filename_or_obj, chunks=grid_chunks, **kwargs)
> except:
>    # If it fails (in most cases, for some files only), use the "netcdf4" engine as backup
>    grid_ds = xr.open_dataset(grid_filename_or_obj, engine="netcdf4", chunks=grid_chunks, **kwargs)
> 
> grid = Grid.from_dataset(grid_ds, use_dual=use_dual)
> ```
> 
> This way, we can make use of faster engines for many files but only back up to "netcdf4" whenever needed.
> 
> If this makes sense to you, please use the same mechanisms in the following code locations as well (Should we consider writing a utility function for all of them?):
> 
> * `api.py`: `open_dataset()` - L193
> * `grid.py`: `from_file()` - L351
> * `dataset.py`: `from_healpix()` - L329
> * `utils.py`: `match_chunks_to_ugrid()` - L72

+1 how about a utility function `open_dataset_with_fallback`? say in `uxarray/core/api.py` this might be the better approach for maintainability.

erogluorhan:
> +1 how about a utility function `open_dataset_with_fallback`? say in `uxarray/core/api.py` this might be the better approach for maintainability.

Sounds good! Please make it internal.

uxarray/core/utils.py
```diff
@@ -3,6 +3,41 @@
...
+    xr.Dataset
+        The opened dataset
+    """
+    try:
+        # Try opening with xarray's default read engine
+        return xr.open_dataset(filename_or_obj, chunks=chunks, **kwargs)
+    except Exception:
+        # If it fails, use the "netcdf4" engine as backup
+        return xr.open_dataset(
+            filename_or_obj, engine="netcdf4", chunks=chunks, **kwargs
```
philipc2: Add a safeguard for the case where `engine` may be provided within the `kwargs` to prevent duplicate parameters.

erogluorhan: Great catch!

uxarray/core/api.py
```diff
@@ -107,6 +111,11 @@ def open_grid(
 
     else:
         # Attempt to use Xarray directly for remaining input types
+        # Force netCDF4 engine for .nc files to avoid scipy reader issues
+        if isinstance(grid_filename_or_obj, (str, os.PathLike)) and str(
+            grid_filename_or_obj
+        ).endswith(".nc"):
+            kwargs.setdefault("engine", "netcdf4")
         grid_ds = xr.open_dataset(grid_filename_or_obj, chunks=grid_chunks, **kwargs)
```
erogluorhan: From Line 114 through Line 119 should be replaced with the brand new utility function:

`grid_ds = _open_dataset_with_fallback(grid_filename_or_obj, chunks=grid_chunks, **kwargs)`

erogluorhan (commented):
Please address a single comment below, then this should be good to go.

PR #1384, opened 2025-10-04
DRAFT: Use contiguous arrays for point containment 

### LINKED ISSUE #1358: Use contiguous arrays for point containment checks
I was also curious about this warning that surfaces to the user:
```
uxarray/grid/point_in_face.py:168: NumbaPerformanceWarning: np.dot() is faster on contiguous arrays, called on (Array(float64, 1, 'C', False, aligned=True), Array(float64, 1, 'A', False, aligned=True))
  hits = _get_faces_containing_point(
```
Did you look at trying to address that? If you decide(d) it's not important or whatever, maybe we can catch and filter it so the user doesn't see it.

_Originally posted by @zmoon in https://github.com/UXARRAY/uxarray/issues/1354#issuecomment-3254500331_

philipc2:
Closes #1358


cmdupuis3:
Per #1358, this fix doesn't really address the underlying problem. Closing for now.

PR #1387, opened 2025-10-10
Divergence 

### LINKED ISSUE #536: Add UxDataArray.divergence()
Compute the divergence of the vector field defined by a UxDataArray and other.

`UxDataArray UxDataArray.divergence(self, UxDataArray other)`

rajeeja:
Closes #536


uxarray/core/dataarray.py
```diff
@@ -1336,6 +1336,87 @@ def gradient(self, **kwargs) -> UxDataset:
...
+        # Compute gradients of both components
+        u_gradient = self.gradient()
+        v_gradient = other.gradient()
+
+        # For divergence: div(V) = ∂u/∂x + ∂v/∂y
+        # In spherical coordinates: div(V) = (1/cos(lat)) * ∂u/∂lon + ∂v/∂lat
+        # We use the zonal gradient (∂/∂lon) of u and meridional gradient (∂/∂lat) of v
+        divergence_values = (
+            u_gradient["zonal_gradient"].values
+            + v_gradient["meridional_gradient"].values
```
erogluorhan: Do we need to extract the numpy arrays out of those `xarray.DataArray`s?

test/core/test_vector_calculus.py
```diff
@@ -0,0 +1,167 @@
+import numpy as np
+import pytest
+
+import uxarray as ux
+import numpy.testing as nt
+
```
erogluorhan, line 6: What do you think about putting divergence cases into `test_gradient.py` (instead of creating a new script)?

rajeeja: Good idea, how about test_vector_calculus.py and put it all there?

erogluorhan: Yeah that makes sense; thanks!

uxarray/core/dataarray.py
rajeeja: Thanks — good catch. We should avoid using .values  this forces eager computation, drops xarray coords/attrs and breaks dask laziness.

PR #1388, opened 2025-10-13
Edge plotting method that extracts visible edges automatically

nils3er:
## Overview

- Introduces `UxGrid.plot_edges`
- extracts visible edges for the extract in an axis.
  - not exact. A edge is plotted if one of the nodes is in the extent.
- Adds a test
- Adds an example to the user guide

## Expected Usage

```Python
import uxarray as ux

center = -50, 0
fig, ax = plt.subplots(
    1,
    1,
    constrained_layout=True,
    subplot_kw={"projection": ccrs.Orthographic(*center)},
)
ax.add_feature(cfeature.LAND)
ax.add_feature(cfeature.COASTLINE)
ax.set_extent([center[0]+20, center[0]-20, center[1]-10, center[1]+10])

uxgrid.plot_edges(ax=ax, color="blue")
plt.show()

```

## PR Checklist

**General**
- [ ] An issue is linked created and linked
- [ ] Add appropriate labels

**Testing**
  - Could be improved with a local axis

**Documentation**
- ~~Docstrings have updated with any function changes~~
- ~~Internal functions have a preceding underscore (`_`) and have been added to `docs/internal_api/index.rst`~~
- [X] User functions have been added to `docs/api.rst`

**Examples**


rajeeja:
> I'm open for discussions here and of course the API needs to be streamlined with the existing one.


You are right, hvplot's "create its own figure" and that approach doesn't fit the YAC workflow you describe. Thanks for the detailed info. I have verified and indeed the speedup is significant. How about we enhance the main grid.plot.edges() interface to automatically detect matplotlib+cartopy contexts and route to your optimized method behind the scenes? Maybe this is also a chance for us to look at a broader plotting ways in other places where we default to hvplot.@erogluorhan

uxarray/grid/grid.py
```diff
@@ -2378,6 +2378,55 @@ def to_linecollection(
 
         return copy.deepcopy(line_collection)
 
+    def plot_edges(
+        self,
+        ax: Optional["cartopy.mpl.geoaxes.GeoAxes"] | None,  # noqa: F821
+        **plot_kwargs,
+    ):
```
philipc2, line 2385: As mentioned by @erogluorhan, if we could design this function similar to how we set up `to_raster()`, that would be preferred to avoid conflicting names with our `.plot` accessors.

`to_raster` also doesn't directly call any `ax.plot` routines. To be consistent with that, having this method return something that can then be externally passed into `ax.plot` would work. 

Let me know if you have any ideas.

philipc2 (changes requested):

uxarray/grid/grid.py
erogluorhan: Maybe something like that, or I'd also be open to discussing whether to support matplotlib plotting functions from our plotting API with a clear distinction that they are not the default `hvplot` functions. Thinking out loud: Maybe through a newer accessor such as `ux.mpl_plot`, or through the existing one with clear function namings such as `ux.plot.mpl_edges()`, or something else. I am curious what you all would think about these.

rajeeja (changes requested):
matplotlib native support and seperate from hvplot sounds good and clearer and more maintainable.
@philipc2 how about `ux.plot.mpl_edges()` as @erogluorhan suggests above?

uxarray/grid/grid.py
erogluorhan: However, I should admit that there might be some confusion with an `ux.mpl_plot` accessor or `ux.plot.mpl_vis_funcs()` because our current plotting functions through `ux.plot` accessor provide `backend` selection that enables `hvplot` to use either matplotlib or bokeh as the plotting engine.

erogluorhan: So, we will need to think about all of these

rajeeja:
@erogluorhan what should we do about this PR?

PR #1389, opened 2025-10-20
Curl

### LINKED ISSUE #935: Add Curl
Add support for computing the finite-volume curl at each vertices using a vector on the cell edges. 

An MPAS specific example can be found [here ](https://github.com/theweathermanda/MPAS_utilities/blob/ab88e21400089eb4a6d9f3416962e0f449103e8c/mpas_calc_operators.py#L377)

### LINKED ISSUE #1300: Vector Calculus Operators
# Overview

This parent issue encompasses the development of vector calculus operators (gradient, divergence, curl, etc.).

rajeeja:
Closes #935. Closes #1300.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -71,12 +71,6 @@ suppress_warnings = [
 # jupyter_execute_notebooks = "off"
 # nb_execution_mode = "off"
 
-# Disable notebook execution on Read the Docs to avoid missing kernel errors
-on_rtd = os.environ.get("READTHEDOCS") == "True"
-if on_rtd:
-    jupyter_execute_notebooks = "off"
-    nb_execution_mode = "off"
-    nbsphinx_execute = "never"
 
 extlinks = {
     "issue": ("https://github.com/uxarray/uxarray/issues/%s", "GH"),
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -67,7 +67,7 @@ These user guides provide detailed explanations of the core functionality in UXa
 `Weighted Mean <user-guide/weighted_mean.ipynb>`_
  Compute the weighted average
 
-`Vector Calculations <user-guide/vector_calculations.ipynb>`_
+`Vector Calculations <user-guide/vector_calculus.ipynb>`_
  Compute vector field operations such as gradients, curl, and divergence
 
 `Tree Structures <user-guide/tree_structures.ipynb>`_
--- a/docs/user-guide/vector_calculus.ipynb
+++ b/docs/user-guide/vector_calculus.ipynb
@@ -765,9 +765,9 @@
  ],
  "metadata": {
   "kernelspec": {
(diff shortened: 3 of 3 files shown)


erogluorhan (approved):
This looks great; thank you! Only thing I have to say is that can we come up with a simple visualization example for Curl and Divergence, maybe using the gradient data that gradient user guide used before?

rajeeja:
> This looks great; thank you! Only thing I have to say is that can we come up with a simple visualization example for Curl and Divergence, maybe using the gradient data that gradient user guide used before?

How about in this PR make only one notebook `vector_calculus.ipynb ` combine gradients/div/curl?

docs/userguide.rst
```diff
@@ -121,6 +124,7 @@ These user guides provide additional details about specific features in UXarray.
    user-guide/topological-aggregations.ipynb
    user-guide/weighted_mean.ipynb
    user-guide/gradients.ipynb
+   user-guide/curl.ipynb
    user-guide/divergence.ipynb
```
erogluorhan: With the newer single notebook, you will need to replace these three.

erogluorhan: ... and also their description above (Lines 70 through 74)

docs/userguide.rst
```diff
@@ -67,14 +67,8 @@ These user guides provide detailed explanations of the core functionality in UXa
...
 
-`Gradients <user-guide/gradients.ipynb>`_
- Compute the gradient of a data variable
-
-`Curl <user-guide/curl.ipynb>`_
- Compute the curl of a vector field
-
-`Divergence <user-guide/divergence.ipynb>`_
- Compute the divergence of a vector field
+`Vector Calculations <user-guide/vector_calculations.ipynb>`_
```
erogluorhan:
```suggestion
`Vector Calculations <user-guide/vector_calculus.ipynb>`_

File: docs/conf.py [hunk 3]
```diff
@@ -71,6 +71,12 @@ def __getattr__(cls, name):
 # jupyter_execute_notebooks = "off"
 # nb_execution_mode = "off"
 
+# Disable notebook execution on Read the Docs to avoid missing kernel errors
+on_rtd = os.environ.get("READTHEDOCS") == "True"
+if on_rtd:
+    jupyter_execute_notebooks = "off"
+    nb_execution_mode = "off"
+    nbsphinx_execute = "never"
```
erogluorhan: I think we should avoid doing this because:
- RTD executions are important for inspecting the notebook execution results for any PRs, e.g. plots etc. 
- Also the actual docs creation is done in the "READTHEDOCS" environment, so notebook executions  should always be ON in this env

erogluorhan: Also, if the devs are pushing their notebook environment metadata (jupyter kernelspec) to the remote, having these executions fail acts as a safeguard

PR #1397, opened 2025-10-30
Add a link checker into CI

### LINKED ISSUE #1395: CI needs a link checker addition
I've caught a broken link in our docs, it should have been there for a long while, and there might be others that we are missing. The CI tests should be able to catch such links easily.  

We should add a link checker into our CI workflow, see [here](https://github.com/NCAR/geocat-comp/blob/3abb43580e354d73d2d1a16218ea2861be4f4da8/.github/workflows/ci.yml#L114) as an example.

erogluorhan:
Closes #1395 

## Overview
Adds a link checker into the CI workflow.

Also fix a broken URL in one of the notebooks.

## PR Checklist

**General**
- [x] Add appropriate labels
- [x] Filled out Overview

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
         - name: Make docs with linkcheck
           uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # v3
           with:
-            timeout_minutes: 15
+            timeout_minutes: 20
             max_attempts: 3
             command: |
               eval "$(micromamba shell hook --shell bash)"


.github/workflows/ci.yml
```diff
@@ -62,7 +62,57 @@ jobs:
...
+          conda list
+
+      - name: Make docs with linkcheck
+        uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08 # v3
+        with:
+          timeout_minutes: 60
+          max_attempts: 3
+          command: |
+            eval "$(micromamba shell hook --shell bash)"
+            micromamba activate
```
rajeeja: Is this needed? 
You're running micromamba activate twice without specifying an environment the first time. Remove it.
```suggestion
```

Also, the eval .. line might be redundant, as at the top bash shell is already specified.

rajeeja (changes requested):
I worry the timeout of 60 min is too long?

.github/workflows/ci.yml
erogluorhan: Oh, thanks for catching that! Also regarding `eval` you are probably right; I haven't removed it yet but we can.

rajeeja (approved):
The change is the great! Thank you so much!

PR #1408, opened 2025-11-20
Fix .sel() wrapper when  a slice indexer and method are used together

### LINKED ISSUE #1406: `.sel(time=slice(...))` fails
### Version

2025.10.0

Source

### What happened?

Trying to slice a UxDataset fails:
```
snowcast_2017 = ux.open_mfdataset('mesh_snowcast.nc',
                                  'output_2017/output_2017.nc').sel(time=slice("2018-04-28T00:00:00","2018-04-28T02:00:00"))
```

```
NotImplementedError: cannot use ``method`` argument if any indexers are slice objects
```

Which I suspect is somewhat self describing! However,`.sel` works as expected without a slice

It select a slice of times, given a single time sel works

```python

```

rajeeja:
Fixes #1406

### DIFF
--- a/test/core/test_api.py
+++ b/test/core/test_api.py
@@ -36,17 +36,21 @@ def test_open_dataset(gridpath, datasetpath, mesh_constants):
     nt.assert_equal(len(uxds_var2_ne30.uxgrid._ds.data_vars), mesh_constants['DATAVARS_outCSne30'])
     nt.assert_equal(uxds_var2_ne30.source_datasets, str(data_path))
 
-def test_open_mf_dataset(gridpath, test_data_dir, mesh_constants):
+def test_open_mf_dataset(gridpath, datasetpath, mesh_constants):
     """Loads multiple datasets with their grid topology file using
     uxarray's open_dataset call."""
 
     grid_path = gridpath("ugrid", "outCSne30", "outCSne30.ug")
-    dsfiles_mf_ne30 = str(test_data_dir) + "/ugrid/outCSne30/outCSne30_v*.nc"
+    dsfiles_mf_ne30 = datasetpath(
+        "ugrid",
+        "outCSne30",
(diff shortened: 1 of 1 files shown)


test/core/test_api.py
```diff
@@ -41,7 +41,7 @@ def test_open_mf_dataset(gridpath, test_data_dir, mesh_constants):
     uxarray's open_dataset call."""
 
     grid_path = gridpath("ugrid", "outCSne30", "outCSne30.ug")
-    dsfiles_mf_ne30 = str(test_data_dir) + "/ugrid/outCSne30/outCSne30_*.nc"
+    dsfiles_mf_ne30 = str(test_data_dir) + "/ugrid/outCSne30/outCSne30_v*.nc"
```
erogluorhan: Could this use `dataset_path` fixture or is it not possible?

test/meshfiles/ugrid/outCSne30/README.md
```diff
@@ -0,0 +1,11 @@
+# outCSne30 UGRID assets
```
erogluorhan, line 1: This very much for adding this! It will be helpful in the future

erogluorhan (approved):
There is a single small comment I have below, but it doesn't stop this; so I've approved it. Thanks very much for the fast solution!

test/core/test_api.py
rajeeja: I thought about that also, fixture doesn’t have a way to emit a glob or directory—only actual files—so we’d have to enumerate each NC file in that folder manually. That would make the test more brittle than the current “grab all var/vortex snapshots” behavior. So sticking with the explicit glob via test_data_dir is the simplest/most flexible option.

erogluorhan: Thanks for the info! Yeah, no need for that then!

rajeeja: wait scratch that, I can get the explicit files as that will be more clearer - thanks for the feedback.

PR #1420, opened 2025-12-08
to_raster: handle unset extent; accept singleton extra dims

### LINKED ISSUE #1393: to_raster producing all the same value
### Version

2025.9.0

Conda

### What happened?

I'm having a problem where `to_raster` is just returning an array of all the same value. I'm not sure if there's something weird about this dataset or what. Here's my example:

```python
b_loc = Path("/glade/work/brianpm/trbelt_derived") / "te_mcs_FLUT_blobs_20250908"
b_file = b_loc / "te_stitchblobs.nc"
grid_file = "/glade/work/brianpm/VRM2112/ne0np4.trbelta.ne30x8/grids/trbelta_ne30x8_np4_MESH.nc"
b_ds = ux.open_dataset(grid_file, b_file)
insta = b_ds['MCS'].isel(time=2468).compute()

# `plot` WORKS:
insta.plot()

# `to_raster` DOES NOT WORK:
proj = ccrs.EckertIV()

fig, ax = plt.subplots(subplot_kw={"projection":proj})
raster, pxmap = insta.to_raster(ax, return_pixel_mapping=True)
ax.set_global()
ax.coastlines()
ax.imshow(raster,origin="lower", extent=ax.get_xlim() + ax.get_ylim())

# `to_polycollection` WORKS:
poly_collection = insta.to_polycollection(rasterized=True)
poly_collection.set_antialiased(False)

fig, ax = plt.subplots(
    1,
    1,
    facecolor="w",
    constrained_layout=True,
    subplot_kw=dict(projection=proj),
)

ax.coastlines()

ax.add_collection(poly_collection)
ax.set_global()
plt.title("PolyCollection")
```
Here are the results:
Image

Image

Image

Expecting a similar result as to_polycollection (but faster and producing a smaller pdf file when saved).

```python

```

rajeeja:
Thanks to @brianpm for reporting #1393. This PR fixes #1393 that axes extent must be set before calling to_raster, and fixes the flat-output issue by auto-detecting default axes limits and setting a reasonable extent from grid lon/lat bounds when needed (with a warning). It also adds support for data with extra length-1 dimensions by allowing singleton dims and squeezing to (n_face,) before rasterization.

### DIFF
--- a/uxarray/core/dataarray.py
+++ b/uxarray/core/dataarray.py
@@ -406,37 +406,21 @@ class UxDataArray(xr.DataArray):
         else:
 
             def _is_default_extent() -> bool:
-                # Default extents can be (0, 1) or projection limits while autoscale stays on.
+                # Default extents are indicated by xlim/ylim being (0, 1)
+                # when autoscale is still on (no extent has been explicitly set)
                 if not ax.get_autoscale_on():
                     return False
-                try:
-                    import cartopy.crs as ccrs
-
-                    extent = ax.get_extent(ccrs.PlateCarree())
-                except Exception:
-                    xlim, ylim = ax.get_xlim(), ax.get_ylim()
-                    return np.allclose(xlim, (0.0, 1.0)) and np.allclose(
-                        ylim, (0.0, 1.0)
-                    )
-                return np.allclose(extent, (-180.0, 180.0, -90.0, 90.0), atol=1.0)
+                xlim, ylim = ax.get_xlim(), ax.get_ylim()
+                return np.allclose(xlim, (0.0, 1.0)) and np.allclose(ylim, (0.0, 1.0))
 
             if _is_default_extent():
                 try:
                     import cartopy.crs as ccrs
 
-                    lon_min = self.uxgrid.node_lon.min(skipna=True)
-                    lon_max = self.uxgrid.node_lon.max(skipna=True)
-                    lat_min = self.uxgrid.node_lat.min(skipna=True)
-                    lat_max = self.uxgrid.node_lat.max(skipna=True)
-
-                    lon_min, lon_max = (
-                        float(lon_min.to_numpy().item()),
-                        float(lon_max.to_numpy().item()),
-                    )
-                    lat_min, lat_max = (
-                        float(lat_min.to_numpy().item()),
-                        float(lat_max.to_numpy().item()),
-                    )
+                    lon_min = float(self.uxgrid.node_lon.min(skipna=True).values)
+                    lon_max = float(self.uxgrid.node_lon.max(skipna=True).values)
+                    lat_min = float(self.uxgrid.node_lat.min(skipna=True).values)
+                    lat_max = float(self.uxgrid.node_lat.max(skipna=True).values)
                     ax.set_extent(
                         (lon_min, lon_max, lat_min, lat_max),
                         crs=ccrs.PlateCarree(),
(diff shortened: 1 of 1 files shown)


rajeeja:
@erogluorhan We can auto-set bounds just for the rasterization (or via an extent kwarg) so users still get a meaningful image without having to call set_global(). What do you think? 

xarray’s .plot() sets xlim/ylim from coord..

uxarray/core/dataarray.py
```diff
@@ -375,11 +375,29 @@ def to_raster(
...
 
+        auto_extent_applied = False
+        if pixel_mapping is None:
+            xlim, ylim = ax.get_xlim(), ax.get_ylim()
+            if np.allclose(xlim, (0.0, 1.0)) and np.allclose(ylim, (0.0, 1.0)):
+                try:
+                    import cartopy.crs as ccrs
+
+                    lon = self.uxgrid.node_lon.values
+                    lat = self.uxgrid.node_lat.values
```
erogluorhan: Please avoid converting the whole node coords to numpy, i.e. only convert the final scalars after finding min -max via xarray objects

uxarray/core/dataarray.py
```diff
@@ -375,11 +375,29 @@ def to_raster(
...
         )
 
-        _ensure_dimensions(self)
+        data = _ensure_dimensions(self)
 
         if not isinstance(ax, GeoAxes):
             raise TypeError("`ax` must be an instance of cartopy.mpl.geoaxes.GeoAxes")
 
+        auto_extent_applied = False
+        if pixel_mapping is None:
```
erogluorhan: Having this if-check separate than and before the existing if-check, i.e. `if pixel_mapping is not None:` on now-Line405, that controlled the main flow of this functions makes readability a bit harder honestly. How about combining them together as follows:

```
if pixel_mapping is not None:
   # existing implementation
else:
   # move implementation under your newer if-check here with slight tweaks like below
   # Discard auto_extent_applied = False variable as we are handling everyhing in a single if-check
   if-check for xlim and ylim
   try:
      auto-extent code here

      warn(
               "Axes extent was default; auto-setting from grid lon/lat bounds for rasterization. "
                "Set the extent explicitly to control this.",
                stacklevel=2,
            )
```

uxarray/core/dataarray.py
```diff
@@ -403,9 +421,15 @@ def to_raster(
                         + input_ax_attrs._value_comparison_message(pm_ax_attrs)
                     )
             pixel_mapping = np.asarray(pixel_mapping, dtype=INT_DTYPE)
+        elif auto_extent_applied:
+            warn(
+                "Axes extent was default; auto-setting from grid lon/lat bounds for rasterization. "
+                "Set the extent explicitly to control this.",
+                stacklevel=2,
```
erogluorhan: This warning is very helpful! Maybe just one thing to add to make it even further useful: Elaborate on how to set extent, i.e. ""Set the extent explicitly to control this, e.g. by calling either of matplotlib's `ax.get_global()`, or `ax.set_extent()`, or `ax.set_xlim()` & `ax.set_ylim()` pair.""

test/test_plot.py
```diff
@@ -108,6 +108,22 @@ def test_to_raster(gridpath):
...
+        subplot_kw={'projection': ccrs.Robinson()},
+        constrained_layout=True,
+        figsize=(10, 5),
+    )
+
+    mesh_path = gridpath("mpas", "QU", "oQU480.231010.nc")
+    uxds = ux.open_dataset(mesh_path, mesh_path)
+
+    da = uxds['bottomDepth'].expand_dims(time=[0])
+    raster = da.isel(time=slice(0, 1)).to_raster(ax=ax)
```
erogluorhan: In this test case specifically, since you are expanding time dim to length 1 only, I think `isel(time=slice(0, 1))` on `da` is unnecessary, but I get that if time dim was longer than 1, it'd make a difference. 

Leaving it up to you if you want to test that.

erogluorhan (commented):
This is looking great and almost there; please see my comments below.

erogluorhan (approved):
Looks good to me! Please fix the link-check and merge

PR #1429, opened 2025-12-15
Add groups configuration for Dependabot

kafitzgerald:
## Overview
Adds group configuration for Dependabot to minimize the number of PRs generated (in case this might be of interest).

We have this configured on most of the GeoCAT repos now and it's pretty nice.


erogluorhan (approved):
This is great; thanks a lot for adding!

PR #1430, opened 2025-12-16
Fix ICON Reader 

### LINKED ISSUE #1425: Naming conflicts on cross_section for ICON grids
### Version

10000.dev2634+g5cdf333b9.d20251212

Source

### What happened?

`cross_section` on ICON-gridded data throws `CoordinateValidationError: coordinate clon has dimensions ('n_face',), but these are not a subset of the DataArray dimensions ('time', 'height', 'steps')`.

The `cross_section` operator should retrieve the data along a given GCA without rising an error. 

Data for reproducing the test can be found here: https://kumulus.iap-kborn.de/index.php/s/qHiQFKzoA5Jwn2X

```python
import uxarray as ux

data_path = "uvwtemp_3d_ML_19790124T000000Z.nc"
grid_path = "icon_grid_0013_R02B04_G.nc"

cross_section_gca = uxds.temp.cross_section(start=(0, -90), end=(0., 90))
```

rajeeja:
- _icon_to_ugrid_dims now derives the mapping from the actual ICON coords and connectivity (including nv→n_max_face_nodes, nc→two, etc.), and falls back to common names.
  - With the full map, _map_dims_to_ugrid cleanly swaps all ICON dims to UGRID for both grid and data, so face coords are consistently normalized. When cross_section rewrites the lon/lat coord keys to the new steps dimension, there’s no leftover clon/clat tied to n_face, and validation passes.

  This makes ICON I/O robust across variants and preventing future coord/dim mismatches.

Fixes #1425 

@tdoerff  please verify.

### DIFF
--- a/test/io/test_icon.py
+++ b/test/io/test_icon.py
@@ -8,3 +8,17 @@ def test_read_icon_grid(gridpath):
 def test_read_icon_dataset(gridpath):
     grid_path = gridpath("icon", "R02B04", "icon_grid_0010_R02B04_G.nc")
     uxds = ux.open_dataset(grid_path, grid_path)
+
+def test_icon_cross_section(gridpath):
+    grid_path = gridpath("icon", "R02B04", "icon_grid_0010_R02B04_G.nc")
+    uxds = ux.open_dataset(grid_path, grid_path)
+
+    # Test cross_section with cell_area variable
+    result = uxds.cell_area.cross_section(start=(0, -90), end=(0, 90))
+    assert result is not None
+    assert len(result) == 100  # 100 steps by default
+    assert result.dims == ('steps',)
+    assert all(result.coords['lon'] == 0.0)  # Constant longitude
+    assert result.coords['lat'].min() == -90.0
+    assert result.coords['lat'].max() == 90.0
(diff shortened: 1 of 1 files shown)


uxarray/cross_sections/dataarray_accessor.py
```diff
@@ -132,7 +132,11 @@ def __call__(
         data = np.moveaxis(filled, -1, dim_axis)
 
         # Build coords dict: keep everything except 'n_face'
-        coords = {d: self.uxda.coords[d] for d in self.uxda.coords if d != "n_face"}
+        coords = {
+            name: self.uxda.coords[name]
+            for name in self.uxda.coords
+            if name != "n_face" and "n_face" not in self.uxda.coords[name].dims
```
erogluorhan, line 138: This addition was needed here because coords like `clon` had a dimension of "n_face" even though they were not named "n_face" themselves, which was the main cause of failure here. 

@tdoerff , your PR helped me figure this fix out, please let us know if this makes sense to you (I've tested it with your data and another ICON grid, and it seems working)

tdoerff: Yes, this version works, thanks a lot!

PR #1437, opened 2026-01-07
Fix bug in _xyz_to_lonlat_rad

### LINKED ISSUE #1438: Fix test_io AssertionError
```
E           AssertionError:
E           Not equal to tolerance rtol=1e-08, atol=0
E           Exodus longitude mismatch for RLL1deg
E           Mismatched elements: 63903 / 64442 (99.2%)
E           Max absolute difference among violations: 192.78769181
E           Max relative difference among violations: 10254.862154
E            ACTUAL: array([ 0.,  0.,  1., ..., -2., -1.,  0.], shape=(64442,))
E            DESIRED: array([0.      , 0.      , 0.017457, ..., 0.06987 , 0.017457, 0.      ],
E                 shape=(64442,))

test/grid/grid/test_io.py:148: AssertionError
```

The error points to  _xyz_to_lonlat_rad function producing incorrect results despite correct inputs, suspecting subtle in-place modifications during array operations—possibly from numpy's handling of expressions like x * x.  Using python 3.14 and numpy 2.26

rajeeja:
Fix #1438 _xyz_to_lonlat_rad normalization to avoid NumPy 2.x aliasing, restoring correct lon/lat in Exodus round‑trip (test failed: test/grid/grid/test_io.py::test_grid_ugrid_exodus_roundtrip) on Python 3.14.

### DIFF
--- a/uxarray/grid/coordinates.py
+++ b/uxarray/grid/coordinates.py
@@ -68,10 +68,6 @@ def _xyz_to_lonlat_rad_no_norm(
 def _xyz_to_lonlat_rad_scalar(x, y, z, normalize=True):
     if normalize:
         x, y, z = _normalize_xyz_scalar(x, y, z)
-        denom = math.hypot(x, y, z)
-        x = x / denom
-        y = y / denom
-        z = z / denom
 
     lon = np.arctan2(y, x)
     lat = np.asin(z)
@@ -121,10 +117,6 @@ def _xyz_to_lonlat_rad(
 
     if normalize:
         x, y, z = _normalize_xyz(x, y, z)
-        denom = np.sqrt(x * x + y * y + z * z)
-        x = x / denom
-        y = y / denom
-        z = z / denom
 
     lon = np.arctan2(y, x)
     lat = np.arcsin(z)


uxarray/grid/coordinates.py
```diff
@@ -68,10 +68,10 @@ def _xyz_to_lonlat_rad_no_norm(
 def _xyz_to_lonlat_rad_scalar(x, y, z, normalize=True):
     if normalize:
         x, y, z = _normalize_xyz_scalar(x, y, z)
-        denom = abs(x * x + y * y + z * z)
-        x /= denom
-        y /= denom
-        z /= denom
+        denom = math.hypot(math.hypot(x, y), z)
```
erogluorhan: So, you are no longer using sum of squares, but instead the hypotenuse, i.e. square root of sum of  squares for `denom`? Please explain.

Also, I think you'd be fine with calling this line as `math.hypot(x, y, z)`, no need for two func calls

rajeeja: the intent is to normalize to a unit vector, so the denominator should be the Euclidean norm (√(x²+y²+z²)), not the squared norm. Dividing by the squared norm would shrink the vector rather than normalize it, leaving z potentially outside [-1, 1] before the asin call. 

Correct — math.hypot(math.hypot(x, y), z) and math.hypot(x, y, z) are mathematically identical (both compute √(x²+y²+z²)), so the result is the same - I'll update it.

erogluorhan: So does it mean that the code here was wrong so far (since `denom` was not equal to the vector norm, instead the squared norm)?

Also, isn't the normalization by vector norm being performed within the `_normalize_xyz(x, y, z)` or `_normalize_xyz_scalar(x, y, z)` call already here?

rajeeja: Yes — the old `denom` logic was redundant here and also incorrect as a normalization factor (`x*x + y*y + z*z `is the squared norm), so I’m removing that extra re-normalization block entirely since _normalize_xyz*() already does the unit-vector normalization.

uxarray/grid/coordinates.py
```diff
@@ -121,10 +121,10 @@ def _xyz_to_lonlat_rad(
 
     if normalize:
         x, y, z = _normalize_xyz(x, y, z)
-        denom = np.abs(x * x + y * y + z * z)
-        x /= denom
-        y /= denom
-        z /= denom
+        denom = np.sqrt(x * x + y * y + z * z)
```
erogluorhan: You can use `math.hypot(x, y, z)` here as well, right?

erogluorhan (commented):
One last comment below

uxarray/grid/coordinates.py
rajeeja: That line is gone in the latest update — I removed the entire extra denom block because _normalize_xyz() already performs the L2 normalization, so there’s no need for another norm computation there.

erogluorhan: Ok, things make more sense now.

PR #1479, opened 2026-03-30
Support one-file open_dataset

### LINKED ISSUE #345: Investigate to allow the user to construct a `UxDataset` from only a grid file
_Originally posted by @philipc2 in https://github.com/UXARRAY/uxarray/pull/216#discussion_r1226969655_

rajeeja:
Closes #345 by allowing ux.open_dataset(file) for combined grid-and-data files.


uxarray/core/api.py
```diff
@@ -406,10 +408,22 @@ def open_dataset(
...
+
+    >>> ux_ds = ux.open_dataset("combined_file.nc")
     """
     if grid_kwargs is None:
         grid_kwargs = {}
 
+    if filename_or_obj is None:
+        if isinstance(grid_filename_or_obj, (str, os.PathLike)):
+            filename_or_obj = grid_filename_or_obj
+        else:
```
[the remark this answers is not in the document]
rajeeja: this is a good one, I'll fix it

uxarray/core/api.py
```diff
@@ -353,7 +353,7 @@ def list_grid_names(
 
 def open_dataset(
     grid_filename_or_obj: str | os.PathLike[Any] | dict | Dataset,
-    filename_or_obj: str | os.PathLike[Any],
+    filename_or_obj: str | os.PathLike[Any] | None = None,
```
erogluorhan: While technically there is nothing wrong with this, I'd like to share a few thoughts on use-case logic:

1. Defaulting `filename_or_obj` to `None` might not be the best idea since, in most cases, it is needed, i.e. single file dataset opening is anticipated to occur much less.
2.  Also, more of a terminology thing, but my observation is that when there is only file that has the grid information in it as well, it is referred to as the data or diagnostics file rather than the grid file. That said, I'd recommend to handle the single file case when `grid_filename_or_obj` is given `None`, and `filename_or_obj` has both grid definition and data in it.

uxarray/core/api.py
```diff
@@ -406,17 +408,50 @@ def open_dataset(
...
 
-    # Construct a Grid, validate parameters, and correct chunks
-    uxgrid, corrected_chunks = _get_grid(
-        grid_filename_or_obj, chunks, chunk_grid, use_dual, grid_kwargs, **kwargs
-    )
+    if filename_or_obj is None:
+        if isinstance(grid_filename_or_obj, (str, os.PathLike)):
+            if os.path.isdir(grid_filename_or_obj):
+                raise ValueError(
+                    "Directory-based grids require a separate data file when calling ux.open_dataset()."
```
erogluorhan: Is this to cover the multiple grid files case?

uxarray/core/api.py
```diff
@@ -406,17 +408,50 @@ def open_dataset(
...
+                )
 
-    # Load the data as a Xarray Dataset
-    ds = _open_dataset_with_fallback(filename_or_obj, chunks=corrected_chunks, **kwargs)
+            filename_or_obj = grid_filename_or_obj
+
+            if "latlon" in kwargs:
```
erogluorhan: Why this check?

uxarray/core/api.py
```diff
@@ -406,17 +408,50 @@ def open_dataset(
...
+                warn(
+                    "'latlon is no longer a supported parameter",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                grid_kwargs["latlon"] = kwargs["latlon"]
+
+            corrected_chunks = match_chunks_to_ugrid(grid_filename_or_obj, chunks)
+            ds = _open_dataset_with_fallback(
```
erogluorhan, line 444: I am a bit confused about `_get_grid` vs. `match_chunks_to_ugrid` plus `_open_dataset_with_fallback`. Please clarify why this switch.

uxarray/core/api.py
```diff
@@ -406,17 +408,50 @@ def open_dataset(
...
     """
     if grid_kwargs is None:
         grid_kwargs = {}
 
-    # Construct a Grid, validate parameters, and correct chunks
-    uxgrid, corrected_chunks = _get_grid(
-        grid_filename_or_obj, chunks, chunk_grid, use_dual, grid_kwargs, **kwargs
-    )
+    if filename_or_obj is None:
+        if isinstance(grid_filename_or_obj, (str, os.PathLike)):
```
erogluorhan, line 435: Limiting what was originally possible (dict, xr.dataset) doesn't look like a great idea. I am not sure if we could support dict with this, but I am pretty sure we should for xr.dataset. I think the main check should be on whether a good Grid object could be opened out of the file object here or not.

uxarray/core/api.py
```diff
@@ -406,17 +408,50 @@ def open_dataset(
...
+            corrected_chunks = match_chunks_to_ugrid(grid_filename_or_obj, chunks)
+            ds = _open_dataset_with_fallback(
+                filename_or_obj, chunks=corrected_chunks, **kwargs
+            )
+            uxgrid = open_grid(ds, use_dual=use_dual, **grid_kwargs)
+        else:
+            raise ValueError(
+                "If filename_or_obj is omitted, grid_filename_or_obj must be a file path."
+            )
+    else:
```
erogluorhan, line 458: Maybe rather than splitting the whole logic into two parts as it is in this if-else block now, we'd only need to split the grid opening task as follows:

1. Check if I have only one file path object given
    - If so, try & catch opening a grid object from that one
    - Else, try opening the grid from the separate grid file
2. In both cases, open dataset from the same file

test/core/test_api.py
```diff
@@ -39,6 +39,32 @@ def test_open_dataset(gridpath, datasetpath, mesh_constants):
     nt.assert_equal(len(uxds_var2_ne30.uxgrid._ds.data_vars), mesh_constants['DATAVARS_outCSne30'])
     nt.assert_equal(uxds_var2_ne30.source_datasets, str(data_path))
 
+
```
erogluorhan, line 42: We need another test case to make sure the single file has proper grid definitions in it (i.e. If not throw exception)

erogluorhan (commented):
Thanks a lot for taking this one. Please find my comments below

uxarray/core/api.py
rajeeja: I went with ux.open_dataset(file) as the smallest backward-compatible public API change and avoids open_dataset(None, file) calling pattern. I agree the naming is a little awkward, but my intent was to add the one-file shorthand without changing the existing two-file semantics.

uxarray/core/api.py
rajeeja: Yes, I  added this to explicit error to avoid a confusing downstream xarray failure

uxarray/core/api.py
rajeeja: Good catch, I can fold that back into a cleaner shared path.

uxarray/core/api.py
rajeeja: this was to avoid opening the same file twice, I will refactor this part of the code to be cleaner.

uxarray/core/api.py
rajeeja: xr.Dataset should also work in the one-input case. I’ll update the implementation

uxarray/core/api.py
rajeeja: Addressed.

uxarray/core/api.py
rajeeja: Addressed.

uxarray/core/api.py
rajeeja: Yes — `ux.open_dataset(combined_file)` (single positional arg) is the intended one-file usage; the file goes in the first slot and both grid and data are read from it. No `None` placeholder needed.

uxarray/core/api.py
```diff
@@ -364,14 +364,17 @@ def open_dataset(
...
+    grid_filename_or_obj : str | os.PathLike[Any] | dict | xr.Dataset
         Strings and Path objects are interpreted as a path to a grid file. Xarray Datasets assume that
         each member variable is in the UGRID conventions and will be used to create a ``ux.Grid``. Similarly, a dictionary
         containing UGRID variables can be used to create a ``ux.Grid``
-    filename_or_obj : str | os.PathLike[Any]
+    filename_or_obj : str | os.PathLike[Any] | xr.Dataset, optional
         String or Path object as a path to a netCDF file or an OpenDAP URL that
-        stores the actual data set. It is the same ``filename_or_obj`` in
-        ``xarray.open_dataset``.
+        stores the actual data set, or an already-open ``xarray.Dataset``. It
```
Sevans711, line 376: I would like to clarify the docs a bit:
- grid_filename_or_obj docstring could start with something like "Grid information for the UxDataset" to help quickly clarify this is always the grid info.
- grid_filename_or_obj docstring should clarify it can be a path to a directory containing grid files (but only if filename_or_obj is provided).
- filename_or_obj should clarify to add something like "Or, it can be omitted to treat the grid as data, for any grid file"

The third point I am saying because these changes enable you to do open_dataset(grid_file) for any grid file, regardless of whether it is "combined" with data or not.

uxarray/core/api.py
```diff
@@ -406,23 +409,60 @@ def open_dataset(
...
 
-    # Construct a Grid, validate parameters, and correct chunks
-    uxgrid, corrected_chunks = _get_grid(
-        grid_filename_or_obj, chunks, chunk_grid, use_dual, grid_kwargs, **kwargs
-    )
+    if filename_or_obj is None:
+        if isinstance(grid_filename_or_obj, (str, os.PathLike)):
+            if os.path.isdir(grid_filename_or_obj):
+                raise ValueError(
+                    "Directory-based grids require a separate data file when calling ux.open_dataset()."
```
Sevans711: What is a "Directory-based grid?" This was my first thought. I explored the code a bit more to see there is one special case of directory-based grids.

I would maybe prefer an error like: "open_dataset(directory) is currently not supported. Consider supplying a path to a grid file instead. In special cases (e.g. for FESOM2 ASCII Dataset) directory-based grids can be recognized, but only if also providing a data file, such as via open_dataset(directory, data_file)."

Although… actually I am not really sure why we can't extend "open a directory-based-grid as a UxDataset" the same way we would be supporting "open a single-file grid as a UxDataset". How confusing is the downstream xarray crash?

Sevans711 (changes requested):
Changes mostly make sense, code looks reasonable, tests cover new cases. Tested locally and behavior seems reasonable. I read through the older comments in the thread on this PR as well. A few lingering thoughts / comments:

I was surprised to discover that this enables ux.open_dataset(grid_file) regardless of whether the grid file also contains "non-grid" data. Added a comment about this, I think it could be cleared up with just a slight change to the docstring.

I was confused by the "Directory-based gridfiles" crash. Nothing in the docstring suggested that the inputs could be directories in the first place, so I didn't realize directory-based gridfiles even existed at first. Suggested rephrasing. Also, I separately may be trying to add support for ux.open_dataset(directory) in some simple cases soon (see #1555). Clarifying this now will help with that.

Finally, I committed a minor change to error message when failing due to typo like ux.open_dataset(data_file). Before the error was confusing; now it is clearer that the issue is being unable to parse uxgrid information from the inputs.

rajeeja:
@Sevans711 Docstrings clarified (grid-vs-data roles, single-arg reads non-grid vars as data) and the directory error message rewritten. Also noted partial / non-UGRID / non-Earth grids can be opened, features degrading as needed.

Sevans711:
@rajeeja Thank you for adding those clarifications! One more thought on the new wording, and then I think the rest looks good and I would be happy to approve. Current wording:

> If omitted,``grid_filename_or_obj`` is also used as the data source, allowing a combined grid-and-data file or ``xarray.Dataset`` to be opened with a single argument. In this single-argument form the file is used both to build the grid and to load the data; any non-grid variables it contains are read as data.

I would change this to something like (bold here to show changes):

If omitted,``grid_filename_or_obj`` is also used as the data source, allowing a combined grid-and-data file or ``xarray.Dataset`` to be opened with a single argument. In this single-argument form the file is used **to build the grid and also treated as data; all variables it contains, including grid variables, will be included as data_vars in the result**.

This clarifies the unexpected behavior I was noticing. E.g., result['node_lon'] exists when building the dataset this way, in addition to things like result['bottomDepth'] or other physical data. This behavior seems reasonable enough, it just wasn't my initial intuition about what would happen, so clarifying the docs should be sufficient.

Sevans711 (approved):
Thank you for all the changes in response to my previous comments! I think it looks good now and recommend merging to main.

erogluorhan:
@rajeeja Looks like we didn't add anything into docs about this. Would you be able to open an issue and do that?

PR #1483, opened 2026-04-01
Add reusable remap weight support

### LINKED ISSUE #172: Applying remap weights from other packages (e.g., ESMF, TempestRemap?)
One of the most useful functions in NCL (for me) that I haven't been able to replicate has been the ESMF wrappers, in particular [ESMF_regrid_with_weights](https://www.ncl.ucar.edu/Document/Functions/ESMF/ESMF_regrid_with_weights.shtml).

A current use case. I am running a large (O(1000)) member ensemble of short hindcasts with ~1deg CAM-SE and would like to "downscale" the data to 10x10deg ~equal-area cells (currently using CS, but icosahedral, etc. also an option) for training a neural net with scikit-learn (so the downstream application is all Python). Since it is O(1000) snapshots on the same grid, it would be highly inefficient to create new map weights every loop -- ideally map weights are generated once and then can just be read and applied to multiple (xarray) slices.

I have actually struggled to find a good way of doing this with Python. There are tools such as xESMF (or digging down to ESMPy) but are highly abstract (i.e., try to take a lot of options out of the user hands and bury the ESMF settings pretty deep) and don't appear to have the exact functionality I want (e.g., take a SCRIP -> SCRIP and provide a weight file for reuse on a DataArray). The one that looks promising is an offshot of [Iris](https://github.com/SciTools-incubator/iris-esmf-regrid) but will have to dig further. Perhaps I'm missing something obvious.

I am not sure this is in uxarray's purview, but I bring it up because I currently have found the best workaround (at this time) is a two-step workflow, where I generate an intermediate file that can be passed into my sklearn code using NCL (or I could also use CDO/NCO I suppose), but ideally, everything could be done on the Python side (with metadata, etc. being preserved, which the NCL wrappers also do an excellent job of).

rajeeja:
Adds support for loading and applying reusable offline remap weights through a generic RemapWeights API, including focused tests and docs for external ESMF and TempestRemap map files. Closes #172.

### DIFF
--- a/docs/user-guide/remap-weights.rst
+++ b/docs/user-guide/remap-weights.rst
@@ -1,7 +1,7 @@
 .. currentmodule:: uxarray
 
-Remap Weights
-=============
+Remap with Weights
+==================
 
 UXarray can apply precomputed offline remapping weights produced outside of UXarray.
 This is useful when weights are generated once with tools such as ESMF or
@@ -11,9 +11,7 @@ slices, or variables.
 The core workflow is:
 
 1. Generate a weight file for a specific source grid and destination grid.
-2. Load the weight file once with :func:`load_remap_weights`.
-3. Reuse the loaded :class:`RemapWeights` object with :meth:`UxDataArray.remap.apply_weights`
-   or :meth:`UxDataset.remap.apply_weights`.
+2. Apply it with :meth:`UxDataArray.remap.apply_weights` or :meth:`UxDataset.remap.apply_weights`.
 
 Basic Usage
 -----------
@@ -25,15 +23,10 @@ Basic Usage
    src = ux.open_dataset("source_grid.nc", "source_data.nc")
    dst = ux.open_grid("destination_grid.nc")
 
-   weights = ux.load_remap_weights("map.nc")
+   remapped_temperature = src["temperature"].remap.apply_weights(dst, "map.nc")
+   remapped_dataset = src.remap.apply_weights(dst, "map.nc")
 
-   remapped_temperature = src["temperature"].remap.apply_weights(
-       weights, destination_grid=dst
-   )
-
-   remapped_dataset = src.remap.apply_weights(weights, destination_grid=dst)
-
-Repeated calls with the same path reuse a cached sparse operator, so loading the
+Repeated calls with the same path reuse a cached sparse operator, so applying the
 same file again in one Python session avoids rebuilding the matrix.
 
 What A Weight File Represents
@@ -89,6 +82,8 @@ This implementation was verified against real files from both families:
 
 In practice, UXarray supports the standard full offline map format used by both
 tools.
+
+Currently, this API applies externally generated sparse remap files. Generating reusable UXarray weight maps can be added as a future extension.
 
 Current caveats:
 
--- a/docs/user-guide/remap-weights.rst
+++ b/docs/user-guide/remap-weights.rst
@@ -117,4 +117,3 @@ This path is useful when:
 - remapping needs to be repeated many times
 - external tools already produce trusted offline maps
 - you want to stay in Python for applying the map and preserving array metadata
-
--- a/docs/user-guide/remap-weights.rst
+++ b/docs/user-guide/remap-weights.rst
@@ -1,0 +1,120 @@
+.. currentmodule:: uxarray
+
+Remap Weights
+=============
+
+UXarray can apply precomputed offline remapping weights produced outside of UXarray.
+This is useful when weights are generated once with tools such as ESMF or
+TempestRemap and then reused many times across multiple ensemble members, time
+slices, or variables.
+
+The core workflow is:
+
+1. Generate a weight file for a specific source grid and destination grid.
+2. Load the weight file once with :func:`load_remap_weights`.
+3. Reuse the loaded :class:`RemapWeights` object with :meth:`UxDataArray.remap.apply_weights`
+   or :meth:`UxDataset.remap.apply_weights`.
+
+Basic Usage
+-----------
+
+.. code-block:: python
+
+   import uxarray as ux
+
+   src = ux.open_dataset("source_grid.nc", "source_data.nc")
+   dst = ux.open_grid("destination_grid.nc")
+
+   weights = ux.load_remap_weights("map.nc")
+
+   remapped_temperature = src["temperature"].remap.apply_weights(
+       weights, destination_grid=dst
+   )
+
+   remapped_dataset = src.remap.apply_weights(weights, destination_grid=dst)
+
+Repeated calls with the same path reuse a cached sparse operator, so loading the
+same file again in one Python session avoids rebuilding the matrix.
+
+What A Weight File Represents
+-----------------------------
+
+A remap weight file represents a linear operator from one grid to another:
+
+.. code-block:: text
+
+   target_values = W @ source_values
+
+If the source grid has ``4800`` elements and the destination grid has ``11000``
+elements, then:
+
+- ``source_values.shape = (4800,)``
+- ``W.shape = (11000, 4800)``
+- ``target_values.shape = (11000,)``
+
+So the weight file necessarily encodes both the source grid and the destination
+grid. It is specific to that grid pair and to the ordering of the source and
+destination degrees of freedom.
+
+Supported File Structure
+------------------------
+
+UXarray currently supports the standard sparse offline-map structure used by
+ESMF-style and TempestRemap-style map files. The essential pieces are:
+
+- ``n_a``: source size
+- ``n_b``: destination size
+- ``n_s``: number of nonzero entries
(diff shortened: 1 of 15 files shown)


uxarray/remap/weights.py
```diff
@@ -0,0 +1,147 @@
...
+        indices = indices - 1
+    elif indices.min() < 0 or indices.max() >= size:
+        raise ValueError(
+            f"{label} indices are out of bounds for size {size}. "
+            f"Found min={indices.min()}, max={indices.max()}."
+        )
+
+    return indices
+
+
```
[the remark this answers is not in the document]
rajeeja: Fixed.

uxarray/remap/weights.py
```diff
@@ -0,0 +1,147 @@
...
+    def from_file(cls, filename_or_obj: str | PathLike[str] | xr.Dataset):
+        """Load a standard sparse remap-weight file into memory once."""
+        if isinstance(filename_or_obj, xr.Dataset):
+            ds = filename_or_obj
+            close_ds = False
+            path = None
+        else:
+            ds = xr.open_dataset(filename_or_obj)
+            close_ds = True
+            path = str(filename_or_obj)
```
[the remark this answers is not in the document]
rajeeja: Fixed.

uxarray/remap/weights.py
```diff
@@ -0,0 +1,147 @@
...
+from dataclasses import dataclass
+from os import PathLike
+from pathlib import Path
+
+import numpy as np
+import xarray as xr
+from scipy import sparse
+
+_WEIGHTS_CACHE: dict[tuple[str, int, int], "RemapWeights"] = {}
+
```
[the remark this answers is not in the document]
rajeeja: Fixed.

uxarray/remap/apply_weights.py
```diff
@@ -0,0 +1,105 @@
...
+            continue
+
+        remapped_any = True
+        other_dims = [dim for dim in da.dims if dim != variable_source_dim]
+        da_t = da.transpose(*other_dims, variable_source_dim)
+        remapped_values = weights_obj.apply(np.asarray(da_t.values))
+
+        coords = {dim: da.coords[dim] for dim in other_dims if dim in da.coords}
+        da_out = uxarray.core.dataarray.UxDataArray(
+            remapped_values,
```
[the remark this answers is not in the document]
rajeeja: Fixed.

uxarray/remap/apply_weights.py
```diff
@@ -0,0 +1,105 @@
...
+            dims=other_dims + [destination_dim],
+            coords=coords,
+            name=da.name,
+            attrs=da.attrs,
+            uxgrid=destination_grid,
+        )
```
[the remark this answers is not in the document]
rajeeja: Fixed.

uxarray/remap/apply_weights.py
```diff
@@ -0,0 +1,105 @@
...
+
+    if source_dim is not None:
+        if source_dim not in da.dims:
+            return None
+        if da.sizes[source_dim] != weights.source_size:
+            raise ValueError(
+                f"Variable {da.name!r} dimension {source_dim!r} has size "
+                f"{da.sizes[source_dim]}, expected {weights.source_size}."
+            )
+        return source_dim
```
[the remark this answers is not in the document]
rajeeja: Fixed.

uxarray/remap/weights.py
```diff
@@ -0,0 +1,189 @@
...
+        return remapped.reshape(values.shape[:-1] + (self.destination_size,))
+
+
+def _cache_key(filename_or_obj: str | PathLike[str]) -> tuple[str, int, int]:
+    path = Path(filename_or_obj).expanduser().resolve()
+    stat = path.stat()
+    return str(path), stat.st_mtime_ns, stat.st_size
+
+
+def load_remap_weights(
```
erogluorhan, line 169: Do we believe there will be need for a separate `ux.load_remap_weights("map.nc")`? i.e. why don't we embed the weight file ingestion into the actual application function: `src.remap.apply_weights(weight_file, destination_grid=dst)`? Also, it wouldn't need to be strictly weight file, and maybe it could support array objects?

If we were to have it for any reason though, how about having it under the `remap` namespace instead of being a direct call from `ux`?

uxarray/remap/accessor.py
```diff
@@ -227,3 +229,48 @@ def bilinear(
...
             )
         return _bilinear(self.ux_obj, destination_grid, remap_to)
+
+    def apply_weights(
+        self,
+        weights,
+        destination_grid: Grid,
+        remap_to: str = "faces",
+        source_dim: str | None = None,
+    ) -> UxDataArray | UxDataset:
```
erogluorhan, line 239: To make the signature similar to the remap functions we have, how about reordering the arguments: `apply_weights(destination_grid, weight_file)`?

docs/user-guide/remap-weights.rst
```diff
@@ -0,0 +1,119 @@
+.. currentmodule:: uxarray
+
+Remap Weights
```
erogluorhan: Title reads a bit vague, how about something like "Remap with Weights", "Remap Applying Weights", etc?

docs/user-guide/remap-weights.rst
```diff
@@ -0,0 +1,119 @@
+.. currentmodule:: uxarray
```
erogluorhan, line 1: I like this document; in addition, having a section in the main remap user guide feels like a must to guide people for when they will do this kind of remap.

uxarray/remap/weights.py
```diff
@@ -0,0 +1,189 @@
...
+                ds.close()
+
+        return cls(
+            matrix=matrix,
+            source_size=source_size,
+            destination_size=destination_size,
+            path=path,
+        )
+
+    def apply(self, values: np.ndarray) -> np.ndarray:
```
erogluorhan: I am not sure about having this. I think application of weights always under the `.remap` namespace would be a good idea to me.

docs/user-guide/remap-weights.rst
```diff
@@ -0,0 +1,119 @@
...
+-----------
+
+.. code-block:: python
+
+   import uxarray as ux
+
+   src = ux.open_dataset("source_grid.nc", "source_data.nc")
+   dst = ux.open_grid("destination_grid.nc")
+
+   weights = ux.load_remap_weights("map.nc")
```
erogluorhan: Not to mention this necesasrily for this PR, but would it be always through reading weights that are externally calculated, i.e. would UXarray's remap functions themselves generate weight maps as well, which would be reused?

erogluorhan (commented):
Great to see the functionality in place. See a few comments below

uxarray/remap/weights.py
rajeeja: Fixed.

uxarray/remap/accessor.py
rajeeja: Fixed.

docs/user-guide/remap-weights.rst
rajeeja: Fixed.

docs/user-guide/remap-weights.rst
rajeeja: Fixed.

uxarray/remap/weights.py
rajeeja: Fixed.

docs/user-guide/remap-weights.rst
rajeeja: Agreed; follow-up scope.

uxarray/remap/weights.py
erogluorhan: I might be missing what was fixed here, please clarify

erogluorhan (commented):
I think this is almost there, thanks much for addressing many things. See just one thing left below

uxarray/remap/weights.py
rajeeja: The user-facing API now lives entirely under the `.remap` namespace — application goes through `uxds["var"].remap.apply_weights(destination_grid, weights)` (accessor → `_apply_weights` in `apply_weights.py`). `RemapWeights.apply()` here is just the internal sparse mat-vec kernel that `_apply_weights` calls; it is not a public entrypoint. Happy to rename it to `_apply` to make its private status explicit if you prefer.

rajeeja: Done — renamed it to `RemapWeights._apply` (commit c669e0dd) to make its private status explicit, and updated the internal call site in `_apply_weights`. The only public way to apply weights remains the `.remap.apply_weights(destination_grid, weights)` accessor. Tests pass.

erogluorhan: Great; thank you!

erogluorhan (approved):

PR #1492, opened 2026-04-16
Fix UxDataset constructor and `to_xarray()` that got broken with xarray==2026.4.0

### LINKED ISSUE #1490: ux.open_dataset and other functions failing with xarray==2026.4.0
### Version

main

Source

### What happened?

Uxarray constructors are now failing with `TypeError('Passing a Dataset as `data_vars` to the Dataset constructor is not supported. Use `ds.copy()` to create a copy of a Dataset.')` . Due to https://github.com/pydata/xarray/pull/11139 in https://github.com/pydata/xarray/releases/tag/v2026.04.0

No failures

You can use the test suite in this project

erogluorhan:
Closes #1490 

## Overview
Avoid `xr.Dataset(xr.Dataset)` calls in our constructor and `to_xarray()` as it got prohibited as of xarray's 2026.4.0.

## Expected Usage
No change to the user API.

## PR Checklist

**General**
- [x] Add appropriate labels

### DIFF
--- a/uxarray/core/dataset.py
+++ b/uxarray/core/dataset.py
@@ -90,7 +90,18 @@ class UxDataset(xr.Dataset):
         else:
             self._uxgrid = uxgrid
 
+        # As of xarray's 2026.4.0, `xr.Dataset(xr.Dataset)` is prohibited;
+        # hence this check, i.e. If we get `xr.Dataset` as input, use its `data_vars`
+        # as `dict` and
+        if args and isinstance(args[0], xr.Dataset):
+            ds = args[0]
+            args = (dict(ds.data_vars),) + args[1:]
+            kwargs.setdefault("coords", dict(ds.coords))
+            kwargs.setdefault("attrs", ds.attrs)
+
         super().__init__(*args, **kwargs)
+
+        # super().__init__(*args, **kwargs)
 
     # declare plotting accessor
     plot = UncachedAccessor(UxDatasetPlotAccessor)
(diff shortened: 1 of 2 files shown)


uxarray/core/dataset.py
```diff
@@ -90,6 +90,17 @@ def __init__(
         else:
             self._uxgrid = uxgrid
 
+        # As of xarray's 2026.4.0, `xr.Dataset(xr.Dataset)` is prohibited;
+        # hence this check, i.e. if we get `xr.Dataset` as input, use its `data_vars`
+        # as `dict` and handle `coords` and `attrs` properly as well
+        if args and isinstance(args[0], xr.Dataset):
+            ds = args[0]
```
rajeeja, line 97: `UxDataset(existing_ds, new_coords)` will fail with the current change. [xr.Dataset takes positional ](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html)data_vars, coords, attrs, so if we pass ds as the first arg, any remaining positional args would map to `coords or attrs`. Might be better to map all 3 out explicitly before the call to super()

rajeeja (changes requested):

rajeeja:
Please add a small test also or something similar: 

```
def test_uxdataset_init_from_xarray_dataset():
    ds = xr.Dataset(
        data_vars={"a": ("x", [1, 2])},
        coords={"x": [10, 20]},
        attrs={"source": "testing"},
    )

    uxds = ux.UxDataset(ds)

    assert "a" in uxds.data_vars
    assert "x" in uxds.coords
    assert uxds.attrs["source"] == "testing"

```

uxarray/core/dataset.py
erogluorhan: Oh, you're right, thanks for catching! Should be fixed now.

rajeeja (approved):

PR #1494, opened 2026-04-18
Keep NumPy unchanged in upstream CI

rajeeja:
Leaves the conda-resolved NumPy in upstream-dev so Numba stays importable while still testing newer upstream packages.


kafitzgerald (approved):
Good catch.

PR #1498, opened 2026-04-23
Fix raster default extent detection

### LINKED ISSUE #1495: CI Upstream Fails on test_plot
Now that we got the CI upstream fixed again, I manually ran it, and it seems to be failing in `test_plot.py`, all with a version of the "test_to_raster..." family. Failure seems to be expected warnings not being emitted with axes extent. 

I think you might have ideas about this as it may be related to the default extent handling you'd done, @rajeeja

rajeeja:
Treat fresh GeoAxes with autoscale enabled as using the default extent so the raster auto-extent path remains stable across upstream Cartopy changes. Closes #1495.

### DIFF
--- a/uxarray/core/dataarray.py
+++ b/uxarray/core/dataarray.py
@@ -465,12 +465,7 @@ class UxDataArray(xr.DataArray):
         else:
 
             def _is_default_extent() -> bool:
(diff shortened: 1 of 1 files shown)


erogluorhan (approved):
Thanks for the fix!

PR #1504, opened 2026-05-11
show ux.open_dataset call in data structures user guide

rajeeja:
Execute ux.open_dataset(grid_path, data_path) in the Opening a Single Data File section, which previously referenced the method in prose but only displayed an already-loaded dataset.


erogluorhan (approved):
Thanks!

PR #1510, opened 2026-05-14
Fix to_polycollection exclude path: always cache antimeridian_face_indices

### LINKED ISSUE #1507: `to_polycollection()` data array method fails when called with `periodic_elements='exclude'`
### Version

v2026.04.1

Conda

### What happened?

With newer UXarray versions (starting with v2026.02), the data array method `.to_polycollection(periodic_elements='exclude')` (which I believe is the default value for this argument) results in an exception being raised:

```
$ python mcve.py 
Traceback (most recent call last):
  File "/scratch3/BMC/hmtb/Michael.Kavulich/MPAS/plotting_scripts/raster_PR/mpas_plot/mcve.py", line 45, in <module>
    plotit(dataset[var].isel(Time=0),ftime_dt)
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch3/BMC/hmtb/Michael.Kavulich/MPAS/plotting_scripts/raster_PR/mpas_plot/mcve.py", line 11, in plotit
    pc=uxda.to_polycollection(periodic_elements='exclude')
  File "/scratch3/BMC/hmtb/Michael.Kavulich/MPAS/plotting_scripts/raster_PR/mpas_plot/conda/envs/mpas_plot/lib/python3.13/site-packages/uxarray/core/dataarray.py", line 327, in to_polycollection
    _data = np.delete(
        self.values,
    ...<3 lines>...
        axis=0,
    )
  File "/scratch3/BMC/hmtb/Michael.Kavulich/MPAS/plotting_scripts/raster_PR/mpas_plot/conda/envs/mpas_plot/lib/python3.13/site-packages/numpy/lib/_function_base_impl.py", line 5444, in delete
    keep[obj,] = False
    ~~~~^^^^^^
IndexError: arrays used as indices must be of integer (or boolean) type
```

Changing the argument to `periodic_elements='split'` or `periodic_elements='ignore'` resulted in a successful plot (though the latter results in an incorrect plot, seemingly identical to the problem described in #1386).

The MCVE below should be runnable on Derecho and Casper; I staged the example data there, though I have seen this with two completely different MPAS datasets so I assume this will occur with any data.

I expected successful plot generation. The example script `mcve.py` runs successfully with v2025.12 (though this older version exhibits the problems from the since-fixed issue #1386).

```python
#!/usr/bin/env python3
from datetime import datetime

import cartopy.feature as cfeature
import cartopy.crs as ccrs

import uxarray as ux

def plotit(uxda: ux.UxDataArray,ftime) -> None:
    pc=uxda.to_polycollection()
    pc.set_antialiased(False)
    pc.set_cmap('viridis')

    fig, ax = plt.subplots(1, 1, figsize=(8,4), dpi=300, constrained_layout=True,
                           subplot_kw=dict(projection=ccrs.PlateCarree()))

    ax.add_feature(cfeature.NaturalEarthFeature(category='cultural', scale='50m',edgecolor='k',facecolor='none',linewidth=0.2, name='admin_0_countries'))
    ax.add_feature(cfeature.NaturalEarthFeature(category='physical',color='k',facecolor='none',
                           linewidth=0.5, scale='10m', name='coastline'))

    pc.set_transform(ccrs.PlateCarree())

    coll = ax.add_collection(pc, autolim=True)
    ax.autoscale()

    plt.title(f'Plot of 2-meter temperature for MPAS forecast, {ftime}',fontsize=8)

    cbar = plt.colorbar(coll,ax=ax,orientation='vertical')
    cbar.set_label('Units: {units}', fontsize=8)

    plt.savefig('example.png',format='png')
    plt.close(fig)
    print("Done saving 'example.png'")


if __name__ == "__main__":

    fn='/glade/work/kavulich/MPAS/tutorial/120km_global/rundir/history.2025-05-14_00.00.00.nc'
    var='t2m'
    timestring='2025-05-14_00:00:00'
    ftime_dt = datetime.strptime(timestring.strip(), "%Y-%m-%d_%H:%M:%S")
    dataset=ux.open_dataset(fn,fn)

    plotit(dataset[var].isel(Time=0),ftime_dt)
```

rajeeja:
Fixes #1507

`to_polycollection(periodic_elements='exclude')` (the default) raised an `IndexError` starting with numpy 2.0 when no cartopy projection was supplied.

**Root cause:** `antimeridian_face_indices` was computed in `_build_polygon_collection` but only written to `_poly_collection_cached_parameters` inside the `if projected_polygon_shells is not None` block. Without a projection that cache entry stayed `None`. `dataarray.to_polycollection` then called `np.delete(data, None, axis=0)` — a silent no-op in numpy < 2.0 but a hard `IndexError` since numpy 2.0 (June 2024).

**Fix:** Move the cache write to before the projection-only block so `antimeridian_face_indices` is always populated regardless of whether a projection is supplied.

### DIFF
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -495,13 +495,6 @@ def _grid_to_matplotlib_polycollection(
         polygon_shells[:, :, 0]
     )
 
-    # Always cache antimeridian_face_indices; dataarray.to_polycollection reads it
-    # regardless of whether a projection is used. Without this, the cache entry
-    # stays None and np.delete raises IndexError on numpy >= 2.0.
-    grid._poly_collection_cached_parameters["antimeridian_face_indices"] = (
-        antimeridian_face_indices
-    )
-
     # Filter out NaN-containing polygons if projection is applied
     non_nan_polygon_indices = None
     if projected_polygon_shells is not None:
@@ -514,9 +507,12 @@ def _grid_to_matplotlib_polycollection(
         does_not_contain_nan = ~np.isnan(shells_d).any(axis=(1, 2))
         non_nan_polygon_indices = np.where(does_not_contain_nan)[0]
 
-        grid._poly_collection_cached_parameters["non_nan_polygon_indices"] = (
-            non_nan_polygon_indices
-        )
+    grid._poly_collection_cached_parameters["non_nan_polygon_indices"] = (
+        non_nan_polygon_indices
+    )
+    grid._poly_collection_cached_parameters["antimeridian_face_indices"] = (
+        antimeridian_face_indices
+    )
 
     # Select which shells to use: projected or original
     if projected_polygon_shells is not None:
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -495,6 +495,13 @@ def _grid_to_matplotlib_polycollection(
         polygon_shells[:, :, 0]
     )
 
+    # Always cache antimeridian_face_indices; dataarray.to_polycollection reads it
+    # regardless of whether a projection is used. Without this, the cache entry
+    # stays None and np.delete raises IndexError on numpy >= 2.0.
+    grid._poly_collection_cached_parameters["antimeridian_face_indices"] = (
+        antimeridian_face_indices
+    )
+
     # Filter out NaN-containing polygons if projection is applied
     non_nan_polygon_indices = None
     if projected_polygon_shells is not None:
@@ -509,9 +516,6 @@ def _grid_to_matplotlib_polycollection(
 
         grid._poly_collection_cached_parameters["non_nan_polygon_indices"] = (
             non_nan_polygon_indices
-        )
-        grid._poly_collection_cached_parameters["antimeridian_face_indices"] = (
-            antimeridian_face_indices
         )
 
     # Select which shells to use: projected or original


uxarray/grid/geometry.py
```diff
@@ -495,6 +495,13 @@ def _grid_to_matplotlib_polycollection(
         polygon_shells[:, :, 0]
     )
 
+    # Always cache antimeridian_face_indices; dataarray.to_polycollection reads it
+    # regardless of whether a projection is used. Without this, the cache entry
+    # stays None and np.delete raises IndexError on numpy >= 2.0.
+    grid._poly_collection_cached_parameters["antimeridian_face_indices"] = (
+        antimeridian_face_indices
+    )
+
```
erogluorhan: Instead of adding this block here, I think you'll need to move original Line #s 510 through 515 out of the if-check and that's it. 

The story behind this issue we are seeing now:

* Some time around May in 2025, we had changed `to_polycollection` to NOT use projections. To see the state of the code before that change, see Lines 497 through 502 [here](https://github.com/UXARRAY/uxarray/blob/7966e2a122392001e821093c3561f8d01c938da9/uxarray/grid/geometry.py) for instance 
* After receiving some reports about problems with those changes, PR #1454 had fixed it and reverted things back. It turns out it missed that Lines 497 through 502 was outside the if-check and out them inside instead.

erogluorhan (commented):
Thanks for the fix, please see my comment below

rajeeja:
You're right — just pushed the fix.

erogluorhan (approved):
Thanks for the fix!

PR #1511, opened 2026-05-18
 Fix broken healpix notebook 

rajeeja:
Empty circles were showing up as the docs build was rendering that HoloViews/GeoViews plot with the default Bokeh backend, which can show only the Orthographic outline/empty circles in the static docs context. I changed the Healpix plotting cell to explicitly `use backend="matplotlib"`


erogluorhan (commented):
Thanks for looking into this! Do we know the reason for bokeh backend not working as before?

rajeeja:
> Thanks for looking into this! Do we know the reason for bokeh backend not working as before?

Some package upgraded the PyViz stack or RTD caused the failure..I was trying to pin it down, but couldn't.

erogluorhan:
I think we need to figure that out before moving forward with this PR as is.

rajeeja:
> I think we need to figure that out before moving forward with this PR as is.

Ok, but I think we can keep this fix as matplotlib is good for static plot; Bokeh is fundamentally an interactive, browser-based, and JavaScript-heavy plotting library.

rljacob (approved):

erogluorhan:
> > I think we need to figure that out before moving forward with this PR as is.
> 
> Ok, but I think we can keep this fix as matplotlib is good for static plot; Bokeh is fundamentally an interactive, browser-based, and JavaScript-heavy plotting library.

I agree. Just wanted to point that having this PR fixed shouldn't lead us to forget the issue, whcih would still show up in user's workflows. Maybe open an issue for that if you're not planning to tackle that and then merge this?

PR #1523, opened 2026-06-11
docs: clarify curl(grad(f)) residual is discretization error

### LINKED ISSUE #1522: Bug: curl(grad(f)) != 0
### Version

v2026.4.2

Source

### What happened?

The [uxarray docs page about vector calculus](https://uxarray.readthedocs.io/en/latest/user-guide/vector_calculus.html#example-1-curl-of-gradient-field) mentions that curl(grad(f)) should be approximately 0 for any scalar field f. However, the example shown demonstrates that it is not close to 0 (probably, depending on what "close" means here). Rather, the example shows that the values range from -4.51 to 4.35. It looks like the data is roughly 1 in order of magnitude (f is a simple Gaussian in this example), and the grid length scale is roughly 0.01 (grad(f) seems to be on the order of 100, when I tried running the code myself).

I expected values of curl(grad(f)) to be closer to 0. Or, maybe not? Do derivatives on unstructured grids have large residuals from discretization? If yes, then the only change needed would be to update that documentation page to clarify.

```python

```

rajeeja:
Fixes #1522.

The docs showed `curl(grad(f))` values of O(1–10) with the caption "Should ≈ 0", with no explanation. This confused a user into filing a bug report.

**Changes:**
- Explain that `curl(grad(f)) = 0` holds exactly only in the continuous setting; UXarray's gradient and curl are independent finite-volume stencils that don't form a discrete de Rham complex
- Note the residual is O(Δx) discretization error that shrinks with grid refinement
- Remove the misleading `Identity verified: {max_curl_grad < 1e-10}` boolean (which was False anyway for a non-trivial field)

### DIFF
--- a/docs/user-guide/vector_calculus.ipynb
+++ b/docs/user-guide/vector_calculus.ipynb
@@ -707,7 +707,7 @@
     "print(\n",
     "    \"Note: non-zero residual is expected discretization error from independent \"\n",
     "    \"finite-volume gradient and curl stencils (not a bug). Shrinks with grid refinement.\"\n",
-    ")\n"
+    ")"
    ]
   },
   {
--- a/docs/user-guide/vector_calculus.ipynb
+++ b/docs/user-guide/vector_calculus.ipynb
@@ -321,7 +321,7 @@
    "source": [
     "### Example 1: Curl of Gradient Field\n",
     "\n",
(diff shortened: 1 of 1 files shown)


erogluorhan (approved):
Thanks very much!

PR #1530, opened 2026-06-23
Add UxDataArray.data_location property

### LINKED ISSUE #547: Add UxDataArray.data_location
Need to provide an enum or similar structure to indicate where data is stored in an UxDataArray.

`UxDataArray.data_location: enumeration {“vertex_centered”, “face_centered”, “face_average”, “edge_centered”, “edge_orthogonal”, “edge_parallel”, “cgll”, “dgll”}`

**Additional context**
Currently there are internal functions `UxDataArray._face_centered()` and `UxDataArray._node_centered()` that work with simply checking the dimensions of data array and return boolean values, which are used for visualization purposes. This effort should address all the need listed above.

rajeeja:
## Overview

Adds a public `UxDataArray.data_location` property that reports where a data
variable is stored on the grid, addressing the long-open issue #547 (opened in
October 2023).

## What it does

`data_location` infers the storage location from the grid dimension present on
the data variable and returns one of:

- `"face_centered"` — data contains the `n_face` dimension
- `"node_centered"` — data contains the `n_node` dimension
- `"edge_centered"` — data contains the `n_edge` dimension
- `None` — data is not mapped to the grid

This builds on the existing internal `_face_centered()` / `_node_centered()` /
`_edge_centered()` helpers and complements the existing `data_mapping` property.

## Scope note

The original issue proposed a richer enumeration including `face_average`,
`edge_orthogonal`, `edge_parallel`, `cgll`, and `dgll` (with polynomial order
for the spectral-element cases). Those locations cannot be inferred from a data
variable's dimensions alone and are not currently representable in UXarray, so
this PR implements the three locations that are reliably determinable today and
documents the rest as not-yet-distinguished. Happy to extend the enum if/when
the underlying metadata becomes available.

## Changes

- Add `UxDataArray.data_location` property (`uxarray/core/dataarray.py`)
- Add it (and the existing `data_mapping`) to the API reference (`docs/api.rst`)
- Add unit tests covering all three locations plus the non-grid / extra-dim
  cases (`test/core/test_dataarray.py`)

Closes #547

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -250,6 +250,8 @@ Grid Accessor
    :toctree: generated/
 
    UxDataArray.uxgrid
+   UxDataArray.data_mapping
+   UxDataArray.data_location
 
 
 UxDataset
--- a/test/core/test_dataarray.py
+++ b/test/core/test_dataarray.py
@@ -134,3 +134,32 @@ def test_isel_invalid_dim(gridpath, datasetpath):
         match=r"Dimensions \{'level'\} do not exist\..*Available dimensions: \('time', 'n_face'\)",
     ):
         uxda.isel(level=0)
+
+
+def test_data_location():
+    """Tests data_location for face/node/edge centered data and non-grid data."""
+    uxgrid = ux.Grid.from_healpix(zoom=1)
+
+    face_da = UxDataArray(
+        np.ones(uxgrid.n_face), dims=["n_face"], uxgrid=uxgrid
+    )
+    node_da = UxDataArray(
+        np.ones(uxgrid.n_node), dims=["n_node"], uxgrid=uxgrid
+    )
+    edge_da = UxDataArray(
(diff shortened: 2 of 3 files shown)


erogluorhan (approved):

Sevans711:
Why are we keeping both `data_location` and `data_mapping`? I just noticed that both of them exist, while debugging some plotting routines. In particular, `UxDataArrayPlotAccesor.points()` uses `data_mapping`. They are returning basically the same information, though the spelling is slightly different (e.g. "faces" versus "face_centered").

Maybe the answer is something like "data_mapping is just a different spelling of data_location, and data_mapping will only ever do faces, nodes, edges, or None, but data_location might eventually be updated to distinguish more possibilities"?

Regardless, it would be helpful to clarify in the docstrings of at least one of these properties, for future reference. When I'm writing new code, should I prefer to use data_mapping or data_location?

PR #1532, opened 2026-06-23
Update Read the Docs Python installer

kafitzgerald:
Changes the Read the Docs Python installer to miniforge ([the mambaforge option was recently deprecated](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-tools-python)).  This should get the docs builds working again for you all.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,7 +3,7 @@ version: 2
 build:
   os: ubuntu-lts-latest
   tools:
-    python: mambaforge-latest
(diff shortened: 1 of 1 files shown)


rajeeja (approved):
Thanks for the quick fix, @kafitzgerald! Confirmed this resolves the RTD failures across all open PRs and main — the discontinued `mambaforge-latest` was failing during environment setup, and `miniforge3-latest` is the right replacement. RTD build now succeeds. Much appreciated!

PR #1538, opened 2026-06-30
Restore matplotlib backend after HoloViews matplotlib plot

### LINKED ISSUE #1537: simple uxarray.plot(backend="matplotlib") seems to break subsequent xarray.plot functionality
### Version

v2025.6.0

Conda

### What happened?

A simple plotting call of a ux data array works fine:
```
ds['weights'].plot(backend="matplotlib").opts(
    cmap="viridis", title="Land Weights"
)
```

But then simple xarray.plot() call won't create a plot (e.g.
```
global_mean[plotVars[i]].plot() 
```

**NOTE** leaving off the `backend="matplotlib"` seems to prevent this issue from occurring
Sample notebook is here https://github.com/wwieder/CESM-Tutorial/blob/land_diags/notebooks/diagnostics/clm_ctsm/basics_clm.ipynb

plotting seamlessly between uxarray and basic matplotlib figures in a workflow

```python

```

rajeeja:
Calling `plot(backend="matplotlib")` runs `hv.extension("matplotlib")`, which switches the active matplotlib backend and clobbers the IPython inline display hook, silently breaking any subsequent native matplotlib/xarray `.plot()` calls. This restores the original matplotlib backend right after the HoloViews extension switch, which is safe because HoloViews objects display through `Store.current_backend` rather than the active matplotlib backend. Verified in real Jupyter kernels that the reported sequence now works, with a new regression test and all plotting tests/relevant docs notebooks passing; closes #1537.

### DIFF
--- a/test/test_plot.py
+++ b/test/test_plot.py
@@ -259,32 +259,59 @@ def test_matplotlib_backend_restored_after_switch(monkeypatch):
         matplotlib.use(original)
 
 
-def test_inline_backend_support_reconfigured(monkeypatch):
-    """Restoring the inline backend must also re-register IPython's display hook."""
+def test_inline_backend_reactivated_via_shell(monkeypatch):
+    """Inside IPython, restoring an inline backend must re-run the shell's own
+    backend activation (the public equivalent of ``%matplotlib inline``), which
+    rebuilds the full display integration that ``hv.extension`` tore down.
+
+    Simply calling ``mpl.use`` is not enough: it restores the backend name but
+    leaves last-line figure auto-display broken (see #1537 / #1538).
+    """
     import sys
     import types
 
     from uxarray.plot.utils import HoloviewsBackend
 
     inline_backend = "module://matplotlib_inline.backend_inline"
-    shell = object()
     calls = []
 
+    class FakeShell:
+        def enable_matplotlib(self, gui):
+            calls.append(("enable_matplotlib", gui))
+
     ipython = types.ModuleType("IPython")
-    ipython.get_ipython = lambda: shell
-    matplotlib_inline = types.ModuleType("matplotlib_inline")
-    backend_inline = types.ModuleType("matplotlib_inline.backend_inline")
-    backend_inline.configure_inline_support = lambda sh, backend: calls.append(
-        (sh, backend)
-    )
+    ipython.get_ipython = lambda: FakeShell()
 
     monkeypatch.setitem(sys.modules, "IPython", ipython)
-    monkeypatch.setitem(sys.modules, "matplotlib_inline", matplotlib_inline)
-    monkeypatch.setitem(sys.modules, "matplotlib_inline.backend_inline", backend_inline)
-    monkeypatch.setattr(matplotlib, "use", lambda backend: calls.append(("use", backend)))
+    monkeypatch.setattr(
+        matplotlib, "use", lambda backend: calls.append(("use", backend))
+    )
 
     be = HoloviewsBackend()
     be.matplotlib_backend = inline_backend
     be.reset_mpl_backend()
 
-    assert calls == [("use", inline_backend), (shell, inline_backend)]
+    # The module:// inline backend must be mapped to the "inline" gui name and
+    # restored through the shell, without falling back to mpl.use.
+    assert calls == [("enable_matplotlib", "inline")]
+
+
+def test_reset_backend_falls_back_to_mpl_use_outside_ipython(monkeypatch):
+    """Outside IPython (get_ipython() is None), restoration falls back to
+    ``mpl.use`` with the stored backend."""
+    import sys
+    import types
+
+    from uxarray.plot.utils import HoloviewsBackend
+
+    calls = []
+    ipython = types.ModuleType("IPython")
+    ipython.get_ipython = lambda: None
+    monkeypatch.setitem(sys.modules, "IPython", ipython)
+    monkeypatch.setattr(matplotlib, "use", lambda backend: calls.append(("use", backend)))
+
+    be = HoloviewsBackend()
+    be.matplotlib_backend = "svg"
+    be.reset_mpl_backend()
+
+    assert calls == [("use", "svg")]
--- a/test/test_plot.py
+++ b/test/test_plot.py
@@ -229,3 +229,29 @@ def test_to_raster_pixel_ratio(gridpath, r1, r2):
     f = r2 / r1
     d = np.array(raster2.shape) - f * np.array(raster1.shape)
     assert (d >= 0).all() and (d <= f - 1).all()
+
+
+def test_matplotlib_backend_does_not_clobber_mpl_state(gridpath):
+    """Regression test for #1537.
+
+    ``plot(backend="matplotlib")`` calls ``hv.extension("matplotlib")`` which
+    switches the active Matplotlib backend and breaks subsequent native
+    Matplotlib/xarray ``.plot()`` calls. UXarray should restore the original
+    Matplotlib backend so plain Matplotlib plotting still works afterwards.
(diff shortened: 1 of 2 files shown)


Sevans711:
I was taking a quick look to see if I could review and get this merged to main, but I ended up with a couple quick questions.

(1) Can you provide a small example that definitely causes the unexpected behavior mentioned in #1537 before this fix, and does not cause it anymore after this fix?

On `main` currently (i.e. without this fix yet), in a Jupyter notebook, I ran the following, putting each plot() call in its own cell. I tried in Python 3.10 and Python 3.13. In both cases, every plot call displays a plot; I can't reproduce the failure to make plots. Each of these calls successfully makes a plot.
```python
import uxarray as ux
uxds["psi"].plot(backend='matplotlib')  # makes a matplotlib-style plot
uxds["psi"].plot()  # makes a matplotlib-style plot
# try switching to bokeh just in case it is related to switching back and forth:
uxds["psi"].plot(backend='bokeh')  # makes a bokeh-style plot
uxds["psi"].plot()  # makes a bokeh-style plot
# switch back again:
uxds["psi"].plot(backend='matplotlib')  # makes a matplotlib-style plot
uxds["psi"].plot()  # makes a matplotlib-style plot
```

(2) Is it desirable for the backend to be changed "permanently" instead of just a temporary change for the current plot? I naively would expect a kwarg being passed to plot() to only affect behavior for that one plot, not to affect global state. E.g., I would expect the following:
```python
import holoviews as hv
hv.extension('bokeh')
uxds["psi"].plot()  # makes a bokeh-style plot
uxds["psi"].plot(backend='matplotlib')  # makes a matplotlib-style plot
uxds["psi"].plot()  # I would expect a bokeh-style plot, but actually it is matplotlib-style.
```
Please let me know if this second point belongs in a separate issue instead.

rajeeja:
Thanks. I’ll keep this separate from #1541. Test added.

Sevans711:
I agree this should stay separate from #1541. I thought both of my points were unrelated to 1541, though.

For (1), I see you added regression tests, with good comments detailing how they aren't actually producing a real Jupyter notebook to see the issue occur (that would be challenging to set up with pytest). Were you able to reproduce the original issue on a real Jupyter notebook before this fix?

Ideally, I would like to run a small piece of code which can reproduce the issue when in `main` in a real Jupyter notebook; then change to this branch, rerun the code, and see that the issue does not occur anymore.

Sevans711:
Ah, I misread the original bug. It isn't saying that uxarray.UxDataArray.plot() fails, it's saying that xarray.DataArray.plot() fails. I believe your changes so far have not fixed this bug. From within this branch, I still see the bug:

Screenshot 2026-07-09 at 11 15 13 AM


What I would expect, once the bug has been fixed, would be for the `xr.DataArray([1,2,3]).plot()` cell to show a plot, like it does if it is run before calling any uxarray plotting routines:

Screenshot 2026-07-09 at 11 11 51 AM

rajeeja:
Yes — I reproduced this in notebook-style execution.

Minimal check:

```python
%matplotlib inline
import matplotlib as mpl
import uxarray as ux

print("initial:", mpl.get_backend())
uxds["psi"].plot(backend="matplotlib")
print("after uxarray:", mpl.get_backend())

plt.figure()
plt.plot([0, 1], [0, 1])
plt.show()
```

On `main`, the UXarray plot changes Matplotlib from `inline` to `agg`, and the later native Matplotlib plot warns that `FigureCanvasAgg is non-interactive`.

On this branch, the backend is restored to `inline`, and the later native Matplotlib plot displays normally.

Sevans711:
Ah, I can confirm, if I explicitly include `%matplotlib inline` then things work on this branch and not on main.

However, when I don't include it, even on this branch, the results are that no plot gets shown:
Screenshot 2026-07-09 at 12 54 42 PM

I basically never explicitly write out `%matplotlib inline` in my notebooks. Can we make a fix that works even when you don't include that line?

Sevans711:
Thank you for working through all these unexpected complications…. I was reviewing and ready to approve, but then somehow discovered another failure case.

First of all, your fix looks good for what we observed already. I confirmed on my end that the following runs and creates a plot, as expected:
```python
# intentionally skipping: %matplotlib inline
import uxarray as ux

uxds["psi"].plot(backend="matplotlib")

plt.figure()
plt.plot([0, 1], [0, 1])
plt.show()
```

Now, for the failure case: it is about what happens when you don't call plt.show() explicitly. When I remove it, I don't see any plot:
```python
import uxarray as ux

uxds["psi"].plot(backend="matplotlib")

plt.figure()  # this line isn't actually necessary but it doesn't hurt.
plt.plot([0, 1], [0, 1])
```
This leads to not displaying any figure.

The issue also persists across all subsequent matplotlib plots in the notebook as well, for example doing plt.plot([1,2,3,2,1]) in another cell will lead to no plot being displayed. Calling plt.show() at a later time will show all plots that have been created since the previous time plt.show() was called.

That's not ideal; normally I expect Jupyter to display plots if they are the last line in the cell. E.g., in a fresh kernel, if I run the following:
```python
plt.plot([0,1],[0,1])
```
Then a plot gets displayed. I usually don't call plt.show() explicitly, because it's more convenient to just end the cell with the plot object.


I'm not sure if this will help with debugging or make things more confusing, but I also noticed that if you call plt.plot() before making your first uxarray plot, then plt.show() remains unnecessary. For example:

```python
import uxarray as ux

plt.plot([1,2,3,2,1])

uxds["psi"].plot(backend="matplotlib")

plt.figure()  # this line isn't actually necessary but it doesn't hurt.
plt.plot([0, 1], [0, 1])
```
This will show two figures, first the [1,2,3,2,1] plot, then the second plot. All subsequent calls to plt.plot() do not seem to require plt.show(), which is good.

rajeeja:
@Sevans711 The no-`plt.show()` case comes from the same root as the earlier ones. `hv.extension("matplotlib")` doesn't just change the active Matplotlib backend — in a Jupyter kernel it also tears down IPython's display integration, and there's no public API to restore it piece by piece. The earlier attempts (restore the backend name, then re-register the inline hook) each reversed one symptom of that global side effect while leaving the next one broken; the last-line figure auto-display was the piece still missing. The fix now up takes a different approach: inside IPython it re-runs the shell's own backend activation (`shell.enable_matplotlib(...)`, the public equivalent of the `%matplotlib` magic), which rebuilds the entire integration in one call, and falls back to `mpl.use` outside IPython. Verified in a real kernel across every case raised here — no-`show()` last-line, subsequent cells, explicit `plt.show()`, plot-before-uxarray, and the uxarray plot itself — all display correctly.

That said, this is still a patch over an upstream problem: the real bug is that `hv.extension` mutates global IPython/Matplotlib display state with no clean way to undo it. Filed upstream as holoviz/holoviews#6955. The alternative is to close this PR, document the `plt.show()` / `%matplotlib inline` workaround, and wait for the upstream fix. Either path is fine — this version makes the default Jupyter flow behave, but if relying on the shell's activation internals isn't worth it, closing and deferring to the upstream fix is the cleaner call.

Sevans711 (approved):
Looks good, and I confirmed it works in the cases we discussed. Thank you for working through the IPython / matplotlib / hvplot / jupyter notebook internals to make this fix! Even though this relies on some of those internals, I suspect jupyter notebook plotting with matplotlib backend is a relatively common use-case, so I think it makes sense to merge to main.

The regression tests can serve as a gate for when the upstream fix is sufficient to avoid poking at those internals directly in uxarray. Maybe when the upstream fix is merged we could revisit this. The comments in the test docstrings probably serve as a sufficient reminder. In the meantime, I think it would be safe to merge and mark the issue as closed!

PR #1541, opened 2026-06-30
Speed up import and load hvplot lazily

### LINKED ISSUE #1539: speed up import and load hvplot lazily

I found a tiny change we can make to speed up `import uxarray`. On my machine the import takes 1.7 seconds before this change, and 0.7 seconds after it. See also: #1260

The change is to import hvplot.pandas and hvplot.xarray lazily in `uxarray.plot.accessor.py`, i.e. only import them the first time any of the `PlotAccessor` objects' `__init__` methods actually get called. These lazy imports should also make it easier to move hvplot to be an optional dependency instead of a required dependency later (see #1224).

This also delays when these strange gray bars appear in Jupyter (they are apparently caused by import hvplot.pandas and hvplot.xarray). Now they only appear the first time you plot something, instead of immediately upon importing uxarray for the first time.

**Previous behavior:**
Image
(with first cell taking ~1.7s, second cell ~0.1s)

**New behavior:**
Image
(first cell faster, second cell slower. Rerunning the second cell takes ~0.1s and does not produce the gray bars.)

Sevans711:
Closes #1539 

## Overview
Speeds up `import uxarray` from ~1.7 seconds to ~0.7 seconds, by making `hvplot.pandas` and `hvplot.xarray` imports lazy, i.e. only import them when actually initializing a PlotAccessor object for the first time.

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

### DIFF
--- a/test/test_dependencies.py
+++ b/test/test_dependencies.py
@@ -1,0 +1,19 @@
+"""
+Purpose: tests related to dependencies
+e.g.: hvplot is optional, so it shouldn't be imported by default.
+Related issues: #1224, #1539
+"""
+import sys
+
+
+def test_hvplot_optional():
+    """Test that hvplot is actually optional and not imported by default.
+    (hvplot in particular is "slow" to import (~1s to import hvplot.pandas),
+    so it should not be imported until actually plotting something.)
+    """
+    assert "hvplot" not in sys.modules  # else, it's a test environment issue..?
+    import uxarray as ux
+    assert "hvplot" not in sys.modules
+
+
+# TODO: similar tests for cartopy, holoviews, and other optional deps.
--- a/test/test_dependencies.py
+++ b/test/test_dependencies.py
@@ -3,7 +3,31 @@ Purpose: tests related to dependencies
 e.g.: hvplot is optional, so it shouldn't be imported by default.
 Related issues: #1224, #1539
 """
+import subprocess
 import sys
+
+
+def _assert_not_imported_after_import_uxarray(module_name):
+    """Run ``import uxarray`` in a fresh interpreter and assert that
+    ``module_name`` was not imported as a side effect.
+
+    A subprocess is used so the check is immune to test ordering: other tests
+    in the same session may have already imported optional deps (e.g. plotting
+    tests import ``hvplot``), which would pollute this process's ``sys.modules``.
+    """
+    code = (
+        "import sys; import uxarray; "
+        f"assert {module_name!r} not in sys.modules, "
+        f"'{module_name} was imported by `import uxarray`'"
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", code],
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, (
+        f"{module_name} should not be imported by default:\n{result.stderr}"
+    )
 
 
 def test_hvplot_optional():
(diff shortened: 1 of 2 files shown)


test/test_dependencies.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_hvplot_optional():
+    """Test that hvplot is actually optional and not imported by default.
+    (hvplot in particular is "slow" to import (~1s to import hvplot.pandas),
+    so it should not be imported until actually plotting something.)
+    """
+    assert "hvplot" not in sys.modules  # else, it's a test environment issue..?
+    import uxarray as ux
+    assert "hvplot" not in sys.modules
```
rajeeja: The first `assert` runs before `import uxarray` and inspects this process's `sys.modules`, which is shared across the whole pytest session — so it trips whenever an earlier test (e.g. `test_plot.py`) has already imported hvplot. Running the check in a fresh subprocess makes it immune to test ordering:

```suggestion
import subprocess
import sys


def _assert_not_imported_after_import_uxarray(module_name):
    """Run ``import uxarray`` in a fresh interpreter and assert that
    ``module_name`` was not imported as a side effect.

    A subprocess is used so the check is immune to test ordering: other tests
    in the same session may have already imported optional deps (e.g. plotting
    tests import ``hvplot``), which would pollute this process's ``sys.modules``.
    """
    code = (
        "import sys; import uxarray; "
        f"assert {module_name!r} not in sys.modules, "
        f"'{module_name} was imported by `import uxarray`'"
    )
    result = subprocess.run(
        [sys.executable, "-c", code],
        capture_output=True,
        text=True,
    )
    assert result.returncode == 0, (
        f"{module_name} should not be imported by default:\n{result.stderr}"
    )


def test_hvplot_optional():
    """Test that hvplot is actually optional and not imported by default.
    (hvplot in particular is "slow" to import (~1s to import hvplot.pandas),
    so it should not be imported until actually plotting something.)
    """
    _assert_not_imported_after_import_uxarray("hvplot")
```

rajeeja: feel free to ignore and apply something similar or better.

Sevans711:
We might want to delay merging; I just discovered that import hvplot.pandas and import hvplot.xarray always change the backend to bokeh. Thus, with these changes, lazily importing hvplot means the first plot you make from uxarray will be forced into having bokeh backend, regardless of the backend you tried to set originally. I am creating a bug report on hvplot repository now about this.

Sevans711:
I'm very confused about why my latest changes would cause those crashes… I only added an additional `import hvplot` and an additional call to `hvplot.extension`. But, the crashes seem to be like ```AttributeError: 'UxDataArray' object has no attribute 'plot'. Did you mean: 'hvplot'?```

Any guesses on how my changes might have led to this? A quick attempt to reproduce it didn't lead me anywhere either, e.g.:
```python
hasattr(uxds, 'plot')
```
gives True. I can chase this down more if I should prioritize it, but wanted to check in first in case you have a guess already. Seems like it might be related to the `__getattr__` architecture?

cmdupuis3 (approved):
Wondering if there's a better way to collect all of the possible optional dependencies in one place, and import the accessors when need.

Doesn't seem necessary atm but would probably be good to use, like, lru_cache or something to force one-time loads across the whole package.

Sevans711:
Thank you for your review and comments, and merging to main! I think we shouldn't need to worry about making our own cache, because Python already handles import caching nicely. We just have to make sure optional dependencies and any "heavy"/"takes a long time to import" dependencies are only imported when we actually need them, rather than immediately upon import uxarray.

Will continue optional dependency updates / discussion here: #1548

rajeeja:
Thanks for looking into this.

PR #1548, opened 2026-07-06
Refactor dependencies to make more optional

### LINKED ISSUE #1224: Dependency Refactor & Organization
There have been many concerns brought up by our users that we could better organize our dependencies, including handling optional dependencies in a more elegant manner.

### LINKED ISSUE #1225: Organize Geometry & Visualization packages as an optional dependency

Sevans711:
Closes #1224
EDIT: also closes #1225 

~Waiting for #1541 before asking to merge (in order to move hvplot from required to "viz").~

## Overview
Addresses user feedback asking to handle dependencies in a more elegant manner. Many previously-required dependencies are now optional. ~Also may speed up "import uxarray" slightly, as now-optional dependencies are moved to no longer be at top-level / top of files; instead, they are imported the first time any relevant method gets called. Imports of some required dependencies were similarly moved, to possibly improve "import uxarray" timing.~ EDIT: movement of imports was already implemented and merged main, to keep this PR smaller; see #1606.

### EDIT: PR Stack Notes:
These optional dependency changes are split into 3 PRs to make it easier to review. The intent is to wait until all 3 PRs are approved, then merge all 3 PRs simultaneously by clicking "merge stack" (important: _must click this from the top PR in order to merge the full stack_). The 3 PRs are:
- This PR, which only updates pyproject.toml and installation-related documentation.
- PR #1654, which adds CI tests for optional dependencies, to confirm the expected functionality works or crashes as expected when installed uxarray with different combinations of optional dependencies.
- PR #1656, which improves the error messages whenever optional dependencies are not yet installed, to suggest installing with relevant extras groups (e.g., something like "holoviews not installed; consider running `pip install "uxarray[viz]"` then try again", instead of just an ImportError saying holoviews isn't installed).

### EDIT: Notes about installation docs updates:
In a meeting on 2026-08-19, people from the uxarray dev team seemed to agree the installation docs page should be given significant readability updates to be extremely clear about the different installation paths and the differences between conda and pip, rather than keeping updates to that page as minimal as possible. The comment in thread below (https://github.com/UXARRAY/uxarray/pull/1548#pullrequestreview-4977198713) is referencing that discussion.

### Notes about dependencies:
Notes about all dependencies which were previously listed as required in pyproject.toml:
- `antimeridian`: important for geometry → **keep as required**. [EFF] Efficiency note: antimeridian was already never being imported at top-level.
- `cartopy`: ~used only in plotting → **viz**~ EDIT: actually, cartopy is being used for plotting _and_ for geometry computations (like `to_geodataframe()`) → **viz** _and_ **geo**
- `dask[dataframe]`: keep as required for now. But, it is only used in core.aggrecation and core.zonal, so maybe could make optional? [EFF] already never at top-level.
- `datashader`: ~not used explicitly anywhere → **removed**.~ EDIT: required implicitly by rasterization operations in plotting routines → **viz**; also add a comment somewhere in the codebase to clarify.
- `geoviews`: used only in examples, for plotting → **viz**
- `holoviews`: used only in examples, tests, and HoloviewsBackend for plotting → **viz**
- `matplotlib`: used only for plotting → **viz**
- `matplotlib-inline`: not used explicitly anywhere → **removed**
- `netcdf4`: not used explicitly, but everyone is reading .nc files → **keep as required**
- `numba`: **keep as required**, of course
- `numpy`: **keep as required**, of course
- `pandas`: important for grid methods, e.g. pd.Series and pd.IntervalIndex used explicitly → **keep as required**
- `pyarrow`: not used explicitly, and dask[dataframe] already depends on it → **removed**
- `requests`: not used explicitly anywhere in uxarray → **removed**
- `scikit-learn`: important for neighbor algorithms, e.g. KDTree → **keep as required**. [EFF] already not at top-level.
- `scipy`: important for delaunay triangulations and some io → **keep as required**. _[EFF] removed from top-level._
- `shapely`: important for geometry → **keep as required**. [EFF] already not at top-level.
- `spatialpandas`: just for spatialpandas.GeoDataFrame conversions; should be optional → **geo**
- `geopandas`: just for geopandas.GeoDataFrame conversions; should be optional → **geo**
- `xarray`: **keep as required**, of course
- `hvplot`: used only for plotting → **viz**. ~Didn't move it yet though; waiting for #1541 to get merged~
- `healpix`: just for healpix grids; should be optional → **geo**
- `polars`: used explicitly for some internal methods → **keep as required**
- `pyproj`: only used in uxarray.cross_sections.sample.sample_geodesic; should be optional → **geo**

### Miscellaneous notes:
- removed `pathlib`; pathlib became a built-in part of the standard library in Python 3.4, and the pyproject.toml requires Python >= 3.10.
- added "all" as alias for "complete" because many libraries use "all"; see example usage below.
- added `pooch` to "dev" dependency group, because pooch was secretly required in order for all tests to pass. Comment in test/io/test_structure.py clarifies why pooch dependency exists despite not being imported explicitly anywhere in uxarray.

## Expected Usage
Users can now install uxarray with optional dependency groups. Simple examples with pip interface:
Required dependencies only:
```
pip install uxarray
```
Also includes all optional dependencies related to geospatial grids (geopandas, healpix, pyproj, spatialpandas)
```
pip install "uxarray[geo]"
```
Also includes everything related to geospatial grids, plus everything related to visualization (cartopy, geoviews, holoviews, matplotlib)
```
pip install "uxarray[geo,viz]"
```
Includes all required plus all optional dependencies:
```
pip install "uxarray[complete]"
```
Equivalent to above: `pip install "uxarray[all]"`

## PR Checklist

**General**
- [X] Add appropriate labels

**Testing**
- Question: should we add something to test suite to ensure you can at least `import uxarray` after doing `pip install uxarray` with minimal dependencies? We could maybe add a few more tests for different combinations of dependencies too?
  - EDIT: answer: yes, definitely. See #1654!


rajeeja:
One thing to flag: anyone doing `pip install uxarray` and then using plotting will now hit an ImportError until they install `[viz]`. This is intended per #1224, but it's a meaningful change to the default install contract, so I think we need a broader discussion before merging. @erogluorhan — can you weigh in on how we want to handle the default UX and messaging here?

Sevans711:
> One thing to flag: anyone doing `pip install uxarray` and then using plotting will now hit an ImportError until they install `[viz]`. This is intended per #1224, but it's a meaningful change to the default install contract, so I think we need a broader discussion before merging. @erogluorhan — can you weigh in on how we want to handle the default UX and messaging here?

I definitely agree, we should make sure to discuss more before merging. I am curious to know what people think.

One thought I had is: something that might help could be to provide hints during those crashes. Something like:
```python
def function_that_uses_holoviews(*args, **kwargs):
    try:
        import holoviews as hv
    except ImportError as err:
        raise ImportError('holoviews import failed; to install all plotting dependencies, could do: pip install "uxarray[viz]".') from err
    # << code from function
```
As opposed to the code presently in the PR, which just does `import holoviews as hv`.

Although, this doesn't match xarray's style. In xarray, if you fail to install optional dependencies you will just see an ImportError crash once you try to use them, there's no hint to pip install xarray[viz] or anything like that.

Another slight downside is that this option might get a bit unwieldy due to the need to repeat it inside every relevant method. A partial workaround could be to handle these hints all in one place:
```python
# --- inside new file: utils.imports.py --- #
def _import_with_hint_if_error(*modules):
    """import modules, but make nice error if crash, providing hint to pip install "uxarray[...]".
    E.g., if 'holoviews' in modules, and import holoviews fails, hint to use: pip install "uxarray[viz]".
    """
    if 'holoviews' in modules:
        try:
            import holoviews as hv
        except ImportError as err:
            raise ImportError('holoviews import failed; to install all plotting dependencies, could do: pip install "uxarray[viz]".') from err
    if 'geoviews' in modules:
        # similar
    # similar for all other optional dependencies.
```
Then functions with optional dependencies can look like:
```python
from uxarray.utils.imports import _import_with_hint_if_error

def function_that_uses_holoviews_and_geoviews(*args, **kwargs):
    _import_with_hint_if_error('holoviews', 'geoviews')
    import holoviews as hv
    import geoviews as gv
```
I actually have a slight personal preference to do something like that; I think user experience is improved when common errors include debugging hints. But, I'm happy to go with whatever we decide together.

Sevans711:
Quick note: #1563 ran into issue with `pathlib` still being in our dependencies and would also remove it, as mentioned in [this comment](https://github.com/UXARRAY/uxarray/issues/1561#issuecomment-4962325120). This PR could have fixed that issue before there was a need to spend efforts debugging it in that branch.

erogluorhan (commented):
Please resolve the merge conflicts so that I can see the test results before a larger review, but here are my initial thoughts about this work:

- Really appreciate the detailed description of this PR; it is extremely helpful!
- Per @rajeeja 's [comment](https://github.com/UXARRAY/uxarray/pull/1548#issuecomment-4937728809), I agree this is a major change and definitely requires being very careful; a few thoughts:
   - conda does not have built-in optional dependencies like pip. Mechanisms such as lazy imports and fail locally rather than fail on installation will be needed in place to make sure there is no divergence-as much as possible- between conda and pip
   - Per @Sevans711 's [suggestion](https://github.com/UXARRAY/uxarray/pull/1548#issuecomment-4938013872), clear guidance messages about the expected optional installs on both conda and pip will be needed moving forward.
- It is possible for some packages, e.g. datashader, that importing them will still be required even if their functionality is not explicitly called in the code (due to how the upstream, e.g. holoviz, dynamically uses them, for instance rasterization methods)
    - Removing such packages might not be viable, so extensive testing of these changes will be needed before a possible merge
- Maybe most importantly, due to the all mentioned above,  I'd recommend to take some time on this, and maybe look into similar but smaller dependency-related tasks first.

Sevans711:
@erogluorhan thank you for looking into this! Replying below with what I hope is a concrete action plan (see bottom; apologies this message ended up being very long!). **Includes a few questions for your consideration, in bold; I will wait for your thoughts on those before I proceed.** We can of course iterate more after it is implemented, but for now I hope to clarify the immediate next steps.


## Minimize divergence between conda and pip:

> Mechanisms such as lazy imports and fail locally rather than fail on installation will be needed in place to make sure there is no divergence-as much as possible- between conda and pip

I believe these mechanisms are already in place in this PR, but the relevant tests are not. I would propose: add these not-all-optional-dependencies pip builds to the autotests:
- (A) pip install uxarray
- (B) pip install "uxarray[viz]"
- (C) pip install "uxarray[geo]"
- (D) pip install "uxarray[viz,geo]"

(we don't need to add pip install "uxarray[complete]", because that's what the current autotest pipeline does already.)

And, run four test cases in each:
1. import uxarray - this should never crash
2. one simple plotting test that doesn't use geo dependencies - this should crash on (A) and (C) but not (B) and (D).
3. one simple test that uses geo dependencies but not plotting (maybe `UxDataArray.to_geodataframe()`) - this should crash on (A) and (B) but not (C) and (D)
4. one plotting test that also relies on geo dependencies - this should crash on (A), (B), and (C), but not (D).

**Would that be a sufficient set of tests, or should we aim for more than that?**


## Clear guidance messages about optional installs

> clear guidance messages about the expected optional installs on both conda and pip will be needed moving forward.

I am reading this as confirmation that **you would maybe appreciate something like my suggested `_import_with_hint_if_error` so we can get clearer crash messages, even though this does not follow xarray's style. Is that correct?** I'd be happy to implement this, it shouldn't take too long, but I wanted to confirm it is at least somewhat likely to be a worthwhile effort.

Beyond that, the PR currently includes updates to the uxarray readthedocs installation page, to help clarify how to do install with none/some/all optional dependencies.


## Silent functionality changes if packages missing

> It is possible for some packages, e.g. datashader, that importing them will still be required even if their functionality is not explicitly called in the code (due to how the upstream, e.g. holoviz, dynamically uses them, for instance rasterization methods). Removing such packages might not be viable, so extensive testing of these changes will be needed before a possible merge.

Thank you for raising these concerns. Considering more closely all the packages being removed:
- `datashader`: this was not being used anywhere explicitly in uxarray, which is why I initially suggested to remove it. However, I just spun up a new environment to figure out when this might be needed. It turns out, holoviews tries to use this as soon as you use the option `rasterize=True`. That option shows up in some of the uxarray example notebooks, and it can help when visualizing large datasets. Some of the examples also explicitly discuss the pros and cons of rasterizing. So:
  - I would now change the recommendation on `datashader` to: → **viz**; do not remove.
  - At least one test case should be added to our autotests which uses rasterize=True. Maybe the plotting test (2) above should use rasterize=True too.
  - A comment can be added on that test case, similar to the comment about `pooch` (see original message in this thread), explaining why the dependency exists despite not being imported anywhere explicitly.
- `matplotlib-inline`: this is not being used explicitly anywhere, and I am pretty sure its use is only for inline plots in places like jupyter/ipython. It should be jupyter/ipython's job to install matplotlib-inline, not uxarray's job to do it. For example, in my test environment with only `pip install "uxarray[viz]"` from this branch, and `pip install ipykernel` (to register this kernel so jupyter can recognize it), I was able to create inline matplotlib plots in jupyter without issue. I believe this should still be removed; I would be surprised (but please correct me if I'm wrong!) if any real workflow is relying on uxarray to get matplotlib-inline installed.
- `pyarrow`: will already be installed in the environment due to dask[dataframe] relying on it, and I believe that is our main (only?) use case for it? Since it isn't used explicitly anywhere and it gets installed anyways, I believe that is a good sign that we can remove it.
- `requests`: this also got installed automatically in my test environment, I suppose that means at least one of the required and/or viz dependencies already relies on it. Since it isn't used explicitly anywhere and it gets installed anyways, I believe that is a good sign that we can remove it.

I think this will reduce the amount of testing needed, as the "surface area" of possible effects of removing packages is decreased with these clarifications and change (moving datashader to viz instead of deleting it entirely).


## Taking time to make sure it works

> Maybe most importantly, due to the all mentioned above, I'd recommend to take some time on this, and maybe look into similar but smaller dependency-related tasks first.

Taking some time and being careful definitely makes sense.

~It is a bit hard for me to come up with "similar but smaller dependency-related tasks." Maybe this would mean splitting this PR into multiple smaller PRs? I'm not really sure how to do that in a way that won't end up duplicating work or be confusing to follow. **Do you have something in mind to that end already?**~

EDIT: ah, I just saw your comment on #1575. That is a good example.

Another option is to open another PR like "move should-be-optional dependency imports inside functions instead of at top of files" which could include all the changes to .py files here, but none of the changes to pyproject.toml and installation.rst. That would duplicate some effort, but it would allow the PR here to instead touch only the pyproject.toml, installation.rst, and yet-to-be-added optional-dependency-installation test files, which seem to be more contentious / needs caution than just relocating some imports. **Do you think the duplicated effort is worth the ability to reduce the size of this PR?**

## Action plan / summary
Summary of changes discussed above / action plan:
1. add tests to ensure installation with optional dependencies is possible, and crashes only when we think it should crash
2. add something like `_import_with_hint_if_error` to make clearer errors when optional dependencies not installed.
3. actually do not remove datashader dependency entirely; but do move it to viz (it is currently required on main). Add comment explaining why it is still included despite no explicit import.
4. add a test somewhere using rasterize=True.
5. EDIT: (maybe, depending on your recommendation) move all changes to .py files into a separate PR, so that this PR can focus only on the changes to installation process and associated test files.

Sevans711:
Notes from uxarray development meeting: it sounds like the preference would be to do item (5) as recommended above, i.e. split "move imports" and other changes to .py files into their own PR(s). Additionally, @erogluorhan suggests it is okay to move forward on #1575 simultaneously, or even beforehand, rather than waiting for those PRs to merge. If I misunderstood anything please clarify below!

Sevans711:
Flagging #1225 as related. Noting that `antimeridian` and `shapely` are both used for geometry / plotting, and should maybe be made optional as well. However, I think further discussion about those two packages should maybe be pushed to a subsequent PR with a narrower scope, after this PR merges, unless there is a clear, noncontroversial answer about how they should be handled.

Sevans711:
Updated action plan, now that #1606 solved point (5) of the plan above. See [comment above](https://github.com/UXARRAY/uxarray/pull/1548#issuecomment-4985842875) for more details.

Looking for confirmation from @erogluorhan that this plan looks good, before I start working on it:

## Action plan - updated
1. Add tests to ensure installation with optional dependencies is possible. Add new tests applied to each environment, asserting failure / success appropriately, depending on which optional dependencies were installed.
2. Add something like `_import_with_hint_if_error` to make clearer errors when optional dependencies not installed. Maybe workshop the name a bit (maybe something more like `_assert_can_import` or `_raise_helpful_error_if_import_fails`)?
3. Add comment somewhere in codebase containing the phrase "import datashader" to clarify it is being used when rasterize=True. Put datashader dependency in viz.
4. Add a test somewhere using rasterize=True (helps to clarify why datashader is required; this would be a good place to add the comment from step (3)).

## Notes / questions:
- There are currently only 2 changed files. This number would increase back to 10+ changed files, if implementing step (2) above. **Would reviewers prefer I split step (2) into its own PR?** Upside would be fewer files changed here, but downside would be that `_import_with_hint_if_error` would never actually raise errors there, because all the relevant packages would still be required dependencies….
  - EDIT to add: actually, [stacked PRs](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs) might provide a clean solution here. If I understand correctly, it would enable a PR like "Add `_import_with_hint_if_error` into dependencies refactor PR", where those changes can be reviewed/approved with respect to how they fit in here, and, once approved, be considered as "approved changes" rather than bloating the number of changed files needing to be reviewed here to 10+. This way also wouldn't be blocked by waiting for pre-approval, because it would be in a separate branch. The main downside would be navigating the complexity of stacked PRs, but I suspect it might be worthwhile here.
- **Are the steps above likely to be sufficient?** I.e., would this PR be considered merge-ready, once those steps are completed and reviewed? Or, are there important missing steps which would continue to block merging, even after completing the steps above?

cmdupuis3:
Just wanted to note again that whether dask is optional or not will have a lot of impact on how we do performance optimization. Atm I'm hedging by using xarray or dask imports in conditionals, but it would be nice to have a decision either way.

erogluorhan:
> Just wanted to note again that whether dask is optional or not will have a lot of impact on how we do performance optimization. Atm I'm hedging by using xarray or dask imports in conditionals, but it would be nice to have a decision either way.

I think having Dask optional and deferring the use of it to xarray (`xarray.Dataset.compute()`) makes sense to me. What do you think?

cmdupuis3:
Yeah, as long as we can route through xarray somehow, that should be fine

Sevans711:
> > Just wanted to note again that whether dask is optional or not will have a lot of impact on how we do performance optimization. Atm I'm hedging by using xarray or dask imports in conditionals, but it would be nice to have a decision either way.
> 
> I think having Dask optional and deferring the use of it to xarray (`xarray.Dataset.compute()`) makes sense to me. What do you think?

Clarifying: do you want dask dependency to become optional? Or just the usage of dask to be optional (but dask always gets installed during `pip install uxarray`)? My intuition is that pushing dask to be optional could be good to consider, but I would maybe prefer to leave that conversation to a different PR later on, instead of bundling it with the changes here.

cmdupuis3:
My impression is that if we're aiming to make it optional in the code, we should be aiming to make it optional eventually at the package level too.

Sevans711:
Testing (in #1654) revealed that cartopy is being used by `to_geodataframe()`, in multiple places (e.g. it is imported in `_build_polygon_shells()` which gets called by `_grid_to_polygon_geodataframe()`). The cartopy package provides geometry methods **and** plotting methods. Maybe one day the geometry methods could be rewritten to use the pyproj package instead, to enable installing "geo" dependencies without any plotting codes.

But, for now, I think the best way to move forward is not to try to rewrite those things, but rather to reflect the actual current structure of the code here, by putting cartopy in both "geo" and "viz" dependency groups. I have updated the original post here to reflect this, and made the corresponding changes in 9f3c17c

Sevans711:
Testing also shows that the test suite is already sensitive to excluding the `datashader` package:

- [Relevant test linked here](https://github.com/UXARRAY/uxarray/actions/runs/31188689250/job/92899672529?pr=1654#step:7:306), from #1654. In that test run, pyproject.toml was missing `datashader`, because I hadn't restored it yet.
- Additionally, upon taking a closer look at the existing test suite, I noticed test_plot.py's `test_engine()` test explicitly includes `rasterize=True`, so it should be sensitive to whether `datashader` is installed as well.

**Clarifying progress on the action plan items proposed above:**

1. **(Completed in #1654)** Add tests to ensure installation with optional dependencies is possible. Add new tests applied to each environment, asserting failure / success appropriately, depending on which optional dependencies were installed.
2. **(Not yet started)** Add something like `_import_with_hint_if_error` to make clearer errors when optional dependencies not installed. Maybe workshop the name a bit (maybe something more like `_assert_can_import` or `_raise_helpful_error_if_import_fails`)?
3. **(Completed in commit 62eaa74)** Add comment somewhere in codebase containing the phrase "import datashader" to clarify it is being used when rasterize=True. Put datashader dependency in viz.
4. **(No changes needed; tests in `main` and #1654 are already sensitive to this)** Add a test somewhere using rasterize=True (helps to clarify why datashader is required; this would be a good place to add the comment from step (3)).

Sevans711:
Step (2) has been completed (see #1656), and I believe now the whole PR stack is ready for review!

rajeeja:
@Sevans711 @erogluorhan after this `pip install uxarray` drops from 24 packages to 11, so existing workflows that plot, call `to_geodataframe()`, or use HEALPix will break on upgrade with an ImportError until users add an extra. One concrete example is [uxarray-mcp-server](https://github.com/UXARRAY/uxarray-mcp-server/), which relies on uxarray to pull in cartopy, healpix, and the geopandas backend without declaring them itself. Might be worth the reverse approach of keeping the current set as the default install instead.

Sevans711:
> after this `pip install uxarray` drops from 24 packages to 11, so existing workflows that plot, call `to_geodataframe()`, or use HEALPix will break on upgrade with an ImportError until users add an extra. One concrete example is [uxarray-mcp-server](https://github.com/UXARRAY/uxarray-mcp-server/), which relies on uxarray to pull in cartopy, healpix, and the geopandas backend without declaring them itself. Might be worth the reverse approach of keeping the current set as the default install instead.

To be clear, yes, indeed, this PR would break those workflows.

However, that cost is mitigated significantly by the fact that the fix is both incredibly simple (those workflows just need to replace `pip install uxarray` with `pip install "uxarray[complete]"` or `pip install "uxarray[geo,viz]"`) and will be clarified via very clear error messages, once this ships (see #1656 on top of this PR stack; the intent is to merge the full stack simultaneously).

The main benefits would be (1) solving the original issues reported here (which report "many concerns brought up by our users that we could better organize our dependencies, including handling optional dependencies in a more elegant manner"), and (2) catching up to what seems to be the standard across all major scientific packages, as far as I'm aware. Those packages all let `pip install packagename` include only the required dependencies by default, but provide additional optional dependencies (e.g., see pyproject.toml in xarray, pandas, dask, scipy, matplotlib, holoviews, just to name a few).

(In fact, it seems like pyproject.toml and pip do not even natively support the reverse approach; they assume you will want to install without optional dependencies by default, but make it easy to request those dependencies if desired.)

I think these benefits are worth the cost of breaking existing workflows, though ultimately I understand it isn't just my decision to make, and I am interested to hear more thoughts from others about this, too.

rajeeja (approved):
Good discussion on this PR today, approving. Please just document as discussed.

Sevans711:
> Good discussion on this PR today, approving. Please just document as discussed.

Sounds good! Just made the documentation changes, there are substantial changes on that page as discussed, so you may want to re-review. Notes:

Installation docs page notes:

1. Completely removed any recommendations to look at Xarray's installation instructions page. In my opinion that just made UXarray's installation look much more daunting, like "don't just follow the simple instructions below; surely you will want to read and learn all about how to install Xarray first!" I would expect most users probably just want the easiest path to get packages installed, not to learn all about how to install them.
2. Moved the statement "UXarray itself is a pure Python package, but its dependencies are not." to the top of the page, instead of in the conda section. When it was in the conda section, I thought it was implying that using pip would cause a less-complete installation due to conda including more dependencies' build wheels than pip. That's not actually the case; both conda and pip lead to full installations (assuming appropriate inclusion of extras during the pip install command).
3. Removed recommendation to use conda. Conda steps are still presented first, but neither is more highly recommended.
4. Clarified differences between pip and conda.
5. Focused on emphasizing/de-emphasizing different parts of the page more effectively to improve readability.
6. Split the `pip install "uxarray[extra]"` commands across multiple blocks / bullet points, instead of putting many commands into a single block.
7. Added a `literalinclude` of the relevant parts of `pyproject.toml`, so that it is possible to see full list of optional dependencies included in each group. Using `literalinclude` prevents any drift between `pyproject.toml` and this docs page.

Also flagging here: @rajeeja is there anywhere in the docs with yac installation notes already? I feel like yac should probably at least get mentioned on the installation docs page. But, that "issue" isn't being introduced by this PR, so maybe it deserves a separate issue to track it, something like "mention yac on installation docs page". The mention could be as simple as saying something like "you can also install yac to improve some remapping routines; instructions are available at [link]".

docs/getting-started/installation.rst
```diff
@@ -5,42 +5,61 @@
...
+The easiest way to get everything installed is to use conda or pip.
 
-Installing with Conda (recommended)
------------------------------------
+.. admonition:: Conda versus Pip
 
-UXarray itself is a pure Python package, but its dependencies are not.
-The easiest way to get everything installed is to use conda.
-To install UXarray with its recommended dependencies using the conda command line tool:
+    Conda installs UXarray with all optional depencies.
```
cmdupuis3: Typo: "depencies"

Sevans711: Fixed!

cmdupuis3 (approved):

Sevans711:
Thank you @cmdupuis3 and @rajeeja for your reviews, it looks like the base PR is now approved, so I just wanted to put a quick reminder here to everyone to **please avoid merging until the whole stack is approved** - the plan is to merge the whole stack simultaneously, once it has all been approved!

PR #1549, opened 2026-07-06
Connectivity benchmarks 

cmdupuis3:
This PR adds benchmarks for basic connectivity routines. 

Helps with Issue #1196

## Overview

      
This PR adds benchmarks for the pairwise connectivity routines (face, edge, node). Two of the nine are apparently unimplemented and are included as TODOs: node/node and edge/edge. We can delete them if they don't make sense.

## Expected Usage

Waiting for merge to see real timings on larger grids on Glade, the smaller ones may be too small to sample reliably. 

Run benchmarks with asv:

`asv run --python=same --bench "bench_connectivity.Connectivity"`

## PR Checklist

**General**
- [ x ] An issue is linked created and linked
- [ x ] Add appropriate labels
- [ x ] Filled out Overview and Expected Usage (if applicable) sections

**Testing**
- [N/A] Adequate tests are created if there is new functionality
- [ x ] Tests cover all possible logical paths in your function
- [N/A] Tests are not too basic (such as simply calling a function and nothing else)

**Documentation**
- [ ] Docstrings have been added to all new functions
- [ ] Docstrings have updated with any function changes
- [N/A] Internal functions have a preceding underscore (`_`) and have been added to `docs/internal_api/index.rst`
- [N/A] User functions have been added to `docs/user_api/index.rst`

Issue #1547, opened 2026-07-06
PR from fork cannot write benchmarking comment


Sevans711:
### Version

v2026.6.1

### How did you install UXarray?

Source

### What happened?

PR #1541 is currently blocked because after the benchmarking job completes, github-actions bot fails to write comment to the PR: https://github.com/UXARRAY/uxarray/actions/runs/28799210394/job/85397701245?pr=1541
The failure includes: `Error: Unhandled error: HttpError: Resource not accessible by integration - https://docs.github.com/rest/issues/comments#create-an-issue-comment`

I suspect this is related to the PR coming from an external fork instead of a branch. Maybe there is a check like "head.repo == base.repo" somewhere?

PR 1541 is trying to merge changes from my fork. Meanwhile, I submitted PR #1546 from a branch (now that I have write access to uxarray branches), and there benchmarking completed successfully and github-actions bot was able to leave a comment.

Sidenote: the two failing benchmark cases (fixed by 1546) do not seem to be related to this issue; other PRs by other people recently have had those benchmarks fail too, but the bot was still able to leave a benchmarking comment successfully.

Possibly related to #1192.

### What did you expect to happen?

I expected github-actions benchmarking to leave a comment after running the benchmarking tests in #1541, similarly to other PRs, regardless of whether to PR came from an external fork or from a branch.

### Can you provide a MCVE to repoduce the bug?

```python

```

rajeeja:
@Sevans711 @cmdupuis3 Yes, this is a long-standing bug (from #831): the benchmark workflow runs on forks with a read-only token but then tries to comment, which forks can't do → 403. Planning a PR to fix via the standard split pattern:

1. asv-benchmarking-pr.yml — keep on: pull_request (read-only, safe for fork code); drop the comment step; upload compare results + PR number as artifacts.
2. New asv-benchmarking-comment.yml — on: workflow_run (runs from main with the repo's write token); downloads the artifact and posts the comment.

PR #1550, opened 2026-07-07
Connectivity benchmarks: gate larger files

cmdupuis3:
This PR is to fix an issue with the connectivity benchmarks. The current github runner doesn't seem to have access to the larger dyamond files, so those benchmarks fail and spam up the benchmark logs.

Continuation of #1549.

## Overview

      
I naively copied the file paths from the dyamond benchmarks without the gate, not realizing that the github runner wouldn't have access to them. The gate there basically pings for file access, which will fail for github's runner, but would succeed for someone running somewhere with transparent access to the data on Glade. 

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**
- [N/A] Tests cover all possible logical paths in your function
- [N/A] Tests are not too basic (such as simply calling a function and nothing else)

**Documentation**

**Examples**

### DIFF
--- a/benchmarks/bench_connectivity.py
+++ b/benchmarks/bench_connectivity.py
@@ -30,14 +30,21 @@ all_paths_exist = True
 for file_path in dyamond_path_dict.values():
     all_paths_exist = all_paths_exist and os.path.exists(file_path)
 
-file_path_dict = oQU_path_dict | dyamond_path_dict
+file_path_dict = oQU_path_dict
+if all_paths_exist:
+    file_path_dict = file_path_dict | dyamond_path_dict
 
 
 class GridBenchmark:
     """Class used as a template for benchmarks requiring a ``Grid`` in this
     module across both resolutions."""
     param_names = ['resolution', ]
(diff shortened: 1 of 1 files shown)


Sevans711:
Note, looks like benchmarks seem to be "failing" here for the reason discussed in issue #1547 (also affecting PR #1541).

Sevans711 (dismissed):
Looks good, though I would maybe suggest tiny changes for readability:
- rename `all_paths_exist` to `all_dyamond_paths_exist` or something that suggests only glade/dyamond paths are being checked.
- (less important) instead of looping with something like: `flag = flag and check`, might be more readable to do, e.g.: `all(os.path.exists(file_path) for file_path in dyamond_path_dict.values())`

cmdupuis3:
Maybe, I was trying to mirror what's in the dyamond benchmarks since it's a copy-paste job, so If we change one we should maybe change both.

Sevans711 (approved):
Looks good, my only previous comments were about style, but this style actually matches the existing style from benchmarks/mpas_dyamond.py

Issue #1542, opened 2026-07-02
Remapping notebook is broken on RTD


alioacar:
## Summary
The remapping notebook currently renders a traceback on RTD, likely solver/env drift in HoloViews/Matplotlib stack. The failure happens at display time in the HoloViews matplotlib backend.

## Details
`holoviews.plotting.mpl.util.fix_aspect` is computing NaN for aspect/height so matplotlib rejects the call to `set_size_inches()`.

## Potential Solution
Explicitly setting the fig_inches may be more robust here ?

<details>
<summary> Full traceback </summary>

```text
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.14/site-packages/holoviews/core/dimension.py:1491, in Dimensioned._repr_mimebundle_(self, include, exclude)
   1484 def _repr_mimebundle_(self, include=None, exclude=None):
   1485     """Resolves the class hierarchy for the class rendering the
   1486     object using any display hooks registered on Store.display
   1487     hooks.  The output of all registered display_hooks is then
   1488     combined and returned.
   1489 
   1490     """
-> 1491     return Store.render(self)

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.14/site-packages/holoviews/core/options.py:1517, in Store.render(cls, obj)
   1515 data, metadata = {}, {}
   1516 for hook in hooks:
-> 1517     ret = hook(obj)
   1518     if ret is None:
   1519         continue

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.14/site-packages/holoviews/ipython/display_hooks.py:330, in pprint_display(obj)
    328 if not ip.display_formatter.formatters["text/plain"].pprint:
    329     return None
--> 330 return display(obj, raw_output=True)

File ~/checkouts/readthedocs.org/user_builds/uxarray/conda/latest/lib/python3.14/site-packages/holoviews/ipython/display_hooks.py:301, in display(obj, raw_output, **kwargs)
    299 elif isinstance(obj, (Layout, NdLayout, AdjointLayout)):
    300     with option_state(obj):
--> 301         output = layout_display(obj)
    302 elif isinstance(obj, (HoloMap, DynamicMap)):
    303     with option_state(obj):

[...]

PR #1554, opened 2026-07-08
Downgrade to matplotlib<3.11 to fix RTD plots

Sevans711:
readthedocs plots are breaking on matplotlib 3.11 but not matplotlib<3.11; see issue #1542. Workaround for now: restrict to matplotlib<3.11 in the pyproject.toml.

This restriction should be removed once things get fixed upstream, either by matplotlib or by geoviews.

## PR Checklist

**General**
- [X] Add appropriate labels

**Testing**


rajeeja (approved):
Verified locally: reproduced the #1542 crash on matplotlib 3.11.0 (figure size must be positive finite not [8. nan]) and confirmed it passes on matplotlib 3.10. Good workaround, and the regression test will tell us when it's safe to lift the pin. Approving.

PR #1557, opened 2026-07-09
Add test: if subset empty, dont crash

### LINKED ISSUE #1285: Bounding box subsetting error
### Version

v2025.05.2

Conda

### What happened?

When parameters are provided to `bounding_box` such that no polygons return (i.e. the box is so small no polygons are included) it fails with a ValueError from some Xarray code (see below).  We're seeing this over in the UXarray Cookbook at the moment as a CI failure: ProjectPythia/unstructured-grid-viz-cookbook#28

Returning an empty UXarray with no error would be ideal I think.

If that's painful to implement, a more descriptive error would be a helpful start.

```python

```

Sevans711:
Closes #1285 

## Overview
If subset is empty, uxarray should return empty array, not crash. This seems to already be the case. But, it also seems to be the primary issue of #1285. This PR adds a test to ensure that is the case.

## PR Checklist

**General**
- [X] Add appropriate labels

**Testing**
- [N/A] Tests cover all possible logical paths in your function

### DIFF
--- a/test/test_subset.py
+++ b/test/test_subset.py
@@ -190,3 +190,20 @@ def test_da_subset(gridpath, datasetpath):
     res3 = uxds['t2m'].subset.nearest_neighbor(center_coord=(0, 0), k=4)
 
     assert len(res1) == len(res2) == len(res3) == 4
+
+
+def test_empty_subset(gridpath, datasetpath):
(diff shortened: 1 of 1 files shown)


rajeeja (approved):
Tested locally — test_empty_subset passes and full test_subset.py is green (11/11). Good fix, confirms the empty-subset contract holds.

Issue #1559, opened 2026-07-10
Dispatcher for optimized connectivity routines


cmdupuis3:
See #1195 

# Proposed Feature

I propose we implement a bit of new architecture to describe full batches of desired connectivity routines, so they can exploit optimized routines without the optimized and non-optimized paths interfering with each other.

# Status Quo

Currently, the architecture for connectivity build routines is designed under the assumption that these routines are orthogonal. However, if we are going to implement optimized routines that apply to multiple connectivity types, it will become important to know which optimization paths are available at a high level. Implementing optimizations with internal grid state checks will impose branch conditions inside the optimized routines, which is counterproductive.

# Possible Solution

I think we need a grid connectivity descriptor, that can be used to either describe which connectivities a grid currently has, or to launch multiple connectivity routines at once. Something like:

(Surface API): Enum for list building -> Connectivity boolean dataclass -> Dispatcher like `_grid_connectivity(conn: Connectivity)` -> branch-free optimized connectivity routines (Low level)

PR #1560, opened 2026-07-10
Optimized connectivity: Simultaneous face_edges and edge_nodes

### LINKED ISSUE #1138: Improve performance of `edge_node_connectivity` construction
For a 3.75KM MPAS Grid, the execution time is over `259s` on my local machine 

Image

### LINKED ISSUE #1196: Poor performancewhen constructing `edge_node`, `face_edge` and `face_face` connectivity

### LINKED ISSUE #1200: Optimize Face Edge Coordinate Construction
Below are timings taken across multiple grid resolutions on a single NCAR Derecho CPU node. 

| Resolution | Nodes      | Faces     | Edges       | Grid Load Time (s) | face_edges_xyz (s) |
|------------|------------|-----------|-------------|--------------------|--------------------|
| 30km       | 1,310,720  | 655,362   | 1,966,080   | 1.921              | 0.498              |
| 15km       | 5,242,880  | 2,621,442 | 7,864,320   | 7.458              | 2.213              |
| 7.5km      | 20,971,520 | 10,485,762| 31,457,280  | 27.46              | 6.764              |
| 3.75km     | 83,886,080 | 41,943,042| 125,829,120 | 113.993            | 28.683             |

While the performance here is not bad, it is noticeably slow for the higher resolutions. 

@hongyuchen1030 pointed out that we could consider caching this variable for re-use, especially since multiple calls to zonal average could take advantage of this. There is an informative discussion about this in #1180 

The current implementation only relies on the `face_node_connectivity` and uses it to derive the edge information. If the `edge_node_connectivity` and `face_edge_connectivity` are available, it may be more efficient to use these to index our coordinates. 

https://github.com/UXARRAY/uxarray/blob/3b7e8c48ab17b7420500c9fe68b8ebca1bac4af5/uxarray/grid/utils.py#L141-L259

cmdupuis3:
Would close #1138, #1196

Related to #1180 

Supercedes #1195 

## Overview

      
This set of changes optimizes face_edge, edge_node, and face_face connectivity. face_edge and edge_node connectivity are combined into one routine, while face_face is optimized stand-alone.

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


cmdupuis3:
I think all the hard parts of the merge are done.

At this point, the only failures seems to be sorting issues.

hongyuchen1030:
> I think all the hard parts of the merge are done.
> 
> At this point, the only failures seems to be sorting issues.

@cmdupuis3  Thanks for your work, I will review it as soon as possible. 

And do you have any idea why the CIs are all failing?

cmdupuis3:
The CI is failing because the new algorithm returns the results in a different order. 

I was thinking we can add some sorting mechanism, at least for legacy behavior. Phillip was evidently aware of this issue too. It depends on if you need to support that... Personally I'd be okay with just changing it, but I think you would be a better judge of the situation.

uxarray/grid/connectivity.py
```diff
@@ -162,77 +162,80 @@ def _populate_edge_node_connectivity(grid):
...
+    # Check edge coordinates already exist, if they do this might cause issues
 
-    edge_node_attrs = ugrid.EDGE_NODE_CONNECTIVITY_ATTRS
-    edge_node_attrs["inverse_indices"] = inverse_indices
+    if "n_edge" in grid.sizes:
+        # TODO: raise a warning or exception?
+        pass
+
+    edge_nodes, face_edges = _build_edge_node_connectivity(
+        grid.face_node_connectivity.values, grid.n_nodes_per_face.values
```
erogluorhan: Please think about if you can get rid of `.values` calls to be chunked-array compatible.

cmdupuis3: On it, I think Philip was trying to scalarize here but it introduced this regression.

cmdupuis3:
Altering `njit` and `.values` here runs the risk of conflicting with #1583. I'm thinking we can scope it so that `.values` cleanup in connectivity.py is allowed on this PR and everywhere *except* connectivity.py belongs to #1583.

uxarray/grid/connectivity.py
cmdupuis3: So the best I could come up with is 

```Python
    face_nodes, n_nodes_per_face = dask.compute(
        grid.face_node_connectivity.data, grid.n_nodes_per_face.data
    )

    edge_nodes, face_edges = _build_edge_node_connectivity(
        face_nodes, n_nodes_per_face, grid.n_node
    )
```

...but I'm unclear on if we're committing to dask or not. Other parts of the repo imply that dask is still sort of an optional load, and doing this would pretty much require a dask import for most grids, unless we have some kind of numpy/dask conditional.

cmdupuis3: Alright, I hacked it a bit to get this:

```Python
    computed = xr.Dataset(
        {
            "face_nodes": grid.face_node_connectivity.variable,
            "n_nodes_per_face": grid.n_nodes_per_face.variable,
        }
    ).compute()

    edge_nodes, face_edges = _build_edge_node_connectivity(
        computed.face_nodes.data, computed.n_nodes_per_face.data, grid.n_node
    )

```

Sevans711: Question for @erogluorhan, does uxarray have any plans to support chunked grids? I was under the impression that data variables can be chunked, but that it is always safe to assume the grid itself can be fully loaded into memory. Is that incorrect?

That said, using more dask-compatible syntax like this doesn't seem like it would have a downside (no need to block merging on this question).

test/grid/geometry/test_centroids.py
```diff
@@ -63,13 +63,16 @@ def test_edge_centroids_from_triangle():
...
-    centroid_z = np.mean(grid.node_z[grid.edge_node_connectivity[0][0:]])
+    edge_nodes = grid.edge_node_connectivity.values
 
-    assert centroid_x == grid.edge_x[0]
-    assert centroid_y == grid.edge_y[0]
-    assert centroid_z == grid.edge_z[0]
+    centroid_x = grid.node_x.values[edge_nodes].mean(axis=1)
+    centroid_y = grid.node_y.values[edge_nodes].mean(axis=1)
+    centroid_z = grid.node_z.values[edge_nodes].mean(axis=1)
+    centroid_x, centroid_y, centroid_z = _normalize_xyz(centroid_x, centroid_y, centroid_z)
```
Sevans711, line 71: This seems like a substantive change to the test. Can you clarify why this change was needed? Reply here is fine, not trying to say it is wrong, just not understanding yet why it was changed here.

test/grid/grid/test_connectivity.py
```diff
@@ -63,6 +65,83 @@ def test_connectivity_build_face_edges_connectivity(gridpath):
     assert np.all(valid_edges >= 0)
     assert np.all(valid_edges < uxgrid.n_edge)
 
+@pytest.mark.parametrize("grid_parts", [("ugrid", "outCSne30", "outCSne30.ug"),
+                                        ("ugrid", "quad-hexagon", "grid.nc"),
+                                        ("ugrid", "geoflow-small", "grid.nc")])
+def test_connectivity_edge_node_canonical_order(gridpath, grid_parts):
+    """Test that constructed edges are numbered in lexicographic node order."""
```
Sevans711, line 105: I didn't realize lexicographic order was supposed to be guaranteed for constructed edges. I see now that the corresponding docstring for _build_edge_node_connectivity has been updated accordingly. Is that change being introduced intentionally by this PR?

If yes, replying with yes here is sufficient and the change looks good to me. In the future if you can mention changes like this too somewhere in the PR overview or as comments in thread, that would have helped with reducing time it takes to review!

uxarray/grid/connectivity.py
```diff
@@ -141,98 +146,170 @@ def _populate_n_nodes_per_face(grid):
...
     and stores it within the internal (``Grid._ds``) and through the attribute
     (``Grid.edge_node_connectivity``)."""
 
-    edge_nodes, inverse_indices, fill_value_mask = _build_edge_node_connectivity(
-        grid.face_node_connectivity.values, grid.n_face, grid.n_max_face_nodes
-    )
+    # Check edge coordinates already exist, if they do this might cause issues
+
+    if "n_edge" in grid.sizes:
+        # TODO: raise a warning or exception?
```
Sevans711: This TODO item should be completed before merging, to avoid introducing a new way for the code to fail silently. I would prefer if an exception is raised, not just a warning. Or (more work but maybe cleaner long-term?) figure out what could actually cause this, and debug it?

Minor style suggestion: change to `"n_edge" in grid.dims` instead, to make it clearer that the actual sizes of the dimensions are not relevant here.

uxarray/grid/connectivity.py
```diff
@@ -141,98 +146,170 @@ def _populate_n_nodes_per_face(grid):
...
+        pass
 
-    edge_node_attrs = ugrid.EDGE_NODE_CONNECTIVITY_ATTRS
-    edge_node_attrs["inverse_indices"] = inverse_indices
+    # HACK: this is lieu of an xarray equivalent to `da.compute(a, b)`
```
Sevans711: Clarifying: what is the purpose of this hack - why not just use something like:

```python
computed_face_nodes = grid.face_node_connectivity.variable.compute()
computed_n_nodes_per_face = grid.n_nodes_per_face.variable.compute()
```

I suppose there could theoretically be a speedup from doing a single compute() call instead of two compute calls?

Or, maybe it is some other reason. Could you add a comment in the code to clarify?

uxarray/grid/connectivity.py
```diff
@@ -125,8 +130,8 @@ def _populate_n_nodes_per_face(grid):
     it within the internal (``Grid._ds``) and through the attribute
     (``Grid.n_nodes_per_face``)."""
 
-    n_nodes_per_face = _build_n_nodes_per_face(
-        grid.face_node_connectivity.values, grid.n_face, grid.n_max_face_nodes
+    n_nodes_per_face = (
+        (grid.face_node_connectivity != INT_FILL_VALUE).sum(axis=1).astype(INT_DTYPE)
```
Sevans711: Briefly calling attention to this change, it looks like a huge speed up in the ASV benchmarking suite, which is awesome! It wasn't mentioned in the PR overview / comment thread so I was surprised to see it.

EDIT: actually, many of the n_nodes_per_face tests in the benchmarking suite seem unaffected by this change or even slightly slower because of it. Is this expected behavior?

Confirming: does it also speed things up for large grids on HPC clusters? (Testing with one large grid would be sufficient.)

Sevans711 (changes requested):
These changes look clean overall! It is great to see the possibility for things that previously took hundreds of seconds to maybe be reduced to a few seconds or less (quoting those numbers from #1195).

I left some inline comments and also have a few bigger questions below:

- (1) Seeing some benchmarks improvements here is encouraging. Though, in order to get better confirmation these changes really solve the linked issues, would you be able to run it on at least one large grid on an HPC system and share the results? Both linked issues referred to large grids.

(You don't necessarily need to reproduce the same exact results or use the same exact grids from the original issues, just looking for at least one large grid showing what the timing looks like on `main` right now and what it looks like after these changes.)

- (2) Are the increases to memory usage in large grids acceptable here? E.g., if `edge_node_connectivity` is requested but user has no plans to utilize `face_edge_connectivity,` this PR leads to increased memory cost.

On `grid=ux.tutorial.open_grid("outCSne30-vortex")`, I checked `grid._ds.nbytes` after a few commands. Upon loading, it is 259236 bytes. On `main` after `grid.edge_node_connectivity` it becomes 432036 bytes, and after `grid.face_edge_connectivity` it becomes 604836 bytes. On this PR, just doing `grid=ux.tutorial.open_grid("outCSne30-vortex"); grid.edge_node_connectivity` leads to 648036 bytes. (The extra ~40000 bytes are from `n_nodes_per_face`, though that is basically just a rounding error when compared to the ~200000 extra bytes from storing face_edge_connectivity.) Of course, the numbers here are very small; this is a tiny example. But, they demonstrate that the concern about memory costs may be somewhat plausible, at least. (Unless everybody who asks for edge_node_connectivity also wants face_edge_connectivity?)

test/grid/grid/test_connectivity.py
cmdupuis3: The sorting was implemented to support compatibility with legacy behavior. If we decide to deprecate that behavior we could throw out all the sorting and revert this a little.

uxarray/grid/connectivity.py
cmdupuis3: Because xarray/dask would end up walking the entire structure twice if you compute them separately. Dask has the multi-arg `dask.compute(a, b)` which allows it to walk the structure once and iterate all the vars it needs simultaneously, but fsr xarray doesn't have native support for that. I opted to hack this structure in to backdoor in the dask behavior while staying dask-agnostic here.

cmdupuis3:
> On this PR, just doing grid=ux.tutorial.open_grid("outCSne30-vortex"); grid.edge_node_connectivity leads to 648036 bytes. (The extra ~40000 bytes are from n_nodes_per_face, though that is basically just a rounding error when compared to the ~200000 extra bytes from storing face_edge_connectivity.) Of course, the numbers here are very small; this is a tiny example. But, they demonstrate that the concern about memory costs may be somewhat plausible, at least. (Unless everybody who asks for edge_node_connectivity also wants face_edge_connectivity?)

Yeah, so I was wondering about that too, which is why I proposed having a dispatcher for combined connectivity routines #1559. However, in practice, I think the most likely situation with the grids we support is that both face/edge and edge/node connectivity are the missing connectivities, which is why they were interested in doing this. There could definitely be a discussion on the design, but I think that's out of scope for this PR.

Sevans711:
> > On this PR, just doing grid=ux.tutorial.open_grid("outCSne30-vortex"); grid.edge_node_connectivity leads to 648036 bytes. (The extra ~40000 bytes are from n_nodes_per_face, though that is basically just a rounding error when compared to the ~200000 extra bytes from storing face_edge_connectivity.) Of course, the numbers here are very small; this is a tiny example. But, they demonstrate that the concern about memory costs may be somewhat plausible, at least. (Unless everybody who asks for edge_node_connectivity also wants face_edge_connectivity?)
> 
> Yeah, so I was wondering about that too, which is why I proposed having a dispatcher for combined connectivity routines #1559. However, in practice, I think the most likely situation with the grids we support is that both face/edge and edge/node connectivity are the missing connectivities, which is why they were interested in doing this. There could definitely be a discussion on the design, but I think that's out of scope for this PR.

Ah, yes that makes sense, and seems like a good solution. I also agree a full design like that should probably be scoped to a different PR. Before this PR merges I would still be interested to hear from other uxarray developers/users to confirm whether the increased memory cost is an acceptable tradeoff for the possible huge speedup in the meantime. E.g., @erogluorhan, @rajeeja, @rljacob, if you get a chance to look into this, what are your thoughts?

cmdupuis3:
The speedups are very real. These are from the new connectivity benchmarks in PR #1633, so this is serial performance. I can add chunked performance to the benchmarks as well. I didn't notice any major memory degradation in glancing at top, if anything it seemed better with this branch. Maybe I can add peakmem benchmarks as well.


| Resolution | merge-OFE time | main time |
|------------|-----------------------|----------------|
|                480km |     1.27±0ms | 10.1±0ms 
|                120km |     3.99±0ms | 220±0ms  
|                 30km |     100±0ms  | 10.00±0s 
|                 15km |     434±0ms  | 44.2±0s  
|                7.5km |     1.97±0s  | 1.58±0m  
|                3.75km |    9.92±0s  | failed

cmdupuis3:
@erogluorhan I'm working on some peakmem benchmarks, here's the results for this PR.

ASV main versus merge-OFE branch---

face_face:

| Resolution | merge-OFE Peakmem | main Peakmem |
|------------|-----------------------|----------------|
|               480km | 101k | 886k |
|                120km | 1.6M | 14.7M |
|                 30km | 57.7M | 351M |
|                 15km | 231M | 1.41G |
|                7.5km | 587M | failed |
|                3.75km | 2.35G | failed |

edge_node:

| Resolution | merge-OFE Peakmem | main Peakmem |
|------------|-----------------------|----------------|
|                480km | 416k | 980k |
|                120km | 6.47M | 15.5M |
|                 30km | 168M | 577M |
|                 15km | 671M | 2.31G |
|                7.5km | 2.35G | 5.68G |
|                3.75km | 9.4G | failed |

uxarray/grid/connectivity.py
erogluorhan: Per @cmdupuis3 's latest comment:

This looks great, and it is more beneficial than just "hacking" I believe (please document it accordingly wherever possible in your code):

- Using `xr.Dataset.compute()` guarantees two major advantages:

   -  Bundled instead of separate `.compute()` calls on a couple DataArrays - uses same upstream dask graph and gets rid of duplicate overhead
   - Doesn't break optional Dask - If given chunked arrays, uses Dask; otherwise, skips compute under the xarray's hood and respects numpy arrays

erogluorhan: Per @Sevans711's latest comment:

The intent with `Grid`'s I/O design was already to have it support chunked features; however; full end-to-end "Grid supports chunkedness" is probably impossible - there are global-topology and spatial-index operations like what we see here in this PR that could probably never avoid materializing the whole array.

1. Grid I/O is already made chunked - there might be things we can further fix though
2. Some of the operations can be made chunk-compatible

I think a smart way forward would be to (1) keep (or make) the chunk-local geometric operations lazy, and (2) leave ( and explicitly document) the global-topology/KDTree/validation operations as materialized (whole array computations) rather than attempting to "fix" every `.values` or `.compute()` - this PR shows there is still room for optimization for even those though

uxarray/grid/connectivity.py
Sevans711: Noting that @erogluorhan clarified the reasoning above, see https://github.com/UXARRAY/uxarray/pull/1560#discussion_r3696045024

Keeping my comment here as "unresolved" for now as a reminder to update comment within the code itself to clarify

test/grid/geometry/test_centroids.py
cmdupuis3: Yeah, so this test was basically passing by coincidence, and whether it passes depends on which edge is at index 0. It also skips normalization. I only found this because prior to implementing sorting, this test was failing, which flagged it as something for Claude to look at, and it just happened to be wrong for another reason. With the new sorting algorithm, the original test would be passing though.

cmdupuis3:
pre-commit.ci autofix

uxarray/grid/connectivity.py
cmdupuis3: Updated

uxarray/grid/connectivity.py
cmdupuis3: Removing `.values` keeps it lazy, and also helps with reducing the graph traversals on line 173

uxarray/grid/connectivity.py
cmdupuis3: Done, I think there could be some downstream effects in `_ugrid` though, but we should fix it there anyway.

cmdupuis3:
pre-commit.ci autofix

uxarray/grid/connectivity.py
Sevans711: I agree removing `.values` is a good idea, but this does more than that: it switches from using the numba `_build_n_nodes_per_face` to a numpy sum. How does this actually affect time and/or memory usage for large grids?

Also, I noticed a potential accuracy error here: should refer to axis by name, not number, because xarray arrays should support any axis order, and face_node_connectivity might be user-supplied, so it might be transposed compared to the order you expected here.

test/grid/grid/test_connectivity.py
Sevans711: Can you clarify a bit further? If you can provide at least one example of a legacy behavior this change was implemented to support, that should make it much easier for me to understand.

erogluorhan:
> > > On this PR, just doing grid=ux.tutorial.open_grid("outCSne30-vortex"); grid.edge_node_connectivity leads to 648036 bytes. (The extra ~40000 bytes are from n_nodes_per_face, though that is basically just a rounding error when compared to the ~200000 extra bytes from storing face_edge_connectivity.) Of course, the numbers here are very small; this is a tiny example. But, they demonstrate that the concern about memory costs may be somewhat plausible, at least. (Unless everybody who asks for edge_node_connectivity also wants face_edge_connectivity?)
> > 
> > 
> > Yeah, so I was wondering about that too, which is why I proposed having a dispatcher for combined connectivity routines #1559. However, in practice, I think the most likely situation with the grids we support is that both face/edge and edge/node connectivity are the missing connectivities, which is why they were interested in doing this. There could definitely be a discussion on the design, but I think that's out of scope for this PR.
> 
> Ah, yes that makes sense, and seems like a good solution. I also agree a full design like that should probably be scoped to a different PR. Before this PR merges I would still be interested to hear from other uxarray developers/users to confirm whether the increased memory cost is an acceptable tradeoff for the possible huge speedup in the meantime. E.g., @erogluorhan, @rajeeja, @rljacob, if you get a chance to look into this, what are your thoughts?

I realize I never responded to this. 

- Dispatcher: Agreed. Out of scope for this PR but we could look into it separately (#1559).
- Memory usage: 
   - The extra bytes in the static output: They seem acceptable to me in this context where we have extreme performance optimization, and the models we support seem to have these two connectivities either going together or lacking together (per @cmdupuis3 's observation) make this combined approavh even more logical.
   - Also, the [peak-mem results @cmdupuis3 provided](https://github.com/UXARRAY/uxarray/pull/1560#issuecomment-5147389118) seem like the approach here even settles peak memory usage. The extra bytes are measured for stored output size, but the peak memory during combined construction seems significantly improved. 

All that said, there seems like a big win on both performance and memory utilization in this PR.

erogluorhan (commented):
@cmdupuis3 a quick observation:
- I believe this PR is not documenting this new way of combined constructions, so please mention it in both `face_edge` and `edge_node` docstrings for the user's reference.

test/grid/grid/test_connectivity.py
cmdupuis3: What I'm talking about is, if you check the test runs on, say, commit [8b37dc7](https://github.com/UXARRAY/uxarray/pull/1560/commits/8b37dc7752b820c3bf313aa235337977367d599b), you will see that five tests consistently fail. These five tests are all failing because the algorithm as it was when I picked up this PR were sorted differently than the legacy tests expect. If we don't care about keeping that legacy behavior, we can remove the sorting algorithm, but I think at this point it's better to keep it and have the sorting because it will be easier to reason about if, in the future, another change to this algorithm changes the sorting order again.

uxarray/grid/connectivity.py
cmdupuis3: It's true that this change is slower at this scope, and the benchmarks show that, but it's relatively small compared to the face_node speedup, and having `.values` here forces a materialization that slows it down.

I reworked it a bit though and got a dask-parallel version that's hopefully faster

test/grid/geometry/test_centroids.py
Sevans711: I still don't understand, why does the test need to be changed? It sounds like the original test passes on main and also passes after your changes. Why is it "wrong"?

You mentioned it depends on which edge is at index 0, but that should be known exactly, right? This test is always just acting on this triangle:
```python
    test_triangle = np.array([(0, 0, 0), (-1, 1, 0), (-1, -1, 0)])
    grid = ux.open_grid(test_triangle, latlon=False)
```
It seems fine to me if part of the test relies on coincidence with that triangle in particular (though it could be nice to have a clarifying comment in that case).

My preference would be to keep the original test unless I can understand why it is wrong. Any of these could work as a way forward: (1) clarify why the original test is wrong, (2) restore the original test but also keep your additions here, so that the test tests more things, or (3) restore the original test and remove your additions here.

benchmarks/mpas_ocean.py
```diff
@@ -89,6 +89,8 @@ def time_to_geodataframe(self, resolution, exclude_antimeridian):
 
 
 class ConnectivityConstruction(DatasetBenchmark):
+    number = 1
```
Sevans711, line 164: What is this doing?

uxarray/grid/connectivity.py
```diff
@@ -125,24 +131,24 @@ def _populate_n_nodes_per_face(grid):
...
-        data=n_nodes_per_face,
+        data=n_nodes_per_face.data,
         dims=ugrid.N_NODES_PER_FACE_DIMS,
         attrs=ugrid.N_NODES_PER_FACE_ATTRS,
     )
 
 
-@njit(cache=True)
-def _build_n_nodes_per_face(face_nodes, n_face, n_max_face_nodes):
+@njit(cache=True, nogil=True)
```
Sevans711, line 152: Can you clarify why is nogil=True here (and not on other numba-compiled functions)? I'm less familiar with the nogil=True flag, so it might be obvious from the code and just need clarification in response thread here. If it isn't obvious from the code, please add a comment to the code to clarify as well.

uxarray/grid/connectivity.py
```diff
@@ -162,77 +168,222 @@ def _populate_edge_node_connectivity(grid):
...
+        dims=ugrid.FACE_EDGE_CONNECTIVITY_DIMS,
+        attrs=ugrid.FACE_EDGE_CONNECTIVITY_ATTRS,
     )
 
 
-def _build_edge_node_connectivity(face_nodes, n_face, n_max_face_nodes):
-    """Constructs the UGRID connectivity variable (``edge_node_connectivity``)
-    and stores it within the internal (``Grid._ds``) and through the attribute
-    (``Grid.edge_node_connectivity``).
+@njit(cache=True, inline="always")
```
Sevans711, line 211: Can you also clarify why is inline="always" here (and not on other numba-compiled functions)?

test/grid/grid/test_connectivity.py
Sevans711: I'm confused. The commit you linked is in this PR's branch, not a commit on main. To clarify, my question here isn't intended to be "what is the reason for tests about sorting?", I am trying to ask "what is the reason for adding sorting at all?"

Are either of these the situation here?:
- (A) "Sorting already exists on `main`, `main` already produces sorted results, but the optimized algorithms here didn't produce sorted results by default, so we need to add sorting if we want to match the behavior of `main`; these tests were added to ensure the results after this PR match the results currently on `main`."
- (B) "Sorting does not exist on `main`; this PR introduces sorting."

uxarray/grid/connectivity.py
Sevans711: (keeping this comment unresolved until at least one check shows faster n_nodes_per_face performance on a large grid)

Sevans711 (commented):
Took a close look at this one just now. It does look like it is getting closer to being ready overall, I have a few small questions and I left some inline comments accordingly.

The main blockers for me would be a couple bigger questions which I do not think are resolved yet. These are: "why does this PR introduce lexicographic ordering?" and "do the changes here actually speed up n_nodes_per_face for large grids?" (See replies to older comment threads for more details).

benchmarks/mpas_ocean.py
cmdupuis3: This has to be here for ASV to only run this once. In ASV's default mode, it will run tests multiple times and average them, and if we're worried about caching, we can't have that. So, ASV sees the magical `number` variable and interprets that as the number of times to run a test.

Smells like Fortran to me, so I get why it raises red flags.

test/grid/geometry/test_centroids.py
Sevans711: Okay so, I played around with this a bit more, and now I understand more about why the new test is a good idea. I still think the old test is fine too, though, and should not be removed, unless there is a more compelling reason to remove it. Keeping it in is sort of like also asserting "uxarray's coincidental ordering of edges has not changed in this case." I understand you want sorting-related tests to be the ones that fail if uxarray edge ordering changes, but I don't really see a problem with other unrelated tests also failing if uxarray edge ordering changes.

test/grid/grid/test_connectivity.py
Sevans711: I played around with this more, and confirmed, for example, that edge_node_connectivity and edge_y remain exactly the same when computed on `main` and on this PR's branch now. I believe that means the situation we are in is (A). The sorting tests here should catch any edge re-ordering changes going forwards, too, so they serve as a guarantee that the sorted order will continue to exist; future changes to order would cause this test to crash.

With that in mind, I think this point is resolved. Sorting exists on main already, but there were no tests about it. This PR introduces tests to that effect, but does not change the behavior/order compared to the behavior currently on `main`

Feel free to mark this one as resolved if you agree!

cmdupuis3: Yes, (A) is basically the case here

test/grid/geometry/test_centroids.py
erogluorhan: I understand that even though the original test passed on main, it was coincidental because of edge 0 connecting nodes (-1, -1, 0) and (-1, 1, 0), giving us raw mean of (-1, 0, 0) ) that already has norm 1.0--no need to be further normalized in that particular case, but a choice of different node values would break on main. So, it tested against unnormalized / raw means, which is an otherwise-shouldn't-be-an-existing-test case. 

I think the current changed version test against the normalization that's actually needed and for all the edges rather than only the first one.

@cmdupuis3 I appreciate your clarification in this thread, it was really helpful. When there is a silent change in PRs though, it is a black-box to the reviewers, so upfront clarification can always go a long way.

Sevans711: Ah, thank you @erogluorhan that comment helped me fully understand that the "coincidentalness" isn't just "only the first edge is being tested" it is also "the norm of the edge being tested happens to be 1 already".

With that in mind, and after playing around with it directly to see the old test still would pass on this PR, I agree it is okay to drop the old test and just move forward with the changed version here.

Marking this as resolved!

benchmarks/mpas_ocean.py
Sevans711: Ah, okay, that makes sense. Indeed ASV does have many "magical" keyphrases, and not always easy to understand them from the docs pages.

Could you leave a small comment in the code near this line to clarify the reason for wanting this test to be run only once? Your explanation here made sense, just hoping to help keep the reasoning clear in the future.

uxarray/grid/connectivity.py
cmdupuis3: Added this comment:

>     # This is in lieu of an xarray equivalent to `da.compute(a, b)`. We traverse the grid once to gather both variables, possibly as chunks if dask is enabled

Is this bit resolved?

benchmarks/mpas_ocean.py
cmdupuis3: I would be disinclined to do that, since it is a standard ASV parameter; it's more that if we do it here, we'd need to do it every time and then keep all those comments in sync. Seems more trouble than it's worth to me.

uxarray/grid/connectivity.py
cmdupuis3: With the new ufunc code, this issue is mooted.

uxarray/grid/connectivity.py
cmdupuis3: `nogil=True` should be used anywhere you can use it, it's just very restricted to low-level parallelizeable code, so you can't use it with most python objects. This is also a standard part of numba, so I see no need to comment on that in the code.

uxarray/grid/connectivity.py
cmdupuis3: Inlining is more for small functions that get called a lot. Most of the time you won't need this because the overhead of calling the function is relatively small, but for this one it's noticeable (around 15%).

benchmarks/mpas_ocean.py
Sevans711: I agree there's no need to document the parameter itself, given that it is an ASV standard parameter. But, even knowing that it means "only run the test once", to me it's not obvious at all from the code itself why you would actually want that here. Something like this would help a lot:
```
number = 1  # rerunning this benchmark gives misleading results due to caching
```
Not sure if that's the best phrasing but hopefully it clarifies my thoughts a bit?

Another possibility is to add a comment about caching anywhere else in that benchmark, it doesn't have to be attached to the `number = 1` line. It just wasn't obvious from the code itself that caching is an important thing to consider here.

uxarray/grid/connectivity.py
Sevans711: Sounds reasonable, marking as resolved!

uxarray/grid/connectivity.py
Sevans711: Okay, I looked into it more because I really wanted to understand why it is being used here _but not elsewhere_.

Advice from the internet / LLMs generally seems to be to only use nogil=True on functions which you actually expect to call from mulithreaded parts of the code. Following that advice means that including nogil=True is a "self-documenting" way to clarify to readers that you expect this to be called from multithreaded code. This feels mostly like a stylistic preference to me, though, so I don't really know whether it is good advice or not. Could you clarify, would you agree or disagree with that advice?

Also, in this particular case, are you including nogil=True as a way to signal that you expect `_build_n_nodes_per_face` to be called from somewhere with multiple threads enabled? Or, are you just including it because this function happens to be compatible with it?

benchmarks/mpas_ocean.py
cmdupuis3: Done

Sevans711: Thank you, this helps and looks good to me!

uxarray/grid/connectivity.py
cmdupuis3: This is not a stylistic preference, this function is called by a ufunc in `_populate_n_nodes_per_face`, which makes it an ideal target for parallelization. In general, Python's GIL is something the language as a whole is trying to move away from. Until they figure that out, we have `nogil=True` for when we can guarantee that parallelization won't create race conditions or corrupt data.

Sevans711: Ah! Okay, that makes sense. Rephrasing to make sure I understand: this function can be reached from parallelized / multithreaded processes via that route, which means the nogil=True might help with efficiency improvements in that case. Marking as resolved!

uxarray/grid/connectivity.py
cmdupuis3: > (keeping this comment unresolved until at least one check shows faster n_nodes_per_face performance on a large grid)

That isn't something I can show, because the performance gain only shows up in the edge_node benchmarks due to the laziness change. The speed difference for `n_nodes_per_face` itself is about the same or slightly worse. The benefit of this ufunc approach shows up more in peakmem (I think about 2x) but the connectivity peakmem benchmarks aren't in yet. But in the broader scheme, we need the laziness change here to enable the speedups on the larger connectivity routines; this one is fairly small.

Sevans711: I'm not really understanding why swapping to different code (here, ufuncs) would make this any less relevant. I have used `n_nodes_per_face` many times without using `face_edges` or `edge_nodes` connectivity variables; `n_nodes_per_face` does not depend on those connectivities. So, I think it is fair to ask that this PR should avoid merging `n_nodes_per_face` performance degradation into `main`. Am I misunderstanding something?

Does the `n_nodes_per_face` peakmem usage decrease in some cases? If there is a tradeoff here between `n_nodes_per_face` speed and `n_nodes_per_face` peakmem, that is a different story.

If you can show at least one realistic check where speed or peakmem usage improve, even if it is a spot check / from a script, rather than something included in the full benchmarking suite, that should be enough to convince me it looks okay.

cmdupuis3:
pre-commit.ci autofix

uxarray/grid/connectivity.py
cmdupuis3: If you look at `_populate_edge_node_connectivity`, you will see that you cannot call it without forcing a materialization of `n_nodes_per_face`. Keeping the computation of `n_nodes_per_face` lazy means that we don't have to force the materialization of the entire grid of `n_nodes_per_face` values, we can instead materialize chunks, which likewise propagates to allowing chunked computations of `_populate_edge_node_connectivity` which is where the materialization would happen.

We can expect about a 2x peakmem reduction of `n_nodes_per_face` in isolation, but I am waiting until the other benchmark PRs are in order until I say anything definitive about this because a good measurement is difficult with the current benchmark suite, and I would need to have the connectivity benchmarks finished and merged. I don't think it's worth holding up the PR for a routine that takes about a 1000th of the time that the optimization targets currently take. If you think there are more potential optimizations of `n_nodes_per_face`, we can always have a follow-up PR.

Sevans711: I'm currently not able to reproduce the peakmem reduction of `n_nodes_per_face`. I copied the `peak_allocated` method from the benchmarking suite and ran:
```python
grid = ux.open_grid(...)
grid.n_nodes_per_face   # warm up the numba cache
grid = ux.open_grid(...)   # get a grid without n_nodes_per_face in it
peak_allocated(grid.n_nodes_per_face)
```
on both `main` and this PR branch, and saw the same memory usage to within 0.1%. This was for a small but not tiny grid, having 157k faces. When I forgot to include the "warm up the numba cache" step I saw roughly double peakmem usage, but that is true on both `main` and on this PR branch.

I also tried copy-pasting the old implementation of `_populate_n_nodes_per_face` and `_build_n_nodes_per_face` into this PR branch locally. When I did that, I saw the same performance for `grid.edge_node_connectivity` compared to the current state of this PR. (Same memory usage to within 0.1%, same runtime to within a few %; the runtime is variable and I'm just doing spot checks not an official benchmarking suite.)

Please provide at least one example where restoring the old `_populate_n_nodes_per_face` and `_build_n_nodes_per_face` implementations causes worse performance than the changes to those methods proposed here.

Alternatively, if other uxarray devs want to move forward on this and accept the `n_nodes_per_face` performance degradation in order to get the other content merged into main faster, I will be happy to drop this point. cc @erogluorhan, @rajeeja, @dylannelson

Please let me know if it would be easier to discuss further offline and report back once we are on the same page!

uxarray/grid/connectivity.py
```diff
@@ -162,77 +168,222 @@ def _populate_edge_node_connectivity(grid):
...
-    edge_nodes, inverse_indices, fill_value_mask = _build_edge_node_connectivity(
-        grid.face_node_connectivity.values, grid.n_face, grid.n_max_face_nodes
-    )
+    # Check edge coordinates already exist, if they do this might cause issues
+
+    if "n_edge" in grid.dims:
+        stale = sorted(n for n in grid._ds if ugrid.EDGE_DIM in grid._ds[n].dims)
+        raise ValueError(
+            f"Constructing 'edge_node_connectivity' on a grid that already has "
+            f"edge-centered variables ({', '.join(stale)}). Constructed edges are "
```
erogluorhan: What do you mean by "edge-centered variables"? It sounds confusing to me as it sounds like data variables that are stored on edge-centers, but it shouldn't be the case since we are talking only about thr grid geometry here, not data. Consider rephrasing.

uxarray/grid/connectivity.py
```diff
@@ -162,77 +168,222 @@ def _populate_edge_node_connectivity(grid):
...
+            f"numbered in lexicographic node-pair order, which need not match the "
+            f"numbering those variables were stored with; they may no longer refer "
+            f"to the same edges."
```
erogluorhan: Can you elaborate on this? 

If the input grid has edges defined but no explicit edge-node connectivity, will we never be able to create that connectivity?

Even if that's the case, UXarray should still be able to construct connectivity that respect the edge ordering in the existing edges, and I believe this has been the case so far with our current connectivity construction.

uxarray/grid/connectivity.py
erogluorhan: I think we can move forward with this; given how significant the performance optimization is with the rest of this PR and the degradation here can be considered negligible. However, I have some asks from @cmdupuis3 along with some notes on this thread:

- The motivation with changes to this is not documented anywhere in this PR; neither in the several linked issues, nor in the description. So, that's not obvious to the reviewers why this is happening. So, please describe such changes and the reason for them in the PR description always.
- "We can expect about a 2x peakmem reduction of n_nodes_per_face in isolation" -> I have a suspicion that this is related to bundled Dask graphs in `_populate_edge_node_connectivity(grid)` but it is really hard to confirm that unless you justify the claim or show it through some tests. None of them happening in this thread makes review a little hard. So, please justify.
- I suspect discussion of the "peakmem" in this thread involves some gaps between how @cmdupuis3 sees it (as mentioned in the previous bullet) and how @Sevans711 interprets it: @Sevans711 's reproduction might not be using chunked grids, and if the previous bullet's suspicion holds, this may be the main disconnect.

erogluorhan (commented):
Please see below a few comments:

uxarray/grid/connectivity.py
Sevans711: Indeed, I did not utilize chunked grids. From the original phrasing of:
> We can expect about a 2x peakmem reduction of `n_nodes_per_face` in isolation

it was unclear to me if this was expected only on chunked grids or on all grids. I'm also not sure whether this was actually tested and observed, or just an estimation/guesswork. If it is the former, i.e. you actually already ran a test like that and saw that improvement, please clarify! (Please confirm it's not just numba compilation/caching though, as I also saw 2x apparent "improvement" before properly accounting for numba caching.)

Another way to justify would be to demonstrate that these changes are actually necessary to improve the connectivity performance. This could be done as follows: (1) make a new branch where you revert the changes to the `n_nodes_per_face` routines, but keep all other changes from this PR, and then (2) run spot checks of connectivity timings. Find at least one case where these spot checks behave worse on the new branch. When I tried this, I wasn't able to see any significant difference between this PR and that branch, but I also didn't test any chunked grids.

All that said, it sounds like @erogluorhan wants to move forward with this PR, so, if you agree, please feel free to mark this conversation as resolved without addressing the points I raised here. I will then do another review (probably will just be a quick "approve", as I think all my other comments were resolved, but I will double-check just to make sure).

uxarray/grid/connectivity.py
cmdupuis3: Yeah the phrasing is off. Updated.

uxarray/grid/connectivity.py
cmdupuis3: So, what's happening here is that if a grid *already* comes with edge_node connectivity, that would be loaded. It would presumably be in some order, but we don't know what order that actually is. If `_populate_edge_node_connectivity` gets called on that `Grid`, it's going to reconstruct the edge_node connectivity in lexicographic order, which may not match the original order. So the idea is that if you have variables indexed by edge_node, recalculating edge_node connectivity could potentially scramble the index order.

I changed this message to hopefully be clearer about what's happening.

uxarray/grid/connectivity.py
erogluorhan: Sounds great!

uxarray/grid/connectivity.py
erogluorhan: That sounds good, but what is concerning to me is that this `if-check` will trigger even when there is only edge coords but not edge connectivities present in the grid, e.g. `edge_lon`, `edge_lat`. What do you think?

Sevans711 (approved):
All of my prior comments have been resolved and no glaring issues popped out at me from a final read-through, so I'm happy to approve!

uxarray/grid/connectivity.py
cmdupuis3: @erogluorhan Maybe `if "edge_node_connectivity" in grid._ds` would be more accurate?

erogluorhan: Yeah, I think that is safer

rajeeja: No, it's not reachable. All three callers check "edge_node_connectivity" not in grid._ds before calling, so a grid that already has edges from a file returns them straight from the property and never gets here.

rajeeja:
Going through this in pieces, will have a few more comments later today. Good PR overall and happy to be reviewing it.

test/grid/grid/test_connectivity.py
```diff
@@ -63,6 +98,83 @@ def test_connectivity_build_face_edges_connectivity(gridpath):
...
+        for cur in range(n_edges):
+            start_node = face_nodes[face_idx, cur]
+            end_node = face_nodes[face_idx, (cur + 1) % n_edges]
+
+            expected = sorted((start_node, end_node))
+            actual = sorted(edge_nodes[face_edges[face_idx, cur]])
+            assert actual == expected
+
+        # Remaining slots stay padded
+        assert np.all(face_edges[face_idx, n_edges:] == INT_FILL_VALUE)
```
rajeeja, line 177: @cmdupuis3 The new `ValueError` in `_populate_edge_node_connectivity` can't be reached from any caller, so it's untested and shows as uncovered. This MPAS mesh already loads with `edge_node_connectivity` in `_ds`, so it pins the guard without a new fixture.

```suggestion
        assert np.all(face_edges[face_idx, n_edges:] == INT_FILL_VALUE)


def test_connectivity_edge_node_refuses_to_overwrite(gridpath):
    """Test that rebuilding edges on a grid that already has them is refused."""
    from uxarray.grid.connectivity import _populate_edge_node_connectivity

    # This mesh supplies verticesOnEdge, so the grid loads with edges already present
    uxgrid = ux.open_grid(gridpath("mpas", "QU", "mesh.QU.1920km.151026.nc"))
    assert "edge_node_connectivity" in uxgrid._ds

    with pytest.raises(ValueError, match="already has"):
        _populate_edge_node_connectivity(uxgrid)
```

uxarray/grid/connectivity.py
erogluorhan: You're right, callers already handle that check.

@cmdupuis3 if you want to handle this check only here, that's okay, but remove the  same check from all the other callers. 

Also on a second thought, should we really raise a `ValueError` or just silently skip? I think the latter.

Sevans711: I don't think any other `_populate...` functions have internal logic like "if check: actually, don't populate anything". It would feel strange to me to add that here, it would not follow my expectations that "calling `_populate...()` should actually populate something". If there is some reason it can't actually run as expected, I would want it to crash (with ValueError if it is fundamentally impossible for the given inputs, or NotImplementedError if it is possible but would just need a cleverer algorithm which doesn't exist yet).

erogluorhan: We are talking about not populating again when there is already `edge_node_connectivity` present in the `Grid` (i.e. in order not to possibly break lexicographic order in this case). It can occur in several cases though, i.e. either when `grid.edge_node_connectivity` is accessed after the first time, or grid.face_edge_connectivity is being populated but `edge_node_connectivity` already exists.). 

I don't agree with throwing an error, i.e. making code crash in that case doesn't make sense since the user has everything ready, and why should avoidance of re-populating an existing property crash the code, shouldn't it be okay with a warning instead?

If we don't want if-check within `_populate...()`, remove it since all of the callers are already doing this check and move the above message to where it is called (only for the case of simultaneous construction with `face_edges` I believe)

rajeeja:
Take outCSne30.ug. Renumber its face_edge_connectivity edge ids, put on a fresh grid with
no edge_node_connectivity. Legal UGRID.

- Touch grid.n_edge. Main: 5400/5400 rows survive. This PR: 5400/5400 overwritten. No error,
  no warning.
- Cause: the PR builds both connectivities in one pass, so _populate_edge_node_connectivity
  now writes two arrays. The guard at connectivity.py:173 still checks one.
- Second write at connectivity.py:200 walks past it.
- The guard's reasoning is right. Constructed edges are lexicographic and won't match file
  order, which is what its own error message says. It just needs to cover both variables the
  function writes now.
- Main isn't correct here either, it keeps the file's array while the new
  edge_node_connectivity is lexicographic, so the two disagree silently.
- Untestable today: no mesh under test/meshfiles has face_edge_connectivity without
  edge_node_connectivity. Same hole as test_connectivity.py:176, one variable over.

Two smaller things in utils.py:529-536. The comment on MIN_ADAPTIVE_SORT_SIZE = 16 says a
bucket that size can't exceed the shift budget, which holds only because
MAX_SHIFTS_PER_EDGE = 8: max inversions are size(size-1)/2 against a budget of 8*size, so
size <= 17. That coupling isn't written down or asserted anywhere, and dropping MAX to 4
makes the comment false at size 9.

And test_connectivity_bucket_sort says the 500 bucket falls back to the heap sort, which it
does, but the test only checks the final order and insertion sort alone would produce the
same. Raise MAX_SHIFTS_PER_EDGE and the heap path goes uncovered with the test still green.
_heap_sort_bucket isn't imported by any test.

uxarray/grid/connectivity.py
cmdupuis3: @rajeeja  Good catch, after looking at the call sites, it looks unreachable to me. On digging deeper though, I found that where Philip left this branch, this block was included but commented out:

```Python
    # if (
    #     "edge_node_connectivity" not in grid._ds
    #     or "inverse_indices" not in grid._ds["edge_node_connectivity"].attrs
    # ):
    #     _populate_edge_node_connectivity(grid)
```

This would have been a case where you'd want to have the check on `if "edge_node_connectivity" in grid._ds:`. But part of the point of this branch is to remove the need for `inverse_indices`, so without that, the check becomes degenerate. 

I think with that, it makes sense to just remove it.

@erogluorhan To me, it's about keeping the state of Grid consistent. Having edge_node_connectivity in a different state than the rest of the Grid object seems like it would get very confusing. But as Rajeev pointed out, this particular guard is unreachable anyway.

cmdupuis3:
@rajeeja Wait, so if I understand correctly, this issue is more that the guards at call sites don't check both connectivities, so without having `if "edge_node_connectivity" in grid._ds or "face_edge_connectivity" in grid._ds:`, we still have the possibility of desynching between these two connectivity variables, right? It would just be a different mode of desynch than I was thinking of.

I'll see about consolidating the two sorting parameters into one and moving it to `constants.py`.

rajeeja:
Yes, that covers both writes now, so the n_edge touch fails loud.

cmdupuis3:
@rajeeja As per a convo yesterday, @erogluorhan and I think the edge_node indexing situation is still a bit confusing, so I want to rethink it slightly before we commit to merging.

erogluorhan (approved):
Great!

PR #1568, opened 2026-07-13
Honor quadrature kwargs in calculate_total_face_area

### LINKED ISSUE #1567: Face area documentation is confusing / wrong
### Version

v2026.6.0

Source

### What happened?

Face area calculations currently do not depend on kwargs as promised in the docstring of `Grid.calculate_total_face_area()`. Also, are face areas returned in meters^2 or in steradians? On the unit sphere these are the same, but I haven't yet checked what happens when the `grid.sphere_radius` is not 1. Either way, it would help to clarify in the docstring, and maybe add a test for this unless one is already present.

The kwargs issue shows up also on the Face Area Calculations docs page (https://uxarray.readthedocs.io/en/latest/user-guide/area_calc.html), where section 2 claims to use a different method for computing areas, and that "For the result above, notice that the area is slightly different than the first calculation we made." However, the result above is not different than the first calculation at all.

Inspecting the code itself reveals:
```python
# --- in grid.py --- #
class Grid:
    def calculate_total_face_area(
        self,
        quadrature_rule: str | None = "triangular",
        order: int | None = 4,
        latitude_adjusted_area: bool | None = False,
    ) -> float:
        """Function to calculate the total surface area of all the faces in a
        mesh.

        Parameters
        ----------
        quadrature_rule : str, optional
            Quadrature rule to use. Defaults to "triangular".
        order : int, optional
            Order of quadrature rule. Defaults to 4.
        latitude_adjusted_area : bool, optional
            If True, corrects the area of the faces accounting for lines of constant lattitude. Defaults to False.

        Returns
        -------
        Sum of area of all the faces in the mesh : float
        """

        return np.sum(self.face_areas.values)
```

This does not appear to incorporate the kwargs as promised.

I expected calculate_total_face_area() to respect the `quadrature_rule` and `order` kwargs as promised, probably giving different results with different values of those kwargs. However, the code does not appear to utilize those kwargs at all in any way, and I got the same results when I varied them in a small test case.

```python

```

vakudo:
Closes #1567.

`Grid.calculate_total_face_area()` documents `quadrature_rule`, `order` and `latitude_adjusted_area` parameters, but the body ignored them and always returned `np.sum(self.face_areas.values)` — the cached face areas computed with the default parameters. So calling it with, for example, `quadrature_rule="gaussian", order=5, latitude_adjusted_area=True` gave exactly the same total as the default triangular call.

### Fix
- Keep the fast path: when the parameters are the defaults (`triangular`, `order=4`, no latitude adjustment) it still returns the cached `face_areas`, which also preserves the theoretical equal-area values used for HEALPix grids.
- Otherwise it routes through `_compute_face_areas(quadrature_rule, order, latitude_adjusted_area)` so the requested quadrature rule, order and latitude adjustment actually take effect.

The default behavior (and the HEALPix path) is unchanged.

### Tests
Added `test_calculate_total_face_area_respects_quadrature_kwargs`, which asserts the corrected result differs from the uncorrected one and matches the expected `TRI_AREA` / `CORRECTED_TRI_AREA` constants to 6 decimals. The existing `test_face_areas_calculate_total_face_area_triangle` used a `decimal=3` tolerance too loose to catch this (the correction for that triangle is ~2.8e-4). The new test fails on the previous code and passes with the fix; the full `test/grid/grid/test_areas.py` suite passes.

Note: the "Face Area Calculations" docs page (section 2) also states the second calculation gives a slightly different area than the first — with this fix that statement becomes true when non-default quadrature settings are used.

### DIFF
--- a/test/grid/grid/test_areas.py
+++ b/test/grid/grid/test_areas.py
@@ -30,6 +30,29 @@ def test_face_areas_calculate_total_face_area_triangle(mesh_constants):
     nt.assert_almost_equal(area_gaussian, mesh_constants['CORRECTED_TRI_AREA'], decimal=3)
 
 
+def test_calculate_total_face_area_respects_quadrature_kwargs(mesh_constants):
+    """calculate_total_face_area must honor its quadrature_rule/order/
+    latitude_adjusted_area kwargs instead of always returning the cached
+    default-parameter face areas."""
+    verts = [
+    [[0.02974582, -0.74469018, 0.66674712],
+    [0.1534193, -0.88744577, 0.43462917],
+    [0.18363692, -0.72230586, 0.66674712]]
+    ]
+
+    grid_verts = ux.open_grid(verts, latlon=False)
+
+    area_default = grid_verts.calculate_total_face_area(
+        quadrature_rule="triangular", order=4)
+    area_corrected = grid_verts.calculate_total_face_area(
+        quadrature_rule="gaussian", order=5, latitude_adjusted_area=True)
+
+    # the corrected result must actually differ from the uncorrected one
+    assert not np.isclose(area_default, area_corrected)
+    nt.assert_almost_equal(area_default, mesh_constants['TRI_AREA'], decimal=6)
+    nt.assert_almost_equal(area_corrected, mesh_constants['CORRECTED_TRI_AREA'], decimal=6)
+
+
 def test_face_areas_compute_face_areas_geoflow_small(gridpath):
     """Checks if the GeoFlow Small can generate a face areas output."""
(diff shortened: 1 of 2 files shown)


Sevans711 (approved):
@vakudo thank you for your contribution! I also confirmed it runs, correctly gives results which actually depend on relevant kwargs, and that the new test fails on `main` while succeeding after your fix.

Let's merge to main!

I may open a separate issue soon related to the other parts of the face areas API. Looking into the existing methods and your fix helped me notice this. For example, uxarray now properly supports kwargs like `quadrature_rule` publicly in `calculate_total_face_area()`, while the only non-deprecated public interface for getting individual face areas (i.e., `Grid.face_areas`) does not support those kwargs. It seems strange to support them publicly only for the total, and not for individual faces.

Issue #425, opened 2023-08-29
Error checks for area calculation


rajeeja:
**Describe the bug**
Area calculation is incorrect?
For mesh `gridfile_fesom = current_path / "meshfiles" / "ugrid" / "fesom" / "fesom.mesh.diag.nc"`

**To Reproduce**
sum of area comes to be 9776.665489773564

**Expected behavior**
Some message and more checks might be needed.

Sevans711:
This seems fine now? On uxarray 2026.6.0, I am getting:
```python
import uxarray as ux
gridpath = ... # path to `"uxarray/test/meshfiles/ugrid/fesom/fesom.mesh.diag.nc"` on my machine
grid = ux.open_grid(gridpath)
area = grid.calculate_total_face_area()  # area in steradians
print(area)
```
Gives `8.378036745330922`. This seems reasonable to me; `grid.plot()` looks like this:

<img width="597" height="294" alt="Image" src="https://github.com/user-attachments/assets/fd459db8-a512-41c0-a43e-5231ecf6c5a7" />

I can certainly believe this grid covers roughly  67% (`== area / (4 * np.pi)`) of the surface.

If this makes sense to you too please mark as closed!

rajeeja:
@Sevans711 Confirmed — the bad value is fixed; the FESOM mesh now gives ~8.38 sr, which is correct. I'll still add the error check this issue asked for: a warning when total face area exceeds the unit-sphere area (4π), so a malformed grid can't return garbage silently. That's in #1577, which will close this issue.

Sevans711:
Clarifying: the error check proposed to be added to #1577 was removed, because it was originally added in a way that does not respect grid.sphere_radius (when not on unit sphere, area can easily be larger than 4π), and in a way that adds inefficiency to every face_areas computation since `face_areas` doesn't otherwise require summing across all faces.

So, what remains to close this issue? I think doing both of the following would be sufficient:
- open a separate PR (sometime after #1577 merges into main), check inside `grid.calculate_total_face_areas()` (which is already doing a sum across all faces) whether the result is too large, and make warning if it is. This check should compute "too large" as "larger than `4 * np.pi * grid.sphere_radius**2`" not just 4π.
- add a regression test for this fesom grid in particular, which will crash if the total area isn't close to ~8.38 sr.

PR #1577, opened 2026-07-15
Clean up and vectorize the face_areas API

### LINKED ISSUE #1571: Clean up face_areas API

The public-facing API for Grid face areas is now inconsistent in its support for quadrature kwargs like `quadrature_rule` and `order`. The `calculate_total_face_area()` method supports these kwargs, but the only public way to get face areas (`Grid.face_areas` property) does not support them. Related: #1567, #1568.

Suggested fix: un-deprecate and improve the public-facing `compute_face_areas()` method. This would provide public support for quadrature kwargs. The public method was originally deprecated in #1379; pinging @rajeeja and @erogluorhan who worked on that PR. More thoughts here:
- Currently, that method returns (face_areas, jacobian); I would suggest to instead just return face_areas by default, since the method name does not mention jacobian. We could adopt numpy style (e.g. see docstring for np.unique) by adding an optional kwarg like `def compute_face_areas(..., return_jacobian=False)`.
- _(new feature)_ Maybe promote the return type to be UxDataArray (with self as the result.uxgrid) by default, or at least provide a flag to enable that, instead of returning numpy array? This would make it easy to visualize the result!
- Remove `_compute_face_areas()` (use public method instead) or rename it to `_compute_face_areas_and_jacobian()` since it also returns the jacobian.

Additional suggestions / questions:
1. Keep the note in `compute_face_areas()` docstring suggesting to use the `face_areas` property instead for caching; clarify this note is only relevant if using default kwargs. 
2. Clarify in `face_areas` property docstring that `compute_face_areas()` can be used to get result with adjusted quadrature behavior.
3. Improve documentation for kwargs: `quadrature_rule`, `order`, and `latitude_adjusted_area`. Minimum: quadrature_rule and order should specify what options are supported. Ideal: explain what each of these things actually means, either directly or by linking to a paper or online documentation somewhere. E.g., what is a quadrature rule? What does "corrects the area of the faces accounting for lines of constant latitude" actually mean?
4. Why do we provide `calculate_total_face_area()` when it is just `np.sum(grid.compute_face_areas())` (or `np.sum(grid.face_areas)` if using default kwargs)? For me personally, the easiest thought process goes like: "I want total face area --> I should get face areas, then sum them." The existence of `calculate_total_face_area()` adds work, because when I see it I also think "wait, should I be using that method instead? Is there some quirk to getting the total area such that np.sum() isn't good enough?" until I inspect and see it really is just sum(). I would suggest to remove this method, or at least add near the top of its docstring something like "equivalent to `grid.compute_face_areas().sum()`"
5. Should the method `get_all_face_area_from_coords` be marked as private (i.e., leading underscore)? It doesn't seem to appear on the public uxarray docs, and it seems to be very low-level (inputs are x, y, z, face_nodes, face_geometry). Similarly, should the other methods in `grid/area.py` by marked as private?

rajeeja:
Closes #1571.

1. Un-deprecate `compute_face_areas()` as the public quadrature-capable entry point — returns areas by default, with `return_jacobian` and `as_uxarray` flags (the latter returns a `UxDataArray` paired with the grid).
2. Rename the internal worker to `_compute_face_areas_and_jacobian` and privatize the low-level `area.py` helpers.
3. Document the `quadrature_rule`/`order`/`latitude_adjusted_area` options; note `calculate_total_face_area` is equivalent to `compute_face_areas().sum()`.
4. Vectorize `get_all_face_area_from_coords` with `prange` and hoist the per-face quadrature setup out of the loop — ~6x faster with identical areas; adds a `FaceAreas` asv benchmark.
5. Plot face areas in `area_calc.ipynb` and the geometric/theoretical ratio in `healpix.ipynb` using the new `as_uxarray` flag.


Sevans711:
Thank you for these changes! I added run-benchmarks label but then realized I'm not sure if there are any benchmarks which actually rely on face_areas. If there's no visible impact, then would you be able to add at least one face_areas benchmark? If the impact is visible already, I don't feel strongly about needing a face_areas specific benchmark.

I'll come back to this soon (probably tomorrow), and if the benchmark impact is visible then I will review the rest!

Sevans711 (changes requested):
Looking through the plan of action discussed in #1571:

1. _(Looks good to me)_ **Un-deprecate and make `compute_face_areas()` the public, kwargs-capable entry point.** Return `face_areas` only by default, with `return_jacobian=False` (numpy-style) for callers that also want the jacobian.
2. _(Looks good to me)_ **Rename `_compute_face_areas` → `_compute_face_areas_and_jacobian`**
3. _(Not completed as planned; see notes below)_ **Add an optional `as_dataarray=False` flag** to return a `UxDataArray` (`result.uxgrid = self`) so per-face areas are easy to plot.
4. _(Looks good to me)_ **Keep `calculate_total_face_area()` but document it as equivalent to `compute_face_areas().sum()`**.
5. _(Looks good to me)_ **Mark the low-level `area.py` routines private**
6. _(Mostly good but I have more notes below)_ **Vectorize the core computation.**
7. _(Mostly good but I have more notes below)_ **Fix the user guide (`area_calc.ipynb`).**
8. _(Looks good, I did a small commit to help)_ improve docstrings for `face_areas` property and `compute_face_areas()`.


More notes / requests / suggestions:
- (3) Originally we discussed returning a UxDataArray object, not an xarray.DataArray. The current version here returns an xarray.DataArray. **Request: make the optional flag return UxDataArray instead.** I think UxDataArray is the way to go; it is more convenient, and I think the whole point for this optional flag is to return something more convenient.
  - For example, plotting result from current version looks something like: `ux.UxDataArray(uxds.uxgrid.compute_face_areas(as_dataarray=True), uxgrid=uxds.uxgrid).plot()`. But, with this requested change, it would become simpler: `uxds.uxgrid.compute_face_areas(as_dataarray=True).plot()`.
  - I wonder if `as_dataarray` is the wrong name here... **Request/suggestion: rename the flag to `as_uxarray`** or `as_uxdataarray` to clarify it is a UxDataArray instead? Of the three names, my preferences would be `as_uxarray` (concise & clear) > `as_uxdataarray` (verbose & clear) > `as_dataarray` (sounds like xarray not uxarray). (Let me know if you want me to take on the task of setting this up!)
- (7a) **Request: plot face_areas somewhere in area_calc.ipynb**, using the new optional flag. Either in a new section, or inside an existing section, e.g. maybe it makes sense to add in the "Calculate Area from Multiple Faces in Spherical Coordinates" section?
- (7b) In the healpix.ipynb guide, there is a section comparing geometric face area calculations versus theoretical healpix areas. This has been updated to now be equivalent to grid.compute_face_faces() / grid.face_areas. There is also some numerical analysis of how the results vary across the grid. **Request: plot this face_areas ratio across the grid.** After making changes in response to point (3) above, it should be as simple as: `(grid.compute_face_faces(as_uxarray=True) / grid.face_areas).plot()`.
- (6) I also measured roughly 6x faster face_areas computations on main versus on this branch. This is a great improvement! Some of the ASV benchmarks show improvements in reducing peak memory usage, but none of them show this significant speedup. **Request: add at least one benchmark that gets improved by these changes.**

Sevans711:
@rajeeja thank you for adding the total face area error check. I don't want to interrupt you if you're still working on things, but wanted to share my thoughts now while I am thinking about it:

1. Maybe this check should only be in calculate_total_face_areas(), or in grid.validate()? It seems unusual to impose inefficiency on the low-level _compute_face_areas_and_jacobian method for a validation check.
2. Flagging that face_areas will be much much larger than 4 * pi for any grids which are not on the unit sphere! From skimming your changes, it is unclear to me whether that would trigger this warning or not. Example non-unit-sphere-grid: 'uxarray/test/meshfiles/mpas/dyamond-30km/gradient_grid_subset.nc'

rajeeja:
@Sevans711 thanks for the review — fixed the items: `compute_face_areas()` now returns a `UxDataArray` via `as_uxarray=True` (renamed from `as_dataarray`); face areas plotted in area_calc.ipynb and the geometric/theoretical ratio in healpix.ipynb using it; and added a `FaceAreas` asv benchmark that captures the speedup.

rajeeja:
@Sevans711 Good catches on both. You're right the check misfires on grids whose `face_areas` are parsed from the file in physical units — e.g. `gradient_grid_subset.nc` returns ~1.5e11 m² and would falsely warn, since the property returns file values, not unit-sphere steradians. There's no reliable way to know the unit at that level, so I've removed the check. #425's actual issue (the wrong value) is already fixed independently; I'll leave #425 open for a proper unit-aware validation later rather than ship a check that false-positives.

Sevans711 (approved):
Looks good! Thank you for implementing all the requests/suggestions I made. I think it is ready to merge!

I did notice a few small issues with type hinting. We could either fix them as part of this PR, or open a separate issue with a more narrowly-scoped PR to avoid blocking this one. Both of these issues already exist on `main`; they aren't being newly introduced by this PR.
1. Type hinting for the result of compute_face_areas() is missing.
2. Type hinting for input arguments is a bit too permissive. For example, it includes `order: int | None = 4,` but there is no support for order=None, so I think it should be `order: int = 4` instead.


Sidenote, I especially enjoyed seeing the face_areas plot in the Face Area Calculations docs page, I think it looks pretty cool:

Image

rajeeja:
Thanks for the approval @Sevans711! On the type-hint items — since both issues already exist on `main` and weren't introduced here, I'd rather not hold up this PR for them. Opened #1590 to track the cleanup separately. Happy to merge whenever Orhan takes a look.

erogluorhan:
Please resolve the conflicts @rajeeja

rajeeja:
@erogluorhan conflicts resolved, ready for another look.

erogluorhan (approved):

Issue #1580, opened 2026-07-16
Nightly builds failing as of 9 July 2026


cmdupuis3:
### Version

2026.6.1

### How did you install UXarray?

Source

### What happened?

Nightly builds began failing on July 9th. 

The first failure on July 9th is a link check error. I'm rerunning it in debug, results are [here](https://github.com/UXARRAY/uxarray/actions/runs/29067207770/job/87744700866).

Relevant errors in logs:
Broken link: https://iopscience.iop.org/article/10.1086/427976 (403 Forbidden)
Timeout: https://www.osti.gov/servlets/purl/10102115 (connect timeout)
Rate-limited checks and other transient failures
Warning during autosummary: ModuleNotFoundError: No module named 'uxarray.RemapWeights' (autosummary couldn't import a submodule)

The CI upstream is also failing, it seems to be something about not finding xarray in the installed env.

### What did you expect to happen?

I expected it to not do that.

### Can you provide a MCVE to repoduce the bug?

```python
Rerun linkcheck for CI and CI Upstream July 9th.
```

erogluorhan:
Thanks for taking this on, folks!

erogluorhan:
Re Upstream CI: 

Looks like the failure started when the [Install upstream versions](https://github.com/UXARRAY/uxarray/blob/b0487c4d1326adba3f7108f0b58abb2f413ab458/.github/workflows/upstream-dev-ci.yml#L43) job in the ci-upstream workflow started to fail finding [a version that satisfies the requirement scikit-learn (from versions: none)](https://github.com/UXARRAY/uxarray/actions/runs/29549818858/job/87789795127#step:5:354).

`scikit-learn`, `scipy`, and `xarray` are the packages that are being installed through scientific python nightly wheels in our upstream ci worflow, so when `scikit-learn` installation fails, `xarray` is not being installed apparently (and `scipy`).

The `scikit-learn` wheels seem to have been somehow removed from here: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/scikit-learn while scipy and xarray are still available.

Sevans711:
See also: #1514

Sevans711:
[...]

PR #1584, opened 2026-07-17
Fix uxarray CI link-check failure

Sevans711:
Address uxarray ci failing due to link-check; see #1580. Does not yet address the upstream-ci failure.

## Overview
The healpix paper link (https://iopscience.iop.org/article/10.1086/427976) started failing recently for github ci. Replaced with doi link (https://doi.org/10.1086/427976). This might be enough to fix the broken link check?

Note: make sure to un-comment commented part of workflow before merging (maybe there was a more elegant way to tell link-check to run on this PR)?

## PR Checklist

**General**
- [X] Add appropriate labels
- [N/A] Filled out Overview and Expected Usage (if applicable) sections

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
           python -m pytest test
 
   link-check:
-      if: github.event_name != 'pull_request'
+      #if: github.event_name != 'pull_request'  # commented for debugging; UNCOMMENT BEFORE MERGE
       runs-on: ubuntu-latest
       defaults:
         run:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
           python -m pytest test
 
   link-check:
-      #if: github.event_name != 'pull_request'  # commented for debugging; UNCOMMENT BEFORE MERGE
+      if: github.event_name != 'pull_request'
       runs-on: ubuntu-latest
       defaults:
         run:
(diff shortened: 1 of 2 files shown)


Sevans711:
@erogluorhan @cmdupuis3 I believe this fixes the link-check issue, and that our nightly uxarray ci would no longer crash if we merge this.

Slight complication: I commented out the "#if: github.event_name != 'pull_request'" rule for this branch so that the link check would actually run here, instead of being skipped. If I push another commit restoring that rule, the bot will rerun checks without link-check, so you won't actually be able to see that tests pass…. My recommendation: take a look, leave review saying it looks good but request change: that line needs to be un-commented before merging. Once we have two "looks good but please un-comment that line" reviews, we can fix it, then the reviewers can approve, then we can merge.

Alternatively, if there is a way to "force run" link-check on this branch, even while that line exists, maybe we could do that instead? I'm not sure how to do that though, please go for it if you know how.

erogluorhan (commented):
The URL fix seems good to me! Please go ahead and revert the commented out part in the workflow, and I'll approve this after that. 

I don't think there is a better way to tell link-check to run in this case because of how we wrote it (i.e. closes all doors to it when it is a PR).

cmdupuis3 (approved):
Just waiting on the comment removal

Sevans711:
@erogluorhan @cmdupuis3 thank you for taking a look! I just un-commented the relevant line, I believe now it is ready for approval and merge.

erogluorhan (approved):
Thanks a lot for the fix!

Issue #1583, opened 2026-07-17
Census of ".values" and cleanup


cmdupuis3:
### Proposed new feature or change:

There are a number of sites in the codebase that call `.values` erroneously, but most of the sites are legitimate. I'm planning on going case-by-case systematically to try to clean them up.

From Claude:

<details>
<summary>PROBLEM SITES (28)</summary>

| File | Line | Should be | Expression | Notes |
|---|---|---|---|---|
| [uxarray/core/aggregation.py](uxarray/core/aggregation.py:106) | 106 | `dask.Array (lazy)` | `data = uxda.values` | dask branch routes here behind `TODO:`; `_apply_*_dask` is a `pass` stub (known gap) |
| [uxarray/core/aggregation.py](uxarray/core/aggregation.py:174) | 174 | `dask.Array (lazy)` | `data = uxda.values` | dask branch routes here behind `TODO:`; `_apply_*_dask` is a `pass` stub (known gap) |
| [uxarray/core/dataarray.py](uxarray/core/dataarray.py:248) | 248 | `int (xarray metadata)` | `if self.values.ndim > 1:` | `.values.ndim` full compute to read rank; use `self.ndim` |
| [uxarray/core/dataarray.py](uxarray/core/dataarray.py:255) | 255 | `int (xarray metadata)` | `if self.values.size == self.uxgrid.n_face:` | `.values.size` full compute to read size; use `self.size` |
| [uxarray/core/dataarray.py](uxarray/core/dataarray.py:292) | 292 | `int (xarray metadata)` | `elif self.values.size == self.uxgrid.n_node:` | `.values.size` full compute to read size; use `self.size` |
| [uxarray/core/dataarray.py](uxarray/core/dataarray.py:294) | 294 | `int (xarray metadata)` | `f"Data Variable with size {self.values.size} does not match the number of f...` | full compute just to format an error-message f-string; use `self.size` |
| [uxarray/core/dataarray.py](uxarray/core/dataarray.py:298) | 298 | `int (xarray metadata)` | `elif self.values.size == self.uxgrid.n_edge:` | `.values.size` full compute to read size; use `self.size` |
| [uxarray/core/dataarray.py](uxarray/core/dataarray.py:300) | 300 | `int (xarray metadata)` | `f"Data Variable with size {self.values.size} does not match the number of f...` | full compute just to format an error-message f-string; use `self.size` |
[...]

Issue #1575, opened 2026-07-15
Reduce number of imports inside functions


Sevans711:
#### Current behavior:
There are some places throughout the uxarray codebase where import is being used inside functions. The benefit is that the imports are "lazy"; they aren't evaluated until that function actually gets called. This is useful for optional dependencies (see also #1548) but also for any "heavy" dependencies which take a long time or a lot of memory to import. However, the downside is that the code is a bit harder to read and a bit harder to extend, as the import statements get repeated across many functions in the same file.

#### Proposed change:
Imports which are not "heavy" or which are already being imported non-lazily _somewhere_ in uxarray should be changed to consistently be imported non-lazily _everywhere_ in uxarray.

For example, many functions inside uxarray.core.api are currently doing:
```python
def core_api_function(...):
    import xarray as xr
    ...
```
However in various places, such as `uxarray.core.dataarray`, xarray is already being imported at the top of file.

Unless there is a compelling reason to avoid importing xarray at the top of `uxarray.core.api`, I would propose to move these imports to the top of the file. I don't think this will reduce efficiency in any meaningful use-case, and it should help make the code easier to extend. Although, if there is a compelling reason to keep the import inside the functions instead, I would instead propose that a comment be added to each such import at least once per file to explain the reasoning.

Note: it probably makes sense to delay this work until after #1548, which also relocates some import statements.

erogluorhan:
Thanks for suggesting this! 

Even though this is a lower priority, I think it could still be taken on before #1548 since that one is huge and would definitely require more thought process. Also, this one could work well for gaining a bit more familiarity with the code  while not taking as much of a risk as it is in #1548.

> #### Proposed change:
> Imports which are not "heavy" or which are already being imported non-lazily _somewhere_ in uxarray should be changed to consistently be imported non-lazily _everywhere_ in uxarray.

I agree, required imports that are already happening at the top of one of the core modules, such as datarray, api, dataset, etc., can be prevented from being locally imported as well.

PR #1588, opened 2026-07-18
Avoid converting chunked data to Numpy arrays (i.e. `.values` calls)

cmdupuis3:
Partly addresses #1583

## Overview

This PR is to resolve suboptimal usage of `.values` throughout the repo, primarily by deferring to lazy xarray and dask operations. See the table of usage sites in Issue #1583.

This specific PR is for the first two tables of usage sites; dead code and bugfixes aren't really included here. Aside from simple cases, some cases were solved by branching on whether the data type was already chunked, so scalability for some parts will depend on if you're using numpy or xarray/dask arrays at those points.

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


cmdupuis3:
pre-commit.ci autofix

cmdupuis3:
There's some spurious asv results from machine variability, but the benchmarks consistenly show peak-mem reductions and some speedups for cross-sections.

cmdupuis3:
pre-commit.ci autofix

uxarray/core/aggregation.py
```diff
@@ -83,38 +83,43 @@ def _node_to_face_aggregation(uxda, aggregation, aggregation_func_kwargs):
...
+            uxgrid=uxda.uxgrid,
+            data=aggregated_var,
+            dims=uxda.dims,
+            name=uxda.name,
+        ).rename({"n_node": "n_face"})
     elif isinstance(uxda.data, da.Array):
-        # apply aggregation on dask array, TODO:
-        aggregated_var = _apply_node_to_face_aggregation_numpy(
+        # apply aggregation lazily on a dask array
+        return _apply_node_to_face_aggregation_dask(
```
Sevans711, line 91: Style/documentation: it may improve readability to make `_apply_node_to_face_aggregation_dask` and `_apply_node_to_face_aggregation_numpy` have the same return type (e.g., both return xarray.DataArray).

If there's a compelling reason to not do that, please leave a small comment explaining why, and add a note in the docstring of each method saying the type of the returned object

uxarray/core/aggregation.py
```diff
@@ -131,9 +136,55 @@ def _apply_node_to_face_aggregation_numpy(
...
+        aggregation_func,
+        aggregation_func_kwargs,
+    )
+
+
+def _apply_node_to_face_aggregation_dask(
+    uxda, aggregation_func, aggregation_func_kwargs
+):
+    """Applies a Node to Face Topological aggregation on a Dask array, lazily"""
+    import xarray as xr
```
Sevans711: Move xarray import to top of file (see #1575)

uxarray/core/aggregation.py
```diff
@@ -146,39 +197,96 @@ def _node_to_edge_aggregation(uxda, aggregation, aggregation_func_kwargs):
...
+            data=aggregation_var,
+            dims=uxda.dims,
+            name=uxda.name,
+        ).rename({"n_node": "n_edge"})
     elif isinstance(uxda.data, da.Array):
-        # apply aggregation on dask array, TODO:
-        aggregation_var = _apply_node_to_edge_aggregation_numpy(
-            uxda, NUMPY_AGGREGATIONS[aggregation], aggregation_func_kwargs
+        # apply aggregation lazily on a dask array
+        return _apply_node_to_edge_aggregation_dask(
```
Sevans711, line 222: Same style/documentation comment as above, but for node to edge case; make numpy and dask have same return type and/or clarify in function docstrings

uxarray/core/aggregation.py
```diff
@@ -146,39 +197,96 @@ def _node_to_edge_aggregation(uxda, aggregation, aggregation_func_kwargs):
...
+):
+    """Applies a Node to Edge topological aggregation on a Dask array, lazily.
+
+    Mirrors the node-to-face dask path: the data field and the edge-node
+    connectivity enter ``apply_ufunc`` as dask inputs (nothing computed
+    eagerly), ``n_node`` is a single (uncrunked) core dimension, and the leading
+    dimensions are processed blockwise. Each edge reduces over its two endpoint
+    nodes.
+    """
+    import xarray as xr
```
Sevans711: (same as before, move xarray import to top of file, see #1575)

uxarray/core/dataarray.py
```diff
@@ -246,14 +246,14 @@ def to_geodataframe(
...
-        if self.values.ndim > 1:
+        if self.ndim > 1:
             # data is multidimensional, must be a 1D slice
             raise ValueError(
                 f"Data Variable must be 1-dimensional, with shape {self.uxgrid.n_face} "
                 f"for face-centered data."
             )
 
-        if self.values.size == self.uxgrid.n_face:
+        if self.size == self.uxgrid.n_face:
```
Sevans711: These changes look good here; no further revision needed in this area, just flagging #1616 as related. Fixing 1616 will lead to these lines being replaced by something like `if self._face_centered()` instead.

uxarray/core/dataarray.py
```diff
@@ -612,24 +612,29 @@ def integrate(
...
-            integral = np.einsum("i,...i", face_areas, self.values)
+        if self.shape[-1] == self.uxgrid.n_face:
+            # dot product between face areas and the face dimension of the data
+            if isinstance(self.data, np.ndarray):
+                # eager data: a direct einsum avoids xr.dot's per-call overhead
+                integral = np.einsum(
+                    "i,...i", self.uxgrid.face_areas.values, self.values
+                )
+            else:
+                # dask-backed data: xr.dot keeps the reduction lazy
```
Sevans711, line 637: Does integration need a regression test which is sensitive to dask versus numpy? Perhaps, assert that the result is a dask array, and that the values equal the numpy version, similar to the tests already added by this PR into test_topological_agg.

test/core/test_topological_agg.py
```diff
@@ -32,3 +33,31 @@ def test_node_to_edge_aggs(gridpath):
...
+    uxds = ux.open_dataset(gridpath("mpas", "QU", "oQU480.231010.nc"), gridpath("mpas", "QU", "oQU480.231010.nc"))
+    uxda = uxds['areaTriangle']
+
+    for agg_func in AGGS:
+        numpy_result = getattr(uxda, agg_func)(destination='face')
+        dask_result = getattr(uxda.chunk(), agg_func)(destination='face')
+
+        assert numpy_result.dims == dask_result.dims
+        assert np.allclose(numpy_result.values, dask_result.values, equal_nan=True)
+
```
Sevans711, line 57: This test currently passes on main, so it does not serve as a regression test yet. One possible fix is to add: `import dask; assert isinstance(dask_result.data, dask.array.Array)`. Then, it fails on main but passes on this branch.

This applies to the node_to_edge case below as well.

Sevans711 (changes requested):
Hi @cmdupuis3, thank you for proposing these changes! Overall these look like good clean changes which should help improve the scalability of uxarray.

I have some notes/suggestions/requested changes. Primarily, I noticed that there are missing regression tests. This may be especially important for places with significant changes (more significant than just replacing `obj.values` with `obj.data`):
- `UxDataArray.integrate()`
- `UxDataArrayCrossSectionAccessor.__call__()`
- `uxarray.plot.matplotlib._nearest_neighbor_resample()`
- `RemapAccessor.apply_weights()`

It may be nice to add a regression test for `UxDataset.to_xarray()` too, even though the changes here are minor, since that feels like a very core part of the functionality.

The tests could be similar to what you added already in test_topological_agg.py (plus my additional request on that file): ensure that numpy and dask inputs ultimately give the same values (plus assert that the dask inputs lead to dask outputs).

Other parts of the code changes here might benefit from similar tests, but I don't know if that should be necessary. For example, the changes in _geos just replace `obj.values` with `obj.data`; does a regression test need to be added for that or no? Curious to hear from @erogluorhan and/or @rajeeja on this question in particular.

Misc. note: I'm not sure if I fully understand the changes to uxarray/cross_sections files. I need to make sure to take a closer look at those during a subsequent review.

test/core/test_topological_agg.py
cmdupuis3: The idea is that if someone is expecting bitwise reproducibility and switching to dask flips bits, these tests will catch it.

Sevans711: That makes sense. Clarifying, I think the bitwise reproducibility part of these tests are important to include and your implementation of them looks good!

Separately, I think there should also be tests to ensure the result is actually a dask array. But, those don't need to be separate test functions, for example it would be sufficient to add them via something like the assert statement above.

uxarray/core/aggregation.py
cmdupuis3: Yeah that's a good idea, I'll move the ux.DataArray type inside

test/core/test_topological_agg.py
erogluorhan: >Separately, I think there should also be tests to ensure the result is actually a dask array. But, those don't need to be separate test functions, for example it would be sufficient to add them via something like the assert statement above.

This makes sense! However, either they need to be in separate tests or the naming of the test case here should slightly be adjusted (currently the name suggests whether input output values and dims are all close)

cmdupuis3:
pre-commit.ci autofix

uxarray/core/dataarray.py
cmdupuis3: Okay, I'll keep this in for coherence, we can sort out the merge conflict later

test/core/test_topological_agg.py
cmdupuis3: I renamed them to `test_node_to_face_dask_reproduces_numpy` and `test_node_to_edge_dask_reproduces_numpy`

uxarray/core/aggregation.py
cmdupuis3: Done

uxarray/core/aggregation.py
cmdupuis3: Moved the ux.DataArray cast inside, like the other one

uxarray/core/aggregation.py
cmdupuis3: Done

cmdupuis3:
@erogluorhan Sam and I were wondering about your philosophy with respect to dask versus numpy routines. I was thinking that if we have a case where the numpy routine could be faster than a dask routine in some situations, we might want to keep both, or maybe you'd want to go all in on dask for simplicity.

```Python
if isinstance(uxda.data, np.ndarray):
    aggregated_var = _apply_node_to_face_aggregation_numpy(uxda, ...)
elif isinstance(uxda.data, da.Array):
    # apply aggregation on dask array, TODO:
    aggregated_var = _apply_node_to_face_aggregation_numpy(uxda, ...)
else:
    raise ValueError
```

This is something in main's aggregate.py. I fleshed out the dask branch, but now we still have this logic, and I left the numpy routine in on this branch.

erogluorhan:
> @erogluorhan Sam and I were wondering about your philosophy with respect to dask versus numpy routines. I was thinking that if we have a case where the numpy routine could be faster than a dask routine in some situations, we might want to keep both, or maybe you'd want to go all in on dask for simplicity.
> 
> ```python
> if isinstance(uxda.data, np.ndarray):
>     aggregated_var = _apply_node_to_face_aggregation_numpy(uxda, ...)
> elif isinstance(uxda.data, da.Array):
>     # apply aggregation on dask array, TODO:
>     aggregated_var = _apply_node_to_face_aggregation_numpy(uxda, ...)
> else:
>     raise ValueError
> ```
> 
> This is something in main's aggregate.py. I fleshed out the dask branch, but now we still have this logic, and I left the numpy routine in on this branch.

If I am understanding this correctly, question is about the use of routines, not data structures. If so, using standard NumPy functions directly on chunked Dask-backed Xarray objects should be completely safe with chunks. Let me know if this helps.

test/core/test_topological_agg.py
Sevans711: Marking as unresolved; still need to add type-checking into these tests. I believe Orhan's comment meant "Either: (1) add type-checking tests separately, or (2) add type-checking here but rename the tests." Unless I am missing something here, it looks like type-checking has not been added anywhere.

uxarray/core/aggregation.py
Sevans711: (Just confirmed, this has been resolved)

uxarray/core/aggregation.py
Sevans711: (Just confirmed, this has been resolved)

uxarray/core/aggregation.py
Sevans711: (Just confirmed, this has been resolved)

uxarray/core/aggregation.py
Sevans711: (Just confirmed, this has been resolved)

test/core/test_topological_agg.py
cmdupuis3: I added typechecks for uxarray objects. At the time it was raised, I was unsure of if these were going to have to be dask objects, but that issue seems to be resolved.

uxarray/core/dataarray.py
Sevans711: Marking as resolved, the new tests in `test/grid/test_basic.py` are sensitive to dask vs numpy, and I confirmed they crash on main but succeed here.

test/core/test_topological_agg.py
Sevans711: (Also still not resolved by latest commit, 180345e, which still passes on main, because it checks `isinstance(dask_result, ux.UxDataArray)`. See comments above, let me know if you have any questions.)

cmdupuis3: Well, these aren't really intended to be regression tests; afaik the main behavior was correct, it's just that *if* there's a question about what the uxarray-with-dask answers are versus uxarray-wtihout-dask, these tests can help. I meant bitwise-reproducible in terms of whether numpy/dask bitwise reproduce on the same commit.

cmdupuis3: I can see ux.DataArray is kind of not testing anything atm though, so I'll go with what you had

Sevans711: Marking as resolved; fixed by commit 1cafb3f, confirmed the tests would now fail on main, but pass here.

Sevans711: (I know they weren't originally intended to be regression tests, but I requested regression tests and the the type-checking is sufficient to allow them to serve as such!)

uxarray/core/dataarray.py
cmdupuis3: New test added to check bitwise repro like the new tests in `test_topological_agg.py`. See `test_basic.py::test_integrate_returns_uxdataarray_matching_numpy`

uxarray/plot/matplotlib.py
```diff
@@ -204,7 +204,11 @@ def _nearest_neighbor_resample(
     # build an array of values for each valid point
     flat_vals = np.full(first_face.shape, np.nan, dtype=float)
     mask_has_face = first_face >= 0
-    flat_vals[mask_has_face] = data.values[first_face[mask_has_face]]
+    # gather only the sampled faces (lazily for dask data): many pixels share a
+    # face, so deduplicate to materialize the minimal set rather than the whole
+    # field, then scatter back via the inverse map
+    unique_faces, inverse = np.unique(first_face[mask_has_face], return_inverse=True)
```
Sevans711, line 210: This was the one remaining case I had noted originally as "should have a regression test" which doesn't have such a test yet. Looking into it further, now I understand the result is going to be a numpy array anyway; this doesn't change inputs or outputs at all, it only changes internal computations a bit for a possible speedup.

I suspect this functionality is already covered by the test suite; a variety of tests call `to_raster()` already, and I confirmed (via adding a print statement) these lines of code are getting run >10 times during the test suite.

So, actually, no need to add any tests here, I think. Feel free to mark this as resolved once you took a quick look, if you agree it doesn't need tests.

test/core/test_topological_agg.py
```diff
@@ -32,3 +34,67 @@ def test_node_to_edge_aggs(gridpath):
         grid_reduction = getattr(uxds['areaTriangle'], agg_func)(destination='edge')
 
         assert 'n_edge' in grid_reduction.dims
+
+
+def test_node_to_face_dask_reproduces_numpy(gridpath):
+    # the numpy (eager) and dask (chunked) branches must agree
+    da = pytest.importorskip("dask")  # dask-backed branch requires dask
```
Sevans711: Remove the `importorskip` calls everywhere; tests should crash if dask is not installed, not be skipped silently (well, mostly silently). Existing tests currently on `main` already use dask without importorskip, and the CI for the test suite installs dask, so it should be safe to assume dask is available.

Sevans711 (changes requested):
This looks almost ready to merge, I just have a few minor comments remaining!

**Requires response:**
- I noticed the `mpas_ocean.Integrate.time_integrate('480km')` benchmark has worsened slightly. Is this expected behavior? Do you think this is real, a rounding error, or something else?
- I would suggest to remove all of the `pytest.importorskip` calls; I left an inline comment with more details.

The rest of this comment does not require response. It includes the notes I wrote down, for future reference.

**Other Benchmarking:**
- The cross sections benchmarks seem to have improved, which is great!
- I'm ignoring peakmem "improvements" (as per #1605)
- Even though other benchmarks didn't show notable improvements, I think that is fine, because they didn't get worse, either. Also, this PR isn't fully closing the original issue, anyway, it is just addressing part of it.

**Testing:**

I checked all the cases where I previously requested tests for correctness (numpy & dask give same result) and regression tests (check dask inputs cause dask outputs); here is what I found:

- (Looks done to me! Confirmed tests pass here, fail on main.) `UxDataArray.integrate()`
- (Looks done to me! Confirmed tests pass here, fail on main. Also, the tests do a good job covering a variety of possible failure cases, by including multiple dimensions and some nan / some non-nan values.) `UxDataArrayCrossSectionAccessor.__call__()`
- (Not done, but now I'm less sure if it's necessary; left an inline comment to discuss.) `uxarray.plot.matplotlib._nearest_neighbor_resample()`
- (Looks done to me! Confirmed tests pass here, fail on main. Also, the tests do a good job covering nontrivial cases, like handling multiple dimensions.) `RemapAccessor.apply_weights()`
- (Not done, but optional; no need to block merging for this.) `UxDataset.to_xarray()`

Previously I noted I need to look closer at cross sections. After doing so and considering the new cross sections tests, I am feeling convinced that the cross sections changes should be working as intended. Noting that the extra dims included in those tests might also help with #1461.

cmdupuis3:
> I noticed the mpas_ocean.Integrate.time_integrate('480km') benchmark has worsened slightly. Is this expected behavior? Do you think this is real, a rounding error, or something else?

It isn't real, you can rerun the benchmarks and it'll probably disappear. There's always a chance with the smaller benchmarks that they'll trigger due to machine variability. 1.1x is pretty suspect.

test/core/test_topological_agg.py
cmdupuis3: If we're assuming that dask is available for these tests though, why can't we assume that in the source code? My strategy lately has been to hedge with the assumption that dask will eventually be optional at the package level, but you're saying here I should be moving in the opposite direction?

rajeeja:
While reviewing this I noticed the numpy aggregation paths drop coords and attrs while your new dask paths preserve them via `apply_ufunc`, so on this branch `.chunk()` decides whether `.sel(time=...)` works on the result. That turned out to be a pre-existing bug on `main`, so I filed #1665 and opened #1666 against `main` rather than adding it here. Merging `main` once that lands should bring your two paths back into agreement.

test/core/test_topological_agg.py
Sevans711: Let's chat offline about this to clear things up and report back here once we're on the same page! I will send you a DM.

Sevans711: Reporting back from our chat: basically, the following are all true, which can be confusing but is technically not self-contradictory:
1. Tests (and benchmarks) assume all dependencies are installed and do not care about import time, so dask should be imported at tops of files there.
2. uxarray source code assumes dask is a required dependency, i.e. that it is installed and always safe to import.
3. uxarray souce code cares about import time, and `import dask` isn't blazingly fast, so dask should only be imported inside functions that use it, instead of at tops of files.

Conclusion: `pytest.importorskip` should indeed be removed throughout the test suite.

cmdupuis3: Dask imports have been moved back to the tops of files for test files

cmdupuis3:
pre-commit.ci autofix

Sevans711 (approved):
Looks to me like this is ready to merge into main!

Any guesses as to why the `mpas_ocean.CrossSections.time_const_lat` benchmarks show improvements now, but didn't show improvements in prior iterations of this PR? (This question is non-blocking, since it's an improvement! Just curious if you have insights about this.)

cmdupuis3:
@Sevans711 They've shown improvements for a while now, that's not recent. There are some changes to how cross-sections are caluculated on chunked grids that served as the `.values` fix, they weren't as trivial as most other changes here.

Sevans711:
> They've shown improvements for a while now, that's not recent. There are some changes to how cross-sections are caluculated on chunked grids that served as the `.values` fix, they weren't as trivial as most other changes here.

Ah, woops! I confused myself by looking at my prior comment mentioning the benchmarks. Thank you for clarifying!

rljacob:
I have copilot access on github and will ask it to do a review.

uxarray/grid/slice.py
```diff
@@ -127,36 +133,30 @@ def _slice_face_indices(
     ds["subgrid_node_indices"] = xr.DataArray(node_indices, dims=["n_node"])
     ds["subgrid_face_indices"] = xr.DataArray(face_indices, dims=["n_face"])
 
-    # Construct updated Node Index Map
-    node_indices_dict = {orig: new for new, orig in enumerate(node_indices)}
-    node_indices_dict[INT_FILL_VALUE] = INT_FILL_VALUE
+    def _build_remap(orig_indices, size):
+        remap = np.full(size, INT_FILL_VALUE, dtype=INT_DTYPE)
+        remap[orig_indices] = np.arange(len(orig_indices), dtype=INT_DTYPE)
+        return remap
```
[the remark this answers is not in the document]
cmdupuis3: I split this into two remapping algorithms. Which path the current algorithm chooses is mainly determined by `_DENSE_REMAP_MAX_RATIO` which seems to have an ideal ratio of 32 or so. At most HPC scales `_remap_searchsorted` will be the predominant method.

uxarray/grid/slice.py
```diff
@@ -167,11 +167,14 @@ def map_edge_indices(i):
...
-            np.vectorize(map_fn, otypes=[INT_DTYPE])(ds[conn_name].values),
-            dims=ds[conn_name].dims,
-            attrs=ds[conn_name].attrs,
+        # Apply remapping (vectorized; stays lazy when the connectivity is dask)
+        ds[conn_name] = xr.apply_ufunc(
+            _remap_indices,
+            ds[conn_name],
+            kwargs={"remap": remap},
+            dask="parallelized",
+            output_dtypes=[INT_DTYPE],
```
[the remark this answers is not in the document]
cmdupuis3: Test added: `test_isel_dask_connectivity`

dylannelson (approved):
Ran a handful of tests locally, and reviewed the discussion. Looks like all good changes. .values -> .data seems to function as it should in all cases, preserving functionality.

Issue #1602, opened 2026-07-21
uxarray/utils/computing.py not used anywhere, and `pyfma` not in pyproject.toml


Sevans711:
uxarray/utils/computing.py internally imports the `pyfma` package, which is not included in uxarray's dependencies, not even as an optional dependency, and it is also not being used anywhere else in uxarray, from what I can tell. Noticed this while working on #1575.

Additionally, none of the methods from uxarray/utils/computing.py are being used anywhere else in the entire uxarray codebase. I cannot tell why this module exists, from looking through the code.

Two possible solutions:
1. Remove the uxarray/utils/computing.py module.
2. Keep the module, but document why it exists, and handle `pyfma` package dependency appropriately.
  - Documentation options include: add a docstring at the top of the file (see #1598); write about it on readthedocs pages; and/or add at least one example to the docs pages which utilize it.
  - Handling the dependency appropriately includes adding it as an optional dependency in pyproject.toml (see also: #1548).

Sevans711:
Realized I should probably ping @hongyuchen1030 on this one, as the person who implemented many (all?) of the methods in this module. Do you know if there are any imminent plans to incorporate uxarray/utils/computing.py functionality elsewhere in uxarray, or if it already being used downstream anywhere?

To clarify, I'm not trying to push towards removing this module, rather, I mostly just want to understand why it exists, and help that reason remain clear for anyone who looks into it in the future. I would only want to remove it if there really is no current/planned use case for it.

hongyuchen1030:
> Realized I should probably ping [@hongyuchen1030](https://github.com/hongyuchen1030) on this one, as the person who implemented many (all?) of the methods in this module. Do you know if there are any imminent plans to incorporate uxarray/utils/computing.py functionality elsewhere in uxarray, or if it already being used downstream anywhere?
> 
> To clarify, I'm not trying to push towards removing this module, rather, I mostly just want to understand why it exists, and help that reason remain clear for anyone who looks into it in the future. I would only want to remove it if there really is no current/planned use case for it.



This has already been addressed in #1513. `uxarray/utils/computing.py` shouldn't be removed, since it serves as the main algorithmic highlight of UXarray, providing the EFT-based floating-point operators. @rajeeja has already cleaned it up in #1513, so no further action is needed here before this branch is merged.

[...]

PR #1603, opened 2026-07-21
Reduce imports inside functions

### LINKED ISSUE #1575: Reduce number of imports inside functions

Sevans711:
Closes #1575

## Overview

There are some places throughout the uxarray codebase where import is being used inside functions. This PR targets imports which are not "heavy" or which are already being imported non-lazily (i.e., at they are imported at "top-level" / top of file), changing them to be consistently imported non-lazily everywhere in uxarray.

Notes about all packages being imported internally in at least one location (e.g. imported inside a function; not just at top of file):
- `antimeridian`: not currently at top-level anywhere. Tiny required dependency, seems trivially-lightweight to import. But, might become optional some day...? → **keep internal**.
- `dask`: not currently at top-level anywhere, and not trivially-lightweight to import → **keep internal**
- `datetime`: trivially lightweight and builtin, no need to import lazily → **move to top**.
- `IPython`: not currently at top-level anywhere; optional (but forgot to list in pyproject.toml at all), and will probably be removed eventually, whenever upstream fix is released (see discussion in #1538 and upstream issue at holoviz/holoviews#6955) → **do not touch in this PR**.
- `numpy`: essential for core functionality and definitions, already at top level in many places → **move to top level**.
- `pyfma`: only appears in an otherwise-unused module? (See #1602) → **do not touch in this PR**.
- `scipy`: non-trivial import size, ~but does exist as top-level import in `weights.py`, which makes this a bit ambiguous. → **did not touch here, but _open for further discussion (should the scipy import in weights.py be moved to be internal instead?)_**.~ EDIT: this would be fully addressed by #1606, which moves that scipy import to be internal instead → **do not touch in this PR**.
- `sklearn`: not currently at top-level anywhere; non-trivial import size → **keep internal**.
- `shapely`: same story as with `antimeridian`: not currently at top-level anywhere. Tiny required dependency, seems trivially-lightweight to import. But, might become optional some day...? → **keep internal**.
- `warnings`: trivially lightweight and builtin, no need to import lazily → **move to top**.
- `xarray`: essential for core functionality and definitions, already at top level in many places → **move to top level**.
- Many packages should be an optional dependency eventually/soon (see #1548) → **do not touch in this PR**; these are already handled by that PR (or maybe will be moved to a separate issue/PR). This applies to:
  - `cartopy`
  - `geopandas`
  - `holoviews`
  - `hvplot`
  - `matplotlib`
  - `spatialpandas`

Notes about importing from uxarray internally to functions:

- In some cases, such as in uxarray.core.accessors, code like `from uxarray.core.dataarray import UxDataArray` must be kept inside functions, to avoid circular imports.
- There are many cases, such as `UxDataArray.isel` containing `from uxarray.core.utils import _validate_indexers`, where uxarray methods are imported internally without a clear compelling reason to do so. These should probably be moved to top-level. However, the original scope of issue #1575 did not clarify this possibility, and uxarray's explicit non-relative imports make me think it could be an intentional design choice, to reduce "cascading" imports. _**I think it makes more sense to open a separate issue to discuss internal imports like "from uxarray... import ...", instead of trying to also consider that here. If PR reviewers agree, let's open that issue before this PR gets merged**_

Misc notes:

- In some cases, imports were gated by "if TYPE_CHECKING" unnecessarily, such as "if TYPE_CHECKING: from xarray import Dataset". There is no reason to do this when xarray is already imported in the file → **cleaned up cases like this**.

## PR Checklist

**General**
- [X] Add appropriate labels

### DIFF
--- a/uxarray/core/api.py
+++ b/uxarray/core/api.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Mapping, Sequence, TypeAlias
+from typing import Any, Mapping, Sequence, TypeAlias
 from warnings import warn
 
 import numpy as np
--- a/uxarray/core/api.py
+++ b/uxarray/core/api.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any, Mapping, Sequence, TypeAlias
 from warnings import warn
 
 import numpy as np
+import xarray as xr
 
 from uxarray.core.dataset import UxDataset
 from uxarray.core.utils import (
@@ -22,9 +23,6 @@ from uxarray.io._scrip import (
     _stack_cell_dims,
 )
 
-if TYPE_CHECKING:
-    from xarray import Dataset
-
 __all__ = [
     "open_grid",
     "open_multigrid",
@@ -35,7 +33,7 @@ __all__ = [
 
 
 def open_grid(
-    grid_filename_or_obj: str | os.PathLike[Any] | dict | Dataset,
+    grid_filename_or_obj: str | os.PathLike[Any] | dict | xr.Dataset,
     chunks=None,
     use_dual: bool | None = False,
     **kwargs: dict[str, Any],
@@ -85,8 +83,6 @@ def open_grid(
 
     >>> uxgrid = ux.open_grid("grid_filename.nc", chunks=-1)
     """
-    import xarray as xr
-
     # Handle chunk-related kwargs first.
     data_chunks = kwargs.pop("data_chunks", None)
     if data_chunks is not None:
@@ -144,9 +140,9 @@ MaskActiveValue: TypeAlias = MaskValue | Mapping[str, MaskValue] | None
 
 
 def open_multigrid(
-    grid_filename_or_obj: str | Path | "Dataset",
+    grid_filename_or_obj: str | Path | xr.Dataset,
     gridnames: list[str] | None = None,
-    mask_filename: str | Path | "Dataset" | None = None,
+    mask_filename: str | Path | xr.Dataset | None = None,
     mask_active_value: MaskActiveValue = 1,
     **kwargs: dict[str, Any],
 ) -> dict[str, Grid]:
@@ -175,8 +171,6 @@ def open_multigrid(
     dict[str, Grid]
         Dictionary mapping grid names to ``Grid`` objects.
     """
-    import xarray as xr
-
     grid_ds_opened = False
     if isinstance(grid_filename_or_obj, xr.Dataset):
         grid_ds = grid_filename_or_obj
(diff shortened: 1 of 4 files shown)


erogluorhan (commented):
The work here seems great to me! Please make sure all the tests are passing. I will get back to your discussion points and approve this then

dylannelson (approved):
The notes and comments here helped a lot, appreciate it! Thanks!

Sevans711:
Thank you both for taking a look!

Re: @erogluorhan's point about tests:
> Please make sure all the tests are passing. I will get back to your discussion points and approve this then

All the tests do seem to pass with the code here, I manually triggered a CI run on this branch (https://github.com/UXARRAY/uxarray/actions/runs/29865868247), after the original run failed one test just due to a network error (link for posterity: https://github.com/UXARRAY/uxarray/actions/runs/29864267121/job/88748409432?pr=1603).  This rerun involved no code changes. I also opened #1604 for further discussion about network failures in the CI autotests.

What is the preferred workflow when this happens? I wasn't able to find a "rerun tests directly in the PR interface" button. One option: I can push a commit with basically no changes (e.g., comment at the single top-level scipy import to say "maybe this should move to be not at top-level? see PR # 1603 for more details"). I will wait for your response before pushing that though, to help the current issue remain clear until then.

Sevans711:
Tests now all passed successfully, thank you for showing me where the "rerun failed jobs in this PR" button lives, @erogluorhan!

I suspect I may have failed to find it before due to looking too early, before all the jobs had finished? Or, maybe I simply didn't know where to look. I left this comment about it: https://github.com/UXARRAY/uxarray/issues/1604#issuecomment-5051185135 for future reference.

erogluorhan (approved):

PR #1606, opened 2026-07-22
Move imports for should-be-optional dependencies

Sevans711:
One step towards solving #1224, but not the full solution.

## Overview
Moves imports for "should-be-optional" dependencies to be inside functions instead of at tops of files. All of these changes were originally proposed in PR #1548, which also includes detailed discussion about which dependencies should be optional. However, the changes here do not include edits to pyproject.toml; the should-be-optional dependencies remain fully required, leaving installation processes fully untouched.

The narrower scope of this PR aims to provide simpler-to-accept changes. It was created in response to concerns about the size of PR 1548.

Recommended action plan:
1. Review this PR. Discuss as needed.
2. When approved, merge to main.
3. Merge main into the original PR (1548). This will reduce the number of changes in PR 1548 significantly.
4. PR 1548 will continue to be responsible for changes to pyproject.toml, optional dependencies testing, etc. Or, maybe some of those details will be split into other PRs, too; that can be discussed in thread on PR 1548 directly.

## PR Checklist

**General**
- [X] Add appropriate labels

### DIFF
--- a/test/io/test_structured.py
+++ b/test/io/test_structured.py
@@ -2,6 +2,9 @@ import uxarray as ux
 import xarray as xr
 import pytest
 
+# import pooch  # not necessary to actually import, but tests would likely fail if this fails.
+# (commented here for future reference, since pooch is included in uxarray[dev]
+#    dependencies, but not imported explicitly anywhere in uxarray.)
 
 @pytest.mark.parametrize("ds_name", ["air_temperature", "ersstv5"])
 def test_read_structured_grid_from_ds(ds_name):
--- a/uxarray/core/dataarray.py
+++ b/uxarray/core/dataarray.py
@@ -5,10 +5,8 @@ from html import escape
 from typing import TYPE_CHECKING, Any, Hashable, Literal, Mapping, Optional
 from warnings import warn
 
-import cartopy.crs as ccrs
 import numpy as np
 import xarray as xr
-from cartopy.mpl.geoaxes import GeoAxes
 from xarray.core import dtypes
 from xarray.core.options import OPTIONS
 from xarray.core.utils import UncachedAccessor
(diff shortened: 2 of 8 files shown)


rajeeja (approved):
Reviewed and tested locally — checked out the branch, ran the full suite (565 passed) plus a pre-commit pass, and traced every moved import (ccrs/GeoAxes in dataarray.py/grid.py, pyproj.Geod in sample.py, antimeridian/shapely in geometry.py, healpix in _healpix.py, holoviews in plot/utils.py, scipy.sparse in remap/weights.py) to confirm no runtime usage was left behind outside the function scope it moved into. Clean, well-scoped split from #1548. Approving.

erogluorhan:
> ... traced every moved import (ccrs/GeoAxes in dataarray.py/grid.py, pyproj.Geod in sample.py, antimeridian/shapely in geometry.py, healpix in _healpix.py, holoviews in plot/utils.py, scipy.sparse in remap/weights.py) to confirm no runtime usage was left behind outside the function scope it moved into.

I was wondering if confirmed we had test coverage for all these imports, so I can ask that to you: Did you trace them yourself by eye, or did you confirm test coverage as well? If the latter, I think we can go ahead and merge this; otherwise, I can dod that check.

cmdupuis3:
I'll trace the dependencies and close this out if I find it's working well.

erogluorhan (commented):
This looks great to me! Once I hear back to my question above, I'll approve this.

cmdupuis3:
Yeah I don't see anything missing from the stack traces.

erogluorhan:
> Yeah I don't see anything missing from the stack traces.

Thanks very much for looking into this, Chris!

erogluorhan (approved):
This is great; thanks Sam!

dylannelson:
Clarifying after a brief conversation with @Sevans711: even though this PR title seems to imply it only affects imports for "should-be-optional" dependencies, it also affected some imports for required packages like scipy (required dependency). Despite the title and description, the actual scope of this PR was to move such imports as well, as defined by the notes in #1548, not just the should-be-optional dependencies. Thank you for the clarification

PR #1609, opened 2026-07-22
ASV better memory benchmarks

### LINKED ISSUE #1605: Some ASV peakmem benchmarks always show the same "improvement"
### Version

2026.6.0

Source

### What happened?

ASV benchmarking is showing the same improvement (to within ~2%) for these three tests, every single time, regardless of whether the PR should have any effect on them:

| Change   | Before    | After    |   Ratio | Benchmark (Parameter)                                                                                                                 |
|----------|----------|--------|---------|---------------------------------------------------------------------------------------------------------------------------------------|
| -        | 578M                 | 391M                |    0.68 | face_bounds.FaceBounds.peakmem_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/ugrid/geoflow-small/grid.nc')) |
| -        | 708M                 | 390M                |    0.55 | face_bounds.FaceBounds.peakmem_face_bounds(PosixPath('/home/runner/work/uxarray/uxarray/test/meshfiles/ugrid/quad-hexagon/grid.nc'))  |
| -        | 498M                 | 384M                |    0.77 | mpas_ocean.Gradient.peakmem_gradient('480km')                                                                                         |

**EDIT: note that recently this problem seems to be occurring for`mpas_ocean.FaceAreas.peakmem_compute_face_areas('480km')` as well.**

Examples of where this occurs:
- https://github.com/UXARRAY/uxarray/pull/1603#issuecomment-5039103216, for a PR which should not affect memory usage in these tests at all.
- https://github.com/UXARRAY/uxarray/pull/1588#issuecomment-5008829970, for a PR which could maybe have some impact on memory usage?
- https://github.com/UXARRAY/uxarray/pull/1577#issuecomment-4986320565, for a PR which could maybe have some impact on memory usage?
- https://github.com/UXARRAY/uxarray/pull/1569#issuecomment-4985121279, for a PR which probably should not affect these tests at all?

Additionally, there are a few slightly older PRs where the numbers look different by more than a few percent, but still show similar trends as above:
- https://github.com/UXARRAY/uxarray/pull/1546#issuecomment-4894742029, for a PR which definitely should not have affected these tests, as it only touched code for unrelated benchmarks.
- https://github.com/UXARRAY/uxarray/pull/1541#issuecomment-4919582171, for a PR which probably should not have affected these tests?

I expected "Benchmarks that have improved" to reflect actual improvements to relevant parts of the code, or at least, to reflect small deviations which occur when running the benchmarking tests multiple times. Instead, these three benchmarks seem to always cost close to the same amount the first time, and "improve" by a similar amount each time, despite the relevant code changes being merged into main.

Pinging @cmdupuis3 for your consideration, when you get the chance!

```python

```

cmdupuis3:
Closes #1605 

## Overview

The current behavior of a few peakmem benchmarks entrains module imports and caching, which dwarf the actual memory usage supposedly being tested. This PR offers benchmarks of low-level memory behavior as well as warm setups to prevent the persistent "every PR improves memory by the same amount" issue in #1605.

## Expected Usage

Most of these benchmarks are useful as-is, but there are also a couple of benchmarks of pure uxarray imports to compare with if the illusory memory improvement ever returns.

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


Sevans711:
Thank you for working on this and opening up this PR! I have a few quick thoughts before reviewing:

1. I wonder if nbytes should be added directly as a property of Grid objects, instead of needing a custom grid_nbytes function? It might be nice to be able to do `uxgrid.nbytes`, since Grid is basically just an xr.Dataset with some extra functionality attached to it, and it seems nice to be able to quickly check how large of a grid you are working with.
2. What is the purpose of benchmarks that measure the size of the loaded Grid objects or arrays? My initial understanding is that the answer should always be the same, and it isn't really measuring runtime efficiency nor total memory usage. Would these work better as unit tests? E.g. "test_face_bounds_array_size" which asserts the size of face_bounds from oQU480.231010.nc is 57.3 kB, with some small tolerance for rounding errors. I might also be not fully understanding the purpose of benchmarks. Also wondering if @erogluorhan or @rajeeja have thoughts on this question in particular?

cmdupuis3:
@Sevans711 as per convo, 1 is really a question for @erogluorhan or @rajeeja. On 2, these are included because the peakmem is not necessarily the same as the total memory usage in a chunked situation. This particular batch of benchmarks is a little verbose, but it basically serves as a regression test against the kinds of spurious improvements in other PRs, so that we have a better idea of which things are bloating the peakmem metrics.

Sevans711:
Just noticed, the `mpas_ocean.FaceAreas.peakmem_compute_face_areas('480km')` is now having a similar problem; would you be able to take a quick look and see if it is easy to fix here in the same way?

benchmarks/mpas_ocean.py
erogluorhan: Can you please clarify why this is not converted to the new way? No `setup_cache` needed for it (since no njit)?

benchmarks/mpas_ocean.py
```diff
@@ -68,16 +69,44 @@ class Gradient(DatasetBenchmark):
...
+        return self.uxds[data_var].gradient().nbytes
+
+    track_nbytes_gradient.unit = "bytes"
 
 
 class Integrate(DatasetBenchmark):
     def time_integrate(self, resolution):
         self.uxds[data_var].integrate()
 
-    def peakmem_integrate(self, resolution):
```
erogluorhan, line 79: Why no `IntegratePeakMem` class defined like others?

erogluorhan:
> Thank you for working on this and opening up this PR! I have a few quick thoughts before reviewing:
> 
> 1. I wonder if nbytes should be added directly as a property of Grid objects, instead of needing a custom grid_nbytes function? It might be nice to be able to do `uxgrid.nbytes`, since Grid is basically just an xr.Dataset with some extra functionality attached to it, and it seems nice to be able to quickly check how large of a grid you are working with.
> 2. What is the purpose of benchmarks that measure the size of the loaded Grid objects or arrays? My initial understanding is that the answer should always be the same, and it isn't really measuring runtime efficiency nor total memory usage. Would these work better as unit tests? E.g. "test_face_bounds_array_size" which asserts the size of face_bounds from oQU480.231010.nc is 57.3 kB, with some small tolerance for rounding errors. I might also be not fully understanding the purpose of benchmarks. Also wondering if @erogluorhan or @rajeeja have thoughts on this question in particular?

1. I agree with this but don't have a strong preference for it to be in this PR (just to be precise though, `Grid` is not a `xr.Dataset`, it just has a composition, i.e. "has a" `xr.Dataset`).
2. Great question! With the benchmarks, you want to know the baseline and then both (1) avoid regression with additions to `Grid` in this case (but can be taken as general) and (2) aim at improving the baseline of the `Grid` object. For instance, for (1), when you're touching `Grid`, maybe adding some calculations or attributes, you may want to at least keep the memory used the same as before. For (2), if Grid is not doing a great job with memory utilization with all its cached variables, connectivities, etc. you may want to see if some of them can be dropped, made optional, etc. and benchmarks help you see the improvements if you have any with your PR. Does this help?

In regards to peak-mem vs static object size (nbytes) though, I think each of them has its own value (the former gives the peak memory used throughout an operation while the latter gives you the static memory size needed for an object, and they should not be considered alternative to each other.

erogluorhan (commented):
This is looking great to me; just a few comments below.

Sevans711:
@erogluorhan Thank you for clarifying these points, that helps, though I do still have a few follow-up thoughts below:

For my question (1), that makes sense, it also sounds good to me to have it go into a different PR. Flagging #1647 as the relevant feature request (created after I left my comment above) for things like `Grid.nbytes`.

> In regards to peak-mem vs static object size (nbytes) though, I think each of them has its own value (the former gives the peak memory used throughout an operation while the latter gives you the static memory size needed for an object, and they should not be considered alternative to each other.

Indeed, it makes sense to me that peak memory is a very different measurement from static memory size needed for an object.

My understanding is that no other benchmarks in the benchmarking suite were measuring static memory size, and that the original issue reported here doesn't really require any static memory size checks in order to solve it. That's the main reason I was surprised these were added here, and wanted to know more about their purpose.

>  2. Great question! With the benchmarks, you want to know the baseline and then both (1) avoid regression with additions to `Grid` in this case (but can be taken as general) and (2) aim at improving the baseline of the `Grid` object. For instance, for (1), when you're touching `Grid`, maybe adding some calculations or attributes, you may want to at least keep the memory used the same as before. For (2), if Grid is not doing a great job with memory utilization with all its cached variables, connectivities, etc. you may want to see if some of them can be dropped, made optional, etc. and benchmarks help you see the improvements if you have any with your PR. Does this help?

I believe I understand what you mean. Is this rephrasing correct?:
- Benchmarking static memory size of objects is now desired, and the main benefits would be:
- (1) help avoid regression where `Grid` code changes cause object sizes to increase.
- (2) help see improvements if `Grid` code changes cause object sizes to decrease.

If that is correct, my suspicion here is still that adding to the pytest `test/` suite might be a better fit. For (1), pytest tests would likely be more reliable in preventing regression, since they are always run, instead of only run when you suspect they might be relevant. It would also be simple and easy to check object size in the pytest test suite, unlike runtime and peakmem which both depend on machine/setup and can produce noisy results. For (2), I suspect any improvements to `Grid` size would also be seen clearly (and may be even be more visible?) if the static size was in the pytest test suite, because then a line of code would need to get "permanently" changed in response to the improvement.

Assuming I understand correctly, those are my last reservations about it. If the overall preference here is still to include these as benchmarks instead of putting them in the test suite, I will get on board with that decision and review with that in mind!

cmdupuis3:
@Sevans711 I think I'm seeing what you're getting at. These benchmarks don't really track how a chunked peakmem metric would actually behave, they basically just measure the total memory consumption. For now though, I think it's probably okay, because we're kind of accepting that the whole grid is going to be materialized elsewhere anyway, and until we have a strong incentive to get an end-to-end chunked grid implementation, the stakes here are low.

That said, I think we might be able to get something like a true peakmem metric with something like this:

```Python
import tracemalloc
__all__ = ["peak_allocated"]


def peak_allocated(build):

    tracemalloc.start()
    try:
        tracemalloc.reset_peak()
        build()
        _, peak = tracemalloc.get_traced_memory()
    finally:
        tracemalloc.stop()
    return peak
    
...

class FaceAreas(GridBenchmark):
    def track_peakmem_face_areas(self, resolution):
        return peak_allocated(lambda: self.uxgrid.face_areas)
```

erogluorhan:
> I believe I understand what you mean. Is this rephrasing correct?:
> 
> * Benchmarking static memory size of objects is now desired, and the main benefits would be:

Let @cmdupuis3 answer all your questions about if static memory measurements needed as part of this PR's work or whether now desired in benchmarking, or if they should into pytest, etc. but I'll just share my thoughts on a few things below:

> * (1) help avoid regression where `Grid` code changes cause object sizes to increase.
> * (2) help see improvements if `Grid` code changes cause object sizes to decrease.
> 
> If that is correct, my suspicion here is still that adding to the pytest `test/` suite might be a better fit. For (1), pytest tests would likely be more reliable in preventing regression, since they are always run, instead of only run when you suspect they might be relevant. It would also be simple and easy to check object size in the pytest test suite, unlike runtime and peakmem which both depend on machine/setup and can produce noisy results. 
> For (2), I suspect any improvements to `Grid` size would also be seen clearly (and may be even be more visible?) if the static size was in the pytest test suite, because then a line of code would need to get "permanently" changed in response to the improvement.

Yeah, `pytest` could be used for that, too, but we picked ASV to handle benchmarks for a number of reasons. A few to exemplify:  When unit tests fail, it means "the code is wrong and needs fixed."; it doesn't/shouldn't apply to performance regression, since there is more than just boolean logic. Performance vs memory tradeoffs are part of benchmarking efforts a lot of times, and looking into them does not necessarily call for a deterministic and binary outcome always, though `pytest` works that way and could create false negatives (e.g. if `Grid` is bigger than this, fail etc). Also, ASV can give us history of benchmarks, i.e. the trend. Benchmarking is, by nature, slower than running unit tests, and we wouldn't want to have it as part of the pytest cycles.

Also, we had ASV always run with PRs until some time when we decided not every PR needed it (and also the above speed cost IIRC). 

For the time being, drawing a line between performance and unit testing, and leaving the former to ASV and mostly @cmdupuis3 's efforts while leaving the latter to `pytest` and everyone's care still makes more sense to me.

We can always decide to get ASV to always run again if that's needed and if the speed cost is worth it.

Sevans711:
> I think I'm seeing what you're getting at. These benchmarks don't really track how a chunked peakmem metric would actually behave, they basically just measure the total memory consumption. For now though, I think it's probably okay

Wait, actually this confused me a bit. I didn't intend to make any claims about the peakmem benchmarks, only the track_nbytes benchmarks. I believe my previous comment is focused entirely on the track_nbytes benchmarks, how it is a completely separate measurement from peakmem, and asking whether track_nbytes should be included in the benchmarking suite.

Can you clarify:
1. What do you mean by "These benchmarks" in your comment? The track_nbytes benchmarks, or the peakmem benchmarks?
2. What do you mean by "the total memory consumption"? (The static object sizes? Or, something else which is also different from peakmem?)
3. Do you believe the peakmem benchmarks implemented here still measure peak memory usage properly for non-chunked grids?

If the answer to (3) is "yes" then I would agree it's probably okay for now, for the purposes of this PR, because the benchmarks aren't using chunked grids (right?).

> Yeah, pytest could be used for that, too, but we picked ASV to handle benchmarks for a number of reasons. A few to exemplify: When unit tests fail, it means "the code is wrong and needs fixed."; it doesn't/shouldn't apply to performance regression, since there is more than just boolean logic. Performance vs memory tradeoffs are part of benchmarking efforts a lot of times, and looking into them does not necessarily call for a deterministic and binary outcome always, though pytest works that way and could create false negatives (e.g. if Grid is bigger than this, fail etc).

> For the time being, drawing a line between performance and unit testing, and leaving the former to ASV and mostly @cmdupuis3 's efforts while leaving the latter to pytest and everyone's care still makes more sense to me.

Thank you for considering my points above and for clarifying this! This explanation helped me understand clearly the motivation for wanting track_nbytes as a benchmark rather than a pytest test. Rephrasing to make sure I understand: the design decisions are (A) to keep performance-related checks in benchmarking, and (B) that "static object size" is performance-related, not an issue of correctness. These seem reasonable enough, and moving forwards I will keep them in mind.

> Also, ASV can give us history of benchmarks, i.e. the trend. Benchmarking is, by nature, slower than running unit tests, and we wouldn't want to have it as part of the pytest cycles.

Oh! I didn't realize there was an easy way to see a graph of how the ASV numbers change across commits, I thought you might need to click one ASV job at a time. A quick google seems to indicate `asv publish` could generate an html page where one might view that… would there be any interest in setting up the asv-benchmarking.yml to create that site, and put it somewhere easy for developers to find?

While looking into that yml file briefly I also noticed it has instructions to post to https://github.com/UXARRAY/uxarray-asv, however that repo seems to have not been updated for over a year.

(Something like `asv publish` and updating the yml file shouldn't go in this PR, but I would be happy to open a feature request/issue report if those ideas sound good to you.)

cmdupuis3:
@Sevans711 I am referring to the `track_nbytes` benchmarks, which are serving as replacements for the `peakmem` benchmarks that are currently failing. The actual peakmem benchmarks cannot work in a way that reveals the true behavior of the particular routines they test, because the peakmem benchmarks generally inherit some overhead from loading libraries and caching numba functions. So, my original thought here was to use `track_nbytes` instead, because that's a more direct measurement. 

What I was saying in my last comment is that what we're measuring with `track_nbytes` is still difficult to compare to the intended peakmem behavior, so my proposal was to use tracemalloc to get to that behavior without the overhead that peakmem entails.

In practice, I think the issue isn't too important because we are currently expecting to materialize the whole grid anyway, so the static object size is actually a reasonable proxy for what the memory usage behavior currently does.

To answer 3, the peakmem benchmarks being replaced here were never measuring memory usage correctly. My goal here is to have them be "less wrong," but there could still be room for improvement.

Sevans711:
> I am referring to the `track_nbytes` benchmarks, which are serving as replacements for the `peakmem` benchmarks that are currently failing.

> To answer 3, the peakmem benchmarks being replaced here were never measuring memory usage correctly. My goal here is to have them be "less wrong," but there could still be room for improvement.

Ah, thank you for clarifying, and this is a great point.

> The actual peakmem benchmarks cannot work in a way that reveals the true behavior of the particular routines they test, because the peakmem benchmarks generally inherit some overhead from loading libraries and caching numba functions. So, my original thought here was to use `track_nbytes` instead, because that's a more direct measurement.

This might be a naive question, but why can't you just run the functions once during setup? I think that's the pattern used by other benchmarks to avoid the transient "cold start" overhead leaking into the tests. For example, something like this:

```python
class FaceBounds:
    def setup(self, grid_path):
        tmp = ux.open_grid(grid_path)
        tmp.bounds   # cold start memory costs of imports, numba caching, etc. will happen here
        self.uxgrid = ux.open_grid(grid_path)  # make a fresh object for the actual tests below

    def peakmem_face_bounds(self, grid_path):
        """Peak memory usage obtain ``Grid.face_bounds."""
        face_bounds = self.uxgrid.bounds
```

Would this not give a proper peakmem measurement of just `Grid.bounds`?

cmdupuis3:
@Sevans711 My understanding is that ASV is relatively simple, and `peakmem` benchmarks are always going to report the peak memory of the process, not the function, so regardless you would still entrain the memory usage of library imports etc. In my experience so far, ASV is kind of an odd package; it's better to think of it as a compiler wrapper with its own semantics, rather than a proper Python library.

erogluorhan:
> > Also, ASV can give us history of benchmarks, i.e. the trend. Benchmarking is, by nature, slower than running unit tests, and we wouldn't want to have it as part of the pytest cycles.
> 
> Oh! I didn't realize there was an easy way to see a graph of how the ASV numbers change across commits, I thought you might need to click one ASV job at a time. A quick google seems to indicate `asv publish` could generate an html page where one might view that… would there be any interest in setting up the asv-benchmarking.yml to create that site, and put it somewhere easy for developers to find?
> 
> While looking into that yml file briefly I also noticed it has instructions to post to https://github.com/UXARRAY/uxarray-asv, however that repo seems to have not been updated for over a year.
> 
> (Something like `asv publish` and updating the yml file shouldn't go in this PR, but I would be happy to open a feature request/issue report if those ideas sound good to you.)

Check this and let me know what you think https://uxarray.github.io/uxarray-asv/. And yeah, we have to restart uploading results there

Sevans711:
> Check this and let me know what you think https://uxarray.github.io/uxarray-asv/. And yeah, we have to restart uploading results there

Ah, something like that is exactly what I was hoping for (once recent results start getting added)… I just opened #1670 to track this.

> My understanding is that ASV is relatively simple, and `peakmem` benchmarks are always going to report the peak memory of the process, not the function, so regardless you would still entrain the memory usage of library imports etc. In my experience so far, ASV is kind of an odd package; it's better to think of it as a compiler wrapper with its own semantics, rather than a proper Python library.

Ah, okay, I see what you're saying. At this point I'm realizing I probably won't be convinced until I try it, so I'm going to see if I can try it locally. I'll report back here on how it goes, and then move forward with review (assuming my local results agree with your understanding).

benchmarks/mpas_ocean.py
cmdupuis3: Probably an oversight, I brought it inline with the others in [1c3d176](https://github.com/UXARRAY/uxarray/pull/1609/commits/1c3d176c434baf1bcdefdb1a060e8f2cc71dd719)

benchmarks/mpas_ocean.py
cmdupuis3: Hm, actually looking at it again, `integrate` doesn't really track anything `face_areas` wouldn't. We would basically be measuring the throughput of `einsum`. I'm thinking we should just remove these two and just have `time_integrate`.

rajeeja (changes requested):
Good find and good fix. How about adding that `peak_allocated` helper you sketched to this PR and using it for the `peakmem_*` cases? It would measure allocation inside the function rather than the whole process, so the import and numba compile overhead drops out and `track_nbytes` gets to stay an honest size check. Also, did @erogluorhan's question about `IntegratePeakMem` get lost?

benchmarks/mpas_ocean.py
erogluorhan: Hmm, I think that makes sense to me re peakmem and nbytes.

erogluorhan (commented):
> @Sevans711 My understanding is that ASV is relatively simple, and `peakmem` benchmarks are always going to report the peak memory of the process, not the function, so regardless you would still entrain the memory usage of library imports etc. In my experience so far, ASV is kind of an odd package; it's better to think of it as a compiler wrapper with its own semantics, rather than a proper Python library.

Is this to clarify the following part of @Sevans711 's previous comment?

>>This might be a naive question, but why can't you just run the functions once during setup? I think that's the pattern used by other benchmarks to avoid the transient "cold start" overhead leaking into the tests. For example, something like this:

```python
class FaceBounds:
    def setup(self, grid_path):
        tmp = ux.open_grid(grid_path)
        tmp.bounds   # cold start memory costs of imports, numba caching, etc. will happen here
        self.uxgrid = ux.open_grid(grid_path)  # make a fresh object for the actual tests below

    def peakmem_face_bounds(self, grid_path):
        """Peak memory usage obtain ``Grid.face_bounds."""
        face_bounds = self.uxgrid.bounds
```

>>Would this not give a proper peakmem measurement of just `Grid.bounds`?

I am also curious about why you had `open_grid()` added into the peak-mem calculations. 

Shouldn't we avoid that "cold start" overhead coming from `open_grid()` call and measure only `.bounds`?

Sevans711:
Following up from earlier. Even something extremely simple has the same issue. I tried the following:
```python
class FaceBounds:

    params = [grid_quad_hex]

    def setup(self, grid_path):
        self.uxgrid = 7

    def teardown(self, n):
        del self.uxgrid

    def peakmem_face_bounds(self, grid_path):
        print('wow, 200+ MBs just for a print statement?')
```
which leads to ASV reporting roughly ~240M for this benchmark.

I explored the [ASV documentation](https://asv.readthedocs.io/en/latest/writing_benchmarks.html#peak-memory) a bit more and found this note:
> The peak memory benchmark also counts memory usage during the setup routine, which may confound the benchmark results. One way to avoid this is to use setup_cache instead.

I tried using `setup_cache`, but the syntax is different and I'm not really sure how to get it working properly.

Given this, the comments above, and the meeting discussion about maybe using something like the tracemalloc you suggested, I will not review just yet, but will keep an eye on this thread. Once you have gotten a chance to look into these things / decide which route you want to take, please clarify here and then I will be happy to review. I don't have a strong preference at this point as long as the PR ends up with some meaningful benchmarks, one way or another.

benchmarks/mpas_ocean.py
cmdupuis3: Is there still something we need here?

cmdupuis3:
These benchmarks have basically been rewritten using the tracemalloc approach to test something like how you'd expect a peakmem benchmark to work without the overhead of library imports and njit caching.

Running these benchmarks show numbers that show stronger scaling with grid resolution, which is what we'd expect from lower overhead.

Re-requesting reviews, since they're basically all different.

> [50.00%] ··· mpas_ocean.FaceAreas.track_peakmem_face_areas                                                                                                                                                                                         ok
[50.00%] ··· ============ =======
              resolution         
             ------------ -------
                480km       791k 
                120km      2.12M 
             ============ =======

benchmarks/face_bounds.py
```diff
@@ -25,6 +34,49 @@ def time_face_bounds(self, grid_path):
...
+    track_peakmem_face_bounds.unit = "bytes"
+
+
+class FaceBoundsPeakMem:
+    """Peak memory of a cold start: import uxarray, open a grid, get its bounds.
+
+    Whole-process ``ru_maxrss``, so the ~250MB uxarray import is part of the
+    number by design -- this is the cold-start cost, not the cost of ``bounds``.
+    For that, see ``FaceBounds.track_peakmem_face_bounds``.
+    """
```
Sevans711, line 70: Thank you for leaving this comment (and similar comment on the peakmem_gradient benchmark). I saw the benchmark results and came to this part of the code specifically to comment that maybe it should be like track_peakmem benchmarks instead of ASV's default peakmem here. But now I see that it is intentional.

So, instead of suggesting to change any implementation, my only suggestion would be to change the class name to make it clearer that this is intentionally for a cold start. Maybe something like `FaceBoundsPeakMemColdStart`?

benchmarks/mpas_ocean.py
```diff
@@ -57,27 +60,91 @@ def teardown(self, resolution, *args, **kwargs):
...
-        integral = self.uxds[data_var].integrate()
+    def track_nbytes_integrate(self, resolution):
+        """Grid footprint after integrating."""
+        self.uxds[data_var].integrate()
+        return grid_nbytes(self.uxds.uxgrid)
+
+    track_nbytes_integrate.unit = "bytes"
+
+
+class GradientPeakMem:
```
Sevans711: Similarly to previous comment, I would suggest to rename to something like `GradientPeakMemColdStart`

benchmarks/mpas_ocean.py
erogluorhan: If you took action to remove them, no I don't think anything else left here.

Sevans711 (changes requested):
With the changes to using the tracemalloc approach, I agree the results are much more intuitive, and benchmark numbers looks good, appropriately scaling with grids. The original problem of peakmem benchmarks always showing the same "improvement" has also been solved, as shown by the benchmarks here correctly not showing any improvements.

Only one tiny suggestion, which is to change names to clarify which peakmem benchmarks are cold start, because it originally confused me to still see a few peakmem benchmarks with 200+ MBs for small grids. I left corresponding inline comments.

rajeeja (approved):
Fixes a lot of existing bugs and looks good to go.

benchmarks/face_bounds.py
cmdupuis3: Renamed to `FaceBoundsTracemalloc` to signal that this is more of a tracemalloc benchmark than a true peakmem benchmark

benchmarks/mpas_ocean.py
cmdupuis3: Likewise `GradientTracemalloc`

benchmarks/face_bounds.py
cmdupuis3: nvm, in working with the other benchmarks PR, it's clear that `ColdStart` is a better description

Sevans711 (approved):
With the ColdStartRss name clarifications in place, to me everything looks ready to merge now!

PR #1621, opened 2026-07-24
Add custom error types in uxarray

### LINKED ISSUE #1556: custom error types in uxarray

It would be convenient to have custom error types in uxarray. This could help with quickly presenting the relevant info, and also improves error handling for packages using uxarray. This could also speed up debugging, and enable cleaner try.. except error handling. For example, we might define:
```python
class DataCenteringError(ValueError):
    """"raised when data is not centered properly, e.g. expected node centered data but got edge centered."""
    pass

class DimensionalityError(ValueError):
    """raise when there are the wrong number of dimensions, or dimensions have wrong names."""
```

Then we can change the relevant code which raises ValueError to instead raise a more specific error, e.g., in plotting routines when there are too many dimensions, we can `raise DimensionalityError(f"Expected data with a single dimension (other axes may be length 1), "but got dims {data.dims} with shape {data.shape}.")` instead of raising a ValueError. Similarly, in aggregation.py, we could `raise DataCenteringError(f"Data Variable must be mapped to the corner nodes of each face")` instead of just a ValueError.

`xarray` doesn't use custom error classes, but other popular scientific packages like `numpy` and `pandas` do. There isn't much cost to setting it up, and it would make the errors easier to read. Another benefit is cleaner error handling in the future, e.g. if you know some code might involve arrays with the wrong number of dimensions, you can `except DimensionalityError` without worrying about accidentally catching an unrelated ValueError. Additionally, it will be fully backwards compatible, e.g. except ValueError would still catch DimensionalityError, due to subclassing.

This would also help me address #1285, for example, which asks for more descriptive error when dealing with empty uxarrays.

Sevans711:
Closes #1556 

## Overview
Adds custom error types in uxarray. Defining custom error types can help with:
- Quickly presenting relevant info (e.g. "DataCenteringError" name quickly clarifies this is about data being centered incorrectly, e.g. centered on nodes but expected it to be on faces). This could speed up debugging for developers and users alike.
- Cleaner error handling within/outside of uxarray (e.g., catch "uxarray.errors.DimensionError" without catching other ValueErrors by accident).

In particular, this PR adds the following error types:
```python
class DataCenteringError(ValueError):
    """data centering issue, such as expecting node-centered data but got edge-centered."""


class DimensionError(ValueError):
    """issue with dimension(s), such as wrong size(s), name(s), or number of dimensions."""


class GridInvalidError(ValueError):
    """data does not correspond to a valid uxarray.Grid.
    E.g., unrecognized format, duplicate nodes, or some faces with area < 0.
    """


class GridsMismatchError(ValueError):
    """attempted to perform an operation involving two incompatible uxarray.Grid objects"""
```

## Small expansion of scope
This PR expands scope slightly beyond that of the original issue as follows:
- move `YacNotAvailableError` from uxarray/remap/yac.py to uxarray/errors.py.

The following justification for this move is provided in the docstring at the top of errors.py:
> Defining all such custom error types in this file, instead of throughout the codebase, helps with maintainability, discoverability, and convenience, as the import syntax will always be "import uxarray.errors.CustomError", and help(uxarray.errors) will list all custom error types in one place.

Though, this could be reverted if PR reviewers disagree or feel that this should be scoped as a separate issue instead.

## Compatibility notes:
This PR should be fully backwards compatible, or nearly fully backwards compatible. Compatibility is maintained via subclassing. For example, any code catching ValueError will still catch any of the newly-defined custom error types, because they inherit from ValueError.

This PR adjusts two tests to check that custom error types are indeed being raised (e.g. `with pytest.raises(DataCenteringError)` instead of `with pytest.raises(ValueError)`). It does not adjust all tests to the new error types because it is desirable to see that the existing tests still pass, to confirm there are no breaking changes here.

Thinking about it a bit further, there are three technically-possible breakage points for existing code, but only for hopefully-very-unlikely code design patterns:
1. Any string-based error processing which checks for string match to error class name, e.g. `repr(err).startswith("ValueError")`
2. Any processing which checks for type equality, e.g. `type(err) == ValueError`
3. Any processing which specifically excludes subclasses, e.g. `isinstance(err, Exception) and not isinstance(err, ValueError)`, could fail in some cases. This PR replaces Exception with a custom error type in some cases, such as using DataCenteringError instead of Exception in a few places in grid.py.

## Other Recommended Changes (Not Implemented Here)

Should custom errors become part of the public API? I would be interested in at least adding `import uxarray.errors` in `uxarray.__init__.py`, so that users who have imported uxarray already can do things like: `ux.errors.DimensionError` without needing to remember to add another import statement like `from uxarray.errors import DimensionError`.

There are a few locations where the new error types would better describe the actual error, but which are not included in this PR because they could break backwards compatibility. These could be added here if PR reviewers want, or split into a separate issue and PR afterwards. Including these here is not necessary to close issue #1556.

- RuntimeError → GridInvalidError, in UxDataArray.get_dual, UxDataset.get_dual, and Grid.get_dual: "Duplicate nodes found, cannot construct dual"
- RuntimeError → DimensionError, in Grid.from_face_vertices: f"Invalid Input Dimension: {face_vertices.ndim}. Expected dimension should be 3: [n_face, n_node, two/three] or 2 when only one face is passed in."
- RuntimeError → GridInvalidError, in Grid.validate: "Mesh validation failed."
- AssertionError → DimensionError, in grid.neighbors._prepare_xy_for_query: "The dimension of each coordinate pair must be two (lon, lat). Did you attempt to query using Cartesian (x, y, z) coordinates?", **and** "The dimension of each coordinate pair must be two (lon, lat).)"
- AssertionError → DimensionError, in grid.neighbors._prepare_xyz_for_query: (similar messages as above)
- RuntimeError → GridInvalidError, in io.utils._parse_grid_type: "Failed to parse uxgrid information from xarray.Dataset."

I also put together a variety of error-type changes I would recommend but which are unrelated to this PR, because they do not use the new error types. _(Example: ValueError → TypeError in grid.utils.make_setter: f"{key} must be an xr.DataArray", which gets raised if not isinstance(value, xr.DataArray).)_ I have opened a separate issue to track those, see #1622.

## PR Checklist

**General**
- [X] Add appropriate labels

**Testing**

**Documentation**
- [X] Docstrings have updated with any function changes
- [N/A] User functions have been added to `docs/user_api/index.rst`

### DIFF
--- a/test/core/test_dataarray.py
+++ b/test/core/test_dataarray.py
@@ -1,5 +1,6 @@
 import numpy as np
 import uxarray as ux
+from uxarray.errors import DimensionError
 from uxarray.grid.geometry import _build_polygon_shells, _build_corrected_polygon_shells
 from uxarray.core.dataset import UxDataset, UxDataArray
 import pytest
@@ -124,7 +125,7 @@ def test_isel_invalid_dim(gridpath, datasetpath):
     uxda = UxDataArray(data, dims=["time", "n_face"], uxgrid=uxds.uxgrid)
 
     with pytest.raises(
-        ValueError,
+        DimensionError,
         match=r"Dimensions \{'invalid_dim'\} do not exist\..*Available dimensions: \('time', 'n_face'\)",
     ):
         uxda.isel(invalid_dim=0)
--- a/test/grid/integrate/test_zonal.py
+++ b/test/grid/integrate/test_zonal.py
@@ -8,6 +8,7 @@ import numpy.testing as nt
 
 import uxarray as ux
 from uxarray.constants import ERROR_TOLERANCE
+from uxarray.errors import DataCenteringError
 
 class TestZonalCSne30:
 
@@ -380,7 +381,7 @@ class TestZonalAnomaly:
         uxda = ux.UxDataArray(
             np.zeros(uxgrid.n_node), dims=["n_node"], uxgrid=uxgrid
         )
-        with pytest.raises(ValueError, match="face-centered"):
+        with pytest.raises(DataCenteringError, match="face-centered"):
             uxda.zonal_anomaly()
 
     def test_invalid_lat_input_raises(self):
--- a/test/test_remap_yac.py
+++ b/test/test_remap_yac.py
@@ -2,7 +2,8 @@ import numpy as np
 import pytest
 
 import uxarray as ux
-from uxarray.remap.yac import YacNotAvailableError, _import_yac
+from uxarray.errors import YacNotAvailableError
+from uxarray.remap.yac import _import_yac
 
 
 try:
--- a/uxarray/core/aggregation.py
+++ b/uxarray/core/aggregation.py
@@ -1,6 +1,7 @@
 import numpy as np
 
 import uxarray.core.dataarray
+from uxarray.errors import DataCenteringError
 from uxarray.grid.connectivity import get_face_node_partitions
 
 NUMPY_AGGREGATIONS = {
@@ -32,7 +33,7 @@ def _uxda_grid_aggregate(uxda, destination, aggregation, **kwargs):
         elif destination == "edge":
             return _node_to_edge_aggregation(uxda, aggregation, kwargs)
         else:
-            raise ValueError(
+            raise DataCenteringError(
                 f"Invalid destination for a node-centered data variable. Expected"
                 f"one of ['face', 'edge' but received {destination}"
             )
@@ -62,7 +63,7 @@ def _uxda_grid_aggregate(uxda, destination, aggregation, **kwargs):
         #     raise ValueError("TODO: ")
 
     else:
-        raise ValueError(
+        raise DataCenteringError(
             "Invalid data mapping. Data variable is expected to be mapped to either the "
             "nodes, faces, or edges of the source grid."
         )
@@ -73,7 +74,7 @@ def _node_to_face_aggregation(uxda, aggregation, aggregation_func_kwargs):
     import dask.array as da
 
     if not uxda._node_centered():
-        raise ValueError(
+        raise DataCenteringError(
             f"Data Variable must be mapped to the corner nodes of each face, with dimension "
             f"{uxda.uxgrid.n_face}."
         )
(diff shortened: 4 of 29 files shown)


cmdupuis3:
> There are a few locations where the new error types would better describe the actual error, but which are not included in this PR because they could break backwards compatibility. These could be added here if PR reviewers want, or split into a separate issue and PR afterwards. 

Can we add them as warnings instead? It should be easy to promote them to errors later. I just lean towards having the messages present in some form.

Sevans711:
> > There are a few locations where the new error types would better describe the actual error, but which are not included in this PR because they could break backwards compatibility. These could be added here if PR reviewers want, or split into a separate issue and PR afterwards.
> 
> Can we add them as warnings instead? It should be easy to promote them to errors later. I just lean towards having the messages present in some form.

Thank you for taking a look! To clarify, the locations specified here are already errors on the `main` branch of uxarray. My suggestion is that their type should be updated. Clarifying the syntax from my original post, for this first example:
- RuntimeError → GridInvalidError, in UxDataArray.get_dual, UxDataset.get_dual, and Grid.get_dual: "Duplicate nodes found, cannot construct dual"

what I mean is: the code currently raises `RuntimeError("Duplicate nodes found, cannot construct dual")` in `UxDataArray.get_dual`, `UxDataset.get_dual`, and `Grid.get_dual`, but I would want that to be replaced with raising `GridInvalidError("Duplicate nodes found, cannot construct dual")` instead.

But, I'm not sure if those replacements should be added in this PR, since they break backwards compatibility in a larger way than the other changes (e.g., GridInvalidError is not a subclass of RuntimeError, so changing the error type in this example would prevent it from being caught by existing `try ... except RuntimeError` statements.)

cmdupuis3:
Ah, yeah, that makes sense.

cmdupuis3:
Alright, I tried out raising error messages with each of them, and they all raise with the correct error type. I'm going to assume that exhaustively triggering all the new callsites is overkill, so I'm approving this.

cmdupuis3 (approved):

rajeeja (approved):
good work

Issue #1587, opened 2026-07-17
Investigate Python thread-scaling of AccuX vs FP64 kernels


rajeeja:
Follow-up from #1513: study how the AccuX/FP64 per-point ratio scales with thread count in the Python/Numba kernels, and whether it approaches the C++ behavior on larger core counts / working sets.

PR #1624, opened 2026-07-27
Better chunking strategies for constlat intersections and zonal routines.

### LINKED ISSUE #1650: Post-accusphere optimized routines (constlat intersections and zonal)
Sub-issue of #1587 to keep track of PR #1624.

cmdupuis3:
This PR contains two post-accusphere optimizations, eliminating low-level hard materializations by using a vector-based masking strategy rather than individual conditionals, and reduced `zonal_mean` peakmem by building only the candidate faces instead of the whole grid.

Partly addresses #1587

Closes #1650

## Overview

> Phase A eliminated the largest peak-memory amplifier in the zonal path: the port materialized the whole-grid (n_face, n_max, 2, 3) face-edge array up front (a 5.8× blow-up over the node coordinates, with a ~23 MB build peak on a 28k-face grid) even though each latitude only touches ~1% of faces. I added an @njit(parallel=True) subset builder, _get_cartesian_face_edge_nodes_array_subset, that constructs edges only for the candidate faces of a given latitude/band — bit-identical to indexing the full array — and rewired _compute_non_conservative_zonal_mean and _compute_face_band_weights to build per-candidate subsets instead of the whole grid. Peak memory for a 180-latitude zonal_mean dropped from 23.66 MB to 0.27 MB (≈88×) and it ran ~10% faster (no full build, no per-latitude fancy-index copies), verified lossless via a git-stash A/B (the only diff was a pre-existing 4.4e-16 parallel-reduction nondeterminism) with the full zonal/cross-section suite passing.

> Phase B moved the six edge/face screeners (constant_lat/lon_intersections_no_extreme, constant_lat/lon_intersections_face_bounds, faces_within_lat/lon_bounds) off @njit and onto plain vectorized NumPy, drawing the boundary so the low-level Numba kernels stay for real geometry while these memory-bound elementwise predicates use NumPy — which is ~2.1× faster here and, unlike an njit kernel that forces a full .values materialization, composes with dask (a single _flatnonzero helper leans on NumPy's __array_function__ protocol so a dask mask reduces block-wise, no explicit dask branch). Call sites pass .data instead of .values so a chunked grid stays lazy, and edge coordinates are gathered positionally (node_z.data[conn.ravel()].reshape(...)) to stay chunk-friendly. Along the way it fixed two real latent bugs: get_edges_at_constant_latitude referenced a nonexistent self.edge_node_z (it raised AttributeError on every call), and both edge paths crashed on chunked grids because xarray can't vindex with a dask indexer. Results are bit-identical to the original per-element loops across 300 randomized trials, dask==numpy, and it's committed as bf0abbec "Lazy intersections".

## PR Checklist

**General**
- [x] Add appropriate labels

**Testing**

**Documentation**


cmdupuis3:
pre-commit.ci autofix

rajeeja:
Thanks for putting this together. I'll take a look at this and post my comments soon

uxarray/core/zonal.py
```diff
@@ -301,11 +310,22 @@ def _compute_face_band_weights(uxgrid, bands):
 
         partial = all_overlapping[~fc_mask]
         partial_pos = np.nonzero(~fc_mask)[0]
-        for pos, f in zip(partial_pos, partial):
-            nedge = n_nodes_per_face[f]
-            weights[pos] = _compute_band_overlap_area(
-                faces_edge_nodes_xyz[f, :nedge], zmin, zmax
+        if partial.size:
```
rajeeja, line 314: New branch (zero partial faces). Add a test for a band with only fully-contained faces.

rajeeja (changes requested):
Verified independently: subset builder is bit-identical to whole-grid builder (tested HEALPix + mixed-polygon MPAS), vectorized screeners match brute-force reference, both dask/AttributeError bugs reproduce on main and are fixed here. Two small comments below, non-blocking on correctness but worth addressing before merge.

uxarray/core/zonal.py
cmdupuis3: Not sure what you mean here... The weights are already computed for all faces overlapping the band, this branch only reweights partial faces, so a strict fully-contained set of faces is already accounted for by the time this branch is reached.

rajeeja: Fair — correctness is unaffected. I just meant the `partial.size == 0` case is untested; a band with only fully-contained faces would cover it. Non-blocking.

cmdupuis3: I think it technically is tested for `bands=[-90, 90]` in `test_conservative_full_sphere_conservation`, but maybe we can add a test on a less trivial band.

On the other hand, this opened a rabbit hole that led to issue #1644, so we might want to fix that first.

erogluorhan (commented):
Overall this looks good to me. Just a couple questions: 
- Should we worry about the couple benchmarks that got worse? I believe, no, they shouldn't be directly related to the changes here, but am curious about your thoughts.
- Could you add test cases to cover the two latent bugs that this PR fixes?

cmdupuis3:
@erogluorhan Pretty sure those benchmark regressions were just variability, they're gone in the new batch. Regression tests are now added for both bugs.

uxarray/core/zonal.py
rajeeja: marking resolved, we can do a followup PR

rajeeja (approved):

rajeeja:
let's wait for asv fixes to be merged before we merge this.

erogluorhan (approved):
Looks good to me!

Issue #1622, opened 2026-07-24
Update uxarray error types for cases with misleading/confusing types


Sevans711:
While addressing #1556 I noticed a variety of changes I would recommend making to error types being used, to make them correspond more closely to the actual error being reported. For example, when an error is raised because an object was expected to be an xr.DataArray or a dask.Array, but it is not either of those types, the error should be a TypeError, not a ValueError.

Here is an exhaustive list of all such cases, split into two sections. First, recommended changes which would not break backwards compatibility; the recommended type is a subclass of the currently-used type. Second, recommended changes where the recommended type is not a subclass of the currently-used type.

EDIT: updated the second section to include changes originally recommended in #1621 but which did not end up getting included in that PR.

EDIT: also note: work should only begin on this issue after at least one other uxarray developer confirms these changes sound reasonable. It will also be necessary to decide if these changes need to be incorporated with some sort of deprecation cycle. I personally would lean towards _not_ having a deprecation cycle, because: (A) these are already error messages anyway, (B) I would be surprised if many people are running code that relies directly on the type of errors being raised by uxarray, and (C) getting a deprecation warning during a crash sounds quite confusing, and might make it harder to quickly understand the actual cause of the crash.

## Proposed changes which do _not_ break backwards compatibility:

1. Exception → ValueError in grid.bounds.insert_pt_in_latlonbox: "Longitude point out of range"
2. Exception → AssertionError or ValueError(?) in grid.bounds.insert_pt_in_latlonbox: "Logic error in longitude box width calculation"
3. Exception → AttributeError in property Grid.inverse_indices: "Grid is not a subset, therefore no inverse face indices exist"

## Proposed changes which _do_ break backwards compatibility:

4. ValueError → TypeError in core.aggregation._node_to_face_aggregation and core.aggregation._node_to_edge_aggregation: (error currently has no message, but is raised to indicate: not isinstance(uxda.data, (np.ndarray, da.Array)).)
5. ValueError → TypeError in UxDataArray.zonal_anomaly: "Invalid value for 'lat'. Must be a tuple (start, end, step) or array-like band edges."
6. ValueError → TypeError in Grid.from_face_vertices: "Input must be either a list, tuple, or np.ndarray"
7. ValueError → NotImplementedError in Grid.boundary_node_indices: (error currently has no message, but seems like it is supposed to indicate that construction of boundary_node_indices is not yet implemented.)
[...]

PR #1628, opened 2026-07-28
Check data location, not `size==n_face`, in UxDataArray.to_geodataframe() and .integrate()

### LINKED ISSUE #1616: to_geodataframe() and integrate() check size==n_faces, not data_location
### Version

2026.6.0

Source

### What happened?

UxDataArray.to_geodataframe() and .integrate() check `self.values.size == self.uxgrid.n_face` instead of `self._face_centered()` (or `self.data_location=='face_centered'` or `self.mapping=='faces'`, I'm not quite sure which of these is the preferred syntax). Similarly, they check `self.values.size == self.uxgrid.n_node` instead of checking if the data is node centered.

I'm not sure if it is possible to have n_faces==n_nodes or n_faces==n_edges, or n_nodes==n_edges, but if it is, this code would lead to very confusing bugs for such a case.

I expected the code to check `self._face_centered()` (or `self.data_location=='face_centered'` or `self.mapping=='faces'`) instead of just comparing values size and n_faces.

```python

```

Sevans711:
Closes #1616

## Overview
Switches to using `self._face_centered()` instead of `self.values.size == self.uxgrid.n_face`, inside `UxDataArray.to_geodataframe()` and `UxDataArray.integrate()`. The comparison to `self.uxgrid.n_face` is brittle; it might get messed up by any grids which happen to have the same number of nodes as faces, or same number of edges as faces.

Entirely removes the `... == self.uxgrid.n_edge` and `... == self.uxgrid.n_node` checks, and rewrites corresponding error messages. Those checks were just being used to provide a different error message based on size-based guess of data location; ultimately the errors were all ValueErrors with different ways to say "data that is not face_centered is not yet supported".

Adds a comment inside the integrate() method suggesting that it should support integration of face_centered data regardless of the n_face dimension position. In the meantime, the method continues to only support integration where the n_face dimension is the last dimension. (Possibly related: #1461.)

~Still draft PR because still need to add regression tests, as discussed in issue 1616.~
EDIT: now ready for review. Added tests for UxDataArray.to_geodataframe() and UxDataArray.integrate() for a node-centered array on a grid with n_face==n_node==6 (with same structure as the grid mentioned here: https://github.com/UXARRAY/uxarray/issues/1616#issuecomment-5093302657). These tests confirm a ValueError must be raised in this case, because these methods are not yet supported. Currently on `main`, these tests silently incorrectly assume face-centered data and provide answers instead of crashing. A crash should be expected here, to prevent silently incorrect results. (Sidenote: should probably be NotImplementedError instead, but changing error types should not be part of this PR; see #1622 instead for further discussion about error types.)

## PR Checklist

**General**
- [X] Add appropriate labels

**Testing**

### DIFF
--- a/test/grid/geometry/test_geometry.py
+++ b/test/grid/geometry/test_geometry.py
@@ -1,11 +1,13 @@
 import numpy as np
-import numpy.testing as nt
 import pytest
+import xarray as xr
 
 import uxarray as ux
 from uxarray.constants import ERROR_TOLERANCE, INT_FILL_VALUE
-from uxarray.grid.coordinates import _lonlat_rad_to_xyz, _normalize_xyz, _xyz_to_lonlat_rad
-from uxarray.grid.geometry import _pole_point_inside_polygon_cartesian
+from uxarray.grid.coordinates import _lonlat_rad_to_xyz, _normalize_xyz
+from uxarray.grid.geometry import haversine_distance, _pole_point_inside_polygon_cartesian
+from uxarray.grid.point_in_face import _face_contains_point
+from uxarray.grid.utils import _get_cartesian_face_edge_nodes_array
 
 
 def test_antimeridian_crossing():
@@ -197,13 +199,6 @@ def test_haversine_distance_creation():
 
     np.testing.assert_allclose(distance, expected_distance, atol=ERROR_TOLERANCE)
 
-from uxarray.constants import ERROR_TOLERANCE
-from uxarray.grid.coordinates import _lonlat_rad_to_xyz
-from uxarray.grid.utils import _get_cartesian_face_edge_nodes_array
-from uxarray.grid.point_in_face import _face_contains_point
-from uxarray.grid.geometry import haversine_distance
-
-
 
 def test_engine_geodataframe(gridpath):
     uxgrid = ux.open_grid(gridpath("ugrid", "geoflow-small", "grid.nc"))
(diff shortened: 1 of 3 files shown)


Sevans711:
Notes about regression tests not added here:
- Did not add a test for the n_edge == n_node case, because that case did not actually produce meaningful differences in results, because the code was checking: "is size == n_face? If yes, provide results, if no, make ValueError." Comparison between n_edge and n_node merely caused the ValueError to have a slightly different message.
- Did not add a test for the n_edge == n_face case, because either grids like that cannot actually exist, or I am just not clever enough to come up with one right now. If you can think of one please let me know, I would be happy to add it!

dylannelson:
> any grids which happen to have the same number of [...] edges as faces

Ok I ended up going down a rabbit hole while reading about `# of ___ = # of Faces` equations. What you're saying about `# nodes = # of faces` seems like a totally valid concern and a great catch. For the 2nd part though I'm not sure if it's possible from what I'm reading. This doesn't impact anything in the code from what I see but wanted to share:

A few searches and a few articles point to [Euler's characteristic formula](https://en.wikipedia.org/wiki/Euler_characteristic), that suggests: 
`V - E + F = 2` (v = #nodes/vertices, e = #edges, f = #faces)
If E and F are equivalent (number of faces = number of edges), they cancel out, and that just leaves V = 2, and if you only have 2 nodes/vertices, you don't have enough nodes to make a polygon, so you likely don't have a grid, or face either.
So the only solution to E = F is if:
- V = 2 and 
- E and F = 0

So you wouldn't really have much of anything

image
*image from the Wikipedia link above*

This does assume that in equation (X = V-E+F) that X (the Euler characteristic) is = 2, which may vary based on the type of shape. X may change if the shape has holes (like a donut), but most closed shapes, like in the image above, have X = 2. Personally not sure if users can have grids that can vary wildly enough to make them have a non 2 Euler characteristic though.

Sevans711:
> Ok I ended up going down a rabbit hole while reading about `# of ___ = # of Faces` equations.
> …
> A few searches and a few articles point to [Euler's characteristic formula](https://en.wikipedia.org/wiki/Euler_characteristic), that suggests: `V - E + F = 2` (v = #nodes/vertices, e = #edges, f = #faces) If E and F are equivalent (number of faces = number of edges), they cancel out, and that just leaves V = 2
> …
> This does assume that in equation (X = V-E+F) that X (the Euler characteristic) is = 2, which may vary based on the type of shape. X may change if the shape has holes (like a donut), but most closed shapes, like in the image above, have X = 2. Personally not sure if users can have grids that can vary wildly enough to make them have a non 2 Euler characteristic though.

Thank you for this clarification! This helped me a lot with understanding that formula, I saw it come up while doing some quick searches but I didn't fully understand it before.

Unfortunately I don't think it applies to all uxarray grids, so the question "can we have n_faces=n_edges" still feels like an open question to me (though probably not super high priority to find a complete answer to, for this PR at least). For example, consider a grid composed of merely a single triangle; such a grid has V=E=3, F=1, so it has χ=1, not 2.

rajeeja (approved):
Approving — this is the right fix. `size == n_face` was a guess at data location, not a check of it, and `_face_centered()` asks the question directly. Good catch on the `n_node == n_face` degenerate grid, and I like that the regression tests use a hand-built mesh with the ASCII diagram rather than a fixture file; it makes the failure mode obvious to the next reader.

Collapsing the three size-based branches into one `DataCenteringError` is a readability win too — as you note, they all said "not face-centered" in three different ways.

One nit, non-blocking: the new `DimensionError` in `integrate()` is missing a space between the sentences, so it renders as `...not yet supported.Got face_centered data...`.

```python
raise DimensionError(
    "Integration of data with n_face not as the final dimension is not yet supported."
    f"Got face_centered data, but the final dimension was {self.dims[-1]}, not 'n_face'."
)
```

Adding a trailing space to the first string fixes it.

Also worth flagging for a follow-up rather than this PR: that branch is now reachable in a case it wasn't before. On `main`, face-centered data shaped `(n_face, time)` hits the old size-based check and fails with a message about node/edge/face counts; with this change it correctly identifies the data as face-centered and reports the real problem (dimension order). That's strictly better, and your TODO already points at the fix. Just noting the error path has real traffic now, so the wording matters a bit more than it did.

Sevans711:
> Approving — this is the right fix. `size == n_face` was a guess at data location, not a check of it, and `_face_centered()` asks the question directly. Good catch on the `n_node == n_face` degenerate grid, and I like that the regression tests use a hand-built mesh with the ASCII diagram rather than a fixture file; it makes the failure mode obvious to the next reader.
> 
> Collapsing the three size-based branches into one `DataCenteringError` is a readability win too — as you note, they all said "not face-centered" in three different ways.
> 
> One nit, non-blocking: the new `DimensionError` in `integrate()` is missing a space between the sentences, so it renders as `...not yet supported.Got face_centered data...`.
> …

Thank you for looking into this closely and for the comments! I had fun creating the ASCII diagram, I initially tried to using an ASCII art generator but then ended up just making it by hand 😄

Good find with this small typo, I just pushed a fix.

dylannelson (approved):

rajeeja:
`E >= 1.5F` always (every face has >=3 edges, every edge borders <=2 faces, so `3F <= 2E`) — no Euler characteristic assumption needed, so `E == F` can't happen for any grid with faces.

PR #1640, opened 2026-08-03
Improve uxgrid=None docs and error-handling for UxDataArray & UxDataset.

### LINKED ISSUE #1620: UxDataset and UxDataArray should not allow uxgrid=None
### Version

v2026.7.0

Source

### What happened?

UxDataset and UxDataArray currently allow uxgrid=None, but their docstrings say "If `None`, it needs to be an instance of `uxarray.Grid`." I'm not really sure what that is supposed to mean.

Additionally, using uxgrid=None leads to confusing errors later, such as `AttributeError: 'NoneType' object has no attribute 'n_face'`. It would be much easier to debug an error like `TypeError: UxDataArray.__init__() missing one required argument: 'uxgrid'` instead.

I expected instantiating UxDataset and UxDataArray to fail if uxgrid is not provided.

Or, I expected a docstring which correctly clarifies what happens if uxgrid is None, perhaps providing at least one example use-case clarifying why uxgrid=None is supported?

```python

```

Sevans711:
Closes #1620.

## Overview
UxDataArray and UxDataset should not function properly if no uxgrid was provided. As discussed in #1620, the ideal solution would probably be to make `uxgrid` a required parameter in `UxDataArray.__init__` and `UxDataset.__init__`. However, subclassing from xarray has some quirks which make that solution especially tricky.

This PR takes the "lower-effort" suggestion discussed in that issue:
1. Keep `uxgrid=None` allowed during `__init__`, but improve the docstring & add comments explaining why
2. Make the `uxgrid` property raise a clear `GridInvalidError` if accessed while still `None`.

This PR also includes a few additional, related tasks:

3. Raise a TypeError if attempting to set the `uxgrid` property to `None`. (Internal methods should set `_uxgrid` instead.)
4. Add type-hinting to the `uxgrid` property.
5. Add regression tests to ensure a `GridInvalidError` gets raised if attempting to access `obj.uxgrid` when it is None.
6. Add a clear internal distinction between `obj._uxgrid` and `obj.uxgrid`. Internal methods should use `_uxgrid` for non-grid-aware operations, such as `_replace()` and `_copy()` which help with xarray subclassing. But, methods should continue to use `uxgrid` for any grid-aware operations.
7. Replace occurrences of `uxgrid` with `_uxgrid` accordingly, as per (6). Without these replacements, some existing tests failed due to having `_uxgrid=None` at some point, despite always having non-None uxgrid whenever any grid-aware operations actually get performed.

**Except for (3)**, this PR should be fully backwards-compatible, with no changes to API. I suspect that setting `obj.uxgrid=None` is not behavior which uxarray ever really intended to support, so think this "breaking change" is an improvement, and not an issue / does not require a deprecation cycle.

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**


dylannelson:
Hey there! Did a handful of tests and got to explore this pretty intesively. Only noticed 2 small things:
## 1) An old error is not needed
On this line: [where there is a custom error for handling a `None` where there should be a grid](https://github.com/UXARRAY/uxarray/blob/012e9def0056bc3fd3eb8ca1ce0cd65d2d6dc590/uxarray/core/api.py#L583) is not needed anymore. At this point, we get the `GridInvalidError` you made. The old lines:
``` python
 ref_uxgrid = getattr(objs[0], "uxgrid", None)
    if ref_uxgrid is None:
        raise AttributeError("The first object does not have a 'uxgrid' attribute.")
```
crash in the first line above, and the 2nd and 3rd line can't be reached
Example:
image
The old `AttributeError` can't be reached, because `GridInvalidError` is reached before. Not really a blocker, but could delete those 3 or so lines I think!

## 2) Pass anything to uxgrid
Around this [line here in init](https://github.com/UXARRAY/uxarray/blob/9507d9536c8b0f618152934fe562c2c92d6ff257/uxarray/core/dataarray.py#L102) in `dataarray.py` there isn't anything stopping any typr of data from being passed to `uxgrid`. It looks like [this line here ](https://github.com/UXARRAY/uxarray/blob/9507d9536c8b0f618152934fe562c2c92d6ff257/uxarray/core/dataarray.py#L94) is supposed to ensure it's actually a grid but maybe it's not working?
Example:
image

Lmk if I missed anything. I tested a variety of other things and it all seems to be in order and working as expected.

Sevans711:
Thank you @dylannelson for taking a detailed look into this!

For point (1), I see what you're getting at, but I believe that error is still necessary because there's not a separate type check for "is every object actually a UxDataArray or UxDataset?" The error can still be raised if the objects are of a different type entirely. Though, I also took a closer look at the `concat` function and noticed many issues… I created a detailed write-up in #1642. I think it probably makes the most sense to move concat-related discussion there, and leave these lines untouched in this PR.

For point (2), thank you, this is a great catch! I committed some changes which should fix that. Now, a TypeError is raised upon any attempts to set `obj.uxgrid = value` for any non-Grid `value`. I also added some tests which would have crashed accordingly before this fix.

uxarray/core/dataarray.py
```diff
@@ -85,17 +91,15 @@ class UxDataArray(xr.DataArray):
...
-        self._uxgrid = None
-
-        if uxgrid is not None and not isinstance(uxgrid, Grid):
-            raise RuntimeError(
-                "uxarray.UxDataArray.__init__: uxgrid can be either None or "
-                "an instance of the uxarray.Grid class"
-            )
-        else:
-            self.uxgrid = uxgrid
-
```
erogluorhan, line 98: With the removal of this whole check block here, what happens when a non-None, non-Grid garbage value is passed to the constructor?

I believe `ux.UxDataArray([1,2,3], dims=['n_face'], uxgrid="garbage")` would construct successfully now, which shouldn't be allowed.

If agreed, adding something like this should work sufficiently with this new implementation I believe:

```
if uxgrid is not None:                                                                                                                                                                                                                                                                                                                          
      self.uxgrid = uxgrid
```

erogluorhan (commented):
This looks good to me, only concerned about a potential regression, see below.

uxarray/core/dataarray.py
Sevans711: Thank you for reviewing! This is similar to point (2) from comments above: https://github.com/UXARRAY/uxarray/pull/1640#issuecomment-5172325706.

Passing non-None, non-Grid values now raises a TypeError. I clarified the error message a bit more in my latest commit. When I run `ux.UxDataArray([1,2,3], dims=['n_face'], uxgrid="garbage")` I see `TypeError: Expected a uxarray.Grid; got value with type=<class 'str'> (while setting UxDataArray.uxgrid = value).` I added corresponding regression tests too (e.g., [line 190 of test_dataarray.py](https://github.com/UXARRAY/uxarray/pull/1640/changes#diff-11f31b6c23850c2cd28bf0e8758fc6206fa6f992a6617174c65668a131f0a756R190))

The check itself occurs inside the uxgrid property, as noted by [comment in `UxDataset.__init__`](https://github.com/UXARRAY/uxarray/pull/1640/changes#diff-f3a9c87d3cc643d5a20d3672948de836c6e8185d78e5eeafa807f72a308eb79aR103) (with same comment in `UxDataArray.__init__` to clarify there too).

Let me know if that makes sense!

Sevans711: Ah, also:
> If agreed, adding something like this should work sufficiently with this new implementation I believe:
>
> ```
> if uxgrid is not None:                                                                                                                                                                                                                                                                                                                          
>      self.uxgrid = uxgrid

Clarifying, this is already present in `__init__` for both UxDataArray and UxDataset. Though it is in the "else" part of an "if uxgrid is None" block, instead.

erogluorhan: Ah okay, I'd wrapped up my review some time yesterday, and submitted it today (overlooking the convo between you two). This now seems fixed. Thanks!

erogluorhan (approved):
Looks great to me

dylannelson (approved):

PR #1651, opened 2026-08-06
Restore cached face_areas fast path in calculate_total_face_area

rajeeja:
The HEALPix branch assigns `result` but no longer returns, so it is always overwritten by the recomputed quadrature sum, which drops the exact equal-area values and contradicts the docstring added in this PR. That drift is also why RTOL had to be loosened, so this restores the `else` branch and puts RTOL back to 1e-7. Adds two regression tests; the first fails on the current branch head and passes with the fix.

### DIFF
--- a/test/grid/grid/test_areas.py
+++ b/test/grid/grid/test_areas.py
@@ -169,3 +169,30 @@ def test_face_areas_fesom(gridpath):
     uxgrid = ux.open_grid(gridpath("ugrid", "fesom", "fesom.mesh.diag.nc"))
     total_area = uxgrid.calculate_total_face_area()
     nt.assert_almost_equal(total_area, 8.3780, decimal=4)
+
+
+def test_total_face_area_healpix_uses_cached_equal_areas():
+    """Default args must reuse the cached ``face_areas``, not recompute.
+
+    HEALPix faces are exactly equal-area, so the cached sum is exactly 4*pi.
+    Recomputing via quadrature drifts and loses that property.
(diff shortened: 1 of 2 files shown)


Sevans711 (approved):
Great catch, thank you! Changes look good. I left a comment to clarify the RTOL choice is mostly arbitrary. You are correct the 1e-7 was causing a warning due to this issue. But, this just pushes the threshold down the line slightly; now there is a warning at 1e-9 but not 1e-8. Should we change the threshold to 1e-8? I think 1e-7 or 1e-6 is fine.

All that to say, "restored 1e-7 RTOL" is misleading; there was nothing to restore, it didn't exist before these changes, it was just the threshold at which errors began. I hope my updated comment clarifies this now. (I usually think every mostly-arbitrary tolerance deserves a comment inline with it, otherwise you come back to it later and worry too much about changing it because you have no idea why it is the number it is.)

Sevans711:
Also, thank you for taking a close look at #1649 and making this PR to the issue that would have been introduced there!

Github was flashing a shiny "want to turn this into a stacked PR?" button at me and I clicked it to see what happened. Now I'm not sure if it was actually worth stacking in this case, since it seems to have the same level of reviewer requirements as merging into main….

Apologies if stacking causes extra confusion here. Though, it maybe it is decent practice with using stacked PRs before applying them to the larger set of changes proposed in #1548 and #1654?

cmdupuis3 (approved):

PR #1654, opened 2026-08-06
Add CI tests for optional dependencies

Sevans711:
Addresses part of #1224, #1225 but does not fully close them.

## Overview
Provides tests for optional dependencies, as discussed in https://github.com/UXARRAY/uxarray/pull/1548#issuecomment-4985842875. Adds a new CI job step to run these tests. Basically the job steps are:
- (A) make and activate new venv; install without any optional dependencies; run corresponding tests
- (B) make and activate new venv; install with only the viz optional deps; run corresponding tests
- (C) make and activate new venv; install with only the geo optional deps; run corresponding tests
- (D) make and activate new venv; install with the viz and geo optional deps; run corresponding tests

The corresponding tests include checks which:
1. should pass everywhere
2. should pass as long as at least viz is installed
3. should pass as long as at least geo is installed
4. should pass only when both viz and geo are installed

And, "run corresponding tests" attempts all four checks, marking appropriately with `with pytest.raises(ImportError)` to ensure checks pass/fail as expected. E.g., case (C) marks (2) and (4) with expecting ImportError, but expects (1) and (3) to pass without any issues.

## PR Checklist

**General**

**Testing & Benchmarking**


rajeeja:
Yac's probably out of scope here — it isn't a pip-installable extra (`pyproject.toml` has no `yac` group; it's built from source with MPI/YAXT), so it doesn't fit this venv-based matrix. It's already exercised separately in `.github/workflows/yac-optional.yml`.

pyproject.toml
```diff
@@ -98,3 +98,6 @@ known-first-party = ["uxarray"]
 
 [tool.ruff.format]
 docstring-code-format = true
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]  # (intentionally excludes test_optional_deps)
```
cmdupuis3: Typo: should be "test"

Sevans711: Great catch, thank you! This would have made `pytest` show failed tests when run without specifying paths (it wasn't making the CI tests fail because the CI explicitly runs `pytest test`), because it would collect and run of the test_optional_deps tests. Fixed, and confirmed locally that `pytest --collect-only` does not contain the test_optional_deps tests anymore.

cmdupuis3 (approved):

rajeeja (approved):

PR #1656, opened 2026-08-07
Add helpful hint error message if optional dependencies not yet installed.

Sevans711:
Addresses part of #1224 #1225 but does not fully close them.

## Overview
Adds `_raise_hint_if_optional_deps_missing` which raises helpful hint upon trying but failing to import optional dependencies due to them not yet being installed.

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**

## AI Disclosure

AI Usage: GitHub Copilot inline code suggestions, some discussion with Claude.


uxarray/utils/imports.py
```diff
@@ -0,0 +1,104 @@
...
+            len(extras) >= 2 for extras in many_extras.values()
+        )  # else wrong format in _OPTIONAL_DEPS_TO_EXTRAS.
+        for pkg, extra in one_extra.items():
+            need_extras.add(extra)  # definitely need to include all of these
+        for pkg, extras in many_extras.items():
+            if any(
+                extra in need_extras for extra in extras
+            ):  # still maybe in case (1) or (2).
+                pass  # this package is already covered by other needed extras!
+            elif len(extras) == 0:  # case (3)
```
rajeeja: Unreachable — the assert above guarantees `len(extras) >= 2`, so this likely wants `>= 2` not `== 0`. Means a cartopy-only miss suggests `[all]` instead of `[geo]` or `[viz]`.

rajeeja (changes requested):
One unreachable branch, see inline.

uxarray/utils/imports.py
Sevans711: Thank you for looking into this! Commit 0f638de fixes this, makes slight improvements to message readability in case of many extras, and adds a test to the CI with expectations about the error messages. That test would have crashed before this fix, and it hard-codes the expected error messages for clarity and readability.

uxarray/cross_sections/sample.py
```diff
@@ -16,6 +18,7 @@ def _fill_numba(flat_orig, face_idx, n_face, n_steps):
 def sample_geodesic(
     start: tuple[float, float], end: tuple[float, float], steps: int
 ) -> tuple[np.ndarray, np.ndarray]:
+    _raise_hint_if_optional_deps_missing("pyproj")
     from pyproj import Geod
```
cmdupuis3, line 10: This pattern looks like it will be difficult to maintain consistently throughout the codebase as we develop it. If we're going to have this, there should definitely be some kind of linting for this so we'll know in the PRs if these are done correctly, and provide an easy way to fix it.

cmdupuis3 (changes requested):

uxarray/cross_sections/sample.py
Sevans711: Thank you for looking into this! Commit c78f80a adds a test to the pytest test suite which should fail if anyone defines a function that imports optional dependencies without all of them being properly included in a call to _raise_hint_if_optional_deps_missing(). The failure mode includes a helpful message which should make it clear enough on how to fix it.

(It also raises clear warnings if bonus optional dependencies are being included, unnecessarily, inside a call to the _raise_hint_... function.)

I chose to implement this as a pytest test instead of full-blown linting because I have a suspicion that this will be easier to maintain (e.g., I'm not yet familiar with building customized linting algorithms). There's no "fix it for me" button like a proper linting algorithm might provide, but I think that this should still be sufficient? I believe it grants the most important benefits of ensuring this pattern gets maintained, and providing clear instructions for how to fix if it if needed.

cmdupuis3: Eh, it's better, but the "fix it for me" button is my ideal in this case.

test_optional_deps/test_installed_with_no_opts.py
```diff
@@ -20,17 +20,113 @@ def test_check_requires_no_opts():
...
+
+    # to fully test the _raise_hint_if_optional_deps_missing() function,
+    # need to check cases with more than 2 extras. Add corresponding "fake packages" here.
+    uxarray.utils.imports._OPTIONAL_DEPS_TO_EXTRAS.update(
+        {
+            "_fakepackage1_": "_fakeextra1_",
+            "_fakepackage2_": ("_fakeextra1_", "_fakeextra2_"),
+            "_fakepackage3_": ("_fakeextra1_", "_fakeextra2_", "_fakeextra3_"),
+        }
+    )
```
cmdupuis3: This will have a side effect... Can we do something like this here?

```Python
    for _name, _extras in {
        "_fakepackage1_": "_fakeextra1_",
        "_fakepackage2_": ("_fakeextra1_", "_fakeextra2_"),
        "_fakepackage3_": ("_fakeextra1_", "_fakeextra2_", "_fakeextra3_"),
    }.items():
        monkeypatch.setitem(
            uxarray.utils.imports._OPTIONAL_DEPS_TO_EXTRAS, _name, _extras
        )
```

cmdupuis3 (changes requested):

test_optional_deps/test_installed_with_no_opts.py
Sevans711: Oh, good call. Fixed! Before this for some reason I suspected pytest might create a fresh python kernel for each test file, but now I see that is not the case; any tests run after this one would have the global dict modified.

(It didn't actually matter in this case yet, because the pytest call to this file was only running this file, in the "no optional deps" case of the test-optional-deps job. But, if anyone ever updated this file later to add more tests, or ran it with a single pytest call that included other files at the same time, they could have encountered those side effects.)

cmdupuis3 (approved):

rajeeja (approved):
Checked the cartopy-only case against the current branch and it now suggests [geo], [viz], or [all], so my earlier comment is resolved. Approving.

PR #1659, opened 2026-08-07
I/O integer overflow bugfix

### LINKED ISSUE #1658: `INT_FILL_VALUE` overflows in downcast, corrupting polygon meshes
### Version

v2026.07.0

Source

### What happened?

Writing a ragged grid (mixed face sizes) to ESMF silently corrupts padded slots instead of preserving them as fill values: _encode_esmf in uxarray/io/_esmf.py adds 1 to every entry of face_node_connectivity, including INT_FILL_VALUE padding, before encoding as int32. INT_FILL_VALUE + 1 (-2**63 + 1) truncates under that narrowing cast to 1, so padding is written as a valid node index (node 0) instead of the declared _FillValue of -1, and the numElementConn fallback (which counts entries != -1) inherits the same error and reports every face at maximum width; a triangle padded to width 4 round-trips back as a quad with an extra vertex at node 0. The reader (_read_esmf) has the mirror-image bug: CF decoding turns the on-disk -1 into NaN, and the code casts to INT_DTYPE before checking for the fill, which only happens to produce INT_FILL_VALUE on platforms where NaN casts to INT64_MIN (x86) — on arm64 it casts to 0, decoding padding to -1, a valid negative index that silently wraps to the last node rather than raising the IndexError INT_FILL_VALUE is designed to trigger. Both directions fail without any exception or warning; only ragged grids are affected, which is why test_esmf_round_trip_consistency (using the all-quad outCSne30 fixture) never caught it.

The fill value should not be able to be cast as a valid value or overflow into a valid value (positive or negative, since Python is like that).

```python
import numpy as np, xarray as xr, uxarray as ux
from uxarray.constants import INT_FILL_VALUE as fv

# one quad, one triangle padded to width 4
fnc = np.array([[0, 1, 2, 3], [1, 4, 2, fv]])
grid = ux.Grid.from_topology(
    np.array([0., 10, 10, 0, 20]), np.array([0., 0, 10, 10, 0]), fnc, fill_value=fv
)
grid.to_xarray(grid_format="ESMF").to_netcdf("out.nc")

raw = xr.open_dataset("out.nc", mask_and_scale=False)
print(raw["elementConn"].values)     # [[1 2 3 4] [2 5 3 1]]  <- expected -1 in last slot
print(raw["numElementConn"].values)  # [4 4]                   <- expected [4 3]

print(ux.open_grid("out.nc").face_node_connectivity.values)
```

cmdupuis3:
Closes #1658

## Overview

Several related bugs in how fill values are used in IO routines. Original case is the first one, where a downcast of `INT_FILL_VALUE` results in valid indices, resulting in spurious extra nodes and incorrect polygons.

| # | File | Line (was) | Function | Error mode |
|---|---|---|---|---|
| 1 | `uxarray/io/_esmf.py` | 148 | `_encode_esmf` | `+1` applied to fill slots, then int32 encoding truncates `-2**63+1` → `1`. Padding written as node 0. |
| 2 | `uxarray/io/_esmf.py` | 96 | `_read_esmf` | Cast to int *before* fill check. Relies on `NaN → INT64_MIN`; arm64 gives `0`, so padding decodes to `-1` — a negative index that wraps. |
| 3 | `uxarray/io/_exodus.py` | 204 | `_encode_exodus` | Matched padding on `== -1`, but padding is `INT_FILL_VALUE`. Never fired → one max-width block, wrong element types, `INT_FILL_VALUE+1` written as an index. |
| 4 | `uxarray/io/_exodus.py` | — | `_encode_exodus` | *(exposed by #3)* Block regrouping reorders faces with no `elem_num_map`; face-centered data silently misaligns. |
| 5 | `uxarray/io/_scrip.py` | 213 | `_encode_scrip` | `NaN` written to padded corners; reader deduplicates it into a real node. `n_node` 5→6, NaN coordinates. |

## PR Checklist

**General**

**Testing & Benchmarking**

- [ NA ] If this PR could affect performance, ran ASV benchmarks and confirmed they show expected behavior (add a new benchmark if necessary)

**Documentation**

- [ NA ] User (public) functions have been added to `docs/api.rst`

## AI Disclosure

Claude Opus 5, for bug search and generated code.

AI Usage:


cmdupuis3:
pre-commit.ci autofix

Sevans711:
Thank you for working on this issue! I'm feeling a little overwhelmed when trying to review this, due to almost 700 lines of changes across 7 files, simultaneously fixing 5 reported bugs, and possibly also including extra changes which are not directly related to fixing the underlying bugs (or I might just not be understanding it clearly just from a quick glance).

To make this easier to review, would you be able to provide a more detailed explanation about which changes relate to which bugs, and also whether the scope of this PR was expanded beyond simply fixing the originally-reported bugs? (For example, I noticed that the implementation of `_scrip.py`'s `_to_ugrid` seems to have changed much more significantly than merely editing parts of the code related to fill values. Is this an intentional expansion of this PR's scope to include "refactor scrip's `_to_ugrid` method", or is it required in order to fix the original bugs?)

A table mapping bugs to sections of the code that needed to be changed to fix that bug and test the fix, would be one way to do this. Another way could be to instead split this PR up into multiple smaller PRs.

Could you also clarify, why did you remove the `test_esmf_round_trip_consistency` test entirely? Is it no longer valid for some reason?

PR #1663, opened 2026-08-10
Normalize the gradient by the dual cell and add the spherical metric terms to curl and divergence

### LINKED ISSUE #1662: gradient divides by primal face area but integrates over the dual cell; curl and divergence drop the spherical metric term
I was putting together a demo of the `curl` feature for a paper and the numbers came out about 2x too big, so I dug in. Two things are going on, and they partly cancel, which is why it looks like a clean factor of 2 on quad meshes but 1.5 on hexagons.

First, `_compute_gradients_on_faces` walks the dual cell (centroids of the faces sharing a node with the face, per its own docstring) but `_normalize_and_project_gradient` divides by the primal face area, so the gradient comes out inflated by `A_dual/A_primal` — about 4 on quads, 3 on hexagons. It isn't truncation error; refining healpix z3 through z6 converges to 3.9278, 3.9811, 3.9952, 3.9988. Second, `UxDataArray.curl` computes `dv/dx - du/dy`, which is the planar formula and is missing the `u*tan(lat)/a` metric term from the spherical expression, and `divergence` is missing the companion `-v*tan(lat)/a` term.

```python
import numpy as np
import uxarray as ux
from uxarray.core.dataarray import UxDataArray

g = ux.Grid.from_healpix(zoom=5)
lat = np.deg2rad(g.face_lat.values)
m = np.abs(lat) < np.deg2rad(60)

# 1) gradient of phi = sin(lat); exact d(phi)/d(lat) = cos(lat)
phi = UxDataArray(np.sin(lat), dims=["n_face"], uxgrid=g, name="phi")
gm = phi.gradient(scale_by_radius=False)["meridional_gradient"].values
print(np.nanmedian((gm / np.cos(lat))[m]))          # 3.9952, expected 1.0

# 2) solid-body rotation u = cos(lat), v = 0; exact zeta = 2 sin(lat)
u = UxDataArray(np.cos(lat), dims=["n_face"], uxgrid=g, name="u")
v = UxDataArray(np.zeros_like(lat), dims=["n_face"], uxgrid=g, name="v")
c = u.curl(v, scale_by_radius=False).values
print(np.nanmedian((c / (2 * np.sin(lat)))[m]))     # 1.998, expected 1.0

# 3) u = 0, v = cos(lat); exact div = -2 sin(lat)
u0 = UxDataArray(np.zeros_like(lat), dims=["n_face"], uxgrid=g, name="u")
d = u0.divergence(v, scale_by_radius=False).values
print(np.nanmedian((d / (-2 * np.sin(lat)))[m]))    # 1.998, expected 1.0
```

Dividing out the 4.0 and then adding the metric term recovers 0.9996 on healpix z5 and 0.9976 on QU480 for both curl and divergence, so the two corrections have to land together — fixing the area alone drives curl to 0.5, and fixing the metric term alone would leave it 4x high. Tested on clean `main` at uxarray `2026.6.1.dev115+gf78f5ffab`.

rajeeja:
Fixes #1662.

Green-Gauss on a contour has to divide by the area that same contour encloses. `_compute_gradients_on_faces` walks the dual cell but `_normalize_and_project_gradient` divided by the primal face area, so gradients came out inflated by `A_dual/A_primal` — about 4 on quads, 3 on hexagons. Separately, `curl` and `divergence` used the planar formulas and dropped the `u*tan(lat)/a` and `-v*tan(lat)/a` metric terms, which costs a factor of two on solid-body rotation and partly cancelled the area inflation. Both corrections have to land together, since fixing the area alone drives curl to 0.5.

Verified against closed-form answers on the unit sphere, median ratio to exact over faces below 60 degrees:

| mesh | gradient | curl | divergence |
|---|---|---|---|
| healpix z4 | 0.9931 | 0.9965 | 0.9965 |
| healpix z5 | 0.9981 | 0.9991 | 0.9991 |
| healpix z6 | 0.9995 | 0.9998 | 0.9998 |
| ne30pg2 (quad) | 0.9994 | 0.9996 | 0.9996 |
| QU480 (hex) | 0.9959 | 0.9975 | 0.9975 |

All were 4.0 / 3.0 / 2.0 / 1.5 before, depending on operator and topology. The quad and hex columns now agree, which they did not previously.

Two behavior changes worth flagging for review. First, `curl` and `divergence` of a constant field are no longer zero — on a sphere they reduce to the metric term, and the new values match `tan(lat)/a` to machine zero, so I updated those two tests rather than the code. Second, a face whose contour is left open by a missing edge neighbor now returns NaN, because there is no enclosed area to divide by; the all-boundary quad-hexagon fixture is therefore all-NaN now. I checked that #1452 stays fixed — ne30pg2 is 1.2% NaN, not 100% — and added a regression guard for it.

There is a cost: gradient on healpix z7 goes from 0.408 s to 1.037 s, about 2.5x, since the dual-cell area is rebuilt per face inside the njit loop. It is not cached the way `Grid.bounds` is. Happy to add caching here or as a follow-up if you would rather not take that hit now.

Also added manufactured-solution tests with non-zero exact answers. The existing suite was all null tests and ordering comparisons, which a constant-factor error passes — that is how this shipped.

`test/core` is green, 137 passed, and pre-commit is clean.


uxarray/core/gradient.py
cmdupuis3, line 365: This entire function looked slow to me, so I had Claude come up with something, now on `cmd/gradient-dual-area`. Supposedly the method it came up with is more numerically accurate as well, but I think some of the speed benefits would still apply if want to keep the same quadrature scheme.

rajeeja: Took it, thanks — 22x on z7 and it erases the regression I flagged, so that concern is gone. You were right to call out the unsigned sum: it is wrong for non-convex cells, worst face sits at ~19% error and does not converge (0.186/0.192/0.196 at z4/z5/z6) while signed gives 0.0127/0.0088/0.0051, so I switched it and added a test bounding the tail. Your commit is on the branch with you as author.

erogluorhan (commented):
This looks good to me as long as @cmdupuis3 's concerns are addressed. Thanks @rajeeja !

cmdupuis3:
Fine with me, I just wanted a second opinion on the new code before I signed off on it

cmdupuis3 (approved):

erogluorhan (approved):
Looks good.

PR #1666, opened 2026-08-10
Preserve coords and attrs in topological aggregations

### LINKED ISSUE #1665: Topological aggregations drop coordinates and attributes
While reviewing #1588 I went digging through `aggregation.py` and came across this, which turns out to be broken on `main` already, independent of that PR.

`_node_to_face_aggregation` and `_node_to_edge_aggregation` rebuild the output from `data`/`dims`/`name` only, so all coordinates and variable attributes are dropped. The result keeps its `time` dimension but loses the `time` coordinate:

```python
uxda = ux.UxDataArray(data, dims=("time", "n_node"),
                      coords={"time": pd.date_range("2000-01-01", periods=6, freq="MS")},
                      uxgrid=grid, attrs={"units": "m"})

r = uxda.topological_mean(destination="face")
r.coords   # []   <- time gone
r.attrs    # {}   <- units gone

r.sel(time="2000-03-01")       # KeyError: 'time'
r.groupby("time.season")       # KeyError: 'time'
r.resample(time="QS")          # TypeError: Only valid with DatetimeIndex...
```

CI misses it because the existing tests use `areaTriangle`, which is 1-D with no leading coordinate, so there is nothing to lose.

This is also relevant to #1588: the dask paths added there go through `apply_ufunc`, which propagates coords and attrs for free, so on that branch the numpy and dask paths disagree and `.chunk()` starts deciding whether `.sel(time=...)` works. Fixing the numpy path on `main` makes them agree.

rajeeja:
## Overview

Fixes #1665.

Node-to-face and node-to-edge aggregations dropped every coordinate and all
variable metadata, so `.sel(time=...)`, `groupby('time.season')` and
`.resample(time=...)` raised `KeyError` on the result, and `units`/`long_name`
were lost for plotting and CF output. The fix carries over any coordinate that
does not span the reduced `n_node` dimension, plus the attrs. Coordinates along
`n_node` are still dropped, since after the reduction they would be the wrong
length.

Per @erogluorhan's review, the same coord-filtering logic was duplicated across
several other sites: both zonal means, the azimuthal mean, both rectilinear
reshape paths, the constant-latitude cross section, and the two remap backends.
Rather than add a new helper, this generalizes the existing
`_preserve_valid_coords` from `remap/structured.py` and moves it to
`uxarray/utils/coords.py` with optional `output_dims` and `exclude` arguments so
every site can share it. It has no uxarray imports, so `core`, `remap` and
`cross_sections` can all use it without import cycles.

One behavior change worth flagging: the YAC weights path keyed its coord dict on
dimension names, so it silently dropped non-dimension coordinates such as an
auxiliary time reference. That is the same bug as #1665 in a second location and
is now fixed, so `remap.yac(...)` output may carry coordinates it previously
discarded. Nothing that was present before is removed.

### DIFF
--- a/test/core/test_topological_agg.py
+++ b/test/core/test_topological_agg.py
@@ -1,5 +1,8 @@
 import uxarray as ux
 
+import numpy as np
+import numpy.testing as nt
+import pandas as pd
 import pytest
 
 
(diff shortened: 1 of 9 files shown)


cmdupuis3 (approved):
Tests pass, and the code isn't complicated and it looks fine to me

erogluorhan (commented):
This looks good to me; thanks for the fix; however, this exact pattern was already duplicated at least three times before (what I can remember from my recent work): in conservative and non-conservative zonal mean and azimuthal mean, see PR #1460 and the later PR #1471 . There might be more.

Would you be interested in coming up with a single shared helper and have `zonal_mean`--conservative &  non-conservative, `azimuthal_mean`, and topological aggregations all call it?

erogluorhan (commented):
I'd already reviewed this, leaving [this comment](https://github.com/UXARRAY/uxarray/pull/1666#pullrequestreview-4917828187).

Please let me know if it makes sense.

rajeeja:
Makes sense — done here. It was actually eight sites, not three: the three in `dataarray.py`, the aggregations, both rectilinear reshape paths, the constant-latitude cross section, and the two remap backends. I reused the existing `_preserve_valid_coords` name from `remap/structured.py` instead of inventing a new one, moved it to `uxarray/utils/coords.py`, and gave it optional `output_dims`/`exclude` args so every site fits; worth flagging that the YAC weights path was keyed on dimension names and silently dropped non-dimension coords, which is the same bug as #1665 and is now fixed too.

rajeeja:
@cmdupuis3 this grew since your approval — please check once.

cmdupuis3:
@rajeeja Looks fine to me, all tests pass too.

erogluorhan (approved):
It looks great to me! Thanks for taking care of the broader refactor!

PR #1667, opened 2026-08-11
Filtering when reading a Grid file, avoid `time` leaking into the Grid

### LINKED ISSUE #1444: `.subset` with da that has time dimension
### Version

2025.11.0

Source

### What happened?

I have a dataset that looks like

```
>>> da = du.sel(time=["2018-04-28 00:00:00", "2023-03-22 00:00:00"])[['Elevation']]
>>> print(da)

<xarray.UxDataset> Size: 1GB
Dimensions:       (met: 2, time: 2, n_face: 28495325)
Coordinates:
  * time          (time) datetime64[ns] 16B 2018-04-28 2023-03-22
  * met           (met) object 16B 'hrdps' 'capa'
    Mesh2_face_x  (n_face) float64 228MB -123.8 -123.4 -123.9 ... -116.3 -116.3
    Mesh2_face_y  (n_face) float64 228MB 46.49 46.61 46.49 ... 52.56 52.33 52.34
Dimensions without coordinates: n_face
Data variables:
    Elevation     (met, time, n_face) float64 912MB dask.array<chunksize=(1, 1, 1000000), meta=np.ndarray
```

which I want to clip to a lat lon box:

```
tmp = da.Elevation.subset.bounding_box(lon, lat)
```
but it errors as:
```
IndexError: dimension coordinate 'time' conflicts between indexed and indexing objects:
<xarray.DataArray 'time' (time: 2)> Size: 16B
array(['2018-04-28T00:00:00.000000000', '2023-03-22T00:00:00.000000000'],
      dtype='datetime64[ns]')
Coordinates:
  * time     (time) datetime64[ns] 16B 2018-04-28 2023-03-22
Attributes:
    standard_name:  time
    long_name:      Time
vs.
<xarray.Variable ()> Size: 8B
array('2016-10-01T00:00:00.000000000', dtype='datetime64[ns]')
Attributes:
    standard_name:  time
    long_name:      Time
```

which I don't totally understand. Weirdly, this code used to work, I want to say 1 or 2 releases ago. 

It to subset across all dim, time included

dylannelson:
Closes #1444 (

## Overview

1) `time` was shown to exist in a grid from a user reading in a grid and data from a single file
2) Later when the user tried to use `.subset` there was a crash related to time being different on the grid vs the data
3) `time` shouldn't realistically be on the grid to start with, so this was a concern, and was causing downstream issues not addressed when reading in the data
4) There was 2 main ideas of where to place filtering to avoid this, `Grid.__init__` or `_read_ugrid`, both were tested extensively
   - `Grid.__init__` is run in multiple scenarios, including when a file is read from disk, but
       -  in other scenarios like `_slice_from_grid` it can cause issues. 
       - This solution also wasn't properly removing the `time` because it wasn't a dimension
   - the path to `_read_ugrid` occurs primarily when a file is read from disk
      - `_read_ugrid(ds)` only solved the the issue for that file type
   - We went back to `Grid.__init__` as it solved the issue in all data types, but required a different approach
5) This has been tested in multiple ways and is shown to:
    - Scalar `time` gone from grid coords and grid variables
    - `node_lon`, `node_lat`, `face_node_connectivity` preserved
    - Grid non-empty (`n_node`, `n_face` > 0)
    - `bounding_box` runs, no error
    - Data `time` preserved
    - Subset returned faces
    - Sliced grid keeps `subgrid_face_indices`
    - `sel(time=...)` still works
    - Dimensional `time` axis + coord stripped
    - Leaked data variable stripped

Changes Made:
1) Added test for `_drop_non_grid_coords`
2) Added function `_drop_non_grid_coords` for cleaning stray coordinates
3) Called `_drop_non_grid_coords` in `Grid.__init__`

## Expected Usage
This cleaning will occur during any and all grid creations

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**

- [N/A] Docstrings have been updated with any function changes

## AI Disclosure

AI Usage:


erogluorhan (commented):
Maybe too early for a review, but I'd like to share some thoughts:

1. > `time` was shown to exist in a grid from a user reading in a grid and data from a single file

   A little more clarity could be helpful here. IIRC, `time` is not shown directly in the `Grid` object but in the `Grid._ds`.

2. That said, I'd like to know what actual purpose we had with that object attribute, i.e. `_ds`. If it was to keep complete track of the original file content through `xarray.Dataset`, I'd say this dropping at the level of `_ds` might break that. 

3. Maybe we'd want to fix it at the `Grid` itself rather than altering `_ds`, but would it be possible?

4. Finally, ugrid (i.e. `_ugrid.py`) is only one of the several formats we support. What about other formats in our I/O, e.g. _mpas.py etc. Their grid files can come in a similar way. Maybe we will need to do this fix at the Grid level rather than particular I/O modules?

dylannelson:
# #1444 — Solution history

This is a bit about each version of the idea, why it's located where it is, and what the final code looks like.

---

## Version 1 — `.values` guard in `_slice_from_grid`

```python
# uxarray/core/dataarray.py — _slice_from_grid()
n_face=sliced_grid._ds["subgrid_face_indices"].values   # was: [...["subgrid_face_indices"]]
```

- **Location:** within `_slice_from_grid` in `core/dataarray.py`.
- **Idea:** Pass the face indexer as `.values` (a plain array) so xarray does positional indexing with no coordinate alignment. It fixes the crash for every subset method, but only guards the symptom — the stray `time` stays on the grid, so it never addresses why the grid carries a `time`.

---

## Version 2 — Name-based clean build in `_read_ugrid`

```python
# uxarray/io/_ugrid.py — end of _read_ugrid()
ds = _keep_only_grid_vars(ds)   # keep only recognized grid vars/coords, drop the rest
```

- **Location:** at the end of `_read_ugrid` in `io/_ugrid.py`.
- **Idea:** Filter the parsed dataset down to recognized grid variables, dropping extras where the grid is built. This removes the stray `time` at its source (the leaky reader), so nothing downstream can collide. It's essentially the final answer — a simpler-looking alternative was proposed next, which prompted the comparison below.

---

## Version 3 — `drop_dims` in `Grid.__init__` (team's idea)

```python
# uxarray/grid/grid.py — Grid.__init__
extra_dims = set(grid_ds.dims) - set(DIM_NAMES)
grid_ds = grid_ds.drop_dims(extra_dims)
```

- **Location:** at the top of `Grid.__init__` in `grid/grid.py`.
- **Idea:** Drop any dimension not in `DIM_NAMES`, centrally for every reader, to remove stray non-grid data by axis. It captured a *dimensional* `time`, but the leaked `time` here is a scalar (0-d) coord with no dimension — so `drop_dims` finds nothing to remove and the bug survives.

---

## Version 4 — "Best of both" (`drop_dims` + 0-d/name drop)

```python
grid_ds = grid_ds.drop_dims(set(grid_ds.dims) - set(DIM_NAMES))              # dimensional extras
drop = [n for n, v in grid_ds.variables.items() if n not in KEEP and v.ndim == 0]
grid_ds = grid_ds.drop_vars(drop, errors="ignore")                          # scalar extras
```

- **Location:** at the top of `Grid.__init__` in `grid/grid.py`.
- **Idea:** Keep V3's `drop_dims` and add a second pass dropping 0-d non-grid vars, catching the scalar `time` it missed. It works, but uses two mechanisms for what one can do — V5's name-based keep-list covers scalar, dimensional, and data-var leaks in a single pass.

---

## Version 5 — Single name-based keep-list in `_read_ugrid` (final)

```python
# uxarray/io/_ugrid.py — end of _read_ugrid()
def _keep_only_grid_vars(ds):
    keep = {"grid_topology", *SPHERICAL_COORD_NAMES, *CARTESIAN_COORD_NAMES,
            *CONNECTIVITY_NAMES, *DESCRIPTOR_NAMES}
    return ds.drop_vars([n for n in ds.variables if n not in keep], errors="ignore")
```

- **Location:** at the end of `_read_ugrid` in `io/_ugrid.py` — different location (see next section).
- **Idea:** One name-based keep-list that drops every kind of non-grid extra (scalar, dimensional, data vars) in a single pass, collapsing V2–V4 into the simplest complete fix. Could be paired with the `.values` option, but didn't seem necessary.

---

## Why V5 lives in `_read_ugrid`, not `Grid.__init__`

Two code paths build a `Grid`. They **share `Grid.__init__` but not `_read_ugrid`**, so why is it in `_read_ugrid`?

**Path A — opening a grid from a file** (where the `time` leaks in):

```
ux.open_grid(...)                   core/api.py
  └─ _open_dataset_with_fallback()  # full xr.Dataset — stray `time` included
  └─ Grid.from_dataset(...)         grid.py
       └─ no source_grid_spec
       └─ _parse_grid_type()
       └─ _read_ugrid(...)          io/_ugrid.py   ← renames in place, returns EVERYTHING (leak enters here)
  └─ Grid.__init__(...)             grid.py   # self._ds = grid_ds
```

**Path B — building a *sliced* grid during a subset** (trying not to disturb this path):

```
uxda.subset.bounding_box(...)
  └─ Grid.isel(...)                    grid.py
       └─ _slice_face_indices(...)     grid/slice.py
            ds = grid._ds.isel(...)    # copy of cleaned grid
            ds["subgrid_face_indices"] = ...
       └─ Grid.from_dataset(...)       slice.py
            └─ grid_ds = ds            (_read_ugrid is skipped)
  └─ Grid.__init__(...)                grid.py   # self._ds holds subgrid_*_indices
```

| | Through `_read_ugrid`? | Through `Grid.__init__`? | `_ds` holds `subgrid_*_indices`? |
|---|:---:|:---:|:---:|
| Path A (file open) | Yes | Yes | No |
| Path B (subset) | No | Yes | Yes (needed by `_slice_from_grid`) |

- **Filter in `_read_ugrid` (V5):** Path A only. `time` is dropped as the grid is read; Path B never enters `_read_ugrid`, so its `subgrid_*_indices` survive, and its sliced grid (copied from the already-clean `grid._ds`) comes out clean for free.
- **Filter in `Grid.__init__`:** both paths. It would strip `subgrid_face_indices` from Path B's sliced grid → `_slice_from_grid` hits a `KeyError`. Fixes the leak but breaks subsetting.

---

## Final code preview

Basically came down to two edits, both in **`uxarray/io/_ugrid.py`**.

**Edit 1 — add one import at the top of the file** (`ugrid` is already imported; only `DESCRIPTOR_NAMES` is new):

```python
import numpy as np
import xarray as xr

import uxarray.conventions.ugrid as ugrid
from uxarray.constants import INT_DTYPE, INT_FILL_VALUE
from uxarray.grid.connectivity import _replace_fill_values
from uxarray.conventions.descriptors import DESCRIPTOR_NAMES  # <---- this line
```

**Edit 2 — filter at the end of `_read_ugrid`, then define the helper directly after it:**

```python
def _read_ugrid(ds):
    """Parses an unstructured grid dataset and encodes it in the UGRID
    conventions."""

    # ... (topology parse, coord/connectivity renames, dim swaps — unchanged) ...

    dim_dict[ds["face_node_connectivity"].dims[1]] = ugrid.N_MAX_FACE_NODES_DIM

    ds = ds.swap_dims(dim_dict)

    # ===== NEW =====
    # Strip non-grid extras (e.g. a stray scalar `time` coordinate, or unrelated data variables)
    ds = _keep_only_grid_vars(ds)
    # ===== end NEW =====

    return ds, dim_dict


# ===== NEW helper — place directly after _read_ugrid =====
def _keep_only_grid_vars(ds):
    """Return ``ds`` with only recognized UGRID grid variables/coordinates.

    Anything else on the dataset (a stray scalar ``time`` coordinate, unrelated
    data variables, etc.) is dropped so it cannot leak onto ``grid._ds``.

    Runs on the file-read path only.
    """
    # uxarray's own canonical grid-variable names (the same lists Grid filters against)
    keep = {"grid_topology"}
    keep.update(ugrid.SPHERICAL_COORD_NAMES)   # node/edge/face lon-lat
    keep.update(ugrid.CARTESIAN_COORD_NAMES)   # node/edge/face x-y-z
    keep.update(ugrid.CONNECTIVITY_NAMES)      # face_node_connectivity, edge_node_connectivity, ...
    keep.update(DESCRIPTOR_NAMES)              # n_nodes_per_face, face_areas, boundary_*_indices, ...

    # drop_vars removes variables/coords by name (never bare dimensions), so grid
    # dims survive with their variables; errors="ignore" tolerates absent names.
    drop = [name for name in ds.variables if name not in keep]
    return ds.drop_vars(drop, errors="ignore")
# ===== end NEW helper =====


# ===== existing code =====
def _encode_ugrid(ds):
    """Encodes an unstructured grid represented under a ``Grid`` object as a
    ``xr.Dataset`` with an updated grid topology variable."""

    if "grid_topology" in ds:
        ds = ds.drop_vars(["grid_topology"])
    # ... (rest of _encode_ugrid, unchanged) ...
```

dylannelson:
@erogluorhan 
>     4. Finally, ugrid (i.e. `_ugrid.py`) is only one of the several formats we support. What about other formats in our I/O, e.g. _mpas.py etc. Their grid files can come in a similar way. Maybe we will need to do this fix at the Grid level rather than particular I/O modules?

Sorry saw this after I was typing all the above content. 

Edit: Here's an update, more focused on # 4 to start

This is a great question and I wish I had realized sooner. Trying to find a solution that was very high level, like in `Grid.__innit__` seemed like a it made the most sense but kept causing problems downstream. Looking to something less high up made it easier, but I totally missed that it would only work for certain file types

So I had to look at 2 questions:
1) Does this leak happen for other file types?
   - If no, we are good, and it was a ugrid specific issue
   - If yes, and it's every other file, then we may have to find some function at a higher level that can handle all the leaks the same
2) Is there somewhere up, that can solve this problem by interacting with all the different file types in the same way, without disturbing other functionality
   - I've looked at this before and was struggling to find a place
   
Looking at # 1, I performed a similar check like before, read in a demo dataset, add time, save, reopen, check if there's time, and see if it crashes when there's a subset. Here's what I got

| Reader | Demo grid tested (`test/meshfiles/…`) | How grid vars are populated | Leaks? |
|---|---|---|:---:|
| MPAS | `mpas/QU/mesh.QU.1920km.151026.nc` | DataArrays derived from `in_ds`| **yes** |
| ESMF | `esmf/ne30/ne30pg3.grid.nc` | `xr.DataArray(<DataArray>, …)` | **yes** |
| ICON | `icon/R02B04/icon_grid_0010_R02B04_G.nc` | `in_ds["…"].T - 1` (DataArray) | **yes** |
| Exodus | `exodus/outCSne8/outCSne8.g` | DataArrays from the source | **yes** |
| SCRIP | `scrip/ne30pg2/grid.nc` | `xr.DataArray(<numpy>, …)` | no |
| GEOS-CS | `geos-cs/c12/test-c12.native.nc4` | `in_ds["…"].values.ravel()` | no |
| UGRID | `ugrid/quad-hexagon/grid.nc` | renames the input dataset in place | no |
| Structured / FESOM2 | `structured/outCSne30_vortex_structured.nc` (FESOM2 untested) | crashes when re-opening | **yes but worse** |

So basically from what I'm seeing right now, we have 3 different outcomes 
   - it's fine as is (good, 3 file types)
   -  it's the same problem as ugrid (bad)
   - it introduces a problem that crashes when reading (worse than what happened with ugrid)

I'll keep looking for a solution for # 2, but worried at it's impact, especially since every file type seems to have a unique way to handle the data when opening.

erogluorhan:
The table in your latest comment showing no leaks for UGRID confused me a bit (From your other words, it sounds like UGRID has the time leaking). Could you clarify?

dylannelson:
@erogluorhan 
For # 1 and # 2 and # 3, I think from what I'm seeing, `_ds` is used for a lot of functionality, and the added `time` being in there causes it to leak out into anything that can call `_ds` (which is a lot, shown below). The "Grid" and "_ds" can't be thought of as separate, because they are so heavily intertwined.

## "the `Grid`" is essentially `_ds`

Almost every `Grid` property is a thin read of `self._ds`. Straight from `uxarray/grid/grid.py` (trimmed to the returns):

```python
@property
def n_face(self):                 return self._ds.sizes["n_face"]          # :878
@property
def face_node_connectivity(self): return self._ds["face_node_connectivity"]  # :1261
@property
def attrs(self):                  return self._ds.attrs                    # :836
@property
def coordinates(self):            # only the NAMES the conventions recognize:
    return {c for c in SPHERICAL_COORD_NAMES if c in self._ds} | \
           {c for c in CARTESIAN_COORD_NAMES if c in self._ds}            # :799
@property
def dims(self):
    return {d for d in DIM_NAMES if d in self._ds.dims}                   # :787
```
Line references (pinned to `17c20f91`):
- [`n_face` — grid.py:878](https://github.com/UXARRAY/uxarray/blob/17c20f91c81859cfc37cc9edcb029096d2b83ec0/uxarray/grid/grid.py#L878)
- [`face_node_connectivity` — grid.py:1261](https://github.com/UXARRAY/uxarray/blob/17c20f91c81859cfc37cc9edcb029096d2b83ec0/uxarray/grid/grid.py#L1261)
- [`attrs` — grid.py:836](https://github.com/UXARRAY/uxarray/blob/17c20f91c81859cfc37cc9edcb029096d2b83ec0/uxarray/grid/grid.py#L836)
- [`coordinates` — grid.py:807](https://github.com/UXARRAY/uxarray/blob/17c20f91c81859cfc37cc9edcb029096d2b83ec0/uxarray/grid/grid.py#L807)
- [`dims` — grid.py:791](https://github.com/UXARRAY/uxarray/blob/17c20f91c81859cfc37cc9edcb029096d2b83ec0/uxarray/grid/grid.py#L791)



The same pattern appears to be true for others as well, like: `n_node`, `n_edge`, `node_lon`, every `*_connectivity`,
`face_areas`, `bounds`, … — all `return self._ds[...]`.

This can be a problem because `time` can leak onto other calls, like what we saw with `subset`. When the subset pulls `_ds["subgrid_face_indices"]` out as an index array, `time` rides along and collides with the data's `time`.

dylannelson:
@erogluorhan 

> The table in your latest comment showing no leaks for UGRID confused me a bit (From your other words, it sounds like UGRID has the time leaking). Could you clarify?

It's no leak for UGRID in _my_ current version, which fixed it. `main` would instead show that yes, there is a leak in UGRID

dylannelson:
and for 
> fixing at the `Grid` *without* altering `_ds`,

that's effectively what I had originally, via `.values`. (Version 1 in the big comment above)

The issue with this is that it fixes a symptom, without addressing the cause. We discussed this early on and it was turned down when we originally discussed it. It's basically a patch in the `.isel` indexer (`_slice_from_grid`) so that when the data is on the way out, we strip any extra data that may cause a clash. This works for our `subset` incident, but it won't permanently remove the leaked `time` and it won't guarantee it won't cause issues in other functions,

erogluorhan:
> and for
> 
> > fixing at the `Grid` _without_ altering `_ds`,
> 
> that's effectively what I had originally, via `.values`. (Version 1 in the big comment above)
> 
> The issue with this is that it fixes a symptom, without addressing the cause. We discussed this early on and it was turned down when we originally discussed it. It's basically a patch in the `.isel` indexer (`_slice_from_grid`) so that when the data is on the way out, we strip any extra data that may cause a clash. This works for our `subset` incident, but it won't permanently remove the leaked `time` and it won't guarantee it won't cause issues in other functions,

You're right; "without altering `_ds`" doesn't seem feasible or ideal, but I think we should discuss if the fix should be in the individual I/O modules (like _mpas etc.) or right before the [Grid object creation](https://github.com/UXARRAY/uxarray/blob/5bfb0c91314d398443f1cb78a31974fab3eba205/uxarray/grid/grid.py#L323), i.e. after the specific grid format reading. Thougts?

rajeeja:
Other readers like MPAS, ESMF, Exodus, ICON, and FESOM2 still leave `time` on `_ds` — MPAS reproduces the same `IndexError` from #1444. Could the filter move to the end of the auto-detect branch in `Grid.from_dataset`, before the `return cls(...)` at grid.py:323, so it covers all formats at once? The keep-list also drops `bounds` and `max_face_radius` when re-reading a uxarray-written grid.

dylannelson:
This new push should have everything we have wanted from the last few discussions
Early on we had the idea of moving the solution to `grid.py` in `Grid.__init__` and doing the filtering there. The filtering would look at the dimensions and remove any that didn't belong. Unfortunately this didn't work, while also introducing downstream issues. It looked like this:
```python
extra_dims = set(grid_ds.dims) - set(DIM_NAMES)
grid_ds = grid_ds.drop_dims(extra_dims)
```
The `time` in #1444 is a scalar (0-d) coordinate — it has no dimension, so it never appears in `grid_ds.dims`, `extra_dims` comes back empty, and `drop_dims` doesn't change it. I moved on from looking in `grid.py` at this point but now I'm back to it, and it seems to be working.

This solution is different in that it is dropping non-grid *coordinates* by name, which catches the scalar one that we've been having trouble with. Here's what this solution looks like instead:
```python
grid_coord_names = set(ugrid.SPHERICAL_COORD_NAMES) | set(ugrid.CARTESIAN_COORD_NAMES)
stray = [c for c in ds.coords if c not in grid_coord_names]
return ds.drop_vars(stray, errors="ignore")
```

## Testing
Before:
image
After:
image

Notebook Preview:
[18-new-fix-demo.html](https://github.com/user-attachments/files/31152082/18-new-fix-demo.html)


## Downstream impact
Performed a few downstream tests with claude code to test what affects it may have, here's the results of what it looked at:

`__init__` runs on every construction path — `from_dataset` (all formats), `from_topology`/`from_healpix`/etc., the subset's sliced grids, and `Grid.copy()`. Static sweep:

| Area | Finding | Verdict |
|---|---|:---:|
| Grid properties (`n_face`, `node_lon`, …) | read by *name*; recognized coords kept | safe |
| Reads of stray coords off `grid._ds` | `grep '_ds.coords'` / `_ds["time"]` → none | safe |
| Grid↔data linking (`open_dataset`) | by **dimensions** (`_map_dims_to_ugrid`), not coords | safe |
| Subset bookkeeping | `subgrid_face/edge/node_indices` are data vars | safe |
| `Grid.copy()` | re-runs `__init__` on a clean `_ds` | safe (idempotent) |
| Encoders / `to_xarray` | no `.coords` use | safe |
| Every `.coords` read in the package | on the data objects / shapely — never `grid._ds` | safe |
| Tests | none assert coords on `grid._ds` | safe |

test/test_subset.py
```diff
@@ -4,7 +4,9 @@
 import uxarray.grid.slice as slice_module
 from uxarray.grid.slice import _remap_dense, _remap_kernel, _remap_searchsorted
 
+import numpy as np
```
Sevans711: remove repeated numpy import

uxarray/grid/grid.py
```diff
@@ -108,6 +108,18 @@
     from uxarray.core.dataarray import UxDataArray
 
 
+def _drop_non_grid_coords(ds):
```
Sevans711: Relocate to uxarray/grid/utils.py or some other file, rather than defining this helper function directly in the main grid.py file

Sevans711 (changes requested):
This looks like a relatively clean fix overall, with lots of tests run by hand to confirm things seem to be working!

I left a few minor comments inline, and I have a couple larger comments here:
1. I had a suspicion that extra dimensions (not just extra coordinates) on input datasets would be an issue, but I tested a few things locally and found that it wasn't a problem (for some reason…). For example, if you had a time dimension on the input dataset, instead of just a time coordinate, it doesn't seem to lead to a crash later (although, it _does_ leak into `Grid._ds`). It isn't really obvious to me why it wouldn't crash…. Can you clarify why this might be the case, and/or add a test to ensure subsetting still works even after adding a time dimension to the grid?
2. Would you also be able to create a file with a time coordinate somewhere in the test suite (in meshfiles folder), and test that, too, instead of only having a test where the time coordinate gets assigned after loading the dataset? That would help me to be more convinced that this fully closes the original issue. It's a bit tricky to know for certain because the original issue report doesn't contain a complete verifiable example. But, the workflow there seemed to be "load from file, then subset (which crashes)" so ideally a test here could include that complete workflow as well.

erogluorhan:
> 1. I had a suspicion that extra dimensions (not just extra coordinates) on input datasets would be an issue, but I tested a few things locally and found that it wasn't a problem (for some reason…). For example, if you had a time dimension on the input dataset, instead of just a time coordinate, it doesn't seem to lead to a crash later (although, it _does_ leak into `Grid._ds`). It isn't really obvious to me why it wouldn't crash…. Can you clarify why this might be the case, and/or add a test to ensure subsetting still works even after adding a time dimension to the grid?

I agree with this, I think we shouldn't let extra dimensions either into the `Grid` regardless of whether they lead to crash or not.

> 2. Would you also be able to create a file with a time coordinate somewhere in the test suite (in meshfiles folder), and test that, too, instead of only having a test where the time coordinate gets assigned after loading the dataset? That would help me to be more convinced that this fully closes the original issue. It's a bit tricky to know for certain because the original issue report doesn't contain a complete verifiable example. But, the workflow there seemed to be "load from file, then subset (which crashes)" so ideally a test here could include that complete workflow as well.

I don't think we should expose a grid file with time coord/dim into our testing suite as it is uncommon to have such a presence and it would be misleading in the future.

erogluorhan (approved):
Thanks very much for this fix! @Sevans711 has already made great points inline and overall ( see (1) [here](https://github.com/UXARRAY/uxarray/pull/1667#issuecomment-5360044326), but I believe (2) should be avoided), so I am approving it to make way once they're addressed.

Sevans711:
> Thanks very much for this fix! @Sevans711 has already made great points inline and overall ( see (1) [here](https://github.com/UXARRAY/uxarray/pull/1667#issuecomment-5360044326), but I believe (2) should be avoided), so I am approving it to make way once they're addressed.

Ah, for (2), I was going based on the file at `gridpath("mpas", "QU", "oQU480.231010.nc")` which is treated as dataset data _and_ grid data during `test_open_dataset_single_combined_xarray_dataset` in `test_api.py` (related: #1479). That file already has a Time dimension in it. So, the testing suite does already include a "grid file" with time dimension, though it isn't really a simple grid file, rather it is a single file grid + data situation.

That was leading me to think it would be reasonable enough to include a second version of this file which is extremely similar but just has a scalar time coordinate instead of a Time dimension, and then call `ux.open_dataset` on it. What would your thoughts be on something like that, @erogluorhan?


Either way, I do still want to see evidence of the full original workflow being reproduced somewhere before I am ready to approve this as fully closing the original issue. Ideally this would involve an example of "load from file, then subset" crashing on `main` but not crashing on this branch. I spent a little time trying to reproduce something like that on my end, but I'm not succeeding quickly and I want to avoid duplicating work if you've already run such a test. @dylannelson my apologies if you have already shared test results like this - would you be able to point me to them if this test was already performed somewhere?

dylannelson:
@Sevans711
For 1, yeah the weirdness of dimensions vs coordinates here is what has taken me a while to understand and work around. It seems to have to do with how dimensions are used during `.isel` indexing. When it's a scalar coordinate it clashes with other existing times, whereas when its a dimension, it only attaches to variables that share a time with it, so time doesn't end up on the index at all.
- this may be why other issues related to this one claim that subsetting didn't work, maybe because it had a time but didn't share the time so it did nothing or something wrong? (**_really_** hard to tell without seeing the data) But I'm not sure, I haven't explore this case much as it was never a dimension, and we don't know if it can be naturally.
- note the time dimension will still be in the `Grid._ds`, it just won't raise conflicts. We can forcefully make it a dim, but it doesn't get carried into subsetting functionality at all.

The summary of a path deep dive through many files and functions basically came down to:

> 
> xarray attaches a coordinate to a variable only when the coordinate's dimensions are a subset of the variable's dimensions.
> 
> When it's scalar time → dims = (). And () is a subset of everything, so it glues onto the `n_face` index array.
> Dimensional time → dims = (time,). (time,) is not a subset of (n_face,), so it stays off the index array.

Which happens about here it xarray: [xarray/core/coordinates.py#L1214](https://github.com/pydata/xarray/blob/main/xarray/core/coordinates.py#L1214)

Filtering dimensions seemed to be distinct from this problem (filtering coordinates) in my mind. Like I don't think we saw anything that indicates a time _dimension_ could appear naturally like this. The downstream affects of filtering dimensions is different than coordinates too. If my understanding seems wrong, and a dimension filter here seems good, maybe we can scope this out over a call because there's lots of threads that have felt tangled while exploring this, and I don't want to overcompensate in this solution because it might introduce new problems.

Maybe of note, that uxarray also won't report any Dims that it has that aren't in DIM_NAMES - which is a static list (in grid.py):
```python
    @property
    def dims(self) -> set:
        """Names of all unstructured grid dimensions."""
        from uxarray.conventions.ugrid import DIM_NAMES

        return set([dim for dim in DIM_NAMES if dim in self._ds.dims])
```
so it shouldn't ever really report anything outside of:
```
DIM_NAMES = [
    "n_node",
    "n_edge",
    "n_face",
    "n_max_face_nodes",
    "n_max_face_edges",
    "n_max_face_faces",
    "n_max_edge_nodes",
    "n_max_edge_edges",
    "n_max_edge_faces",
    "n_max_node_faces",
    "n_max_node_edges",
    "two",
]
```
2) Was about to respond why we didn't add local files, but @erogluorhan got there first 
Basically all my local tests read a file, made a alternate version with time, saved, reopened, check if it has time, and run subset to see if it crashes. This effectively does the same because the same `Grid.__init__` should be called either way. If you're interested in seeing that we could hop on a call and I can demo. I have a few notebooks dedicated to this.

test/test_subset.py
dylannelson: No clue how I missed that twice, thanks for the catch, should be gone

uxarray/grid/grid.py
dylannelson: Got it, should be moved there now

dylannelson:
@Sevans711 I think you're looking for the notebook I made in this comment here:
[https://github.com/UXARRAY/uxarray/pull/1667#issuecomment-5319132521](https://github.com/UXARRAY/uxarray/pull/1667#issuecomment-5319132521)

There's a link in the middle to a notebook. It should be what I did and referenced above, and what I think what you're suggesting you wanted to see

rajeeja (changes requested):
This also silently drops ICON's `clon/clat/vlon/vlat/elon/elat` and FESOM2's `lon/lat` — since only 0-d coords collide, can you narrow it to `and ds[coord].ndim == 0`? That still strips the scalar `time` from #1444 but leaves every 1-d native coordinate in place, so ICON and FESOM2 grids round-trip unchanged.

Sevans711:
> I think you're looking for the notebook I made in this comment here: [#1667 (comment)](https://github.com/UXARRAY/uxarray/pull/1667#issuecomment-5319132521)
> 
> There's a link in the middle to a notebook. It should be what I did and referenced above, and what I think what you're suggesting you wanted to see

Ah, thank you! This is much closer to what I wanted. Though, I do think it isn't fully there yet (but I also might not be fully understanding). I'm hoping that the full workflow implied in https://github.com/UXARRAY/uxarray/issues/1444#issuecomment-5144980717 can be followed as closely as possible. I understand the workflow to be something like:
1. `uxds = ux.open_dataset(datafile, datafile)`   (or just `ux.open_dataset(datafile)` now that one-file `open_dataset` is supported)
2. `uxda = uxds.sel(time=[val1, val2])[varname]`
3. `uxda.subset.bounding_box(...)`

The tests you pointed to are subtly different; you manually attached the uxgrid to an entirely different Dataset, rather than getting the values for both from the same source. But, as per comment linked above, it sounds like the original issue used the same exact data for the grid and for the underlying dataset data. Are you still able to reproduce the issue when utilizing the same exact data in both cases?


I believe your clarifications to my original question (1) make sense; it sounds like the original issue could maybe only occur specifically from a time coordinate, but not from a time dimension. Is that correct? Assuming it is true, and there is separately a desire to avoid time dimension leaking into the Grid (as per https://github.com/UXARRAY/uxarray/pull/1667#issuecomment-5360044326), I would lean towards scoping the "strip extra dimensions from Grid" fix towards a separate PR to avoid too many things getting tangled here.

dylannelson:
@rajeeja Good catch, I implemented the change and everything works the same on my end so looks good. Pushed the changes.
@Sevans711 I will try to make a custom notebook for you, may take a bit

erogluorhan:
> This also silently drops ICON's `clon/clat/vlon/vlat/elon/elat` and FESOM2's `lon/lat` — since only 0-d coords collide, can you narrow it to `and ds[coord].ndim == 0`? That still strips the scalar `time` from #1444 but leaves every 1-d native coordinate in place, so ICON and FESOM2 grids round-trip unchanged.

Rajeev, this is being called after our I/O modules (e.g. `_read_icon()`) work, and I believe `grid_ds` shouldn't have those grid model-specific coords, but did you test this?

uxarray/grid/grid.py
```diff
@@ -192,8 +192,8 @@ def __init__(
         self.source_grid_spec = source_grid_spec
 
         # internal xarray dataset for storing grid variables.
-        # drop stray coordinates (e.g. a `time` carried in from the source file) so they
-        # can't leak onto the grid and collide during subsetting (see #1444).
+        # drop stray scalar coordinates (e.g. a `time` carried in from the source file)
```
erogluorhan, line 195: What if the dataset file ( that has the grid definition embedded into it) has a non-scalar time?

erogluorhan (changes requested):
As far as I could understand from the whole conversation in this PR, especially the last parts of it, maybe we got a little too focused/concerned on fixing the reported case only, i.e. scalar-only coords colliding  during `isel()` when `.subset` is called? I think the fix should be different, more comprehensive, than that, and I have a few thoughts on what that pathway should look like:

1. The main problem is when there is a non-spatial coord leakage into the `Grid._ds`, whether scalar or dimensional, we should clean it up before wrapping up the Grid object.
2. We should avoid trying to fix this at the low-level I/O modules (`_mpas`, _icon`, etc.) That is to say, the current approach (`_drop_non_grid_coords` being called in `Grid` constructor instead) makes sense to me.
3. I don't agree with dropping only scalar coords. The actual overarching fix here should rather cover dimensional coords as well. It should actually avoid the unwanted removal of any spatial coords, e.g. `vlon` etc. in ICON. That said, `_drop_non_grid_coords`'s current version focusing only on scalar coords seems incorrect to me. Also, avoiding the unwanted removal of spatial coords should easily be achievable because such coords should have the size of the number of either geometric element (nodes, faces, or edges). 

To wrap up the above thoughts, I've put together a notebook ([pr_1667_grid_filter.html](https://github.com/user-attachments/files/31292758/pr_1667_grid_filter.html)) where I test our `main` branch for: 
(1) a MPAS case with a single-file that has both grid and data in it, 
(2) Quad Hexagon grid and data files, 
(3) A single-file ICON case

All of these cases use files from our testing suite. I used the previous version of `_drop_non_grid_coords` that removed dimensional coords (i.e. not only scalar ones).  I put explanatory comments into each cell, but in summary: 

(1) has no leakage for this particular MPAS case, but since I was able to see Time leakage in the other MPAS case @dylannelson was pointing out priorly, it seems safe to solve potential problem at the Grid level as this PR currently attempts, rather than making model output-based assumptions.
(2) With the expanded Quad Hexagon case, I showed that in a single file scenario where data variables have dimensional coords but not necessarily the grid variables, we might see dimensional Time coord leaking into Grid. So, fixing only scalar coords maynot be a complete fix.
3. Spatial coords such as `vlon` etc. are also dropped after `_drop_non_grid_coords` call. Solution should be check for each coord's size, and if it is equal to the number of either faces, edges, or nodes, don't remove it.

dylannelson:
@Sevans711 
here's a custom notebook for you, with a full process and my best attempt at rebuilding everything the way the original user may have had things set up on his end. Please keep in mind, I don't have the original data, or have any clue what it contained or how it was read in, and what happened to after. I had to find a way to construct the data that could produce similar outputs, and more strangely, finding a way to read in the data from a single file that causes the times to be split in the weird way that it was. This was like trying to reproduce someone's cookie recipe, but only knowing they used eggs.

[21-original-workflow-demo.ipynb](https://github.com/user-attachments/files/31309294/21-original-workflow-demo.ipynb)

[21-original-workflow-demo.html](https://github.com/user-attachments/files/31309304/21-original-workflow-demo.html)

Sevans711:
> @Sevans711 here's a custom notebook for you, with a full process and my best attempt at rebuilding everything the way the original user may have had things set up on his end. Please keep in mind, I don't have the original data, or have any clue what it contained or how it was read in, and what happened to after. I had to find a way to construct the data that could produce similar outputs, and more strangely, finding a way to read in the data from a single file that causes the times to be split in the weird way that it was. This was like trying to reproduce someone's cookie recipe, but only knowing they used eggs.
> 
> [21-original-workflow-demo.ipynb](https://github.com/user-attachments/files/31309294/21-original-workflow-demo.ipynb)
> 
> [21-original-workflow-demo.html](https://github.com/user-attachments/files/31309304/21-original-workflow-demo.html)

Thank you for working through this! I agree it is quite difficult to reproduce without more details about the original issue. This notebook might be as close as possible to reproducing without having more information. I am still a little suspicious about these lines:
```python
ds = xr.open_dataset("demo_single_file.nc").isel(time=0)
uxds = ux.open_dataset(ds, "demo_single_file.nc")
```
To me it feels unlikely that something like "call .isel() on the grid dataset then pass that into the uxarray open_dataset call" would have happened in a user workflow. But, at this point I also can't say that that _isn't_ what happened, either. For the purposes of "does this PR close the original issue?" your efforts here have convinced me it is reasonable to move forward assuming the answer is yes. (If not, it is always possible for another issue to be posted later, after all.)

That said, what would your thoughts be on opening a new issue which this PR would close as well, something like "Grid should include only grid-related information"? I think that could be a good way to clarify why this PR is going through so many iterations, to me it is sounding like @erogluorhan is asking for an issue like that to be solved here too. Is that accurate?

I also have a minor note about this:

> Solution should be check for each coord's size, and if it is equal to the number of either faces, edges, or nodes, don't remove it.

I suspect it would probably be cleaner to check for each coord's dims instead of sizes; only keep coords which have grid dims (e.g. n_face). This could avoid spawning another issue like #1616 in the future. Is there any case where this wouldn't work out?

dylannelson:
After further discussion, the team proposed a solution that covers more cases to be certain no other cases of data leaking could occur as well, and not be specific to scalar. The approach looks like:

- Remove any data variables and dimensions with non-grid dims 
- Read through all the Data Variables and Dimensions, check their dimensions, if any are not in DIM_NAMES, drop that data variable/dimension

Basically we check to make sure everything is related to: 
``` python
['n_edge', 'n_face', 'n_max_edge_edges', 'n_max_edge_faces', 
'n_max_edge_nodes', 'n_max_face_edges', 'n_max_face_faces', 
'n_max_face_nodes', 'n_max_node_edges', 'n_max_node_faces', 
'n_node', 'two']
```
and if not, it gets removed. Testing this logic on a variety of files with grids, we get some outliers:
```bash
mpas/QU/480/grid.nc   [spec=MPAS]
   Variable   'boundary_node_indices' would be REMOVED  : non-grid dim(s) ['n_boundary_nodes']   (dims=('n_boundary_nodes',))

mpas/QU/mesh.QU.1920km.151026.nc   [spec=MPAS]
   Variable   'boundary_node_indices' would be REMOVED  : non-grid dim(s) ['n_boundary_nodes']   (dims=('n_boundary_nodes',))

mpas/QU/oQU480.231010.nc   [spec=MPAS]
   Variable   'boundary_node_indices' would be REMOVED  : non-grid dim(s) ['n_boundary_nodes']   (dims=('n_boundary_nodes',))

ugrid/fesom/fesom.mesh.diag.nc   [spec=UGRID]
   Variable   'nod_area' would be REMOVED  : non-grid dim(s) ['nz']   (dims=('nz', 'n_node'))
   Variable   'edge_node_connectivity' would be REMOVED  : non-grid dim(s) ['n2']   (dims=('n_edge', 'n2'))
   Variable   'edge_face_connectivity' would be REMOVED  : non-grid dim(s) ['n2']   (dims=('n_edge', 'n2'))
   Variable   'nod_in_elem2D' would be REMOVED  : non-grid dim(s) ['N']   (dims=('N', 'n_node'))
   Variable   'edge_cross_dxdy' would be REMOVED  : non-grid dim(s) ['n4']   (dims=('n4', 'n_edge'))
   Coordinate 'nz' would be REMOVED  : non-grid dim(s) ['nz']   (dims=('nz',))
   Coordinate 'nz1' would be REMOVED  : non-grid dim(s) ['nz1']   (dims=('nz1',))
```
Some of these look relevant, like `edge_node_connectivity` or `edge_face_connectivity` getting removed because it has dimension `n2`, but there's nothing in the [ugrid conventions](https://github.com/UXARRAY/uxarray/blob/main/uxarray/conventions/ugrid.py) that we are using that mentions n2 or any of the others.
@erogluorhan , @Sevans711 , @rajeeja 
So how do we feel about the variables/coordinates removed above? Is the function doing what it's supposed to, or should we extend the whitelist for these instances?

Sevans711:
I think we are on the same page but just to clarify, I think the steps are:

- Remove any data variables which have any non-grid dims (i.e. any dims not in DIM_NAMES).
- Remove any coordinates which have any non-grid dims.
- Remove any scalar coordinates.
- (Do not remove any scalar data variables)

Would you be able to also open a corresponding issue to track this? It might be easier to refer back to this in the future if these details get discussed on an issue focused on this topic, to make sure they don't end up getting buried inside this PR discussion (if the PR discussion continues to grow).

> Some of these look relevant, like `edge_node_connectivity` or `edge_face_connectivity` getting removed because it has dimension `n2`, but there's nothing in the [ugrid conventions](https://github.com/UXARRAY/uxarray/blob/main/uxarray/conventions/ugrid.py) that we are using that mentions n2 or any of the others. … So how do we feel about the variables/coordinates removed above? Is the function doing what it's supposed to, or should we extend the whitelist for these instances?

Thank you for checking this ahead of time, these feel like important cases to consider! Clarifying, is this function being applied directly to the data from the file, or after the data has already been parsed by the relevant uxarray io routines? For example, I would have guessed that `edge_node_connectivity` variable should have dimension "two", not "n2", after being parsed by existing io routines. I think the non-grid-dims-stripping should only occur after io routines parse the data, not directly on data from files. (If this was already from after the io routines parse the data, then I will defer to others' thoughts about these data variables, I don't really have any particular direct intuitions about them yet.)

PR #1669, opened 2026-08-11
Init AGENTS.md

### LINKED ISSUE #1668: Adding an `AGENTS.md` file

An `AGENTS.md` file helps LLMs, including Claude, work more effectively in repos, helping LLMs to consider repo conventions and assisting with basic tasks like searching and testing.

The linked draft PR should be taken as a starting point of what an `AGENTS.md` file could look like. 

Any ideas or discussions about what we'd want out of it, approaches, philosophies, and any general read-throughs, nitpicks, etc. are welcome!

cmdupuis3:
Closes #1668

## Overview

This PR adds an `AGENTS.md` markdown file to help orient LLMs, hopefully yielding more efficient and better outputs.

A `CLAUDE.md` pointer markdown is also included, pointers for other LLMs are welcome as well.

This draft should be taken as a starting point for what an `AGENTS.md` file could look like. All community input is welcome in regards to reviews, concerns, philosophies, etc.

## PR Checklist

**General**

**Testing & Benchmarking**

- [ ] Tests cover all major paths in your new functions
- [ ] If this PR could affect performance, ran ASV benchmarks and confirmed they show expected behavior (add a new benchmark if necessary)

## AI Disclosure

AI Usage: Claude Opus 5/Sonnet 5

- [ ] I take responsibility for all AI-generated content in my PR.
- [ ] I have tested all AI-generated content in my PR.


cmdupuis3:
pre-commit.ci autofix

rljacob:
Did you use ask claude or other to write this or is it by hand?

cmdupuis3:
Mostly I used Claude to generate it, I edited some parts by hand on parts of uxarray I'm more familiar with. I'm hoping others can help review it so I'm not missing any blind corners.

Sevans711:
Upon quick skimming, I noticed a few things:

1. The number of tests is wrong (it's 800+ not 546)
2. The file refers to Claude.md despite the file name being AGENTS.md
3. ```This file is a snapshot, not an oracle. Where it contradicts the repository, the repository wins, but establish that by dates, not by instinct.``` I am confused why the request is to establish that by dates. There are some things in this file which are probably currently incorrect, and I would want an LLM to always use the repository as ground truth rather than this file, I think?
4. ```JIT functions must be **module-level** and take/return plain arrays and scalars. `self`, `xr.DataArray`, and dicts cannot cross the boundary``` Is this really necessary? I would think that LLMs should be smart enough to "know" this generic statement about numba, without instructing them.
5. ```on failure raise a **typed error from `errors.py`**``` I don't think this is correct, you shouldn't always raise a custom error type. Lines above this clarified that already, this contradicts those lines.
6. I don't really know how to properly review something like this, especially since LLMs are not deterministic. Is there any way to test whether it is actually causing LLMs to produce improved results? Even if it's something you've run by-hand and not included anywhere in the automatic test suite, I think before this PR gets approved I would really like to see at least one example of "I gave an LLM the same prompt, without AGENTS.md, and with AGENTS.md, and it did a better job responding in the latter case." (To make sure order doesn't matter, also try doing in the reverse order. Maybe there is a need to disable "remember things across chats" somehow too.) This would at least provide some direct evidence that the file is plausibly providing an improvement for LLM usage. Alternatively, maybe there is some other way to show that evidence, such as existing tools which might review AGENTS.md files?

(Clarifying the following because you already requested review, which feels more official than just asking for help looking through the content) I'd be happy to properly give this a full review once you have indicated that you take responsibility for the AI generated content. I think the intention of those boxes is that an actual review (which is different from just providing suggestions or working together in thread) should not occur until you are ready to take responsibility for the content, in order to save reviewers' time.

cmdupuis3:
@Sevans711:

1 + 2) I'll fix these.

3) Well, my hope is that at some point, the `AGENTS.md` would be about as trustworthy as anything else. Since these are instructions about how to understand the repo, I think it makes sense that an up-to-date `AGENTS.md` file would take into consideration the state of the whole repo at the time it was last edited. Files that are older can't have insight to what's in the `AGENTS.md` file, and aside from documentation, aren't really supposed to help LLMs anyway. LLMs tend to trust these files more than the actual source code, so the failure mode is usually the exact opposite scenario, where the LLMs trust the older `AGENTS.md` files more than the source code.

4) It's cheaper to just tell the LLM how to deal with these situations than it is for them to figure it out for themselves at every prompt. That's part of the point of adding this file. Ideally, we'd get higher quality outputs too.

5) Mmm, yeah I'll fix that too.

6) I think that's going to be really subjective. IMO the only way to get a sense if it's working is to include it and see if it feels like the outputs are better in general. That should probably be an iterative process. 


I didn't want to check the AI responsibility boxes yet because I was hoping to get input on some of the later sections. IMO it's better to have a starting point that someone else can look at than to leave it blank and expect someone else to come along and add stuff. My intent here is for this to be a conversation for now, I don't expect to have this merge in anytime soon. If there are remaining bits nobody can vouch for, I'll delete them.

rljacob (approved):

Issue #1648, opened 2026-08-05
Optimization: don't construct many tiny numpy arrays inside numba routines


Sevans711:
Allocating lots of tiny numpy arrays is expensive compared to most other operations being performed in numba routines. Discovered during #1566, which also shows one example where this optimization caused >10x speedup: https://github.com/UXARRAY/uxarray/issues/1566#issuecomment-5185614746. (Some part of the speedup quoted there might be from fastmath=True, which should generally be avoided in uxarray, but the majority of it is due to avoiding construction of many tiny numpy arrays.)

Here are a few examples where this occurs in uxarray right now:
- `_compute_gradients_on_faces` (in gradient.py) constructs np.zeros(3) and result of np.cross for many pairs of faces. Each of those is a tiny numpy array being constructed.
- `_sort_points_by_angle` (in zonal.py) constructs np.array for every point.
- `_small_angle_of_2_vectors` (in grid/utils.py) calls np.linalg.norm, which will create a numpy array if inputs weren't a numpy array already. It doesn't have a loop internally, but it is used inside loops in other functions (see thread of issue linked above), where the inputs are just vectors with 3 components, like [ux, uy, uz], with no extra dimensions (it isn't being used in a vectorized manner).

There may be more examples throughout the code. A complete fix to this issue should attempt to find and address all of them.

**Suggested fix:**
- Avoid constructing tiny numpy arrays inside numba functions. For example, work with ux, uy, uz components directly, using `value = (ux**2 + uy**2 + uz**2)**0.5`, instead of using `value = np.linalg.norm(u)`.
- Perhaps add a few helper functions, e.g. `value = _numba_norm3(ux, uy, uz)` instead of `value = (ux**2 + uy**2 + uz**2)**0.5`. That could make the changes easier to implement (closer to existing syntax) and maintain (one source of truth for the norm function, easier to avoid easy-to-miss typos like `(ux**2 + uy**2 + ux**2)**0.5`).

Tagging @cmdupuis3 in case you want to claim this one after it gets triaged! (I'd also be happy to do it, but would certainly end up asking for you to review any related PRs.)

cmdupuis3:
Yeah, this seems to be a common failure mode here, my advice would be to file PRs at a fine grain, so it's not too overwhelming. I can definitely review if you have something specific though.

Sevans711:
[...]

PR #1626, opened 2026-07-27
Clarify in docs: UXarray supports only spherical 2D grids with convex faces

Sevans711:
Closes #1597
Closes #1615 

## Overview
Clarifies in readthedocs pages: UXarray supports only spherical 2D grids with convex faces.

As per discussion in issue 1615:
- (1) Adds the question, "What does UXarray assume about grids?" and its answer to the FAQs page
- (2) Adds the new section, "Additional Assumptions of UXarray" on the grid-formats page

As per discussion in issue 1597:
- Adds `.. note::` clarifying these assumptions in the "Why UXarray?" overview page.
- Clarifies "spherical" assumption.

Labeling this PR with "bug" in addition to "documentation", as in 1615, because uxarray currently silently handles non-convex cases incorrectly.

This PR does not update the UXarray README as originally proposed, because these details might be too specific to include in the README itself. But it should still solve the fundamental problem of lack of clarity in the docs regarding these assumptions, so it should be safe to close the underlying issues when this PR merges.

## PR Checklist

**General**
- [X] An issue is linked created and linked
- [X] Add appropriate labels
- [X] Filled out Overview and Expected Usage (if applicable) sections

**Documentation**
- [N/A] Docstrings have been added to all new functions
- [N/A] Docstrings have updated with any function changes
- [N/A] Internal functions have a preceding underscore (`_`) and have been added to `docs/internal_api/index.rst`
- [N/A] User functions have been added to `docs/user_api/index.rst`

Issue #1578, opened 2026-07-15
Visualize face_node values somehow - wedge decomposition?


Sevans711:
### Proposed new feature or change:

It may be helpful to be able to visualize "face_node" data (I'm not sure if it has an official name, but what I mean is: any data which has "one value for each node, for each face", which has dimensions n_faces, n_max_face_nodes). For example, if any angles are surprisingly small or surprisingly large during `face_node_angles` (see #1566), being able to visualize the results could help speed up debugging.

One way to do this might be to decompose into wedges. For convex faces, these could be built by connecting the midpoints of each edge to the point at the center of each face. For example, a hexagon would be decomposed into six wedges, and the value in each wedge would be the value originally at that node for that face (image adapted from claude outputs):

<img width="210" height="208" alt="Image" src="https://github.com/user-attachments/assets/87aa4c70-c44b-480a-b445-bb6e19d9702a" />

This image corresponds to wedges created from a single hexagon face, with original node values of "green" for nodes 1,2,3,4, "red" for node 5, and "yellow" for node 6.

The grid of wedges could be represented as a uxarray.Grid; each wedge is just a face in the new grid. Although, there are many more faces in the wedge grid than in the original grid, so it might be too expensive to store the result for very large grids. Even if it is limited to small grids, it could still be applied to small regions/subsets of large grids to help with debugging.

PR #1672, opened 2026-08-12
Implements `Grid.compute_face_node_angles()`

### LINKED ISSUE #1566: Compute `face_node_angles` from uxarray.Grid?

This would have the same shape as `face_node_connectivity`: for every face, for every node in that face, this tells the angle at that node.

Computing angles is necessary for computing one common measure of grid skewness. Related: #1305.

There are a few different possible ways to get the result, so I'm not sure if this should be a cached property or a function with a few optional flags:
- should the result be in radians or degrees?
- should the computation use 'euclidean' / flat geometry (sum of angles of a triangle is always 180 degrees) or respect the 'spherical' geometry (sum of angles of a triangle is always larger than 180 degrees)?

What are the best defaults for those flags?

Also, perhaps `grid.face_node_angles(geometry=...)` should dispatch to `face_node_angles_flat` or `face_node_angles_spherical` properties (whose results should be cached)?

Sevans711:
Closes #1566 

## Overview
Implements new function: `Grid.compute_face_node_angles()`, which computes the internal angles at each node, for each face in a grid. Intentionally excludes many options initially suggested in #1566, as the discussion there revealed:
- There is no need to support non-spherical geometry or non-convex faces (see also #1626).
- The implementation here is extremely fast (e.g., roughly 3 seconds for 3.75km MPAS grid, even when numba has only 4 threads), and the result's size is nontrivial (`size=n_faces*n_max_face_nodes`), so it doesn't really make sense to include a caching option right now.

Also adds corresponding tests, checking that the angle computations give correct values in a roughly 30,60,90 degree triangle and a hexagon grid. Also checks nan handling in a grid with both pentagons and hexagons (result should be nan wherever `n_max_face_nodes` dimension index is larger than `n_nodes_per_face`). For all three cases, checks that the generalized spherical excess formula holds, by comparing to `Grid.face_areas` results.

**Within scope but possibly unexpected change included this PR:** refactored `_small_angle_of_2_vectors` to improve efficiency when inputs are not numpy arrays, and rewrote docstring accordingly. (See also: #1648.) This was necessary to improve efficiency for `compute_face_node_angles()` without inlining all of the `_small_angle_of_2_vectors` logic. This is unlikely to affect efficiency elsewhere in the code, because other regions of the code seem to always construct tiny numpy arrays before calling `_small_angle_of_2_vectors`. Running ASV benchmarks here to check, just in case (but it would be acceptable if benchmarks show no improvements).

Also related (may help visualize face node angle results, if implemented): #1578

## Expected Usage

```Python
import uxarray as ux

grid = uxds.uxgrid   # or just do: grid = ux.open_grid(grid_path)

# internal angle [in radians] at each node, for each face, as an xr.DataArray
angles_rad = grid.compute_face_node_angles()

# internal angle [in degrees] at each node, for each face, as an xr.DataArray
angles_deg = grid.compute_face_node_angles(degrees=True)

# internal angle [in radians] at each node, for each face, as a ux.UxDataArray
angles_rad_uxarr = grid.compute_face_node_angles(as_uxarray=True)
```

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**

- [X] User (public) functions have been added to `docs/api.rst`

## AI Disclosure

AI Usage: discussions with Claude and ChatGPT, inline code suggestions from GitHub Copilot. Also (not in this PR, but contributed to decisions, see original issue for details) used Claude to make a numpy-only method for efficiency comparisons, and other people used Claude to iterate and run tests for more efficiency improvements.


Sevans711:
Looks like this PR is having a nice side effect: the optimization to `_small_angle_of_2_vectors` did in fact improve the `mpas_ocean.ConstructFaceLatLon.time_welzl` benchmarks. Following the code reveals that it does in fact get called there, so this is probably a real improvement, not just a rounding error.

(The chain of calls is deep, so I'm putting it here for reference: `construct_face_centers(method='welzl')` --> `_populate_face_centerpoints` --> `_construct_face_centerpoints` --> `_smallest_enclosing_circle` --> `_welzl_recursive` --> `_circle_from_two_points` or `_circle_from_three_points` --> `_small_angle_of_2_vectors`)

Ignoring the apparent peakmem improvements (see #1605).

Most importantly, none of the benchmarks show performance degradation, which helps demonstrate that the `_small_angle_of_2_vectors` are at best an improvement, at worst causing no change.

erogluorhan (approved):
This looks great to me! Thanks for all the invaluable discussion throughout this!

uxarray/grid/utils.py
```diff
@@ -12,24 +12,63 @@ def _small_angle_of_2_vectors(u, v):
...
+        v_norm_times_u[1] + u_norm_times_v[1],
+        v_norm_times_u[2] + u_norm_times_v[2],
+    )
+    norm_vec_minus = _numba_norm3(vec_minus)
+    norm_vec_sum = _numba_norm3(vec_sum)
+    angle_u_v_rad = 2 * np.arctan2(norm_vec_minus, norm_vec_sum)
     return angle_u_v_rad
 
 
+# TODO: move _numba_norm3 to a higher-level utils file. For more details, see issue #1648.
```
erogluorhan, line 47: FYI: Since you are already working on #1648 I wasn't too worried about this TODO being missed later.

erogluorhan: Feel free to resolve this once saw it.

Sevans711: Indeed! #1674 solves part of #1648, including updates to `_small_angle_of_2_vectors` (functionally equivalent to the updates here) and putting `_numba_norm3` in a more appropriate location. Ultimately my goal is to keep the version from #1674, but I didn't want to block this work on that!

Marking as resolved, as you suggested.

test/grid/geometry/test_angles.py
```diff
@@ -0,0 +1,67 @@
...
+    angles_deg = grid.compute_face_node_angles(degrees=True)
+    assert np.allclose(np.rad2deg(angles_rad), angles_deg)
+    angles_uxarr = grid.compute_face_node_angles(as_uxarray=True)
+    assert isinstance(angles_rad, xr.DataArray)
+    assert isinstance(angles_uxarr, ux.UxDataArray)
+    assert np.all(angles_rad == angles_uxarr)
+    angle_at_n0 = angles_deg.isel(n_face=0, n_max_face_nodes=0)
+    angle_at_n1 = angles_deg.isel(n_face=0, n_max_face_nodes=1)
+    angle_at_n2 = angles_deg.isel(n_face=0, n_max_face_nodes=2)
+    assert angle_at_n0 == 90  # this turns out to be exact... from arctan2(any_value, 0)
```
cmdupuis3: This should still be FP math with `np.isclose` IMO. Relying on python to provide exact integer values from atan2 is pretty sketchy to me, even if that's a standard behavior. It certainly wouldn't happen like this in the actual C/C++ backends.

test/grid/geometry/test_angles.py
```diff
@@ -0,0 +1,67 @@
...
+    assert np.abs(angle_at_n1 - 60) < 1e-2   # basically 60 degrees
+    assert np.abs(angle_at_n2 - 30) < 1e-2   # basically 30 degrees
```
cmdupuis3: These can be `np.isclose` with `60.0` and `30.0` also

test/grid/geometry/test_angles.py
```diff
@@ -0,0 +1,67 @@
...
+    spherical_excess = angles_rad.sum('n_max_face_nodes') - np.pi
+    face_areas = grid.compute_face_areas()
+    assert np.allclose(spherical_excess, face_areas, atol=0, rtol=1e-12)
+
+def test_face_node_angles_hexagons_and_pentagons():
+    """ensure face_node_angles works as expected on grids with hexagons and pentagons"""
+    grid = ux.tutorial.open_grid('quad-hexagon')  # has multiple faces, all hexagons.
+    angles_deg = grid.compute_face_node_angles(degrees=True)
+    # every hexagon in this grid is close to regular (all 120 degree angles):
+    regular_hex_deviation = angles_deg - 120
```
cmdupuis3: Probably want to have the decimal place on `120` if your intent is to do FP math, otherwise it's a bit harder to tell what's going on.

cmdupuis3 (changes requested):

test/grid/geometry/test_angles.py
Sevans711: Done!

Clarifying, atan2((value>0), 0) is actually specified in the C standard as being exactly pi (see e.g. the numpy arctan2 page: https://numpy.org/doc/stable/reference/generated/numpy.arctan2.html), so even C backends should give exactly pi there.

Though, I do agree it might not be guaranteed that "converting exactly pi into degrees must give exactly 90.0 degrees with no floating point rounding errors."

test/grid/geometry/test_angles.py
Sevans711: Done! Writing out the math feels slightly more readable to me when I only want an absolute comparison, but I don't really have a strong preference, so I'm happy to swap to `np.isclose`. (In fact, I made a typo when converting, and used rtol by accident when I meant atol… but made another commit to fix this.)

test/grid/geometry/test_angles.py
Sevans711: Done! Swapped to 120.0 because either way is fine in this case.

But, briefly clarifying here that 120 is exact, and using exact integers has the tiny benefit of allowing integer math to occur if the other values are all integers (which will never be the case here, hence why it doesn't matter here), without silently type-casting everything to floats, and no numerical downsides as far as I know.

My intent wasn't to do floating point math in particular, but it also wasn't to not do floating point math…. As of Python 3.0 all usual math operations behave the way you would expect a calculator to perform them, without needing to think about object types (including division, e.g. 7/2 == 3.5, not 3). Because of that, I don't usually think about "is this floating point math", just "are the numbers here maybe all integers? If yes, I should make sure my constants are integers too", which leads me to slightly prefer integer constants since then I don't need to consider this question at all.

test/grid/geometry/test_angles.py
cmdupuis3: What I mean is more that with FP math, 90.0 is a slight misnomer, it's not exactly 90.0, it's just very close. 90, on the other hand, is exactly 90. Even if the backend is computing hi-fi values for pi or trig funcs, the FP representation is inherently slightly off.

test/grid/geometry/test_angles.py
cmdupuis3: It could just be a Python thing, dynamic typing is increasingly odd to me. A lot of the most recent languages have very hard type checks in these cases, so there's no mystery about what the compiler's going to do.

cmdupuis3 (approved):
LGTM, tests pass and the tests and numba parallelization look reasonable. The vector arithmetic will probably need to sync up with other incoming PRs in the future though

PR #1673, opened 2026-08-12
Change substring matching to use of constants in SpatialCoordsRemapper

### LINKED ISSUE #1671: Use our canonical constants in SpatialCoordsRemapper instead of lowercase substring matching
### Version

v2026.07.0

Source

### What happened?

Nothing happened, but I just noticed a small fragility worth fixing in `SpatialCoordsRemapper`: `_get_element_type_from_dimension` does lowercase substring matching ("face" in dim_lower) to guess element type, but `uxarray/conventions/ugrid.py` already defines canonical `NODE_DIM`, `FACE_DIM`, `EDGE_DIM` constants that internal dimension names always match exactly.

Use constants instead of substring matching.

```python

```

erogluorhan:
Closes #1671 

## Overview

We already have global constants for geometric element dim names, but I realized we used lowercase substring matching in `SpatialCoordsRemapper`. This PR fixes that.

## PR Checklist

**General**

## AI Disclosure
AI Usage: N/A


uxarray/remap/spatial_coords_remap.py
Sevans711: Please update docstring to clarify that only "n_face", "n_node", "n_edge" dimensions lead to "faces", "nodes", "edges", respectively. A dimension name like "nMesh2_face" would now lead to returning None.

Sevans711 (changes requested):
Noting that this change breaks backwards compatibility for anyone who was using SpatialCoordsRemapper with nonstandard dimension names like "nMesh2_face". Can you clarify here, is this actually the intended behavior?

I noticed that `_yac_remap` seems to intentionally support dimension inputs like "face centers" (see `LABEL_TO_COORD`). On skimming, it looks like those names might all be converted to the standard dimension names like "n_face" before getting passed to `_construct_remapped_ds` (which uses `SpatialCoordsRemapper`), so it might be fine for this case. Just wanted to confirm the intention here is to drop support for nonstandard dimension names in this remapper.

uxarray/remap/spatial_coords_remap.py
erogluorhan: Done! Actually even before, we'd never get a dimension name like "nMesh2_face", as this class functions are used only internally with the `Grid` object, so docstrings should have not mentioned "nMesh2_face" in the first placed.

erogluorhan:
> Noting that this change breaks backwards compatibility for anyone who was using SpatialCoordsRemapper with nonstandard dimension names like "nMesh2_face". Can you clarify here, is this actually the intended behavior?
> 
> I noticed that `_yac_remap` seems to intentionally support dimension inputs like "face centers" (see `LABEL_TO_COORD`). On skimming, it looks like those names might all be converted to the standard dimension names like "n_face" before getting passed to `_construct_remapped_ds` (which uses `SpatialCoordsRemapper`), so it might be fine for this case. Just wanted to confirm the intention here is to drop support for nonstandard dimension names in this remapper.

That's technically correct; however, in practice, it'd be fair to expect no one to have used this outside its implicit use under remap functions, which should already input either of "n_face", "n_node", or "n_edge" only. The call trace so far should have been:

either of (`nearest_neighbor`, `bilinear`, `inverse_distance_weighted`, `yac`) -> `uxarray/remap/utils/_construct_remapped_ds()` -> `remap/spatial_coords_remap/construct_output_coords()` -> `remap/spatial_coords_remap/_get_element_type_from_dimension()`

Does that make sense?

Sevans711 (approved):
With function docstring update, and clarifications in comments that this was almost certainly only being used inside internal methods, I think it looks good and is ready to merge!

dylannelson (approved):
Straightforward and clean change. Makes more sense to handle it by checking for exact match as opposed to checking if a word is in the input. Looks good!

PR #1674, opened 2026-08-13
add numba_math.py; optimize coordinates.py and `_small_angle_of_2_vectors`

### LINKED ISSUE #1680: Avoid tiny numpy arrays in numba: create numba helper routines file
Sub-issue of #1648 to track: should create a file like numba_math.py with helper routines acting on tuples.

As mentioned in the original issue, the idea is to put helper functions here, e.g. `_numba_norm3(ux, uy, uz)` which allows `value = _numba_norm3(ux, uy, uz)` instead of `value = (ux**2 + uy**2 + uz**2)**0.5`. These helper functions will be used throughout various regions of the code while addressing #1648.

Although, after looking into it a bit, I have learned that using fixed-length tuples such as `u = (ux, uy, uz)` is just as efficient as using individual components directly; the inefficiencies come into play if you ever create a list or numpy array. So, I believe helper functions working on tuples would be better, e.g. `_numba_norm3(u)` to handle `value = (u[0]**2 + u[1]**2 + u[2]**2)**0.5`.

### LINKED ISSUE #1681: Avoid tiny numpy arrays in numba: optimize coordinates.py
Sub-issue of #1648 to track: should apply this optimization throughout the coordinates.py file: don't construct many tiny numpy arrays inside numba routines.

Sevans711:
Closes #1680 and closes #1681 (both are sub-issues of #1648).

## Overview
(For #1680) adds `uxarray/utils/numba_math.py` to provide numba routines primarily for 3-vectors. Applying these directly to tuples to avoid constructing lots of tiny numpy arrays inside numba routines can cause huge speedup in some cases.

(For #1681) Optimizes numba routines in `uxarray/grid/coordinates.py`, and the `_small_angle_of_2_vectors` function (which is called directly by some routines in coordinates.py). As per discussion in #1648, this PR does not yet attempt to make these optimizations more broadly, but instead just focuses on improving a single area of the code.

After these changes, locally I am seeing grid.construct_face_centers('welzl') time decrease from ~14 s to ~150 ms, for a grid with 157k faces. ASV benchmarks similarly show huge improvement (>=20x speedup) there.

**Sidenote**: attempted to make a cleverer / more convenient interface, such as `def _numba_add3(*vectors)` which could add multiple vectors together, instead of just `def _numba_add3(u, v)` but that leads to extremely confusing numba errors in case of any type mismatch between values in tuples. Seems like a better idea to keep this simple (yet maybe slightly less convenient) interface to avoid those confusing errors.

**Tiny expansion of PR scope**: also fixes type-hinting for `_lonlat_rad_to_xyz` and `_xyz_to_lonlat_rad`. 

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**

## AI Disclosure

AI Usage: GitHub Copilot's inline code suggestions; discussions with Claude Sonnet and ChatGPT to help understand and debug numba routines and confusing numba errors.


uxarray/grid/coordinates.py
```diff
@@ -9,13 +9,17 @@
...
+    _numba_norm3,
+)
 
 
 @njit(cache=True)
 def _lonlat_rad_to_xyz(
     lon: np.ndarray | float,
     lat: np.ndarray | float,
-) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+) -> tuple[np.ndarray | float, np.ndarray | float, np.ndarray | float]:
```
cmdupuis3: This type is too broad, we should have `tuple[np.ndarray, np.ndarray, np.ndarray] | tuple[float, float, float]` so there's no chance of mixed types.

uxarray/grid/coordinates.py
```diff
@@ -94,7 +102,7 @@ def _xyz_to_lonlat_rad(
     y: np.ndarray | float,
     z: np.ndarray | float,
     normalize: bool = True,
-) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+) -> tuple[np.ndarray | float, np.ndarray | float]:
```
cmdupuis3: Same issue here with the tuple types

cmdupuis3 (changes requested):

cmdupuis3:
Also, I think if we're going to go this route with `numba_math`, we should probably have a standard vector type for the whole repo, and then for these binops we can use operator overloads like `@overload(operator.sub)` so we don't have to explicitly call a long-named function when we just want to subtract things. Either way seems like a big change to the repo philosophy though.

cmdupuis3:
Ope, wait, there's an issue with operator overloads. If you disable the numba jit, the decorator skips and Python goes back to concatenating with `+` instead of adding.

The joys of Python...

uxarray/grid/coordinates.py
Sevans711: Good catch - fixed!

uxarray/grid/coordinates.py
Sevans711: Similarly, fixed!

Sevans711:
Thank you @cmdupuis3 for reviewing!

> Also, I think if we're going to go this route with `numba_math`, we should probably have a standard vector type for the whole repo, and then for these binops we can use operator overloads like `@overload(operator.sub)` so we don't have to explicitly call a long-named function when we just want to subtract things. Either way seems like a big change to the repo philosophy though.

> Ope, wait, there's an issue with operator overloads. If you disable the numba jit, the decorator skips and Python goes back to concatenating with `+` instead of adding.
> 
> The joys of Python...


Yeahhh, while Python objects would be happy to support a convenient interface with overloads like that, it seems to me like for performance-optimized numba it might be better to just keep the clunkier interface. Numba compilations make things harder to reason about, so keeping it simple seems to be the way to go here.

(E.g. I also originally tried `def _numba_add3(*vectors, scalar=None)` as a way to allow adding any number of 3-vectors plus maybe a scalar if you want, but this led to extremely confusing behavior, such as `_numba_add3((1,2,3),scalar=10)` crashing while `_numba_add3((1,2,3),10)` unexpectedly ran without complaint and returned (11,12,13); and `_numba_add3((1,2,3),(4,5,6),(7,8,9))` working okay but `_numba_add3((1,2,3),(4,5,6),(7,8,9.0))` crashing with an extremely confusing numba typing error)

To me it doesn't feel like a change to the whole repo philosophy, only a change to the philosophy of numba-compiled functions (wherein this directly causes huge speedup). Outside of njit-decorated functions, I believe uxarray should certainly still use numpy/dask/xarray arrays for dealing with vectors!

uxarray/grid/coordinates.py
```diff
@@ -82,6 +86,10 @@ def _xyz_to_lonlat_rad_scalar(x, y, z, normalize=True):
         lat = math.copysign(math.pi / 2, z)
         lon = 0.0
 
+    # TODO: constructing tiny numpy array inside numba function is sub-optimal,
+    #  if function gets called many times. (see issue $1648). But, as of 2026-08-13,
+    #  it looks like this function isn't being used anywhere, so maybe it should
+    #  just be removed entirely, instead of being optimized?
```
erogluorhan: Maybe you should just go ahead and remove it? @rajeeja what do you think (this seems unused)?

erogluorhan (approved):
This looks good to me, so I approve it; just a single comment below though.

cmdupuis3 (approved):

uxarray/grid/coordinates.py
Sevans711: @hongyuchen1030 mentioned remembering this was an artifact left over from an old codebase vectorization refactor, and to feel free to remove it if it is no longer being used. I am going to remove it and mark this conversation as resolved once tests pass!

PR #1684, opened 2026-08-18
`isel()`: Clean up logic, improve docs, and fix `UxDataset.isel(..., ignore_grid=True) crash`

### LINKED ISSUE #1683: `UxDataset.isel(..., ignore_grid=True)` crash, also `.isel()` logic cleanup and docs improvements
I encountered a variety of issues/suggested improvements with `.isel()` when starting work on #1641. I believe the way I will want to solve 1641 will involve offloading the grid indexing to `.isel()`, and just applying xarray's sel() to the remaining, non-grid indexers. It took me longer than expected to figure out how to get that started because the `.isel()` logic was confusing, and I also found a bug along the way.

This issue tracks the bug, as well as some possible ways to clean up the `.isel()` internal logic and improve documentation. The issue looks somewhat long but the actual corresponding changes will likely be small enough to fit reasonably into a single PR.

### Bug: `UxDataset.isel(..., ignore_grid=True)` crash:

`UxDataset.isel(..., ignore_grid=True)` crashes due to failing to provide a uxgrid object.

Example:
```python
import uxarray as ux
ds = ux.tutorial.open_dataset("outCSne30-timeseries")
ds.isel(n_face=0, ignore_grid=True)
# --> GridInvalidError: Expected a uxarray.Grid; got UxDataset.uxgrid = None. ...

ds.isel(time=0, ignore_grid=True)
# --> GridInvalidError: Expected a uxarray.Grid; got UxDataset.uxgrid = None. ...
```

Similar code does not crash from UxDataArray.isel() instead, e.g. put `ds['psi'].isel(...)` in the examples above and there is no issue.

### Cleanup: confusing logic and redundant error message handling

Right now the internal logic of `UxDataArray.isel()` and `UxDataset.isel()` looks something like this:
```python
indexers, grid_dims = _validate_indexers(...)  # grid_dims means "grid dims being indexed"
try:   # this try.. except error handling exists in UxDataArray.isel but not UxDataset.isel
    if not ignore_grid:
        if len(grid_dims) == 1:
            apply the one grid dim indexer to both the grid and the data
            apply any remaining indexers to the data
            construct result using sliced data and sliced grid  # (P0)
        else:
            construct result using sliced data and original grid   # (P1)
    else:
        construct result using sliced data and original grid  # (P2)
except ValueError as err:
    if err looks like a dimension-related error:
        raise DimensionError(str(err) + highly_redundant_addition_to_error_message) from err
    else:
        raise
```
I would suggest these improvements:
1. Restructure the if..else checks slightly to remove the need for repeating the P1 code at P2.
2. Clarify: at P1, why should we use the original, unsliced grid when `len(grid_dims)>1 and not ignore_grid`? (The answer is: it turns out the `len(grid_dims)>1` case is actually handled by crashing in `_validate_indexers`. This wasn't obvious originally. I think there should be a comment and/or assert statement to clarify.)
3. Remove the error handling logic entirely from UxDataArray.isel(), making its behavior consistent with UxDataset.isel(). Allow xarray to raise its own messages; the addition to the message here was not adding anything, and the upcasting to DimensionError was introduced extremely recently (see #1621). To clarify, the highly redundant addition was something like `". Available dimensions: ('n_face', 'time', 'lev')"` being appended to an error message which already looked like `"Dimensions {'badnonexistentdim'} do not exist. Expected one of ('n_face', 'time', 'lev')"`

### Docstring improvements

I would suggest these improvements to the docstrings of `UxDataArray.isel()` and `UxDataset.isel()` methods:
1. Improve consistency: both isel() docstrings should be more consistent with each other. E.g., use the same exact text for describing how `missing_dims="ignore"` behaves. Additionally, both isel() docstrings should be more consistent with their xarray.DataArray and xarray.Dataset counterparts, where appropriate.
2. Clarify a very cool feature which I didn't know about until digging into this: you can `.isel()` along any grid dimension, regardless of where the data is currently located. E.g., even if the data has "n_face" dim, you can `.isel(n_edge=7)` to get a result with "n_face" dim, containing only the two faces touching edge 7.
3. Further clarify the `ignore_grid` behavior and put a much clearer note/flag that using `ignore_grid=True` will cause the result's data to no longer live on the result's uxgrid; they will be desynched, if any grid dims were sliced.

Sevans711:
Closes #1683 

## Overview
Implements all fixes and suggestions from #1683.

### Bug fix:
The following lines used to crash, but now succeed:
```python
import uxarray as ux
uxds = ux.tutorial.open_dataset("outCSne30-timeseries")
uxds.isel(n_face=0, ignore_grid=True).uxgrid   # now succeeds
uxds.isel(time=0, ignore_grid=True).uxgrid   # now succeeds
```
The originally-reported bug was slightly misleading in that it claimed `uxds.isel(..., ignore_grid=True)` crashed. In fact, that method does not actually crash directly. But, it does lead to a resulting `UxDataset` without a `uxgrid`, so when it is the last line in a Jupyter notebook cell and Jupyter attempts to display it, it crashed upon attempting to access `uxgrid`. This crash no longer occurs in this PR.

Added a regression test which fails on main but succeeds after this fix.

### Cleanup of confusing logic and redundant error message handling:

With numbers here corresponding to numbers in original issue report:
1. Restructured if..else checks slightly to remove repeated "construct result using sliced data and original grid" code.
2. Clarified with comment and with assert False, in the case of `len(grid_dims)>1 and not ignore_grid` (which should be unreachable due to `_validate_indexers` crashing earlier on in that case).
3. Removed the error handling logic from UxDataArray.isel() which was converting some xarray ValueErrors into uxarray.errors.DimensionErrors, and adding highly redundant details.

### Docstring improvements
With numbers here corresponding to numbers in original issue report:
1. Improved consistency between UxDataArray.isel() and UxDataset.isel() docstrings.
2. Clarified (near top of docstrings) the special handling of grid dimensions and the ability to .isel() any grid dimension, regardless of where the data is currently located.
3. Clarified `ignore_grid=True` behavior in docstring, including the phrase: `CAUTION: using ignore_grid=True will cause the result's data to be inconsistent with its underlying grid, if any grid dimensions were sliced.`

## PR Checklist

**General**

**Testing & Benchmarking**

- [N/A] Tests cover all major paths in your new functions

**Documentation**

## AI Disclosure

AI Usage: just GitHub Copilot's inline code suggestions.

### DIFF
--- a/test/core/test_dataarray.py
+++ b/test/core/test_dataarray.py
@@ -125,14 +125,14 @@ def test_isel_invalid_dim(gridpath, datasetpath):
     uxda = UxDataArray(data, dims=["time", "n_face"], uxgrid=uxds.uxgrid)
 
     with pytest.raises(
-        DimensionError,
-        match=r"Dimensions \{'invalid_dim'\} do not exist\..*Available dimensions: \('time', 'n_face'\)",
+        ValueError,
+        match=r"Dimensions \{'invalid_dim'\} do not exist\. Expected one or more of \('time', 'n_face'\)",
     ):
         uxda.isel(invalid_dim=0)
 
     with pytest.raises(
         ValueError,
-        match=r"Dimensions \{'level'\} do not exist\..*Available dimensions: \('time', 'n_face'\)",
+        match=r"Dimensions \{'level'\} do not exist\. Expected one or more of \('time', 'n_face'\)",
     ):
         uxda.isel(level=0)
 
--- a/test/core/test_dataset.py
+++ b/test/core/test_dataset.py
@@ -129,6 +129,20 @@ def test_sel_method_forwarded(gridpath, datasetpath):
         np.array(uxds["time"].values[2], dtype="datetime64[ns]"),
     )
 
+def test_isel_ignore_grid():
+    """ensure UxDataset.isel(..., ignore_grid=True) still attaches result.uxgrid.
+    Regression test for issue #1683.
+    """
+    uxds = ux.tutorial.open_dataset("outCSne30-timeseries")
+    result = uxds.isel(time=0, ignore_grid=True)
+    result.uxgrid  # (will cause crash if uxgrid not properly attached to result)
+    assert result.uxgrid == uxds.uxgrid
+
+    result = uxds.isel(n_face=0, ignore_grid=True)
+    result.uxgrid  # (will cause crash if uxgrid not properly attached to result)
+    assert result.uxgrid == uxds.uxgrid   # ignore_grid means grid never gets sliced here
+
+
 def test_uxdataset_init_from_xarray_dataset():
     ds = xr.Dataset(
         data_vars={"a": ("x", [1, 2])},
--- a/uxarray/core/dataarray.py
+++ b/uxarray/core/dataarray.py
@@ -2062,7 +2062,7 @@ class UxDataArray(xr.DataArray):
             # no other dims, return the grid‐sliced da
             return da
         else:  # len(grid_dims)>1; _validate_indexers should have crashed.
-            assert False, "internal implementation error if reached this line"
+            raise AssertionError("internal implementation error if reached this line")
 
     @classmethod
     def from_xarray(cls, da: xr.DataArray, uxgrid: Grid, ugrid_dims: dict = None):
--- a/uxarray/core/dataarray.py
+++ b/uxarray/core/dataarray.py
@@ -1966,30 +1966,40 @@ class UxDataArray(xr.DataArray):
         inverse_indices: bool = False,
         **indexers_kwargs,
     ):
-        """
-        Return a new DataArray whose data is given by selecting indexes along the specified dimension(s).
+        """Return a new UxDataArray indexed along the specified dimension(s).
+        The data is indexed, as well as the underlying grid when applicable.
 
-        Performs xarray-style integer-location indexing along specified dimensions.
-        If a single grid dimension ('n_node', 'n_edge', or 'n_face') is provided
-        and `ignore_grid=False`, the underlying grid is sliced accordingly,
-        and remaining indexers are applied to the resulting DataArray.
+        Grid dimensions ('n_node', 'n_edge', 'n_face') are treated specially
+        when `ignore_grid=False`. Providing one of them will slice to the specified
+        nodes, edges, or faces, regardless of data location. If the data does not
+        contain the specified dimension, the result will have the minimal grid
+        region containing everything specified. For example, using n_edge=7 for data
+        on 'n_face' makes a result with 'n_face' with just the two faces on edge 7.
 
         Parameters
         ----------
         indexers : Mapping[Any, Any], optional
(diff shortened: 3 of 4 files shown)


rajeeja:
Logic and docs are a clear improvement, and matching xarray's error messages verbatim is the right call — users shouldn't get a different exception from us than from xarray for the same mistake. One nit: `assert False` vanishes under `python -O`, so raise an explicit error for that unreachable branch. cc @erogluorhan

Sevans711:
> Logic and docs are a clear improvement, and matching xarray's error messages verbatim is the right call — users shouldn't get a different exception from us than from xarray for the same mistake. One nit: `assert False` vanishes under `python -O`, so raise an explicit error for that unreachable branch.

@rajeeja Thank you for taking a look! I just updated it to `raise AssertionError(message)` syntax, which shouldn't vanish anymore under `python -O`, and follows style of other parts of uxarray which have `raise AssertionError(message)` syntax.

dylannelson:
From the issue:

> Similar code does not crash from UxDataArray.isel() instead...

Was the changes to `DataArray` necessary then?

Both `DataArray` and `DataSet` got similar treatment in the form of:
```python
if ignore_grid or len(grid_dims) == 0:
    # no grid dims, or ignore_grid=True --> just call xarray's isel
    return type(self)(
        super().isel(
            indexers=indexers or None,
            drop=drop,
            missing_dims=missing_dims,
        ),
        uxgrid=self.uxgrid,          # <-- is this doing anything on UxDataArray?
    )
```

I had claude run through the function call paths for all the scenarios for `isel` and it got something like this
```
UxDataArray  basic indexer -> uxgrid kept | hit UxDataArray._replace
UxDataArray  fancy indexer -> uxgrid kept | hit UxDataArray._replace
UxDataset    basic indexer -> uxgrid *lost* | hit UxDataset._construct_direct
UxDataset    fancy indexer -> uxgrid kept | hit UxDataset._replace, UxDataset._construct_direct
```
I think your fix solves the 3rd scenario, but the 1st and 2nd (UxDataArray) may not need it? `xr.DataArray.isel` returns via `self._replace` on both exits, running a similar line mentioned above: 
```python
if isinstance(da, UxDataArray):
   da._uxgrid = self._uxgrid
```
[_replace reference](https://github.com/UXARRAY/uxarray/blob/07e8097524965ef2168547e41c0aad7558808524/uxarray/core/dataarray.py#L148-L149)

Everything else ran fine and tests passed locally.

Sevans711:
@dylannelson Thank you for taking a look!

> Was the changes to `DataArray` necessary then?

From a pure bugfix standpoint, no, there were no bugs in `UxDataArray.isel()` which were discovered/fixed by this issue/PR.

But, I do still think the changes to `UxDataArray.isel()` here are worthwhile, because they improve readability / documentation / error messages, and consistency with the `UxDataset.isel()` implementation here. (See original issue write-up for more details.)

If you have any follow up questions or if I misunderstood your question, please let me know!

dylannelson (approved):

rajeeja (approved):
thanks for the fix.

PR #1690, opened 2026-08-19
Merge coincident pole and antimeridian nodes in structured grids

### LINKED ISSUE #1689: `Grid.from_structured` does not merge coincident pole and antimeridian nodes
- `_read_structured_grid` builds its KDTree on raw `(lon, lat)` pairs, so nodes that are the same point on the sphere are never merged.
- At the pole all `nlon+1` nodes sit at `lat=90` with distinct lons; at the seam `-180.5` and `179.5` differ by exactly 360.
- Result: `n_node` is exactly `(nlon+1)*(nlat+1)`, zero merges. 899 nodes should collapse at 1 deg, 3599 at 0.25 deg.
- The mesh is topologically torn: the pole node should have valence `nlon` (360 at 1 deg, 1440 at 0.25 deg) but reports 4.
- Only affects the native `to_rectilinear` backend; YAC uses `reg_2d_new` with a cyclic flag and never enumerates nodes.
- Fix: merge in Cartesian space inside `_read_structured_grid`.
- Related to #865, which raises the same pole/antimeridian concern for duplicate nodes generally.

rajeeja:
Closes #1689

- `_read_structured_grid` matched nodes in the lon/lat plane, so the `nlon+1` nodes at each pole and the two sides of the antimeridian were never merged.
- Now matches in Cartesian space, so nodes that are the same point on the sphere collapse: 899 nodes at 1 deg, 3599 at 0.25 deg.
- Merging leaves each polar quad with a repeated corner, so those faces are now stored as triangles padded with the fill value.
- Pole valence is `nlon` as expected (36 at 10 deg, 360 at 1 deg) and `V - E + F == 2` confirms a closed sphere.
- Regional grids that touch neither pole nor the seam are unchanged.
- Two regression tests added; the global one fails on main.

### DIFF
--- a/test/io/test_structured.py
+++ b/test/io/test_structured.py
@@ -74,3 +74,52 @@ def test_from_xarray_with_grid_from_latlon(ds_name):
     subset = uxds["air"].isel(time=0).subset.bounding_circle((-100.0, 40.0), 5)
     assert "n_face" in subset.dims
     assert subset.sizes["n_face"] > 0
+
+
+def test_global_structured_grid_merges_poles_and_seam():
+    """Nodes coincident on the sphere must be merged, even though their
+    (lon, lat) pairs differ. Regression test for issue #1689."""
+    import numpy as np
+
+    n_lon, n_lat = 36, 18
+    d_lat = 180.0 / n_lat
+    lon = np.linspace(-180, 180, n_lon, endpoint=False)
+    lat = np.linspace(-90 + d_lat / 2, 90 - d_lat / 2, n_lat)
+
+    uxgrid = ux.Grid.from_structured(lon=lon, lat=lat)
+
+    # Every duplicated pole node and antimeridian node must be gone.
+    assert uxgrid.n_node < (n_lon + 1) * (n_lat + 1)
+    assert np.isclose(uxgrid.node_lat.values, 90.0).sum() == 1
+    assert np.isclose(uxgrid.node_lat.values, -90.0).sum() == 1
+
+    # A closed sphere: V - E + F == 2.
+    assert uxgrid.n_node - uxgrid.n_edge + uxgrid.n_face == 2
+
+    # The pole is now a real singularity touching every longitude column, and
+    # its faces are triangles rather than quads with a repeated corner.
+    face_nodes = uxgrid.face_node_connectivity.values
+    n_nodes_per_face = uxgrid.n_nodes_per_face.values
+    assert (n_nodes_per_face == 3).sum() == 2 * n_lon
+
+    for face, n_nodes in zip(face_nodes, n_nodes_per_face):
+        nodes = face.tolist()[:n_nodes]
+        assert len(set(nodes)) == n_nodes
+
+    pole = int(np.flatnonzero(np.isclose(uxgrid.node_lat.values, 90.0))[0])
+    assert (face_nodes == pole).any(axis=1).sum() == n_lon
+
+
+def test_regional_structured_grid_is_unchanged():
+    """A grid that touches neither pole nor the antimeridian must keep every
+    node and stay entirely quadrilateral."""
+    import numpy as np
+
+    lon = np.linspace(-50, -10, 20)
+    lat = np.linspace(10, 40, 15)
+
+    uxgrid = ux.Grid.from_structured(lon=lon, lat=lat)
+
+    assert uxgrid.n_node == 21 * 16
+    assert uxgrid.n_face == 20 * 15
+    assert (uxgrid.n_nodes_per_face.values == 4).all()
--- a/test/io/test_structured.py
+++ b/test/io/test_structured.py
@@ -1,3 +1,4 @@
+import numpy as np
 import uxarray as ux
 import xarray as xr
 import pytest
@@ -79,8 +80,6 @@ def test_from_xarray_with_grid_from_latlon(ds_name):
 def test_global_structured_grid_merges_poles_and_seam():
     """Nodes coincident on the sphere must be merged, even though their
     (lon, lat) pairs differ. Regression test for issue #1689."""
-    import numpy as np
-
     n_lon, n_lat = 36, 18
     d_lat = 180.0 / n_lat
     lon = np.linspace(-180, 180, n_lon, endpoint=False)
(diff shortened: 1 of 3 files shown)


rajeeja:
Checked the docs and user guide for behavior changes: both datasets in `user-guide/structured.ipynb` are unaffected (`air_temperature` is regional, `ersstv5` spans 0-358 deg so the seam never closes), and neither collapses a single node. Also fixed `tol` to stay in degrees rather than silently becoming a chord length, so the documented parameter keeps its meaning. Only grids that actually touch a pole or the antimeridian change.

uxarray/io/_structured.py
```diff
@@ -79,11 +79,25 @@ def _read_structured_grid(lon, lat, tol=1e-10):
...
     nodes = np.column_stack((node_lon, node_lat))
 
+    # Match nodes on the sphere rather than in the lon/lat plane, so that the poles
+    # (many lon values, one point) and the antimeridian seam (lon differing by 360)
+    # are recognized as coincident.
+    lon_rad = np.deg2rad(node_lon)
+    lat_rad = np.deg2rad(node_lat)
+    cos_lat = np.cos(lat_rad)
+    node_xyz = np.column_stack(
+        (cos_lat * np.cos(lon_rad), cos_lat * np.sin(lon_rad), np.sin(lat_rad))
```
Sevans711: Why not re-use the helper function `_lonlat_rad_to_xyz` suggested here: https://github.com/UXARRAY/uxarray/issues/865#issuecomment-2269741156?

uxarray/io/_structured.py
```diff
@@ -21,7 +21,7 @@ def _read_structured_grid(lon, lat, tol=1e-10):
     lat : array_like
         1D array of latitude coordinates in degrees.
     tol : float, optional
-        Tolerance for considering nodes as identical (default is `1e-10`).
+        Tolerance in degrees for considering nodes as identical (default is `1e-10`).
```
Sevans711: This comment: https://github.com/UXARRAY/uxarray/issues/865#issuecomment-2269741156 makes me think that the default tolerance should be derived from `uxarray.constants.ERROR_TOLERANCE` (maybe exactly that value, maybe np.rad2deg of that value, or maybe the input tolerance should just be documented as a distance relative to unit sphere instead?)

uxarray/io/_structured.py
```diff
@@ -138,6 +152,20 @@ def _read_structured_grid(lon, lat, tol=1e-10):
     # Stack the node indices to form face_node_connectivity
     face_node_conn = np.vstack((n1, n2, n3, n4), dtype=INT_DTYPE).T
 
+    # Merging the poles leaves their quads with a repeated corner; drop it so those
+    # faces are stored as the triangles they are, padded with the fill value.
```
Sevans711: I don't really understand this. Could you clarify further, why is it necessary to add faces, when all you are doing is merging coincident nodes? A diagram would probably help significantly, even if it is an extremely rough sketch.

Sevans711 (changes requested):
Thank you for your work on this! Requesting minor changes and clarifications. I will take a closer look at the tests after understanding things a bit better.

Can you clarify, what does this mean from your original comment?
> - Now matches in Cartesian space, so nodes that are the same point on the sphere collapse: 899 nodes at 1 deg, 3599 at 0.25 deg.

I was also confused about the following, but I left an inline comment about it, feel free to respond there if that is easier:
> - Merging leaves each polar quad with a repeated corner, so those faces are now stored as triangles padded with the fill value.

erogluorhan (approved):
Wow, it is great to see this fix! It's always been unsettling to see those lons unmerged at the pole with UXarray's structured grids.

uxarray/io/_structured.py
```diff
@@ -79,11 +94,22 @@ def _read_structured_grid(lon, lat, tol=1e-10):
...
+    lat_rad = np.deg2rad(node_lat)
+    node_xyz = np.column_stack(_lonlat_rad_to_xyz(lon_rad, lat_rad))
+
     # Build KDTree
-    tree = KDTree(nodes)
+    tree = KDTree(node_xyz)
+
+    # ``tol`` is an angle in degrees; on the unit sphere the matching radius is the
+    # chord subtended by that angle, so the threshold keeps its documented meaning.
+    chord_tol = 2.0 * np.sin(np.deg2rad(tol) / 2.0)
```
Sevans711: This conversion is now redundant with the changes to default `tol` above.

test/io/test_structured.py
```diff
@@ -74,3 +74,52 @@ def test_from_xarray_with_grid_from_latlon(ds_name):
     subset = uxds["air"].isel(time=0).subset.bounding_circle((-100.0, 40.0), 5)
     assert "n_face" in subset.dims
     assert subset.sizes["n_face"] > 0
+
+
+def test_global_structured_grid_merges_poles_and_seam():
+    """Nodes coincident on the sphere must be merged, even though their
+    (lon, lat) pairs differ. Regression test for issue #1689."""
+    import numpy as np
```
Sevans711: minor but please move numpy import to top of file to match style with other testing suite files. Similar for `import numpy` call from test below.

uxarray/io/_structured.py
Sevans711: Latest changes clarified with comments inline, thank you! Marking as resolved.

uxarray/io/_structured.py
Sevans711: Latest changes swapped to deriving tol from `ERROR_TOLERANCE` so I am marking this as resolved

uxarray/io/_structured.py
Sevans711: Latest changes swapped to using that helper function so I am marking this as resolved.

Sevans711 (changes requested):
Looks almost ready to merge now, thank you for the changes! Left inline comments with minor style request and noting one typo. Other than that I think it looks good!

uxarray/io/_structured.py
rajeeja: Fixed — `tol` now defaults to `None` and uses `ERROR_TOLERANCE` as the radius directly, so the degrees round-trip is gone. Explicit `tol` still converts from degrees.

test/io/test_structured.py
rajeeja: Fixed — numpy import moved to top of file, both tests.

uxarray/io/_structured.py
Sevans711: Looks good! Confused me slightly at first when starting from this comment thread, but the docstring itself explains things very clearly now and the implementation looks good. Marking as resolved!

Sevans711 (approved):
Looks ready to merge, thank you for the implementation here and for the back and forth with addressing all of my comments!

PR #1692, opened 2026-08-19
Merge coincident nodes when constructing any Grid

### LINKED ISSUE #865: Handling Duplicate Node Indices

rajeeja:
Closes #865

- Grids with duplicate node indices were rejected outright; the dual is now built after canonicalizing those indices in the face-node connectivity.
- `geoflow-small` previously could not produce a dual at all and now yields 3840 faces; the test asserts this and fails on main.
- `_find_duplicate_nodes` is vectorized with `np.unique` instead of a per-node dict.
- `Grid.get_dual(check_duplicate_nodes=...)` is now ignored and deprecated rather than removed, so existing callers keep working.
- `UxDataArray.get_dual` and `UxDataset.get_dual` still raise `GridInvalidError`, since node-centered data cannot be remapped onto a merged node set.
- The search is skipped for formats whose faces index an already-unique vertex list (MPAS, HEALPix), where it cannot find anything to merge. Formats storing per-face corners (SCRIP, Exodus, GEOS-CS) still run it, and UGRID is deliberately not skipped since it permits either layout.
- Nodes coincident on the sphere but differing in `(lon, lat)`, such as poles and the antimeridian, are handled by #1690, now merged into main.
- A user-facing merge tolerance is tracked separately in #1770.


uxarray/grid/validation.py
```diff
@@ -77,30 +77,19 @@ def _check_area(grid):
...
-        if len(indices) > 1:
-            source_idx = indices[0]
-            for duplicate_idx in indices[1:]:
-                duplicate_dict[duplicate_idx] = source_idx
-
-    return duplicate_dict
+    """Map duplicate node indices to the first index with the same coordinates."""
+    node_coordinates = np.column_stack((grid.node_lon.values, grid.node_lat.values))
+    _, first_indices, inverse_indices = np.unique(
+        node_coordinates, axis=0, return_index=True, return_inverse=True
```
Sevans711: Is "exact equality" the correct way to go here? My intuition originally was that there should probably be some sort of tolerance here, e.g. if values agree to within 1e-12, they are probably the same node, right?

uxarray/grid/validation.py
```diff
@@ -77,30 +77,19 @@ def _check_area(grid):
...
+    )
+
+    duplicate_indices = np.flatnonzero(
+        np.arange(grid.n_node, dtype=INT_DTYPE) != first_indices[inverse_indices]
+    )
+    return {
+        INT_DTYPE(index): INT_DTYPE(first_indices[inverse_indices[index]])
```
Sevans711: I would guess that creating a dict here is extremely inefficient… maybe that doesn't matter if there are only ever a tiny number of duplicate nodes. Not necessarily blocking, but have you looked into how long this takes to run for any larger grids containing duplicate nodes? (Or, do you expect a very limited number of duplicate nodes in most cases? I would guess it is probably not worth worrying about if there are less than ~1000 duplicates or so.)

uxarray/grid/connectivity.py
```diff
@@ -299,6 +299,14 @@ def _build_face_edge_connectivity(inverse_indices, n_face, n_max_face_nodes):
     return inverse_indices
 
 
+def _remap_node_connectivity(connectivity, duplicate_node_indices):
+    """Return a copy of connectivity with duplicate node indices canonicalized."""
+    remapped_connectivity = connectivity.copy()
+    for duplicate_index, source_index in duplicate_node_indices.items():
+        remapped_connectivity[remapped_connectivity == duplicate_index] = source_index
```
Sevans711: (There is also probably some cleverer way to do this using numpy indexing without a loop through a dictionary… but maybe not necessary, see my comment related to efficiency in validation.py)

test/grid/grid/test_core.py
```diff
@@ -129,7 +129,14 @@ def test_dual_mesh_mpas(gridpath):
 
 
 def test_dual_duplicate(gridpath):
-    """Test dual mesh creation with duplicate grids."""
-    dataset = ux.open_dataset(gridpath("ugrid", "geoflow-small", "grid.nc"), gridpath("ugrid", "geoflow-small", "grid.nc"))
+    """Test dual mesh creation with duplicate node indices."""
```
Sevans711: Could you include something in this test to assert there are actually duplicate nodes in the original grid? That would help to prove this test is actually testing what it claims to be testing. Right now I just have to trust that geoflow-small grid happens to contain duplicates, but I can't see from these lines if that's really true, or how many duplicates there are.

Can you also clarify with a comment where the number 3803 comes from?

Extra helpful, but not necessarily required, would be if you are able to construct a tiny example inline here which clearly has some duplicate nodes, something small enough to directly reason through how they should be handled.

uxarray/grid/validation.py
Sevans711: In fact, looking back at the original issue thread, it looks like you were the person who originally suggested there be a tolerance in the first place! So, I would actually now assert that exact inequality is not the desired implementation, there should be a tolerance, as clarified in thread of #865.

test/grid/grid/test_core.py
```diff
@@ -129,7 +129,14 @@ def test_dual_mesh_mpas(gridpath):
 
 
 def test_dual_duplicate(gridpath):
```
Sevans711, line 136: Please add tests which also check the other grids mentioned in the original issue report: `geos-cs/c1/test-c12.native.nc4` and `esmf/ne30/ne30pg3.grid.nc`. It would be fine to add them as part of this function, unless you prefer to make new test functions for them.

Sevans711 (changes requested):
Mostly looks like a good fix! I think it still needs a little bit of extra work, and I left some inline comments accordingly.

Sevans711:
Actually, apologies for not including this during the original review, comment but one more thought: does this actually fully close the original issue? The issue writeup makes it sound to me like duplicate nodes should be handled immediately upon constructing the grid, not just during one functionality (get_dual). Is there a reason that duplicate nodes should be handled only during get_dual, instead of immediately? (Do all other current/planned functions work properly regardless of whether there are duplicate nodes?)

test/io/test_structured.py
```diff
@@ -74,3 +74,52 @@ def test_from_xarray_with_grid_from_latlon(ds_name):
     subset = uxds["air"].isel(time=0).subset.bounding_circle((-100.0, 40.0), 5)
     assert "n_face" in subset.dims
     assert subset.sizes["n_face"] > 0
+
+
+def test_global_structured_grid_merges_poles_and_seam():
+    """Nodes coincident on the sphere must be merged, even though their
+    (lon, lat) pairs differ. Regression test for issue #1689."""
+    import numpy as np
```
Sevans711: Please move numpy imports to top of file; numpy imports should always be at top of file throughout all parts of uxarray.

uxarray/grid/validation.py
```diff
@@ -76,31 +87,85 @@ def _check_area(grid):
...
+    canonical = np.arange(n_points, dtype=INT_DTYPE)
+
+    pole_mask = np.isclose(np.abs(points_xyz[:, 2]), 1.0, atol=tolerance)
+    mergeable_indices = np.flatnonzero(~pole_mask)
+
+    if len(mergeable_indices) < 2:
+        return canonical
+
+    tree = KDTree(points_xyz[mergeable_indices])
+    pairs = tree.query_pairs(r=tolerance, output_type="ndarray")
```
Sevans711, line 119: Why does this tolerance use ERROR_TOLERANCE for a radius, but other tolerances convert to chord_tol?

uxarray/io/_structured.py
```diff
@@ -79,11 +94,22 @@ def _read_structured_grid(lon, lat, tol=1e-10):
...
+    lat_rad = np.deg2rad(node_lat)
+    node_xyz = np.column_stack(_lonlat_rad_to_xyz(lon_rad, lat_rad))
+
     # Build KDTree
-    tree = KDTree(nodes)
+    tree = KDTree(node_xyz)
+
+    # ``tol`` is an angle in degrees; on the unit sphere the matching radius is the
+    # chord subtended by that angle, so the threshold keeps its documented meaning.
+    chord_tol = 2.0 * np.sin(np.deg2rad(tol) / 2.0)
```
Sevans711: This computation is redundant with the `_DEFAULT_STRUCTURED_TOL_DEG` computation above; the result will be incorrect because the formula is being applied twice.

Sevans711 (changes requested):
Suggestion: please change the title of this PR to reflect the full scope. Duplicate nodes are now handled directly whenever constructing a `Grid`, not just when constructing the dual mesh. I was confused about why `_check_duplicate_nodes_indices` checks were removed despite `construct_dual()` being unchanged. I think the reason is because all `Grid` objects are now guaranteed to not have duplicate nodes.

I tried leaving a full review but I kept getting the feeling that something weird was happening, a nagging feeling like "hey I think I've looked at this code before and left an inline comment, why am I reviewing it again?" Then I realized many of the changes here are also in #1690 which hasn't been merged yet. That led to duplicating review work and will probably lead to needing to apply fixes multiple times. I'm not sure the cleanest way forwards at this point, but I might suggest waiting to merge this until after #1690 gets merged. At least, I will want to do another close review of this PR after that PR merges, because I lost track of which things I already looked at closely and which I need to consider again.

uxarray/grid/validation.py
rajeeja: Fixed — matching goes through `_coincident_node_canonical_indices`, which is a KDTree query within `ERROR_TOLERANCE` on the sphere, not exact equality.

uxarray/grid/validation.py
rajeeja: The dict only holds duplicates, not every node, so it stays small. The real cost was `_check_duplicate_nodes_indices` looping over every face in Python; that is now one `np.isin` over the connectivity array.

uxarray/grid/connectivity.py
rajeeja: Fixed — `_remap_node_connectivity` builds a lookup array and indexes into it once, no per-duplicate array scan.

test/grid/grid/test_core.py
rajeeja: Added `assert grid.n_node == 6000` and `len(duplicates) == 2150`, and derived 3840 in the test: 3850 distinct nodes remain after the merge, ten touched by one face only, so no dual cell. Also added `test_duplicate_nodes_minimal_example` — two quads, eight nodes at six locations, asserting the map is `{6: 1, 7: 2}`.

uxarray/grid/validation.py
rajeeja: `ERROR_TOLERANCE` is already a chord distance on the unit sphere, so it is the radius directly. Conversion is only needed where the tolerance arrives in degrees, as in `_read_structured_grid`.

test/io/test_structured.py
Sevans711: Marking as resolved because this PR's diff from main (now that 1690 has merged) no longer shows any changes to test_structured.py

uxarray/grid/validation.py
Sevans711: Is there any reason to avoid using the same meaning for tolerance here as in `_read_structured_grid`? I would be less confused by it if it meant the same thing in both places. It seems like both routines are using the input tolerance in a similar way, while building a KDTree to query for everything within a certain distance of a given point.

uxarray/io/_structured.py
Sevans711: Marking as resolved because this PR's diff from main (now that 1690 has merged) no longer shows any changes to _structured.py

uxarray/grid/validation.py
Sevans711: This fix looks reasonable. One more nitpick here though: your comment on the original thread https://github.com/UXARRAY/uxarray/issues/865#issuecomment-2259306428 suggested you wanted the user to be able to specify the tolerance.

Right now, there is no way to specify the "tolerance for merging duplicate nodes" from any public-facing functions. To specify it as a larger value (e.g. 1e-4) a user would need to call `_merge_coincident_grid_ds_nodes` directly on `grid_ds` before passing to `Grid.__init__`, while there is presumably no way to properly specify it as a smaller value (e.g. 1e-12) because the `_merge_coincident_grid_ds_nodes` call inside `Grid.__init__` would always rerun the process with a tolerance of ERROR_TOLERANCE=1e-8.

Should this option be added somewhere (e.g. something like a `duplicates_tolerance` parameter in `Grid.__init__`) or is "allow the user to specify the tolerance" no longer desirable?

uxarray/grid/validation.py
Sevans711: I think that makes sense. Added the run-benchmarks label because this PR adds a call to `_merge_coincident_grid_ds_nodes` into `Grid.__init__` so it might affect uxarray performance. I believe my original comment here is resolved, but I am keeping this thread open for now as a reminder to myself to consider the benchmarking results before approving this PR.

uxarray/grid/connectivity.py
Sevans711: Looks reasonable, I'm not certain whether the new implementation is the most efficient way to do things, but it avoids the dictionary looping, is easy enough to read, and looks correct, and my guess is that it is more efficient than looping through the dictionary. No need to optimize further unless this is a real bottleneck somewhere; marking as resolved.

test/grid/grid/test_core.py
Sevans711: The new version of the test looks much clearer and the comments do a good job explaining the steps overall.

I still don't think the number 3850 is explained clearly. In particular, why should we expect len(duplicates)==2150 for this case? I am extra suspicious because the implementation changes have changed the number from 3803 to 3850.

One way forward would be to somehow separately figure out the number you would expect here. Although, I do think that the `test_duplicate_nodes_minimal_example` might be a reasonable enough test for ensuring the code is identifying duplicates correctly. If you want to avoid finding a justification for the len(duplicates)==2150 number, it would be sufficient to leave a comment there to say something like "The source file just happened to have 2150 duplicates according to uxarray, but there is not necessarily any reason to suspect it must be 2150. This test does not check for correctness of which duplicates are identified, the number is hard-coded just to prevent unexpected regressions. For a correctness test of the duplicate-identifying algorithm, see test_duplicate_nodes_minimal_example."

test/grid/grid/test_core.py
```diff
@@ -129,7 +134,170 @@ def test_dual_mesh_mpas(gridpath):
...
+def test_duplicate_nodes_minimal_example():
+    """Two quads that share an edge, but whose shared corners are stored twice.
+
+    Nodes 2 and 3 are repeated as nodes 6 and 7, so the file describes 8 nodes at
+    6 distinct locations. Node 6 must canonicalize to node 2 and node 7 to node 3,
+    leaving the second face pointing at the first face's corners.
+
+        3---2---7      lat 1   nodes 2,3 are the shared edge
+        |   |   |              nodes 7,6 are their duplicates
+        0---1---6      lat 0
```
Sevans711: I don't think this diagram corresponds to the words/code… did a robot draw this? Please fix! (Or, please clarify how it corresponds to the data, if I misunderstood….) Where are nodes 4 and 5?

Sevans711: Also, the repeated nodes in the code below are 1,2 mapping to 6,7, not 2,3 mapping to 6,7 as claimed in the test comments above.

test/grid/grid/test_core.py
```diff
@@ -129,7 +134,170 @@ def test_dual_mesh_mpas(gridpath):
...
+    """
+    node_lon = np.array([0.0, 1.0, 1.0, 0.0, 2.0, 2.0, 1.0, 1.0])
+    node_lat = np.array([0.0, 0.0, 1.0, 1.0, 0.0, 1.0, 0.0, 1.0])
```
Sevans711, line 209: The test suite is missing a (non-pole) test demonstrating that error tolerance is being handled properly. Adding a tiny number (1e-9?) to one of the duplicates here would be sufficient to spot check this. E.g. node 1 could keep node_lon = 1.0, but set node 6's node_lon to 1.0+1e-9.

uxarray/grid/connectivity.py
```diff
@@ -510,6 +510,161 @@ def _build_face_edge_connectivity(
...
+        )
+    elif "node_lon" in grid_ds and "node_lat" in grid_ds:
+        points_xyz = np.column_stack(
+            _lonlat_rad_to_xyz(
+                np.deg2rad(grid_ds["node_lon"].values),
+                np.deg2rad(grid_ds["node_lat"].values),
+            )
+        )
+    else:
+        return grid_ds
```
Sevans711, line 638: Why return grid_ds unchanged if it doesn't have x,y,z or lon,lat values? My concern is that this is a good way to silently cause incorrect results, where you expect duplicates to have been merged but can't trust that they actually were merged. Maybe crash or at least raise a warning in this case?

uxarray/grid/connectivity.py
```diff
@@ -510,6 +510,161 @@ def _build_face_edge_connectivity(
...
+    if "face_node_connectivity" in grid_ds:
+        canonical_values = np.unique(
+            np.fromiter(
+                duplicate_node_map.values(),
+                dtype=INT_DTYPE,
+                count=len(duplicate_node_map),
+            )
+        )
+        grid_ds["face_node_connectivity"] = grid_ds["face_node_connectivity"].copy(
+            data=_collapse_repeated_face_corners(
```
Sevans711, line 674: Why is something like `_collapse_repeated_face_corners` only relevant in the face_node_connectivity case? Maybe it is true but could you clarify with comment reasoning through the other connectivity combinations and justifying why it wouldn't happen for them?

Sevans711 (changes requested):
Making good progress! Still a few inline comments to resolve, see below and also I left some follow-up comments on previous inline comment threads.

Also, looking at ASV benchmark results, it looks like quite a few benchmarks have slowed down significantly, especially dual mesh construction being ~50% slower. Why would dual mesh construction slow down so much, if the search for duplicates primarily occurs when building the original grid? Tagging @cmdupuis3 to maybe comment on the benchmarks if you have additional insights. See also #1710 as a sanity check that the benchmark changes represent real changes to performance, rather than quirks of the benchmarking suite.

uxarray/grid/validation.py
rajeeja: Can be a good edition, but not in this PR. I'll open a separate issue for a user-facing tolerance.
Adding this means threading it through open_grid and the from_* constructors..

rajeeja:
> Making good progress! Still a few inline comments to resolve, see below and also I left some follow-up comments on previous inline comment threads.
> 
> Also, looking at ASV benchmark results, it looks like quite a few benchmarks have slowed down significantly, especially dual mesh construction being ~50% slower. Why would dual mesh construction slow down so much, if the search for duplicates primarily occurs when building the original grid? Tagging @cmdupuis3 to maybe comment on the benchmarks if you have additional insights. See also #1710 as a sanity check that the benchmark changes represent real changes to performance, rather than quirks of the benchmarking suite.

Thanks, the dual mesh slowdown was the duplicate check running on every get_dual call and rescanning every node each time. It's cached on the grid now.

test/grid/grid/test_core.py
rajeeja: 2150 is 2111 nodes exactly equal in lon/lat plus 39 within ERROR_TOLERANCE of an earlier node, and 10 pole nodes are held out of merging by design. The 3803 you're comparing to was the old dual.n_face assertion, not a node count; the comparable number now is 3840, and 3850 is distinct nodes.

uxarray/grid/validation.py
rajeeja: Only the input unit differs: _read_structured_grid takes tol in degrees from the user so it converts, while ERROR_TOLERANCE is already a chord on the unit sphere. Both query the KDTree with the same chord radius by default.

test/grid/grid/test_core.py
rajeeja: Redrew the diagram and fixed the node numbering in 77b5f836.

test/grid/grid/test_core.py
rajeeja: Added test_duplicate_nodes_tolerance in c968d017, node 6 at lon 1.0+1e-9.

uxarray/grid/connectivity.py
rajeeja: Warns with a RuntimeWarning now instead of returning silently, in f1e6e5c5. Kept the unchanged return so datasets with no coordinates still round-trip.

uxarray/grid/connectivity.py
rajeeja: Added the reasoning as a comment in e61e1fd9.

uxarray/grid/validation.py
rajeeja: The cache wasn't the whole story - dual mesh is still ~1.6x slower because get_dual builds a new Grid, which re-runs the coincident merge on the dual. Fixing that next, so don't go by the current benchmark numbers.

test/grid/grid/test_core.py
rajeeja: Correction: that offset was 1e-9 degrees but the cutoff is a 1e-8 chord (~5.7e-7 degrees), so the test passed far inside the boundary. Parametrized in 4d580fc1 over 1e-9 and 1e-7 (merge) and 1e-5 (must not).

uxarray/grid/connectivity.py
rajeeja: Correction: node_node_connectivity can gain a repeat too ([1, 1, 3, FILL] when a row lists both a duplicate and its canonical node). Rewrote the comment in 4d580fc1 to cover all three groups.

uxarray/grid/connectivity.py
rajeeja: Added test_merge_warns_when_node_locations_are_unknown in 4d580fc1; the branch had no coverage.

uxarray/grid/validation.py
Sevans711: (Keeping this unresolved as a reminder to open that issue. Feel free to mark as resolved as soon as you opened that issue!)

test/grid/grid/test_core.py
Sevans711: Sounds good, and seeing now that you added a comment clarifying the 2150 is hard-coded to guard against regressions rather than test correctness, and pointing to the `test_duplicate_nodes_minimal_example` for a correctness check. Marking as resolved!

uxarray/grid/validation.py
Sevans711: I still might have chosen for them to have the same meaning, but I see that you have added a comment in the docstring of `_coincident_node_canonical_indices` to clarify, which is good enough for me! Marking as resolved.

test/grid/grid/test_core.py
Sevans711: Looks clear now, I double-checked the diagram and words/code now match. Marking as resolved!

test/grid/grid/test_core.py
Sevans711: The new test looks good! Marking as resolved.

uxarray/grid/connectivity.py
Sevans711: Warning and test coverage looks good here now. Marking as resolved!

uxarray/grid/connectivity.py
Sevans711: The comment clarifies and makes the logic much easier for me to follow now, thank you! Marking as resolved.

uxarray/grid/validation.py
Sevans711: That fix definitely helped, but it still looks like >40% slowdown in dual_mesh construction. Any insights?

Though, now that the performance improvements are expanding significantly beyond the original lines I flagged here, I think it might make more sense to continue the benchmarking discussion in the main thread below instead. Feel free to mark this as resolved once you read it, I will leave a comment soon during re-review.

uxarray/grid/connectivity.py
```diff
@@ -510,6 +510,192 @@ def _build_face_edge_connectivity(
...
+        return grid_ds
+
+    n_node = points_xyz.shape[0]
+    canonical = _coincident_node_canonical_indices(points_xyz, tolerance)
+    duplicate_node_map = {
+        INT_DTYPE(index): INT_DTYPE(canonical[index])
+        for index in np.flatnonzero(canonical != np.arange(n_node, dtype=INT_DTYPE))
+    }
+    if not duplicate_node_map:
+        return grid_ds
```
Sevans711, line 679: Minor nit: could you indent the subsequent lines to make it clear that the logic here looks like this:
```
if no duplicates exist:
    no need to handle any duplicates
else:
    handle duplicates
```
Alternatively, could swap the order if you want:
```
if not duplicate_node_map:
    handle duplicates
else:
    return grid_ds
```

With the current code, I had to scroll around a bit to find this line and understand that is what is happening.

Alternatively, if you added a small comment in this function's docstring to emphasize "if there are no duplicates, returns grid_ds, unchanged", that would be sufficient. (Or, could add a small comment below, or above when defining `_NODE_INDEX_CONNECTIVITY_TO_REMAP` and `_DERIVED_CONNECTIVITY_TO_INVALIDATE`, to clarify that these things only get applied if actually detected any duplicates.)

Sevans711 (changes requested):
One minor inline comment. Aside from that, my only remaining concern is about performance.

> Thanks, the dual mesh slowdown was the duplicate check running on every get_dual call and rescanning every node each time. It's cached on the grid now.

This seems to have helped, but the latest benchmarks still shows a >40% slowdown. Additionally, the slowdown is worse for the larger grid (120km resolution reported 45% slowdown, 480km resolution reported 34% slowdown), which suggests the slowdown might be much worse for larger, production-scale grids.

This led me to trying a larger grid as a sanity check. I have a 15km MPAS grid on my local machine (it's the dyamond 15km grid mentioned in the benchmarks suite) so I used that. Running `ux.open_grid()` a few times, I'm seeing it takes roughly 600 ms on `main`, and roughly 3.6 s on this branch. That's a 6x slowdown! Considering the possibility for nonlinear scaling, and that people often use even better resolution than 15km (e.g. 7.5km or 3.75km), this may be a significant performance hit for workflows with large grids.

Any ideas for how to improve this? I suspect that merging a 6x slowdown into open_grid() might not be worth always enforcing this extra validation, as many grids do not have any duplicates at all. (If there's really no way to improve the performance, maybe this check during `Grid.__init__` should be disabled by default, with a flag to enable it?)

cc: @cmdupuis3

uxarray/grid/connectivity.py
cmdupuis3: To add, is there a way to say that some file formats will be guaranteed to not have duplicate nodes? If we know when we can skip this completely, that would be a nice thing to have.

Issue #1617, opened 2026-07-23
Make error messages clearer and more informative


Sevans711:
### Proposed new feature or change:

Error messages throughout uxarray could be made clearer and more informative.

For example:
```python
    if not uxda._node_centered():
        raise ValueError(
            f"Data Variable must be mapped to the corner nodes of each face, with dimension "
            f"{uxda.uxgrid.n_face}."
        )
```
could be improved to something like:
```python
    if not uxda._node_centered():
        raise ValueError(
            f"Expected node-centered data, got {uxda.data_location}, for node-to-edge aggregation. "
            f"(Face-centered data has n_face dimension of size {uxda.uxgrid.n_face})."
        )
```
The latter has the following benefits:
- uses more common terminology ("node-centered" instead of the current version's "mapped to the corner nodes of each face")
- gives relevant information about the actual state of objects which caused the failure (uxda.data_location indicates "face_centered" or "edge_centered", as opposed to the current version which does not say where the data is actually centered)
- briefly clarifies what was being attempted when the error occurred ("node-to-edge aggregation")
- clarifies what the `uxda.uxgrid.n_face` number means.
  - (The original message was quite confusing to me. For example, I don't really know what to think about the "4730" when just seeing "Data Variable must be mapped to the corner nodes of each face, with dimension 4730." And, if I had to guess, I would guess "ah, 4730 must be the expected size of the data variable, i.e. the number of nodes" but that is actually incorrect, it is the expected size of the result after mapping from nodes to faces.)

This is merely one example. There are many error messages throughout uxarray which could be improved. Fully addressing this issue would involve looking through all uxarray error messages (there are over 250 of them), which may be infeasible to do or review all at once. Perhaps one reasonable way to address this issue could be over the span of multiple PRs, each focused on a single section of the code, with PRs like "Make uxarray/grid/ error messages clearer and more informative" focused on the uxarray.grid subpackage, and similar PRs for "core", "plot", etc.

This is complementary to #1556, which proposes to swap to using custom error types, but should avoid editing error messages directly.

PR #1694, opened 2026-08-19
Raise the underlying cause when geospatial data cannot be read

### LINKED ISSUE #1693: `_read_geodataframe` raises `UnboundLocalError` when the file cannot be read
- `_read_geodataframe` catches every read failure, prints it, and falls through, so `gdf` is never bound and the next line raises `UnboundLocalError` instead of the real cause.
- Reproduced by reading any shapefile or GeoJSON without `pyogrio`/`fiona` installed; the actual message ("requires the 'pyogrio' or 'fiona' package") only reaches stdout.
- Applies to any read failure: bad path, corrupt file, or missing driver all produce the same misleading error.
- Fixed in #1694 by removing the try/except entirely, letting geopandas raise its own exception directly.

rajeeja:
Closes #1693

- `_gpd_read` caught every read failure, printed it, and fell through, leaving `gdf` unbound so the next line raised `UnboundLocalError` instead of the real cause; dropped the try/except entirely so geopandas raises directly (also surfaces `FileNotFoundError` for free).
- `geopandas` was unpinned and pre-1.0 releases don't require `pyogrio`, so an install could end up with no file-IO backend at all; pinned `geopandas>=1.0`.
- `_open_dataset_with_fallback` now chains the fallback engine's error onto the default engine's, so a file neither engine can open reports both reasons.
- FESOM2 ASCII parsers raised `FileNotFoundError("TODO: ")`; now name the missing file/directory.
- Assuming WGS84 for CRS-less data and skipping an unsupported geometry type are now `warnings.warn` instead of `print`; the latter previously produced a grid silently missing a face.
- Removed `_is_structured`'s speculative stdout diagnostics, which fired on every valid MPAS/Exodus/SCRIP file.

### DIFF
--- a/test/io/test_fesom.py
+++ b/test/io/test_fesom.py
@@ -4,6 +4,7 @@ import numpy.testing as nt
 import pytest
 
 import uxarray as ux
+from uxarray.io._fesom2 import _parse_elem2d, _parse_nod2d
 
 
 
@@ -66,15 +67,11 @@ def test_open_mfdataset_pi_path(test_data_dir):
 
 def test_parse_nod2d_missing_file_names_the_file(tmp_path):
     """A missing 'nod2d.out' must say which file is missing and where."""
-    from uxarray.io._fesom2 import _parse_nod2d
-
     with pytest.raises(FileNotFoundError, match="nod2d.out"):
         _parse_nod2d(str(tmp_path))
 
 
 def test_parse_elem2d_missing_file_names_the_file(tmp_path):
     """A missing 'elem2d.out' must say which file is missing and where."""
-    from uxarray.io._fesom2 import _parse_elem2d
-
     with pytest.raises(FileNotFoundError, match="elem2d.out"):
         _parse_elem2d(str(tmp_path))
--- a/test/io/test_fesom.py
+++ b/test/io/test_fesom.py
@@ -62,3 +62,19 @@ def test_open_mfdataset_pi_path(test_data_dir):
     assert "n_node" in uxds.dims
     assert "n_face" in uxds.dims
     assert len(uxds) == 3
+
+
+def test_parse_nod2d_missing_file_names_the_file(tmp_path):
+    """A missing 'nod2d.out' must say which file is missing and where."""
+    from uxarray.io._fesom2 import _parse_nod2d
+
+    with pytest.raises(FileNotFoundError, match="nod2d.out"):
+        _parse_nod2d(str(tmp_path))
+
+
+def test_parse_elem2d_missing_file_names_the_file(tmp_path):
+    """A missing 'elem2d.out' must say which file is missing and where."""
+    from uxarray.io._fesom2 import _parse_elem2d
+
+    with pytest.raises(FileNotFoundError, match="elem2d.out"):
+        _parse_elem2d(str(tmp_path))
--- a/test/io/test_geopandas.py
+++ b/test/io/test_geopandas.py
@@ -43,15 +43,23 @@ def test_load_xarray_with_from_file(gridpath):
 
 
 def test_read_failure_raises(tmp_path):
-    """A read failure must surface its cause, not an UnboundLocalError.
-    Regression test for issue #1693."""
+    """A read failure must surface the backend's own error rather than being
+    printed and swallowed into an UnboundLocalError.
+
+    Regression test for issue #1693. Requires a geopandas file-IO backend,
+    otherwise read_file raises ImportError and the test would pass for the
+    wrong reason.
+    """
     import pytest
 
-    from uxarray.errors import GridInvalidError
+    pytest.importorskip("pyogrio")
+
     from uxarray.io._geopandas import _gpd_read
 
     not_geospatial = tmp_path / "not_geospatial.shp"
     not_geospatial.write_text("this is not a shapefile")
 
-    with pytest.raises(GridInvalidError, match="Could not read"):
+    with pytest.raises(Exception) as excinfo:
         _gpd_read(str(not_geospatial))
+
+    assert not isinstance(excinfo.value, UnboundLocalError)
--- a/test/io/test_geopandas.py
+++ b/test/io/test_geopandas.py
@@ -4,6 +4,7 @@ import pytest
 from shapely.geometry import Point, Polygon
 
 import uxarray as ux
+from uxarray.io._geopandas import _extract_geometry_info, _gpd_read, _set_crs
 
 def test_read_shpfile(test_data_dir):
     """Read a shapefile."""
@@ -52,8 +53,6 @@ def test_read_failure_raises(tmp_path):
 
     Regression test for issue #1693.
     """
-    from uxarray.io._geopandas import _gpd_read
-
     not_geospatial = tmp_path / "not_geospatial.shp"
     not_geospatial.write_text("this is not a shapefile")
 
@@ -65,8 +64,6 @@ def test_read_failure_raises(tmp_path):
 
 def test_set_crs_warns_when_crs_is_missing():
     """Assuming WGS84 for CRS-less data is a guess and must be announced."""
-    from uxarray.io._geopandas import _set_crs
-
     gdf = gpd.GeoDataFrame(
         geometry=[Polygon([(0, 0), (1, 0), (1, 1)])], crs=None
     )
@@ -80,8 +77,6 @@ def test_set_crs_warns_when_crs_is_missing():
 def test_unsupported_geometry_is_reported():
     """Dropping a geometry silently would yield a grid missing a face with no
     indication that anything was skipped."""
-    from uxarray.io._geopandas import _extract_geometry_info
-
     gdf = gpd.GeoDataFrame(
         geometry=[Polygon([(0, 0), (1, 0), (1, 1), (0, 0)]), Point(5, 5)],
         crs="EPSG:4326",
--- a/test/io/test_geopandas.py
+++ b/test/io/test_geopandas.py
@@ -40,3 +40,18 @@ def test_load_xarray_with_from_file(gridpath):
     nc_filename = gridpath("scrip", "outCSne8", "outCSne8.nc")
     uxgrid = ux.Grid.from_file(nc_filename, backend="xarray")
     uxgrid.validate()
+
+
+def test_read_failure_raises(tmp_path):
+    """A read failure must surface its cause, not an UnboundLocalError.
+    Regression test for issue #1693."""
+    import pytest
+
+    from uxarray.errors import GridInvalidError
+    from uxarray.io._geopandas import _gpd_read
+
+    not_geospatial = tmp_path / "not_geospatial.shp"
+    not_geospatial.write_text("this is not a shapefile")
+
+    with pytest.raises(GridInvalidError, match="Could not read"):
+        _gpd_read(str(not_geospatial))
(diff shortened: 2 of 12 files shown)


Sevans711 (changes requested):
Good find, I definitely agree with this premise of needing better error handling here; the version on main is not using best practices for error handling in a few different ways: (1) try multiple things at once; (2) catch error and print but don't re-raise, and (3) then continue with the code which will crash with confusing UnboundLocalError if an error occurred previously.

I am not 100% certain that GridInvalidError is the way to go here. What if the file simply doesn't exist, for example? Then it should be a FileNotFoundError. I think at least this case needs to be handled separately. Perhaps changing "except Exception" to "except ValueError" would be a cleaner way to do this. I tend to be skeptical of "except Exception" unless there is a clear reason for wanting to catch every error type.

More on that point, though: what is the motivation for catching errors here at all? Why not just allow gpd.read_file(...) to fail directly with its own error message?

rajeeja:
I hit this by accident — testing something unrelated and saw four failures in `test_geopandas.py`. Looked like a broken env at first, but `geopandas` was unpinned in `pyproject.toml`, and pre-1.0 geopandas doesn't require `pyogrio`, so my install landed on geopandas 0.9 with no file-IO backend at all — neither `pyogrio` nor `fiona` is mentioned anywhere in the repo or `ci/environment.yml`. Real message was `the 'read_file' function requires the 'pyogrio' or 'fiona' package, but neither is installed`; the old code printed it and carried on, which is how it became an `UnboundLocalError`.

You're right that we shouldn't be catching here at all — none of the other readers do; `_esmf`, `_scrip`, and `io/utils` only raise `GridInvalidError` once the data is parsed and actually wrong. Dropped the try/except and let geopandas raise, which also gets `FileNotFoundError` for free. `except ValueError` wouldn't have helped either way, since the real failures are `ImportError` or pyogrio's `DataSourceError` / fiona's `DriverError` — moot now that the catch is gone.

Pinned `geopandas>=1.0` and fixed my test too — it was passing on the `ImportError` rather than a parse failure, green for the wrong reason.

test/io/test_fesom.py
```diff
@@ -62,3 +62,19 @@ def test_open_mfdataset_pi_path(test_data_dir):
     assert "n_node" in uxds.dims
     assert "n_face" in uxds.dims
     assert len(uxds) == 3
+
+
+def test_parse_nod2d_missing_file_names_the_file(tmp_path):
+    """A missing 'nod2d.out' must say which file is missing and where."""
+    from uxarray.io._fesom2 import _parse_nod2d
```
Sevans711: I would have a slight preference for imports like this to be at top of file, rather than within each test, for the test suite, especially for a small test suite file like this! I'm not sure what the benefit is of putting these imports directly inside the test functions.

test/io/test_geopandas.py
```diff
@@ -40,3 +40,65 @@ def test_load_xarray_with_from_file(gridpath):
...
+
+def test_read_failure_raises(tmp_path):
+    """A read failure must surface the backend's own error rather than being
+    printed and swallowed into an UnboundLocalError.
+
+    Regression test for issue #1693. Requires a geopandas file-IO backend,
+    otherwise read_file raises ImportError and the test would pass for the
+    wrong reason.
+    """
+    import pytest
```
Sevans711: pytest import should definitely go at top of file; that would match the style of all other pytest imports throughout uxarray test suite.

test/io/test_fesom.py
Sevans711: (This applies to a variety of imports throughout your new test functions, but I'm only leaving this one comment rather than making a new comment thread for each one.)

test/io/test_geopandas.py
```diff
@@ -40,3 +40,65 @@ def test_load_xarray_with_from_file(gridpath):
...
+        _gpd_read(str(not_geospatial))
+
+    assert not isinstance(excinfo.value, UnboundLocalError)
+
+
+def test_set_crs_warns_when_crs_is_missing():
+    """Assuming WGS84 for CRS-less data is a guess and must be announced."""
+    import pytest
+    gpd = pytest.importorskip("geopandas")
+    from shapely.geometry import Polygon
```
Sevans711: shapely import should probably go at top of file for pytest test suite, but isn't as necessary to move it. It would avoid the need to re-import in multiple tests at least.

test/io/test_geopandas.py
```diff
@@ -40,3 +40,65 @@ def test_load_xarray_with_from_file(gridpath):
...
+    """A read failure must surface the backend's own error rather than being
+    printed and swallowed into an UnboundLocalError.
+
+    Regression test for issue #1693. Requires a geopandas file-IO backend,
+    otherwise read_file raises ImportError and the test would pass for the
+    wrong reason.
+    """
+    import pytest
+
+    pytest.importorskip("pyogrio")
```
Sevans711: Don't introduce importorskip to the test suite; no other tests are using importorskip, and, like you mentioned in your comment, you expect pyogrio to be installed when uxarray has been installed with all dependencies. The test suite only gets run in an environment with all dependencies installed.

test/io/test_geopandas.py
```diff
@@ -40,3 +40,65 @@ def test_load_xarray_with_from_file(gridpath):
...
+        out = _set_crs(gdf)
+
+    assert out.crs is not None
+
+
+def test_unsupported_geometry_is_reported():
+    """Dropping a geometry silently would yield a grid missing a face with no
+    indication that anything was skipped."""
+    import pytest
+    gpd = pytest.importorskip("geopandas")
```
Sevans711: (See previous comments, don't introduce importorskip)

uxarray/core/utils.py
```diff
@@ -32,11 +32,18 @@ def _open_dataset_with_fallback(filename_or_obj, chunks=None, **kwargs):
     try:
         # Try opening with xarray's default read engine
         return xr.open_dataset(filename_or_obj, chunks=chunks, **kwargs)
-    except Exception:
+    except Exception as default_engine_error:
```
Sevans711, line 35: Apologies if you already clarified this in comment thread but could you clarify why this change is being added as part of this PR? (Might not need to change anything in response to my comment here, just not fully understanding yet how this helps to fix the original issue.)

uxarray/io/_fesom2.py
```diff
@@ -89,7 +89,10 @@ def _parse_nod2d(grid_path):
 
     file_path = os.path.join(grid_path, "nod2d.out")
     if not os.path.isfile(file_path):
-        raise FileNotFoundError("TODO: ")
+        raise FileNotFoundError(
+            f"Expected a FESOM2 ASCII grid directory containing 'nod2d.out', "
+            f"but no such file exists under {grid_path!r}."
```
Sevans711: Minor suggestion: here and below, I recommend to use os.path.abspath(grid_path) in the error messages. Avoids confusion for any users with file structures like: `run1/gridfolder/...` and `run2/gridfolder/...` just showing "gridfolder" in the error message.

Sevans711 (changes requested):
Looks pretty reasonable as an overall set of changes to make, but I left a variety of inline comments to be addressed, including one point where I just needed a bit more clarification on the reason for the changes. Thank you for doing this work to improve the clarity and error handling throughout the code!

Flagging #1617 as related (may want to delay any work there on uxarray/io error messages until this merges, since this touches multiple errors in uxarray/io).

test/io/test_fesom.py
Sevans711: This was fixed in test_fesom.py, but not in test_geopandas.py (which has lines like `from uxarray.io._geopandas import ...`), so I am leaving this comment unresolved for now.

test/io/test_geopandas.py
Sevans711: Looks like this was fixed; marking as resolved!

test/io/test_geopandas.py
Sevans711: Looks like this was addressed; marking as resolved!

test/io/test_geopandas.py
Sevans711: Looks like this was fixed; marking as resolved!

uxarray/io/_fesom2.py
Sevans711: Looks like this was addressed; marking as resolved!

test/io/test_fesom.py
rajeeja: Fixed — imports moved to top of test_geopandas.py.

uxarray/core/utils.py
rajeeja: Same bug as `_gpd_read`: the default engine's error was discarded, so the failure reported only the netcdf4 reason. Chained so both show.

Sevans711 (approved):
Looks ready to merge, thank you for implementing this fix and for all the back and forth while addressing my comments!

test/io/test_geopandas.py
```diff
@@ -40,3 +45,45 @@ def test_load_xarray_with_from_file(gridpath):
...
+        out = _set_crs(gdf)
+
+    assert out.crs is not None
+
+
+def test_unsupported_geometry_is_reported():
+    """Dropping a geometry silently would yield a grid missing a face with no
+    indication that anything was skipped."""
+    gdf = gpd.GeoDataFrame(
+        geometry=[Polygon([(0, 0), (1, 0), (1, 1), (0, 0)]), Point(5, 5)],
```
cmdupuis3, line 109: I would think you'd want this to be a hard error? Are there grids that have loose points in the geometry like this?

rajeeja: Fair point, but most of the times these are no for actual analysis type data/mesh, they are for viz etc. 

We only handle Polygon and MultiPolygon, so what gets skipped is the annotation layer normal GIS files carry alongside polygons, like city markers and coastlines. 

Making that a hard error means one stray point renders a 10,000-polygon shapefile unreadable and you have to go pre-filter in geopandas, while the warning still gets you the grid and says what was dropped.

cmdupuis3 (commented):
I skimmed the code, and aside from the question above, nothing raised any immediate red flags.

I generated some adversarial tests to trigger the deeper traceback stacks that should be expected with this PR, and I can confirm that `_gpd_read` shows tracebacks through geopandas and pyogrio.

A couple of things make the `_set_crs` and `_extract_geometry_info` tracebacks a little obscure though. For one, the `stacklevel=2` hardcode stops partway down the traceback... is that intended?

Another thing is that these two don't include the name of the file they're raising warnings about:

```Python
import os, tempfile, warnings
import geopandas as gpd
from shapely.geometry import Polygon
import uxarray as ux

shp = os.path.join(tempfile.mkdtemp(), "no_crs.shp")
gpd.GeoDataFrame(geometry=[Polygon([(0, 0), (1, 0), (1, 1), (0, 0)])], crs=None).to_file(shp)

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    ux.Grid.from_file(shp, backend="geopandas") 

w = [x for x in caught if "CRS" in str(x.message)][0]
assert "no_crs.shp" in str(w.message), "warning does not name the file it is about"
assert "uxarray" not in w.filename, f"warning blames uxarray internals: {w.filename}"
```

rajeeja:
Thanks for the pointers. I will make these changes

cmdupuis3 (approved):

PR #1700, opened 2026-08-21
Cached benchmark file I/O

### LINKED ISSUE #1698: Need a holistic approach to file I/O for benchmarks
This issue is two-pronged:

1. File I/O for benchmarks is implemented on a per-benchmark-file basis, so it's difficult to know at a glance which files are currently being used for benchmarks. This also promotes copy-pasting of I/O code, so consolidating the I/O would be in line with the DRY principle.
2. Partly as a consequence of (1), the fragmented I/O regime basically makes ASV retrieve the needed files at every single benchmark run. For benchmark suites on shared files, this can be enormously expensive. 

Instead, I'm proposing that we have one benchmark helper file to support all needed file I/O, which will consolidate I/O code and enable single-retrieval for batches of benchmarks.

cmdupuis3:
Closes #1698

## Overview

 
This PR is a large-scale overhaul of the benchmarks to support single-loads of files per benchmark suite.

Currently, the ASV benchmark suite runs a new process per benchmark, requiring full-scale loads of files for every single benchmark. There are a limited number of files that are actually used to benchmark, so we end up loading many files dozens of times, running up wallclock time on the benchmark suite as a whole. This PR instead has a unified approach, where a parent process is launched for the whole benchmark suite, uncached files are retrieved into `/uxarray/benchmarks/_io_cache/`, and benchmarks are spawned on forked processes that inherit visibility into the file cache and some other properties. 

Simultaneously, this reorganization of file I/O will help to expand the scope of what most non-I/O benchmarks are able to test, since some of them were artificially restricted to smaller grids. That raises another question about how many benchmarks really need to be run at every resolution, but I think that's out of scope for this PR, since it would alter the actual benchmark suite, rather than just changing how they are run.

General CI runs should see a wallclock improvement for whole benchmark suites. CI benchmark runs are showing improvements of about 19 minutes so far (38min -> 19min). Runs on HPC can expect larger improvements. Larger benchmark suites will benefit more from cached file I/O, while benchmarks run individually won't see much improvement.

Other general speed improvements:
1. Test processes forking from warm-cache parent process
2. Async dispatch of tests on file arrival, so large files don't stall tests of smaller files.
3. Sample generation on parent process for reuse in forked processes; sacrifices a bit of robustness for significant speed gains in geometry routines.

This PR has merged in PR #1708 already, since it is a precondition for the optimizations here to be thread-safe.

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**

## AI Disclosure

AI Usage: Claude Opus 5

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,15 +1,22 @@
-<!--  The PR title should summarize the changes, for example "Add `Grid._build_face_dimension` function".
+<!--  Thank you for opening a PR! To help us review your contribution, please follow instructions below
+      while filling out this form, and read the etiquette reminders at the bottom. -->
+
+<!--  Please ensure the PR title summarizes the changes, e.g. "Adds `Grid._build_face_dimension` function".
       Avoid non-descriptive titles such as "Addresses issue #229". -->
 
-<!--  Replace XXX with the issue number resolved by this PR, if this PR fully resolves an issue.
-      If it does not fully resolve any issues, replace with something like "Related to #XXX",
-          or "Fixes part of #YYY but does not fully close it."
-      If it resolves multiple issues, repeat "closes" for each, like "Closes #XXX, closes #YYY." -->
 Closes #XXX
+<!--  Replace XXX with the issue number resolved by this PR, if this PR fully resolves an issue.
+      If it resolves multiple issues, repeat "closes" for each, like "Closes #XXX, closes #YYY."
+      If it does not fully resolve any issues, replace with something like "Related to #XXX",
+          or "Fixes part of #YYY but does not fully close it." -->
 
 ## Overview
-<!--  Please provide a few bullet points summarizing the changes in this PR. This should include
-      points on any bug fixes, new functions, or other changes that have been made. -->
+<!--  Please summarize the changes in this PR. How does it solve the original issue? -->
+
+
+<!--  Does the scope of this PR do anything aside from just solving the original issue?
+      And/or, does it not fully solve the issue as originally reported? If so, please clarify. -->
+
 
 ## Expected Usage
 <!--  If this PR adds a new feature, please provide a short example of it in action.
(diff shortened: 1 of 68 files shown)


cmdupuis3:
In timing the `bench_connectivity.Connectivity` suite on derecho, I'm finding that with this PR and a full `_io_cache`, the whole suite takes almost exactly an hour. On the main branch without this PR, just `bench_connectivity.Connectivity.time_edge_face` and `time_face_edge` take about 45 minutes each. The `Connectivity` suite has six tests in total, so the IO cache is actually not being used as extensively as it would be for a full benchmark suite.

dylannelson (approved):
This branch's benchmarks can run in around 20 minutes as opposed to previous PR's runtime of around 50+ minutes, so this is a significant boost ([looking at past and current runs here](https://github.com/UXARRAY/uxarray/actions/workflows/asv-benchmarking-pr.yml)).

The cached grids/datasets match what a fresh read in every time would provide, while saving a significant amount of time. Once again ran into the SCRIP non-determinism problem again while looking into this so maybe we should get that done soon, ha.

Tried to run locally and had a few Window's related issues but I don't think that matters much for this

The only thing that popped up was how a single test seems to be slower now
Before ([grabbed from my old PR](https://github.com/UXARRAY/uxarray/pull/1736)):

Image

After:

Image


It was suggested it may be caused by it now now measuring numba dispatch overhead. Don't think this is a blocker by any means, just strange.

Thanks!

cmdupuis3:
@dylannelson Hmm, I'll take a look. It could be that since these timings are so small, they may get swamped by something else.

Sevans711 (approved):
This looks great to me, happy to see these improvements!

Seeing that @dylannelson already took a closer look including running benchmarks locally and comparing benchmark suite timings before and after these changes, and considering that this code isn't getting run by uxarray users directly, I didn't investigate more closely than just reading through the changes. But, I did read through them, and everything looks reasonable as far as I can tell.

In addition to the benchmark suite speedup here, moving code into the central helpers files like `_fixtures.py` also makes the suite easier to read, and hopefully easier to maintain in the future, because there is now a single source of truth, e.g. filenames are defined in one place, and more ASV quirks handled in one place, instead of repeated throughout the suite.

Approving as-is!

PR #1705, opened 2026-08-25
Improve error messages in uxarray/grid

### LINKED ISSUE #1704: Error messages improvements in uxarray/grid
Sub-issue of #1617 to track improvements to error messages in uxarray/grid. (Clarifying: this is for everything in uxarray/grid, not just uxarray/grid/grid.)

Sevans711:
Closes #1704 (sub-issue of #1617)

## Overview
Makes error messages clearer and more informative throughout uxarray/grid. As suggested by #1617 this is attempting to be an exhaustive fix for all desired error message improvements in uxarray/grid, while touching nothing outside of uxarray/grid. (Clarifying, this is for all files in uxarray/grid, not just files inside uxarray/grid/grid.)

**Small expansion of scope: fixes confusing/misleading error types in a few places:**
1. Exception → AssertionError in `insert_pt_in_latlonbox`: "Logic error in longitude box width calculation". (Missed this while fixing #1622, which did mention this case but it actually appeared twice and the fix missed the second occurrence.)
2. DataCenteringError → NotImplementedError in `Grid.isel()` messages like "Inverse indices are not yet supported for node selection, please use face centers". DataCenteringError was also misleading because these errors depended only on input kwargs, not the actual location of any data.
3. ValueError → NotImplementedError in `_slice_face_indices`: "Exclusive slicing is not yet supported." (Missed this while fixing #1622 despite it being noted there explicitly.)

**Misc. small expansions of PR scope:**
- Removes warning message when initializing a Grid with source_grid_spec None. The message text and comments imply that the intent here was to warn whenever grid_ds is not a minimum representable UGRID dataset, whenever source_grid_spec is None. But, that case is already being handled by the `_validate_minimum_ugrid` check which was directly above it (which will crash when invalid). I suspect that for a while only the warning existed, and then whatever PR added the check forgot to also remove the warning.
- Improves docstring of `Grid.from_face_vertices` to clarify expected shape and content of face_vertices, and the meaning of latlon=False.
- Tiny efficiency improvement in `_compute_face_areas_and_jacobian`: actually utilize `min_jacobian` (computed as `np.min(self._face_jacobian)`), by checking `min_jacobian < 0`, instead of computing it then ignoring it to instead check `np.any(self._face_jacobian < 0)`. Additionally, do not compute `max_jacobian` anymore; previously it was always being computed despite only being used if raising an error, and now it is no longer included in the error message at all.

Aside from these small expansions of scope there should be no functionality changes; the only differences in this PR are the changes to error message contents.

**Benchmarks:** ran benchmarks (see comments below for more details) to confirm there wasn't any performance hit. (Highly unlikely, but could happen if errors are used as part of control flow in a hot loop and any objects now placed into error message strings take non-negligible time to convert to string.)

## PR Checklist

**General**

**Testing & Benchmarking**

- [N/A] Tests cover all major paths in your new functions

## AI Disclosure

AI Usage: GitHub Copilot's inline code suggestions, some small chats with Claude

### DIFF
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -447,8 +447,8 @@ def _grid_to_matplotlib_polycollection(
     # Handle unsupported configuration: splitting periodic elements with projection
     if periodic_elements == "split" and projection is not None:
         raise ValueError(
-            "Explicitly projecting lines is not supported. Please pass in your projection "
-            "using the 'transform' parameter"
+            'Must provide `projection` when periodic_elements=="split" '
+            "while attempting to create polycollection, but got projection=None."
         )
 
     # Correct the central longitude and build polygon shells
@@ -716,7 +716,7 @@ def pole_point_inside_polygon(pole, face_edges_xyz, face_edges_lonlat):
     """Determines if a pole point is inside a polygon."""
 
     if pole != 1 and pole != -1:
-        raise ValueError("Pole must be 1 (North) or -1 (South)")
+        raise ValueError(f"Pole must be 1 (North) or -1 (South); got pole={pole!r}")
 
     # Define constants within the function
     pole_point_xyz = np.empty(3, dtype=np.float64)
@@ -838,7 +838,11 @@ def pole_point_inside_polygon(pole, face_edges_xyz, face_edges_lonlat):
         return ((north_intersections + south_intersections) % 2) != 0
 
     else:
-        raise ValueError("Invalid pole point query.")
+        # (location will always be 1, -1, or 0 from _classify_polygon_location,
+        #  so it should always be handled by cases above.)
+        raise AssertionError(
+            "Internal coding/implementation error: invalid `location`."
+        )
 
 
 @njit(cache=True)
@@ -1272,8 +1276,10 @@ def barycentric_coordinates_cartesian(polygon_xyz, point_xyz):
 
                 return weights, nodes
 
-        # If the point doesn't reside in the polygon, raise an error
-        raise ValueError("Point does not reside in polygon")
+        raise ValueError(
+            "Point does not reside in polygon, during "
+            "barycentric_coordinates_cartesian(polygon_xyz, point_xyz)"
+        )  # (can't do str(float) in numba --> can't include numbers here.)
 
 
 @njit(cache=True)
--- a/uxarray/grid/geometry.py
+++ b/uxarray/grid/geometry.py
@@ -716,7 +716,8 @@ def pole_point_inside_polygon(pole, face_edges_xyz, face_edges_lonlat):
     """Determines if a pole point is inside a polygon."""
 
     if pole != 1 and pole != -1:
-        raise ValueError(f"Pole must be 1 (North) or -1 (South); got pole={pole!r}")
+        raise ValueError("Pole must be 1 (North) or -1 (South)")
+        # (numba complains about f-strings, so don't put `pole` value in message.)
 
     # Define constants within the function
     pole_point_xyz = np.empty(3, dtype=np.float64)
--- a/uxarray/grid/grid.py
+++ b/uxarray/grid/grid.py
@@ -17,7 +17,7 @@ from uxarray.conventions import ugrid
 # Import the utility function for opening datasets with fallback
 from uxarray.core.utils import _open_dataset_with_fallback
 from uxarray.cross_sections import GridCrossSectionAccessor
-from uxarray.errors import DataCenteringError, DimensionError, GridInvalidError
+from uxarray.errors import DimensionError, GridInvalidError
 from uxarray.formatting_html import grid_repr
 from uxarray.grid.angles import _compute_face_node_angles_convex
 from uxarray.grid.area import _get_all_face_area_from_coords
@@ -172,18 +172,9 @@ class Grid:
             if not _validate_minimum_ugrid(grid_ds):
                 raise GridInvalidError(
                     "Grid unable to be represented in the UGRID conventions. Representing an unstructured grid requires "
-                    "at least the following variables: ['node_lon',"
-                    "'node_lat', and 'face_node_connectivity']"
+                    "at least the following variables: ['node_lon', 'node_lat', 'face_node_connectivity'],"
+                    f"\nbut got grid_ds with data_vars: {list(grid_ds.data_vars)}"
                 )
-
-        # grid spec not provided, check if grid_ds is a minimum representable UGRID dataset
-        if source_grid_spec is None:
-            warnings.warn(
-                "Attempting to construct a Grid without passing in source_grid_spec. Direct use of Grid constructor"
-                "is only advised if grid_ds is following the internal unstructured grid definition, including"
-                "variable and dimension names. Using ux.open_grid() or ux.from_dataset() is suggested.",
-                Warning,
-            )
             # TODO: more checks for validate grid (lat/lon coords, etc)
 
         # mapping of ugrid dimensions and variables to source dataset's conventions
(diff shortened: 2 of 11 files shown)


Sevans711:
@cmdupuis3 Do you have any guesses about whether the increased peakmem values for face_bounds are likely to be real performance hits, or just an artifact of how the benchmarks were run? Similar question for the slowdown of `mpas_ocean.ZonalAverage.time_zonal_average('480km')`.

I don't see any obvious routes in the codebase where these changes would have any significant impact on memory usage. (I would only really expect that to occur if the error message is being generated as part of the standard control flow, something like `try: if_it_works_we_are_in_case_A(); except Exception: handle_case_B()`, but it doesn't look like that is occurring here? It's hard to know for sure though, which is the main reason I ran benchmarks.)

Less relevant to this PR, but also, do you have any guesses about why the bot edited the message 5 times after posting it? (Is there any change it is related to the benchmarking suite taking roughly 6 times longer than it took a few weeks/days(?) ago?)

cmdupuis3:
@Sevans711 I've been seeing some odd behaviors from the FaceBounds benchmarks lately. I'm rerunning it to see if we get the same 38k -> 50k behavior in a second round

Sevans711:
Based on how the benchmarks look in #1710 (with no code changes), which also shows performance degradation on face_bounds memory usage, I believe it is safe enough to conclude the degradation here is not a real performance hit. The apparent slowdown to `mpas_ocean.CheckNorm.time_check_norm('480km')` might also just be noise, as it is on the order of 0.2 ms, and in the edit history of the benchmarks message, earlier versions didn't show that slowdown. I am guessing the small apparent improvement to `mpas_ocean.RemapUpsample.time_inverse_distance_weighted_remapping` is probably also not "real", just noise.

So, I am checking the checklist item for "benchmarks show expected behavior", and requesting reviews!

rajeeja (changes requested):
Missed these two new f-strings are inside @njit functions — they print type placeholders, not values.
- bounds.py:416 — insert_pt_in_latlonbox is @njit(cache=True); f"Longitude point out of range (<0 or >2*pi): {lon_pt}" renders as ...: <object type:float64>.
- arcs.py:~72 — point_within_gca is @njit(cache=True); renders gca_a_xyz=<object type:array(float64, 1d, C)>, gca_b_xyz=<object type:...>.

uxarray/grid/grid.py
```diff
@@ -2627,7 +2663,8 @@ def to_linecollection(
         """
         if periodic_elements not in ["ignore", "exclude", "split"]:
             raise ValueError(
-                f"Invalid value for 'periodic_elements'. Expected one of ['ignore', 'exclude', 'split'] but received: {periodic_elements}"
+                "Invalid periodic_elements. Expected one of ['include', 'exclude', 'split'], "
```
rajeeja: Needs to be  ['ignore', ...] ?

Sevans711: This was a github copilot hallucination which I didn't catch because it "looked right" at a quick glance 😢 Thank you for finding this, it should now be fixed!

Sevans711:
@rajeeja Thank you for reviewing and finding these typos! I believe they are now all fixed. Please let me know if you see any other issues!

uxarray/grid/grid.py
rajeeja: These models make us hallucinate also :) sometimes so smart - sometimes so dumb :)

rajeeja:
> @rajeeja Thank you for reviewing and finding these typos! I believe they are now all fixed. Please let me know if you see any other issues!

Yes, reviewing. I think I found a few more - big changes on this one and some place functionality is changed .. a little bit. I'll point out by the end of the day.

dylannelson:
I think I see two errors that may not be working as intended:
Notebook:
[04-blockers-testing1705.html](https://github.com/user-attachments/files/32298546/04-blockers-testing1705.html)

Examples:
## 1) `to_polycollection` with `periodic_elements="split"`
```python
grid = ux.open_grid(r"..\..\uxarray-pr1705\test\meshfiles\ugrid\quad-hexagon\grid.nc")
grid.to_polycollection(periodic_elements="split", projection=ccrs.Robinson())
```
results in 
```bash
ValueError: Must provide `projection` when periodic_elements=="split" while attempting to create polycollection, but got projection=None.
```
yet `grid.to_polycollection(periodic_elements="split")` works perfectly fine with no error, guessing that it shouldn't from the error above. Are these errors swapped? Providing a projection says it received none, and giving it none seems to pass?

## 2) Grid.from_dataset() responding to invalid types
```python
ux.Grid.from_dataset(None)
```
doesn't give `GridInvalidError: Unsupported Grid Format` anymore and instead gives
```bash
TypeError: _path_normpath: path should be string, bytes or os.PathLike, not NoneType
```
or 
```ux.Grid.from_dataset('text')``` giving 
```bash 
UnboundLocalError: cannot access local variable 'grid_ds' where it is not associated with a value
```
(maybe this one's intended?)

Not sure if these last two matter as much, I feel like none is more likely than a string if someone is using this in an automated script, but given the 2nd error says to provide strings specifically, thought I'd add that too since this error isn't the most clear either.

uxarray/grid/geometry.py
```diff
@@ -447,8 +447,8 @@ def _grid_to_matplotlib_polycollection(
     # Handle unsupported configuration: splitting periodic elements with projection
     if periodic_elements == "split" and projection is not None:
         raise ValueError(
-            "Explicitly projecting lines is not supported. Please pass in your projection "
-            "using the 'transform' parameter"
+            'Must provide `projection` when periodic_elements=="split" '
+            "while attempting to create polycollection, but got projection=None."
```
dylannelson: I think this comment is what I'm running into in the comment above, it checks for "is not None" but the error suggests the projection "is None"

dylannelson (changes requested):

Sevans711:
@dylannelson Thank you for the review! I believe both points are fixed now. Notes:

(1) I originally misunderstood the code that caused this error. Updated the error message to clarify that you must **not** provide `projection` when periodic_elements=="split", and may want to use the `transform` kwarg instead.

(2) The ux.Grid.from_dataset(None) issue here was that I incorrectly assumed `dataset` was a path-like object at this point. The `ux.Grid.from_dataset('text')` crash also occurs on `main` in the same way, but it is confusing. I updated the code to fix both cases by splitting into two try..except blocks, one around the `os.path.isdir` call (which, if it crashes, now raises a TypeError instead of a GridInvalidError) and another around the `_read_fesom2_asci` and `cls(grid_ds, "FESOM2", source_dims_dict)` call (which still converts any TypeError into a GridInvalidError). Additionally, now raises FileNotFoundError when `dataset` is a path-like object but the path doesn't exist, and a clearer GridInvalidError when `dataset` points to a file instead of a directory. Added corresponding regression tests into `test_grid_from_dataset_errors_when_invalid`, in test/grid/grid/test_initialization.py.

dylannelson (approved):
Looked at everything again and it looks to be working as expected and all prior concerns are resolved and working. 

Looking at the transform/split message looks to be more complicated than I think we all originally thought. Not a blocker but here's a few scenarios where it either does/doesn't work depending on the projection. This may be good for a separate issue if we think it's something that we should elaborate on. 

| call | result on a Robinson map |
|---|---|
| `to_polycollection(periodic_elements="split", projection=ccrs.Robinson())` | error |
| `to_polycollection(periodic_elements="split", transform=ccrs.Robinson())` (the hint) | empty map |
| `to_polycollection(periodic_elements="split")` | works |
| `to_polycollection(periodic_elements="split", transform=ccrs.PlateCarree())` | works (same as the default) |
| `to_polycollection(periodic_elements="exclude", projection=ccrs.Robinson())` | works, antimeridian faces dropped |
| `to_linecollection(periodic_elements="split", projection=ccrs.Robinson())` | works |

Sevans711:
> Looked at everything again and it looks to be working as expected and all prior concerns are resolved and working.

Awesome, thank you!

> Looking at the transform/split message looks to be more complicated than I think we all originally thought. Not a blocker but here's a few scenarios where it either does/doesn't work depending on the projection. This may be good for a separate issue if we think it's something that we should elaborate on.

From my current understanding of this part of the code, I don't have any strong feelings about it one way or another. Happy to leave it up to your judgement here!

rajeeja (approved):
The Numba error-message issue from my earlier review is fixed: the affected jitted functions now use static messages rather than f-strings that render type placeholders. The other review feedback has also been addressed, the new initialization regression tests look good, and CI is green. Looks good to me.

PR #1708, opened 2026-08-25
Lazy neighborhood filter kernel compilation

### LINKED ISSUE #1706: Neighborhood filter reductions are making uxarray imports slow
The new neighborhood filters require support for vectorized reduction kernels:

```Python
_MEAN_KERNEL = _make_kernel(lambda window, _: np.mean(window))
_SUM_KERNEL = _make_kernel(lambda window, _: np.sum(window))
_MIN_KERNEL = _make_kernel(lambda window, _: np.min(window))
_MAX_KERNEL = _make_kernel(lambda window, _: np.max(window))
_PTP_KERNEL = _make_kernel(lambda window, _: np.max(window) - np.min(window))
_MEDIAN_KERNEL = _make_kernel(_median)
_VAR_KERNEL = _make_kernel(_variance)
_STD_KERNEL = _make_kernel(lambda window, ddof: np.sqrt(_variance(window, ddof)))
# ``percentile`` is ``quantile`` on a 0-100 scale, so both methods rescale onto
# this one kernel rather than compiling a near-duplicate.
_QUANTILE_KERNEL = _make_kernel(lambda window, q: np.quantile(window, q))
```

However, this block is resulting in severe performance degradations of uxarray imports, notably affecting the benchmark suites.

In exploratory work with PR #1700, I found that this block can affect thread pooling as well. The strategy seems to be to lazily generate these kernels, which would defer the overeager thread pooling and limit the burden on uxarray imports.

cmdupuis3:
Closes #1706

## Overview

This PR should solve an issue introduced with PR #941, wherein the neighborhood filter gufunc-based kernels would compile eagerly, forcing a major slowdown of basic `import uxarray` imports, as well as creating threadpools that would eventually block safe forking that is needed in PR #1700.

The idea is to have lazy compilation of gufuncs introduced with neighborhood filters for performant reductions. This prevents the otherwise eager compilation of every one of these kernels upon `import uxarray`.

Combined, this PR and PR #1700 can bring the total benchmark suit runtime down to about half of the pre-neighborhood filters runtime of ~40min, yielding about 20min without neighborhood filter benchmarks and 26-40ish minutes with them. 

On its own, this PR can at least mitigate some of the benchmark performance regression, on the order of about 30-50% or so.

## PR Checklist

**General**

**Testing & Benchmarking**

**Documentation**

## AI Disclosure

AI Usage: Claude Opus 5

### DIFF
--- a/uxarray/grid/neighbors.py
+++ b/uxarray/grid/neighbors.py
@@ -1512,14 +1512,20 @@ class Neighborhood:
     _sum_kernel = staticmethod(_make_kernel(lambda window, _: np.sum(window)))
     _min_kernel = staticmethod(_make_kernel(lambda window, _: np.min(window)))
     _max_kernel = staticmethod(_make_kernel(lambda window, _: np.max(window)))
-    _ptp_kernel = staticmethod(_make_kernel(lambda window, _: np.max(window) - np.min(window)))
+    _ptp_kernel = staticmethod(
+        _make_kernel(lambda window, _: np.max(window) - np.min(window))
+    )
     _median_kernel = staticmethod(_make_kernel(_median))
     _var_kernel = staticmethod(_make_kernel(_variance))
-    _std_kernel = staticmethod(_make_kernel(lambda window, ddof: np.sqrt(_variance(window, ddof))))
+    _std_kernel = staticmethod(
+        _make_kernel(lambda window, ddof: np.sqrt(_variance(window, ddof)))
+    )
 
     # ``percentile`` is ``quantile`` on a 0-100 scale, so both methods
     # rescale onto this one kernel rather than compiling a near-duplicate.
-    _quantile_kernel = staticmethod(_make_kernel(lambda window, q: np.quantile(window, q)))
+    _quantile_kernel = staticmethod(
+        _make_kernel(lambda window, q: np.quantile(window, q))
+    )
 
     def mean(self, uxda):
         """Mean of each neighborhood."""
--- a/uxarray/grid/neighbors.py
+++ b/uxarray/grid/neighbors.py
@@ -1,4 +1,4 @@
-import threading
+import functools
 import warnings
 from typing import Callable
 
@@ -1247,43 +1247,6 @@ def _make_kernel(reduce_fn):
     return kernel
 
 
-class _LazyKernel:
-    """A kernel that builds itself the first time it is called.
-
-    ``guvectorize`` compiles at decoration time when it is given explicit
-    signatures, so calling :func:`_make_kernel` at module scope would compile
-    every reduction during ``import uxarray``
-    """
-
-    __slots__ = ("_reduce_fn", "_kernel", "_lock")
-
-    def __init__(self, reduce_fn):
-        self._reduce_fn = reduce_fn
-        self._kernel = None
-        self._lock = threading.Lock()
-
-    def __call__(self, *args):
-        kernel = self._kernel
-        if kernel is None:
-            # Checked again under the lock: dask's threaded scheduler can call
-            # one reduction from several workers at once
-            with self._lock:
-                kernel = self._kernel
-                if kernel is None:
-                    kernel = self._kernel = _make_kernel(self._reduce_fn)
-        return kernel(*args)
-
-    def __getstate__(self):
-        # Only the reducer travels. A compiled gufunc is a ``numpy.ufunc``,
-        # which pickle cannot address by name and so refuses outright.
-        return self._reduce_fn
-
-    def __setstate__(self, reduce_fn):
-        self._reduce_fn = reduce_fn
-        self._kernel = None
-        self._lock = threading.Lock()
-
-
 # Reducers take ``(window, param)``; those without a parameter ignore the
 # second argument. Numba keys its cache by code object rather than qualified
 # name, so the identically-named lambdas below do not collide.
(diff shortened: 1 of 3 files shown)


uxarray/grid/neighbors.py
```diff
@@ -1509,45 +1548,45 @@ def __repr__(self) -> str:
 
     def mean(self, uxda):
         """Mean of each neighborhood."""
-        return self._apply_kernel(uxda, _MEAN_KERNEL, 0.0)
+        return self._apply_kernel(uxda, _mean_kernel, 0.0)
```
Sevans711: I'm pretty sure this isn't equivalent anymore; the equivalent syntax if sticking with functools here would be `_mean_kernel()`, right?

I would have a slight preference for a solution that doesn't use functools.cache, to avoid this confusion. Something like this would be more readable and less likely to cause typo, in my opinion:

```python
_NUMBA_KERNELS = {}
def _numba_kernel(kernel_name):
    """returns numba kernel object, such as _make_kernel(lambda window, _: np.mean(window)).
    kernels are cached after initially being created.
    Supported kernels are: mean, sum, ...
    """
    if kernel_name in _NUMBA_KERNELS:
        return _NUMBA_KERNELS[kernel_name]
    # else, need to make kernel for the first time and cache it:
    if kernel_name=='mean':
        kernel = _make_kernel(lambda window, _: np.mean(window))
    elif kernel_name=='sum':
        kernel = _make_kernel(lambda window, _: np.sum(window))
    # ... similar pattern for all supported names
    else:
        raise ValueError(f'unrecognized kernel_name: {kernel_name!r}')
    _NUMBA_KERNELS[kernel_name] = kernel
    return kernel
```

Sidenote: upon initially reading the code I actually had the feeling that something like _NUMBA_KERNELS probably should actually be attached to the Neighborhood class for now, since these kernels are specifically used by the Neighborhood class and nowhere else. If you think these will always just be specific to the Neighborhood class, I'd recommend something like:

```python
class Neighborhood():
    _NUMBA_KERNELS = {}

    @classmethod
    def _numba_kernel(cls, kernel_name):
        # same as above, but use cls._NUMBA_KERNELS instead.
```

uxarray/grid/neighbors.py
```diff
@@ -1592,7 +1631,7 @@ def run(block, arrays):
             # path does too by writing into a float64 output.
             if block.dtype not in (np.float64, np.float32):
                 block = block.astype(np.float64)
-            return kernel(block, *arrays, param)
+            return kernel()(block, *arrays, param)
```
Sevans711: Ah, I see now that you added an extra call here… probably to deal with the issue I noted in my previous comment? I think this is a confusing abstraction though; basically, with this syntax it means that kernel isn't actually a compiled kernel at all, it is a "function factory which returns a compiled kernel function". This feels like it is returning to functional programming abstractions which I recall you mentioning you wanted to avoid.

If you dislike my previous suggestion and really want to keep the functools solution, I would really want to see the calls used above, e.g. `self._apply_kernel(uxda, _mean_kernel(), 0.0)` instead of here.

(If you really like the code as-is, the minimal change I would want to see would be to add clearer comments / docstrings to clarify for future developers that `kernel` isn't actually the kernel, but rather a function which returns a kernel function.)

Sevans711 (changes requested):
Thank you looking into this! Assuming you wanted me to give an initial review now because you requested review, but not 100% sure since it is still marked as draft.

The regression test is clever, and I confirmed it crashes on `main`. This should be one good way to guard against `import uxarray` slowdowns in the future!

I left inline comments about the caching. Basically, it feels strange to me to use functools.cache in this situation, and I think it leads to some more-complicated-than-necessary code logic at the kernel call sites. I think it would be easier to maintain/understand quickly in the future if using a single cache which directly contains the cached kernels themselves, instead of caching on a per-function basis. That might also help with introspection/debugging later if you ever want to quickly check which kernels have been compiled so far. For example, there could even be a test which asserts `_NUMBA_KERNELS` is empty after `import uxarray`.

uxarray/grid/neighbors.py
cmdupuis3: That would be okay, either way is fine. 

To be specific, I was talking about getting away from the functional-style API, rather than necessarily the underlying mechanics. I think considering how numba and `@guvectorize` are interacting here, a functional approach at some level is unavoidable.

uxarray/grid/neighbors.py
cmdupuis3: So, the first commit on this branch took an OO approach that should be equivalent to the current approach, but I thought it was messier than the `functools.cache` implementation here. I think this approach has some of the same clarity issues, where we need a lot of OO and control flow to get a relatively simple result.

I'll move the kernels inside `Neighborhood`, and go from there.

cmdupuis3:
pre-commit.ci autofix

uxarray/grid/neighbors.py
Sevans711: Marking as resolved/outdated due to the new factory pattern introduced by defa8d6

uxarray/grid/neighbors.py
Sevans711: Marking as resolved/outdated due to the new factory pattern introduced by defa8d6

Sevans711 (approved):
After the factory pattern rework in defa8d6, I am happy with how this looks. The `_make_kernel` function itself looks a bit complicated but the docstring and comments clearly explain what is happening, and now it actually returns "a kernel function" instead of returning "a kernel factory which must be called to get the kernel function" so the way to use it is aligned with my intuitions based on the function names and documentation. I recommend merging to main!

erogluorhan (approved):
This looks good to me; thanks for the fixes!

Issue #1713, opened 2026-08-26
`UxDataset.isel()` silently fails to slice when providing grid dim not in the dataset


Sevans711:
### Version

v2026.8.0

### How did you install UXarray?

Source

### What happened?

Calling `UxDataset.isel()` to select along a grid dimension other than the grid dimension where the data is currently located successfully returns a `UxDataset` with the grid sliced properly, but the data not sliced at all. For example, for data with "n_face" dimension, slicing at `n_edge=7` as mentioned in the docstring (added in #1684) leads to a result with uxgrid having only the two faces touching edge 7, but with "n_face" data dimension untouched.

I confirmed that this bug was not introduced by #1684; it exists on `main` even before that commit was merged. The behavior is inconsistent with `UxDataArray.isel()`, which instead also slices the "n_face" data dimension appropriately to keep the uxgrid and data in synch.

### What did you expect to happen?

I expected the data to sliced appropriately such that the result's data contains precisely the edges/nodes/faces corresponding to the result's uxgrid.

### Can you provide a MCVE to repoduce the bug?

```python
import uxarray as ux
ds = ux.tutorial.open_dataset("outCSne30-vortex")
assert "n_face" in ds.dims
result = ds.isel(n_edge=7)
print((result.sizes["n_face"], result.uxgrid.n_face))
# expected (2, 2), got (5400, 2).

# Note: when using UxDataArray, there is no issue. This works just fine:
arr = ds["psi"]
result = arr.isel(n_edge=7)
print((result.sizes["n_face"], result.uxgrid.n_face))
# prints (2, 2) as expected.
```

Issue #1639, opened 2026-08-03
isel() fails with `slice` object, in UxDataArray and UxDataset


Sevans711:
### Version

v2026.7.0

### How did you install UXarray?

Source

### What happened?

Slicing UxDataArray or UxDataset using `.isel()` and providing a `slice` object fails with `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'slice'`.

### What did you expect to happen?

I expected `.isel()` to properly select the indicated indices even when a slice object is used to indicate them.

Combining .isel() with .plot() seems like a decent way to visualize "which faces are which?" For example, I might use `uxarr.isel(n_face=slice(0, 100, 5))` to understand how faces 0, 5, 10, ... are laid out in the grid.

### Can you provide a MCVE to repoduce the bug?

```python
import uxarray as ux
ds = ux.tutorial.open_dataset("outCSne30-vortex")
ds.isel(n_face=slice(None, None, 10))  # should get every 10th face.
# but actually, crashes with TypeError.

arr = ds['psi']
arr.isel(n_face=slice(2, 15, 3))  # should get faces 2, 5, 8, 11, 14
# but actually, crashes with TypeError.

# maybe helpful to note, the following seems to work just fine:
arr.isel(n_face=[2,5,8,11,14])
```

Sevans711:
Update: discovered that this bug only applies when providing slice object for "n_face" indexer (regardless of data location).

For example, the following works just fine:
```python
arr = ux.tutorial.open_dataset("outCSne30-vortex")['psi']
arr.isel(n_node=slice(None, None, 100))
arr.isel(n_edge=slice(None, None, 100))
```

(Not showing UxDataset in example here because slicing UxDataset by a dimension which is not one of the data dimensions is currently bugged, as reported in #1713, and this example has data along dimension "n_face".)

PR #1715, opened 2026-08-26
Fix: `.sel()` now actually slices the uxgrid too, for UxDataArray and UxDataset

### LINKED ISSUE #1641: `.sel()` unexpectedly forgets to slice the uxgrid too, for UxDataArray and UxDataset
### Version

2026.7.0

Source

### What happened?

When calling `.sel()` on a UxDataArray or UxDataset, the underlying uxgrid does not get sliced at all.

An example workflow that might want to utilize `.sel()` instead of `.isel()` is if you are slicing multiple times in a row, or doing subsetting, and want to utilize faces' original labels. In that case, it would be reasonable to do something like:
```python
arr = ux.tutorial.open_dataset("outCSne30-vortex")['psi']
arr1 = arr.assign_coords(n_face=np.arange(arr.n_face.size))   # make sure to track the faces' original labels
arr2 = arr1.isel(n_face=range(0, 100, 5))
arr3 = arr2 + 7   # or some other math or analysis operations
# then realize "oh I want to check what the results look like on what were originally faces 20, 30, and 40"
arr4 = arr3.sel(n_face=[20,30,40])
# and "I want to remind myself where the data is located so I want to plot this"
arr4.plot()
```
This currently works up until the last line, when there is a crash because the grid wasn't also sliced to just faces 20, 30, and 40.

I expected `.sel()` to work similarly to `.isel()`; the underlying grid should be sliced appropriately. Or, at the very least, I expected to receive a warning message or a crash to suggest that `.sel()` is not yet fully supported by uxarray, rather than a confusing error later on when the data no longer aligns with the grid (for example, the code above makes `ValueError: Length of values (3) does not match length of index (20)`).

```python
arrA = ux.tutorial.open_dataset("outCSne30-vortex")['psi']
arrB = arrA.sel(n_face=[20,30,40])
arrB.uxgrid.n_face   # Got 5400; expected 3.
```

Sevans711:
Closes #1641 

## Overview

Adds implementations for `UxDataset.sel()` and `UxDataArray.sel()` which actually slice the uxgrid too, ensuring the result's uxgrid and the result's data do not get out of synch. Followed the plan described in #1641, i.e., offload grid dimension indexing to `.isel()` (if there are coordinate labels for a grid dim, convert labels into indices first), and just utilize xarray's `.sel()` for non-grid dims.

Adds regression tests (see `test_sel_indexes_grid()`, which ensures the originally-reported bug is fixed, and `test_sel_uses_grid_dim_labels()`, which ensures sel() is using the actual coordinate labels for grid dimensions if there is a corresponding coordinate value). Also adds a test to ensure sel() still works even for grid dimensions which are not part of the data.

Tiny expansion of scope: fills out the docstring for the `_validate_indexers` function. Aside from that, there is no expansion of scope, the changes here all relate to fixing the original bug.

## PR Checklist

**General**

**Testing & Benchmarking**

## AI Disclosure

AI Usage: GitHub Copilot's inline code suggestions, and probably some small discussions with Claude

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -244,6 +244,7 @@ Selection & Indexing
    :toctree: generated/
 
    UxDataArray.isel
+   UxDataArray.sel
    UxDataArray.where
 
 
--- a/test/core/test_indexing.py
+++ b/test/core/test_indexing.py
@@ -1,0 +1,85 @@
+"""
+Purpose: tests related to indexing Grid, UxDataArray, and/or UxDataset,
+e.g. UxDataArray's and UxDataset's .isel() and .sel() methods.
+
+(Some .isel() and sel() tests are in test_dataarray.py and test_dataset.py,
+but could maybe be moved here? Having test_indexing.py as its own file helps
+to ensure consistency between UxDataArray and UxDataset indexing.)
+"""
+import numpy as np
+import uxarray as ux
+
+def test_sel_indexes_grid():
+    """ensure obj.sel({grid_dim: ...}) actually indexes the result.uxgrid, too,
+    for UxDataArrays and UxDatasets. Regression test for #1641.
+    """
+    # extremely simple case:
+    uxds = ux.tutorial.open_dataset("quad-hexagon")
+    result = uxds.sel(n_face=0)
+    assert result.sizes['n_face'] == result.uxgrid.n_face == 1
+    # (similar check for UxDataArray)
+    uxarr = uxds['t2m']
+    result = uxarr.sel(n_face=0)
+    assert result.sizes['n_face'] == result.uxgrid.n_face == 1
+
+    # more complicated case:
+    uxds = ux.tutorial.open_dataset("outCSne30-timeseries")
+    assert uxds.sizes['n_face'] > 5000
+    result = uxds.sel(time=['2018-04-28T00', '2018-04-28T03'], n_face=range(0, 5000, 100))
+    assert result.sizes == {'time': 2, 'n_face': 50}
+    assert result.uxgrid.n_face == 50
+    # (similar check for UxDataArray)
+    uxarr = uxds['psi']
+    result = uxarr.sel(time=['2018-04-28T00', '2018-04-28T03'], n_face=range(0, 5000, 100))
+    assert result.sizes == {'time': 2, 'n_face': 50}
+    assert result.uxgrid.n_face == 50
+
+def test_sel_uses_grid_dim_labels():
+    """ensure obj.sel({grid_dim: ...}) actually utilizes coordinate labels on that grid dim,
+    for UxDataArrays and UxDatasets. Regression test for #1641.
+    TODO: fix #1714 then uncomment the UxDataset tests below
+    """
+    # test corresponding to the workflow described in #1641, but for UxDataset
+    uxds = ux.tutorial.open_dataset("outCSne30-vortex")
+    # (uncomment the next few lines after fixing #1714)
+    # uxds1 = uxds.assign_coords(n_face=np.arange(uxds.n_face.size))
+    # uxds2 = uxds1.isel(n_face=range(0, 100, 5))
+    # uxds3 = uxds2 + 7
+    # # "check what the results look like on what were originally faces 20, 30, and 40"
+    # uxds4 = uxds3.sel(n_face=[20,30,40])
+    # assert uxds4.sizes['n_face'] == uxds4.uxgrid.n_face == 3
+
+    # test corresponding to the workflow described in #1641, for UxDataArray
+    uxarr = uxds['psi']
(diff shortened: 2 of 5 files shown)


dylannelson (approved):
Ran it through some tests, got it to install and run cleanly on my end. 
The ~70 line docstring seems like a bit much, but not a blocker.
I noticed the mix of comments like
>   (uncomment the next few lines after fixing #1713)

So I'll start taking a look at those too to get an idea for the bigger picture here, but it looks like you're planning ahead. Thanks!

Sevans711:
Thank you for the review! Small follow-up:

> The ~70 line docstring seems like a bit much, but not a blocker.

It also felt a bit long to me too… though most of it is just a copy-paste of xarray's `sel()` docstrings, to match uxarray's style for `isel()` docstrings. Maybe eventually (perhaps as a followup issue/PR) it would better to rewrite all of these docstrings to instead just clarify uxarray-specific behaviors and then just refer directly to the docs for the corresponding xarray functions?

cmdupuis3 (changes requested):
A few things:

1) Adding `.sel` for `UxDataArray` adds to the API separately from the `UxDataset.sel` version, so `UxDataArray.sel` should be added to api.rst.

2) `da.sel` crashes when using slice objects.

```Python
import numpy as np
import uxarray as ux

uxgrid = ux.Grid.from_healpix(zoom=0)  # 12 faces
uxda = ux.UxDataArray(
    np.arange(uxgrid.n_face, dtype=float), dims="n_face", uxgrid=uxgrid
)

# Works: integer-array indexer on the grid dim
print("sel(n_face=[0, 1, 2]) ->", uxda.sel(n_face=[0, 1, 2]).uxgrid.n_face)

# Works: a slice, once the grid dim has coordinate labels
labeled = uxda.assign_coords(n_face=np.arange(uxgrid.n_face))
print("labeled.sel(n_face=slice(0, 2)) ->", labeled.sel(n_face=slice(0, 2)).uxgrid.n_face)

# Crashes: the same slice with no coordinate labels (the default)
print("sel(n_face=slice(0, 3)) ->", uxda.sel(n_face=slice(0, 3)).uxgrid.n_face)
```

3) method/tolerance are dropped on grid dims without coords: `arr.sel(n_face=3, method="nearest")` does positional selection. Plain xarray raises `ValueError: cannot supply selection options ... for dimension 'd' that has no associated coordinate or index. The docstring documents the index-based fallback, but silently ignoring an explicitly-passed method diverges from xarray; raising would match. This is an issue with `UxDataset` as well, so maybe a bit out of scope.

Sevans711:
Thank you @cmdupuis3 for the review and for finding these bugs! I believe I have addressed them all in the latest commit. More details:

> 1) Adding `.sel` for `UxDataArray` adds to the API separately from the `UxDataset.sel` version, so `UxDataArray.sel` should be added to api.rst.

Done! Good catch; I hadn't thought to check this because `UxDataArray.sel()` already existed before this PR, it was just being inherited from xarray.DataArray.

> 2) `da.sel` crashes when using slice objects.

This one is deeply connected to issue #1639 (`.isel()` fails with slices for both `UxDataArray` and `UxDataset`) but I've added a temporary workaround for this PR (inside `sel()` logic, convert `slice` to `range` before passing to `isel()`) along with a comment that the workaround should be removed after 1639 gets fixed (using `slice` objects properly may be more efficient than using `range`, and conceptually this should be handled by `isel()` logic, not directly in `sel()`).

I added `test_indexing.test_sel_can_use_slice()` as a regression test for this.

> 3) method/tolerance are dropped on grid dims without coords: `arr.sel(n_face=3, method="nearest")` does positional selection. Plain xarray raises `ValueError: cannot supply selection options ... for dimension 'd' that has no associated coordinate or index.` The docstring documents the index-based fallback, but silently ignoring an explicitly-passed method diverges from xarray; raising would match. This is an issue with `UxDataset` as well, so maybe a bit out of scope.

Good point, I didn't realize xarray was doing this. I think it is reasonable to scope this within this PR (sel() on main would do this currently since it basically just calls xarray's sel() directly), so I implemented a fix for this here.

I added `test_indexing.test_sel_crash_if_provided_selection_options_with_coordless_dims()` as a regression test for this. That test includes both 1D and 2D examples (therefore it is maybe tangentially relevant to #1461?), and demonstrates the behavior is analogous to xarray's behavior by repeating the same tests on `obj.to_xarray()` objects. (Checking the xarray behavior here should help mitigate drift between uxarray and xarray on this topic.)

Please let me know if you notice any other issues or if it looks okay now!

cmdupuis3:
I'll check it out. If the slice crash is already documented with #1639, we can drop that issue here.

cmdupuis3 (approved):

Issue #1718, opened 2026-08-28
Math with multiple uxarray objects with different grids silently succeeds, when data dims are otherwise compatible


Sevans711:
### Version

v2026.8.0

### How did you install UXarray?

Source

### What happened?

Mathematical operations with multiple uxarray objects having data with compatible dimensions (i.e., compatible via broadcasting, such that xarray wouldn't complain if performing the same operation on [obj.to_xarray() for obj in objs]) succeeds regardless of whether the underlying grids are the same, and keeps only one of the objects' uxgrids.

This might not have been noticed before because xarray crashes if the data dimensions aren't compatible; this only occurs when the dimensions are compatible.

### What did you expect to happen?

I expected any operations combining multiple uxarray objects to require all of those objects to have equivalent uxgrids (compared as `objA.uxgrid == objB.uxgrid`). When having compatible dimensions but incompatible uxgrids, I expected a GridsMismatchError would be raised.

### Can you provide a MCVE to repoduce the bug?

```python
objA = ux.tutorial.open_dataset('quad-hexagon')['t2m']
_tmp = ux.tutorial.open_dataset('outCSne30-timeseries')['psi']
objB = _tmp.isel(n_face=[10,20,30,40])
print((objA.sizes, objB.sizes))
# prints (Frozen({'n_face': 4}), Frozen({'time': 6, 'n_face': 4}))
print(arrA.uxgrid == arrB.uxgrid)
# prints False

# this should fail but it succeeds:
result = objA + objB

# result's uxgrid == objA.uxgrid; the objB.uxgrid was silently ignored:
print((result.uxgrid == objA.uxgrid, result.uxgrid == objB.uxgrid))
# prints (True, False)

# similar issue if objB is a UxDataset instead (i.e. remove ['psi'] above)
# although that actually leads to result.uxgrid == objB.uxgrid instead.
```

Issue #1573, opened 2026-07-15
Utilize keyword-only arguments, at least in public API


Sevans711:
### Proposed new feature or change:

Public API should use "*" in function signatures to enforce "everything after this is a keyword-only argument". This has (at least) three major benefits:
1. **Enforce clarity**: no more calls like `ux.open_dataset(gridfile, datafile, None, False, True)`. What do the None, False, and True mean?
2. **Prevent silent bugs from mismatched/misremembered parameters**: if I accidentally forgot the order for the 4th and 5th parameters above, I would have accidentally done `ux.open_dataset(gridfile, datafile, None, True, False)` instead, and uxarray would not give me any warnings about my typo.
3. **Improve code extensibility**: I am planning to add a new optional argument soon, something like `ux.open_dataset(gridfile, datafile, dirname=path_to_files)` (see #1555). Currently, because we are not using any keyword-only arguments, to support full backwards compatibility I am forced to put dirname last in the signature, after all of the potentially-positional arguments like those taking None, False, and True values above. That's not ideal; conceptually, dirname is very close to gridfile and datafile since they all specify file paths; I also hope it might be a convenient option for many users, so I would prefer to put it earlier to make it easier to find.

Additionally, **this would match xarray style**. For example the start of the corresponding function's signature there looks like: `xarray.open_dataset(filename_or_obj, *, engine=None, ...)`, meaning `filename_or_obj` is the only allowed positional argument in xarray's case.

## Deprecation cycle
This change would break backwards compatibility, so it would need an associated deprecation cycle. For that, I would propose to utilize something similar to xarray's `_deprecate_positional_args` wrapper, with slight customizations for uxarray's deprecation cycle.

Example: current version (hiding type annotations here for brevity):
```python
def open_dataset(grid_filename_or_obj, filename_or_obj,
    chunks=None, chunk_grid=True, use_dual=False, grid_kwargs=None, **kwargs):
    ...
```
New version (during deprecation cycle):
```python
@_deprecate_positional_args('2026.7.0', end='Jan 2027')
def open_dataset(grid_filename_or_obj, filename_or_obj, *,
    chunks=None, chunk_grid=True, use_dual=False, grid_kwargs=None, **kwargs):
    ...
```
[...]

PR #1719, opened 2026-08-28
Fixes `uxarray.concat` bugs and adds docstring

### LINKED ISSUE #1642: uxarray.concat implementation incomplete/buggy and needs a docstring
### Version

v2026.7.0

Source

### What happened?

uxarray defines `concat()` in its public API, however the implementation is confusing/incomplete/buggy and does not have any docstring. Examples of issues (aside from missing docstring):
1. The implementation seems to attempt to support objects other than UxDataArray and UxDataset; instead of explicit type-checking it only checks "do the objects all have a `uxgrid` attribute? If yes, proceed." Is this intentional? There is no documentation to clarify so I can't tell, but it's not necessarily what I would have guessed. Relevant to comment: https://github.com/UXARRAY/uxarray/pull/1640#issuecomment-5172325706
2. Using objects with equal uxgrids which are not exactly the same object (via `is`) incorrectly cause crash.
3. UxDataArrays are not actually supported; the method always attempts to convert to UxDataset even if inputs are all UxDataArrays. This leads to a confusing error: `TypeError: unhashable type: 'UxDataArray'`.
4. `uxarray.concat()` does not appear anywhere in the test suite. (It also isn't used anywhere internally throughout uxarray, which might be why these issues haven't been reported yet…?)

0. `uxarray.concat` needs a docstring. EDIT: the function signature should also be updated to clarify that `dim` is a required parameter.
1. Use explicit type-checking (e.g. `isinstance(obj, (UxDataArray, UxDataset))`). Or (probably less likely to be the desired solution), clarify in docstring that this method attempts to support other objects too, as long as they have a certain combination of characteristics (one of which is "has a `uxgrid` attribute")
2. uxgrids should be compared via `==` instead of `is`. This would match the existing pattern from UxDataArray's calculus methods. 
  - (2a) If efficiency is a concern, consider adding a check like `if self is other: return True` inside `Grid.__eq__`.
  - (2b) Similarly to those comparisons, the error should be changed to a GridsMismatchError if the grids are not equal. Missed this during #1621, possibly because it looked like an object id comparison rather than a grid equality test.
3. UxDataArrays should be supported in a way that naturally extends how xarray.concat handles DataArrays. E.g., if the inputs are all UxDataArrays, the output should be as well.
4. As apart of public API, there should definitely be at least a few tests to ensure `uxarray.concat` works as expected.

(EDIT: example below actually demonstrates issue 1 now, not issue 0.)

```python
import uxarray as ux

# -- issue 1 --
class Foo():
    def __init__(self, uxgrid):
        self.uxgrid = uxgrid
foo1 = Foo(7)
foo2 = Foo(7)  # also 7 because: want to use the same uxgrid value in both cases.
ux.concat([foo1, foo2], dim='anything')
# >> TypeError: can only concatenate xarray Dataset and DataArray objects, got <class '__main__.Foo'>
# (expected something like: "TypeError: expected UxDataArray or UxDataset; got type Foo")


# -- issue 2 --
arrA = ux.tutorial.open_dataset("outCSne30-vortex")['psi']
arrB = arrA.copy()
ux.concat([arrA, arrB], dim='new_dim')
# >> ValueError: Object at index 1 has a different 'uxgrid' attribute.
# (expected: should be able to concat these; the grids are definitely equivalent...)


# -- issue 3 --
arrA = ux.tutorial.open_dataset("outCSne30-vortex")['psi']
arrB = arrA + 10
ux.concat([arrA, arrB], dim='new_dim')
# >> TypeError: unhashable type: 'UxDataArray'
# (expected: should be able to concat these... result should be a UxDataArray with
#  arrA values at result.isel(new_dim=0) and arrB values at result.isel(new_dim=1).)
```

Sevans711:
Closes #1642

## Overview

Addresses all bugs mentioned in #1642. This PR does the following (with numbers here corresponding to numbers in the original issue report):

0. Adds a clear docstring to `uxarray.concat`. Updates the function signature to clarify that `dim` is a required parameter (instead of hiding it inside `*args`).
1. Uses explicit type checking to ensure objects are either all UxDataArrays or all UxDatasets.
2. Compares uxgrids using `==` instead of `is` (well, technically, uses `!=`, but the important part is that the thing being compared is whether the grids are equal, not whether they refer to literally the same object).
3. Now supports concatenating UxDataArray objects, not just UxDatasets.
4. Adds regression tests for (1), (2), and (3), and a few tiny checks of correctness.

**Minor expansions of scope:**
- Adds a check for `self is other` in `Grid.__eq__` to possibly speed up comparison in a relatively common case. Spot checks currently seem to show negligible performance improvements for pre-existing code (e.g., in calculus operations like curl() which compare uxgrids). Might become relevant to performance in concat, now that it compares grids using `==` instead of `is`, or possibly in the future when fixing issue #1718. Even if impact on performance remains negligible in all normal use-cases, it definitely does not hurt to add this check, as it has negligibly tiny costs (`self is other` is blazingly fast compared to array comparison).
- Adds type hinting for UxDataArray.to_xarray() output type (xr.DataArray).
- Moves top-level `import uxarray.core.dataset` in uxarray/remap/utils.py to be inside the relevant function instead. It was causing a circular import error when adding explicit import of UxDataArray into api.py (see notes below). This solution feels like a better style than preventing UxDataArray import in api.py; I don't think it makes sense for internal utils files to eagerly import (at top of file) from such a core class file like dataset.
  - Also, uses explicit import of UxDataArray in `remap.utils._to_dataset` instead of brittle namespace population behavior (it previously assumed `uxarray.core.dataarray` existed after calling `import uxarray.core.dataset`, but without ever explicitly importing `uxarray.core.dataarray`).
- `concat()` now utilizes keyword-only arguments after the second possibly-positional argument (i.e. `concat(objs, dim, third_arg)` is no longer allowed, everything after `dim` must be passed as keyword argument); see #1573 for motivation. Normally I might hesitate to remove the ability to pass positional args without any deprecation cycle, however the numerous bugs with `concat`, lack of appearance in any documentation examples, and complete lack of a docstring, lead me to believe `concat` was not being used much (if at all) before this.

Misc notes:
- The new import of UxDataArray in api.py cannot degrade `import uxarray` performance, because UxDataset is already being imported in api.py, and the dataset.py file already imports UxDataArray.
- Similarly, the new import of pandas in api.py cannot degrade performance, because the dataarray.py file imports UxDataArrayPlotAccessor, and the plot/accessor.py file already imports pandas.
- Maybe it could be nice to support concat() with a mix of xarray and uxarray objects eventually? Though, users could also just turn them all into uxarray objects first, so that would probably not be a high priority.

## PR Checklist

**General**

**Testing & Benchmarking**

## AI Disclosure

AI Usage: GitHub Copilot's inline code suggestions, and a brief Claude chat about docstring style.

### DIFF
--- a/test/core/test_api.py
+++ b/test/core/test_api.py
@@ -284,3 +284,67 @@ def test_open_multigrid_missing_grid_error(gridpath):
 
     with pytest.raises(ValueError, match="Grid 'land' not found"):
         ux.open_multigrid(grid_file, gridnames=["land"])
+
+
+def test_concat_various_inputs():
+    """Ensure concat() requires uxarray objs and raises clear TypeError otherwise,
+    and that concat() works in basic cases (exactly two UxDataArrays or two UxDatasets)
+    Includes regression test for first and third bugs mentioned in #1642.
+    """
+    # ensure reasonable crash when providing no inputs
+    with pytest.raises(ValueError, match="requires at least one object"):
+        ux.concat((), dim='anything')
+    # ensure crash when providing no uxarray objects
+    with pytest.raises(TypeError, match="expected either all UxDataArray or all UxDataset"):
+        ux.concat((7, "not a uxarray object"), dim='anything')
+    # ensure crash when providing some non-uxarray objects
+    uxds0 = ux.tutorial.open_dataset('quad-hexagon')
+    uxds1 = uxds0 + 10
+    with pytest.raises(TypeError, match="expected either all UxDataArray or all UxDataset"):
+        ux.concat((uxds0, "not a uxarray object", uxds1), dim='new_dim')
+    # ensure crash when providing a mix of UxDataArray and UxDataset objects
+    uxarr0 = uxds0['t2m']
+    with pytest.raises(TypeError, match="expected either all UxDataArray or all UxDataset"):
+        ux.concat((uxds0, uxds1, uxarr0), dim='new_dim')
+    with pytest.raises(TypeError, match="expected either all UxDataArray or all UxDataset"):
+        ux.concat((uxarr0, uxds0), dim='new_dim')
+    # ensure concat works in basic cases: two UxDatasets or two UxDataArrays
+    ux.concat((uxds0, uxds1), 'new_dim')  # also ensures can provide dim as positional arg.
+    uxarr1 = uxds1['t2m']
+    assert uxarr0.uxgrid == uxarr1.uxgrid
+    result = ux.concat((uxarr0, uxarr1), 'new_dim')
+    # ^^ also serves as regression test for third bug in #1642,
+    # i.e.: ensures can actually concat UxDataArray objects.
+    # include a few quick tests that the result looks reasonable:
+    assert isinstance(result, ux.UxDataArray)
+    assert result.uxgrid == uxarr0.uxgrid
+    assert result.sizes == {**uxarr0.sizes, 'new_dim': 2}
+
+    # regression test for first bug in #1642: using non-uxarray objects with equal uxgrid
+    #    should raise error message mentioning uxarray objects, not xarray objects.
+    class Foo():
+        def __init__(self, uxgrid):
+            self.uxgrid = uxgrid
+    foo0 = Foo(7)
+    foo1 = Foo(7)  # also 7 because: want to use the same uxgrid value in both cases.
+    with pytest.raises(TypeError, match="expected either all UxDataArray or all UxDataset"):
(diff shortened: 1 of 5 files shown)


dylannelson (approved):
Reviewed locally, made an environment, and tested a handful of operations. The changes to dataarray, grid, api, and utils preserve functionality as expected. Looks good, thanks!

rajeeja (approved):

PR #1688, opened 2026-08-19
Scalarized `gca_gca_intersection`

cmdupuis3:
Tangent issue to #1623, also related to #1648

## Overview

The existing `gca_gca_intersection` uses four `np.ndarray` objects of size 3. The overhead incurred by using these python objects is njit compiled loops is significant, and we get about a 2x speedup in there by scalarizing `gca_gca_intersection`, `_try_gca_gca_intersection`, and `_accux_gca`, and using `_on_minor_arc_xyz` instead of `on_minor_arc`. That precipitates a ~10% speedup overall of `_gca_gca_intersection`.

This PR mainly provides scalarized versions of `_accux_gca_scalar` and `_try_gca_gca_intersection`. The vector versions are preserved, as established with `_accux_constlat`, but I think it's reasonable to delete all the vector versions if they aren't being used otherwise. Benchmarks for scalarized routines are added separately, but we could also delete the vector benchmarks if the vector routines go.

## PR Checklist

**General**
- [x] An issue is created and linked
- [x] Added appropriate labels (if your uxarray repo permissions allow it)
- [x] Filled out Overview and Expected Usage (if applicable) sections

**Testing & Benchmarking**

- [NA] Adequate tests are created if there is new functionality
- [NA] Tests are not too basic (such as simply calling a function and nothing else)
- [NA] Tests cover all major paths in your new functions
- [x] If this PR could affect performance, ran ASV benchmarks and confirmed they show expected behavior (add a new benchmark if necessary)

**Documentation**

- [x] Docstrings have been added to all new functions
- [x] Docstrings have been updated with any function changes
- [x] Internal (private) function names start with an underscore (`_`)

## AI Disclosure

AI Usage: Claude Opus 5 / Sonnet 5

- [x] I take responsibility for all AI-generated content in my PR.
- [x] I have tested all AI-generated content in my PR.

PR #1727, opened 2026-09-01
Avoid tiny numpy arrays in numba: optimize intersections.py (and refactor: "un-scalarize")

### LINKED ISSUE #1587: Investigate Python thread-scaling of AccuX vs FP64 kernels

### LINKED ISSUE #1726: Avoid tiny numpy arrays in numba: optimize intersections.py
Sub-issue of #1648 to track: should apply this optimization throughout the intersections.py file: don't construct many tiny numpy arrays inside numba routines.

Additionally, refactor the intersections.py routines to restore prior behavior of using 3-vector inputs and outputs, but to maintain performance improvements just accept tuples instead of requiring the allocation of tiny numpy arrays. Methods were recently "scalarized" via PR #1688, but [the discussion there](https://github.com/UXARRAY/uxarray/pull/1688#issuecomment-5443658259) clarified that the intent was to restore to the pre-scalarized function interfaces where possible, as long as performance gains remain intact.

Sevans711:
Closes #1726 (sub-issue of #1648)

## Overview

Optimizes numba routines in `uxarray/grid/intersections.py` to avoid constructing many tiny numpy arrays inside numba routines, as discussed in #1648. ASV benchmarks seem to show roughly 2x speedup of intersections algorithms, and no performance degradations. Hard to know for sure how gca_gca intersection performance has been affected because those benchmarks are failing on `main` right now (see discussion on #1688 for details), but the results here at least seem to be the right order of magnitude (roughly 1 μs; latest successes of these benchmarks on main were roughly 1 to 1.5μs).

Additionally, refactors the intersections.py routines to restore prior behavior (from before #1688 merged) of using 3-vector inputs and outputs where possible. To maintain performance improvements, just use tuples instead of allocating tiny numpy arrays. The discussion in PR 1688 clarified the intent was to restore to the pre-scalarized function interfaces where possible without sacrificing performance.

A few changes are within scope of the original issue but the reasoning is not necessarily obvious directly from inspecting the code diff; clarifying here:
- Previously, `gca_gca_intersection` output (len(result) == number of intersections) did not match its docstring, which promised shape (2,3) result but filling unused rows with NaNs. This PR updates the implementation to fill unused rows with NaNs, and _always_ return a tuple of 2 length-3 tuples, like ((x1,y1,z1),(x2,y2,z2)), filling with NaNs to represent nonexistent intersection points. (The consistent shape is necessary for the performance optimization here. Numba complains that it "Can't unify return type" if the output length depends on the number of intersections.) The test suite has been updated accordingly, to check for nans instead of len(result).
  - Also updated the docstring of `get_number_of_intersections` to clarify that it could now be applied to results from `gca_gca_intersection` too.
- Removes scalarized versions of methods where non-scalarized versions now suffice. I.e., removed `_accux_constlat_scalar` (use `_accux_constlat` instead), and `_snap_const_lat_endpoint_xy` (use `_snap_const_lat_endpoint` instead).  By using tuples instead of tiny numpy arrays, the "non-scalar" versions also avoid allocation costs, so there is no performance-related need to continue maintaining the scalarized versions.
- Similarly, updated arcs.py, removing `_on_minor_arc_xyz` by moving its logic into `on_minor_arc`, since the scalarized version wasn't being used anywhere anymore (except in benchmarking suite, which has been updated appropriately to use `on_minor_arc`). Updated `on_minor_arc` docstring to clarify inputs don't need to be numpy arrays; tuples of length 3 also work just fine.
- Improved various docstrings, updating docstrings where needed to explain new behavior, but also adding to existing docstrings to clarify meanings and expectations about inputs and outputs, and removing now-obsolete notes about preferring scalarized versions.

EDIT: discussions below revealed the need to incorporate the following changes, too:
- update the geometry_samebody.py and geometry_samebody_gcagca.py benchmarks files to keep the fp64 function performance in line with the optimizations to the corresponding functions from uxarray.
  - While doing that change, I also noticed and fixed: `_fp64_try_gca_gca_intersection` was using "and" instead of multiplying together 0s and 1s, so fp64 vs accux wasn't actually the only difference between it and intersections.py's `_try_gca_gca_intersection`.

EDIT: **minor expansions of scope** (requested in [comment below](https://github.com/UXARRAY/uxarray/pull/1727#pullrequestreview-5118426007)):
- Swapped np.max to np.nanmax in geometry_samebody_gcagca.py:274, to match the const-lat sibling at geometry_samebody.py:296.
- Swapped njit decorators for all _fp64 methods in geometry_samebody and geometry_samebody_gcagca benchmarks to exactly match decorators from intersections.py. This fix may help to resolve #1587(?)

## PR Checklist

**General**

**Testing & Benchmarking**

## AI Disclosure

AI Usage: GitHub Copilot's inline code suggestions, plus some Claude for numba debugging questions

### DIFF
--- a/benchmarks/geometry_samebody_gcagca.py
+++ b/benchmarks/geometry_samebody_gcagca.py
@@ -31,6 +31,9 @@ def _fp64_gca(w0, w1, v0, v1):
     analogue of _accux_gca (intersections.py) with accucross/accucross_pair
     replaced by naive FP64 cross products.  Same allocation shape (two np.empty(3))
     so the twin's allocation profile matches the real kernel exactly.
+
+    (Actually longer identical; same operations but here allocates tiny numpy arrays,
+    while intersections.py avoids that for improved efficiency; see issue #1648.)
     """
 
     n1x = w0[1] * w1[2] - w0[2] * w1[1]
@@ -62,6 +65,9 @@ def _fp64_try_gca_gca_intersection(w0, w1, v0, v1):
     """
     L2 (FP64 body) -- byte-for-byte identical logic to _try_gca_gca_intersection
     (intersections.py), only the L1 call differs.
+
+    (Actually longer identical; same operations but here allocates tiny numpy arrays,
+    while intersections.py avoids that for improved efficiency; see issue #1648.)
     """
     pos, neg = _fp64_gca(w0, w1, v0, v1)
 
@@ -102,6 +108,9 @@ def _fp64_gca_gca_intersection(gca_a_xyz, gca_b_xyz):
     """
     L3 (FP64 body) -- identical dispatcher to gca_gca_intersection
     (intersections.py), same np.empty((2, 3)) + res[:count] slice profile.
+
+    (Actually longer identical; same operations but here allocates tiny numpy arrays,
+    while intersections.py avoids that for improved efficiency; see issue #1648.)
     """
     if gca_a_xyz.shape[1] != 3 or gca_b_xyz.shape[1] != 3:
         raise ValueError("The two GCAs must be in the cartesian [x, y, z] format")
@@ -224,8 +233,8 @@ def _batch_accux_gca_dispatch(ga, gb):
     acc = 0.0
     for i in range(ga.shape[0]):
         res = gca_gca_intersection(ga[i], gb[i])
-        if res.shape[0] > 0:
-            acc += res[0, 0]
+        if math.isfinite(res[0][0]):
+            acc += res[0][0]
     return acc
 
 
@@ -265,7 +274,11 @@ def main(n_cases=100_000, seed=20251104):
     for i in range(n_check):
         r_fp = _fp64_gca_gca_intersection(ga[i], gb[i])
         r_ax = gca_gca_intersection(ga[i], gb[i])
-        if r_fp.shape[0] != r_ax.shape[0]:
+        # r_fp gives tiny numpy array result, with len(r_fp) == number of intersections,
+        # while r_ax always gives ((x1,y1,z1), (x2,y2,z2)) tuple result,
+        #   filling with nans if less than 2 intersections.
+        n_intersections = math.isfinite(r_ax[0][0]) + math.isfinite(r_ax[1][0])
+        if len(r_fp) != n_intersections:
             row_mismatch += 1
         elif r_fp.shape[0] > 0:
             max_out_diff = max(max_out_diff, float(np.max(np.abs(r_fp - r_ax))))
--- a/benchmarks/geometry_samebody_gcagca.py
+++ b/benchmarks/geometry_samebody_gcagca.py
@@ -1,7 +1,7 @@
 """Same-body FP64-vs-AccuX diagnostic for the GCA x GCA path.
 
 Companion to ``geometry_samebody.py``, which only covers the GCA/constant-latitude
-stack.  T
+stack. For more details, see description at top of that file.
 
 gca_gca is the kernel invoked per edge from the njit point-in-polygon crossing
 loop (``uxarray/grid/geometry.py`` ``_check_intersection``), so its allocation
@@ -20,8 +20,15 @@ import time
 import numpy as np
 from numba import njit
 
+from uxarray.errors import DimensionError
 from uxarray.grid.arcs import on_minor_arc
 from uxarray.grid.intersections import _accux_gca, gca_gca_intersection
+from uxarray.utils.numba_math import (
+    _numba_add3,
+    _numba_allfinite3,
+    _numba_mul3_scalar,
+    _numba_neg3,
+)
 
 
 @njit(cache=True, inline="always")
@@ -29,11 +36,7 @@ def _fp64_gca(w0, w1, v0, v1):
     """
     L1 (FP64 body) -- plain double-precision cross-product triple, the direct
     analogue of _accux_gca (intersections.py) with accucross/accucross_pair
-    replaced by naive FP64 cross products.  Same allocation shape (two np.empty(3))
-    so the twin's allocation profile matches the real kernel exactly.
-
-    (Actually no longer identical; same operations but here allocates tiny numpy arrays,
-    while intersections.py avoids that for improved efficiency; see issue #1648.)
+    replaced by naive FP64 cross products.
     """
 
     n1x = w0[1] * w1[2] - w0[2] * w1[1]
@@ -49,14 +52,8 @@ def _fp64_gca(w0, w1, v0, v1):
 
     vn = math.sqrt(vx * vx + vy * vy + vz * vz)
     inv = 1.0 / vn if vn != 0.0 else np.inf
-    pos = np.empty(3)
-    pos[0] = vx * inv
-    pos[1] = vy * inv
-    pos[2] = vz * inv
-    neg = np.empty(3)
-    neg[0] = -pos[0]
-    neg[1] = -pos[1]
-    neg[2] = -pos[2]
+    pos = _numba_mul3_scalar((vx, vy, vz), inv)
+    neg = _numba_neg3(pos)
     return pos, neg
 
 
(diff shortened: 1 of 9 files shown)


Sevans711:
Marking as ready for review because tests passed and ASV benchmarks look good. Pinging @cmdupuis3 regarding the `gca_gca` non-constant-latitude benchmarks, since they are failing on main so it is hard to compare. Do you think another intersections-related benchmark needs to be added as part of this PR or is it okay to move forward as-is?

cmdupuis3:
@Sevans711 We should be able to compare it to main commits prior to #1688, the latest of which is 7eff90d. I can do that and report the results here if you'd like.

Sevans711:
@cmdupuis3 Ah, yes if you are able to do that comparison and share results here that would be great!

uxarray/grid/intersections.py
```diff
@@ -300,57 +306,68 @@ def _accux_gca(w00, w01, w02, w10, w11, w12, v00, v01, v02, v10, v11, v12):
...
+    pos : tuple of length 3
+        The positive candidate intersection point (antipodal to neg).
+    neg : tuple of length 3
+        The negative candidate intersection point (antipodal to pos).
     """
-    px, py, pz, ngx, ngy, ngz = _accux_gca(
-        w00, w01, w02, w10, w11, w12, v00, v01, v02, v10, v11, v12
-    )
+    pos, neg = _accux_gca(w0, w1, v0, v1)
+    _numba_allfinite3(pos)
```
dylannelson: Is this line needed? The line below assigns this to `pos_fin`

dylannelson (changes requested):
Looking around here, one small comment

uxarray/grid/intersections.py
Sevans711: Good catch, nope it definitely isn't needed! This was leftover from while I was debugging, just forgot to remove it. Latest commit fixes this!

dylannelson (approved):
Everything I ran came through clean. One minor line was fixed. I didn't see any blockers. Hoping tests done with Chris turn out well. Looks good on my end! Thanks!

cmdupuis3:
The timings are pretty small, so some wide variability is expected, but this is what I see:

| Change   | Before [7eff90d1] <v2026.08.1~2>   | After [4e34008c] <sevans/tinynumpynumba-optimize-intersections.py>   |   Ratio | Benchmark (Parameter)                                             |
|----------|------------------------------------|----------------------------------------------------------------------|---------|-------------------------------------------------------------------|
| -        | 667±0.5ns                          | 417±0ns                                                              |    0.63 | geometry_kernels.GCAGCAIntersection.time_gca_gca_intersection     |
| -        | 1.08±0.06μs                        | 500±20ns                                                             |    0.46 | geometry_kernels.GCAGCAIntersection.time_try_gca_gca_intersection |
| -        | 833±40ns                           | 375±0ns                                                              |    0.45 | geometry_kernels.GCAGCAIntersection.time_accux_gca_kernel         |

benchmarks/geometry_samebody_gcagca.py
```diff
@@ -31,6 +31,9 @@ def _fp64_gca(w0, w1, v0, v1):
     analogue of _accux_gca (intersections.py) with accucross/accucross_pair
     replaced by naive FP64 cross products.  Same allocation shape (two np.empty(3))
     so the twin's allocation profile matches the real kernel exactly.
+
+    (Actually no longer identical; same operations but here allocates tiny numpy arrays,
+    while intersections.py avoids that for improved efficiency; see issue #1648.)
```
cmdupuis3: I'd prefer if the docstrings reflect the current state of the code base, I think it would be okay to remove this and just take the word "identical" out of the original logic (it sounds like Claude-ish to me, so it was probably overeager about that).

cmdupuis3 (changes requested):
All tests pass, and the benchmarks show improvements. I had Claude scan for accuracy divergences and none were found, although LLMs aren't great for that. Considering that this is a sort-of reversion of #1688, that should be expected.

A few benchmarks still need to be migrated though:


- `geometry_samebody.py:270` — import renamed _accux_constlat_scalar → _accux_constlat, but the px, py, nxo, nyo = ... 4-way unpack was kept; the new kernel returns two 3-tuples. TypingError, which takes down the asv class SameBodyConstLat.

- `geometry_samebody.py:292` — res[0, 0] 2D-indexes the now-tuple return of gca_const_lat_intersection.

- `geometry_samebody.py:345` — same 2D indexing in main()'s accuracy cross-check.

benchmarks/geometry_samebody_gcagca.py
Sevans711: Upon taking a closer look, I am wondering… was the intention behind redefining those functions in the benchmarking suite just to compare their performance with the versions from uxarray (which includes AccuSphere calculations)? If yes, it would mean I should also be updating the function bodies similarly here as well.

But also, maybe ASV benchmarking isn't really a great way to test that? The benchmarking output message doesn't directly compare the fp64_kernel and accux_kernel results, you have to separately know that you should consider comparing the two (I didn't realize that until just now). It feels strange to me to have some parts of the benchmarking suite be benchmarking functions that are defined entirely independently of uxarray.

I think I also recall you mentioning something about maybe wanting to drop some of the benchmarks to make the benchmarking suite more concise/meaningful. Would any of these be getting dropped?

I'd be happy to update just the docstrings if that still feels sufficient, after considering these points, just let me know!

Sevans711:
> A few benchmarks still need to be migrated though: …

@cmdupuis3 Thank you for reviewing and pointing these out! I now see that they were originally reported as "failed" in the ASV benchmarking suite results comment above. I believe I have fixed all of these with the latest commit.

benchmarks/geometry_samebody_gcagca.py
cmdupuis3: You mean with #1688? With 1688, the intention was only to change to scalarization, but that forced a rename to `_scalar` suffixes if we wanted to have both the scalarized and non-scalarized versions. That seemed to be a convention, but I think it's just an artifact and we can just have whichever version is better.

If you mean why we have the fp64 benchmarks, the idea is that the AccuX versions should converge to fp64 behavior as multithreading increases, because it's SIMD parallelizable. At the moment, I don't think that issue is fully resolved (#1587), so having the benchmarks is still a good way to note the divergence between current AccuX scaling behavior in Python versus the original.

Sevans711: Oh, no, sorry, to clarify my "intentions" question, I meant to say these functions are getting defined in both the benchmarking suite and in the main codebase, and ask what is the the intention behind that choice? For example, the function at the lines highlighted here is `_fp64_gca`, which is defined only in the benchmarking suite, and the docstring (before I messed with it) seemed to indicate it is intended to just be a redefinition of `_accux_gca` but with naive math (like naive cross product) instead of accuX methods (like accucross).

Based on the rest of your comment above it sounds like the idea is to keep these benchmarks around so that people can compare the timing of fp64 and accuX methods using the ASV benchmarking suite, while investigating #1587.

If that's the goal, then these benchmarks would need to be maintained properly, for a fair comparison. I.e., they should be similarly optimized during this PR; construction of tiny numpy arrays should be dropped here too in favor of using tuples.

Does it make sense to put the effort in to properly optimize these `_fp64` benchmarking functions? Checking in first before I make those changes to ensure we're on the same page - it wouldn't be trivial, but it should be simple enough, just some extra bookkeeping and will lead to more changes needing to be reviewed.

cmdupuis3: Yeah, that's the intent with the fp64 functions, they are supposed to differ only in the compensated arithmetic portion, but again, there are different benchmarking expectations between the two, so that's why it's important to have both. We don't need the fp64 versions in the main codebase.

I see what you mean about keeping the tupling consistent to guarantee a fair comparison. That can possibly be another issue, or if you want, you can append that work here and I'll review that too.

Sevans711: Sounds good, I will append that work here, and will ping you once it is ready for review!

Sevans711: Okay, I believe that work is complete! Details (updated original post in this PR to include this info for future reference, too)

- Updated the geometry_samebody.py and geometry_samebody_gcagca.py benchmarks files to keep the fp64 function performance in line with the optimizations to the corresponding functions from uxarray.
  - While doing that change, I also noticed and fixed: `_fp64_try_gca_gca_intersection` was using "and" instead of multiplying together 0s and 1s, so fp64 vs accux wasn't actually the only difference between it and intersections.py's `_try_gca_gca_intersection`.

cmdupuis3 (changes requested):
Two issues:

1) geometry_samebody_gcagca.py:274 uses np.max; the const-lat sibling correctly uses np.nanmax at geometry_samebody.py:296. Since a 1-intersection result is now (point, (nan,nan,nan)), np.max returns NaN — and max(0.0, nan) returns 0.0, because NaN loses every comparison.

2) geometry_samebody_gcagca.py:61 _fp64_try_gca_gca_intersection is plain @njit(cache=True), but production _try_gca_gca_intersection is @njit(cache=True, inline="always", ...). This means the FP64 version will show different behavior in benchmarks, so the FP64 version should have `inline="always"` as well.

Sevans711:
> Two issues:
> 
>  1. geometry_samebody_gcagca.py:274 uses np.max; the const-lat sibling correctly uses np.nanmax at geometry_samebody.py:296. …
> 
>  2. geometry_samebody_gcagca.py:61 _fp64_try_gca_gca_intersection is plain @njit(cache=True), but production _try_gca_gca_intersection is @njit(cache=True, inline="always", ...)…

Note that both of these are currently issues on `main`, not being introduced by this PR. I do think it is reasonable to solve them here though. I've updated the original post of this PR to clarify that the scope has expanded to include these fixes.

Both should be fixed now! Note that I also updated the decorators for the other _fp64 benchmarks too, to exactly match their production counterparts (many were missing error_model="numpy"; not sure if this matters, but I think it makes sense to make them exactly the same.)

cmdupuis3 (approved):

Issue #1712, opened 2026-08-26
`isel()` with xr.DataArray indexer along grid dimension fails to include indexer's coordinates in result


Sevans711:
### Version

v2026.8.0

### How did you install UXarray?

Source

### What happened?

When using `.isel()` from a `UxDataArray` or `UxDataset`, and supplying an `xr.DataArray` as the indexer for a grid dimension, the `xr.DataArray`'s coordinates are not included in the result.

### What did you expect to happen?

I expected the `xr.DataArray`'s coordinates to be included in the result where possible, consistently with `xarray`'s behavior. See, e.g.: https://docs.xarray.dev/en/stable/user-guide/indexing.html#more-advanced-indexing

### Can you provide a MCVE to repoduce the bug?

```python
import uxarray as ux
obj = ux.tutorial.open_dataset('quad-hexagon')
result = obj.isel(n_face=xr.DataArray(0, coords={'newcoord': 7}))
print(result.coords)
# expected 'newcoord' to be in result.coords, but it was not.

# to see the corresponding xarray behavior, can try:
result = obj.to_xarray().isel(n_face=xr.DataArray(0, coords={'newcoord': 7}))
print(result.coords)
# (in this case, 'newcoord' is indeed in result.coords)
```

Sevans711:
Copy-pasting desired behavior as mentioned in comment in #1729:
> - The dims/coords of the Grid object should never incorporate indexer's dims/coords.
> - The dims/coords of the data object (UxDataArray or UxDataset) should not incorporate the indexer's dims when indexing along a grid dim (e.g. 'n_face') (this is already true), but should probably incorporate its coords (this isn't true yet; see issue 1712).

Here "incorporating the indexer's dims" would mean renaming the grid dim accordingly. For example, in pure xarray,
```
original = xr.DataArray(['a','b','c','d'], dims='letter')
indexer = xr.DataArray([0,2], dims='newdim')
result = original.isel(letter=indexer)
result.equals(xr.DataArray(['a','c'], dims='newdim'))  # this is True
```
That makes sense in xarray. However, if "letter" was a grid dim like "n_face", and `original` was a UxDataArray, then result should still continue to use dim "n_face", not swap to "newdim".

Meanwhile, "incorporating the indexer's coords" would mean that all coords found in the indexer should be included in the result.
[...]

PR #1729, opened 2026-09-02
Fix isel() bugs: accept slices and boolean arrays, and do not smash input dict

### LINKED ISSUE #1639: isel() fails with `slice` object, in UxDataArray and UxDataset

### LINKED ISSUE #1711: `.isel()` modifies indexers dict, causing inconsistent behavior across repeated calls
### Version

v2026.8.0

Source

### What happened?

When using `UxDataArray.isel(indexers)` or `UxDataset.isel(indexers)` with a pre-defined `indexers` dict, the dict itself is edited, so repeating the same `.isel()` code multiple times leads to different results.

I expected the indexers dict to not be modified directly.

The fix should be simple; inside `.isel()` methods, use a copy of `indexers` instead of popping grid_dim from the original.

```python
obj = ux.tutorial.open_dataset('quad-hexagon')  # obj is a UxDataset
chosen_faces = {'n_face': 0}
resultA = obj.isel(chosen_faces)
resultB = obj.isel(chosen_faces)
print((resultA.sizes['n_face'], resultB.sizes['n_face']))
# expected (1, 1), got (1, 4)
print(chosen_faces)
# expected {'n_face': 0}, got {}

# same issue occurs with UxDataArray, e.g. try above with obj=obj['t2m']
```

### LINKED ISSUE #1728: `isel()` silently mishandles boolean indexer arrays
### Version

v2026.08.1

Source

### What happened?

Boolean indexer arrays (or array-like objects) along grid dimensions are not handled properly by UxDataArray.isel() and UxDataset.isel(). Providing boolean indexes leads to them all being converted to 0s and 1s, so the result just has many copies of elements 0 and 1.

I expected boolean indexes to be treated as instructions for whether each element should be included; only elements where the indexer's value is True should be kept in the result (consistent with numpy array indexing and xarray indexing).

```python
import uxarray as ux
arr = ux.tutorial.open_dataset("quad-hexagon")['t2m']
assert arr.sizes['n_face'] == 4
arr.isel(n_face=[True, False, False, False])
# expected UxDataArray with size 1 along n_face; got size 4.

# the following works as expected:
arr.to_xarray().isel(n_face=[True, False, False, False])
```

Sevans711:
Closes #1639, closes #1728
Closes #1711

## Overview

Mapping from changes in this PR to solved issues:
- The majority of the (non-docstring) code changes here contribute directly towards solving both issues #1728 and #1639; both were ultimately caused by the `_slice_face_indices` function in `uxarray/grid/slice.py` incorrectly attempting to convert all `indices` inputs via `np.atleast_1d(np.asarray(indices, dtype=INT_DTYPE))`. The `np.asarray` doesn't make sense for `slice` objects, while the `dtype=INT_DTYPE` doesn't make sense for boolean inputs.
- The fix for #1711 was simply to add the line `indexers = indexers.copy()  # don't modify the original dict` in the appropriate locations in UxDataArray.isel() and UxDataset.isel().

This PR also updates isel() and sel() docstrings to clarify something I previously misunderstood. Indexing grid dimensions does not actually care about data location at all; it instead always selects the minimal grid containing all nodes, edges, or faces specified. For example, if your data lives on "n_edge", indexing by n_edge=7 will not lead to a result with only edge 7. Instead, the result would be on the minimal grid containing edge 7, which is just the two faces touching edge 7 along with all of their edges and nodes. So, the result from selecting n_edge=7 will actually contain all edges from both of those two faces. The previous docstrings were not technically wrong, they were just maybe a bit misleading about this.

Added regression tests for all issues solved in this PR, and confirmed that all new regression tests fail on main but pass here.

**Minor expansions of PR scope:**
- Clarified in isel() and sel() docstrings that grid dimension indexers cannot have more than 1 dimension (such as a 2D DataArray). Using multi-dimensional indexers is supported along other dimensions, and can be a powerful xarray feature (toy example: `selected = data.isel(timesteps=xr.DataArray([[1,3,5],[4,5,6]], dims=['experiment_name', 't_index'], coords={'timestep_group_name': ['odd_timesteps_only', 'just_4_5_6']}; selected.mean('t_index')` gets the mean value of the data for each timestep group). However, supporting it for grid dimensions is not possible, due to the limit of only one Grid object attached to a given UxDataset or UxDataArray.
- Also created a clear DimensionError error message which occurs in case of 2D+ indexers (previously was getting `IndexError: Unlabeled multi-dimensional array cannot be used for indexing: n_face` but that was particularly misleading if providing a well-labeled xarray.DataArray). Added a test to ensure this case raises DimensionError.
- Removed the promise from Grid.isel docstring of "Support for more methods, such as exclusive and clipped indexing is in the works." Git blame shows this promise is more than 2.5 years old. Searching uxarray github issues with the word "inclusive" reveals no open (or closed!) issues about this topic.
- Similarly, removed the `inclusive=True` kwargs from the internal `_slice_node_indices`, `_slice_edge_indices`, and `_slice_face_indices` functions. Passing `inclusive=False` was never actually supported (it just raised NotImplementedError). This could be restored if `inclusive=False` indexing ever gets implemented, or maybe would implement entirely separate functions instead, but in the meantime there doesn't seem to be a compelling reason to keep it.

**Tiny expansion of PR scope:**
- Added leading underscore for values being cached to the grid's `_ds`: `_subgrid_node_indices`, `_subgrid_face_indices`, and `_subgrid_edge_indices`, along with comment to clarify these are only being used by `_slice_from_grid`, and that if multiple isel() operations are performed then these only reflect the indices from the latest one, not the indices relative to the original Grid.

## PR Checklist

**General**

**Testing & Benchmarking**

## AI Disclosure

AI Usage: Some Claude chats, plus GitHub Copilot's inline code suggestions

### DIFF
--- a/test/core/test_indexing.py
+++ b/test/core/test_indexing.py
@@ -154,9 +154,11 @@ def test_isel_can_use_bool():
 def test_indexing_by_dataarray():
     """ensure isel() and sel() with indexer=xr.DataArray(...) both work as expected.
     The dims/coords of the Grid object should never incorporate indexer's dims/coords.
-    The dims/coords of the data object (UxDataArray or UxDataset)
-    should not incorporate the indexer's dims (this is already true),
+    The dims/coords of the data object (UxDataArray or UxDataset) should not incorporate
+    the indexer's dims when indexing along a grid dim (e.g. 'n_face') (this is already true),
     but should probably incorporate its coords (this isn't true yet; see issue #1712).
+
+    Regression test for bug in branch (fixed before merging to main) for PR 1729.
     """
     # ensure grid's dims/coords do not incorporate indexer's dims/coords:
     indexer0 = xr.DataArray(0, coords={"newcoord": 7})
--- a/test/core/test_indexing.py
+++ b/test/core/test_indexing.py
@@ -133,6 +133,23 @@ def test_sel_can_use_slice():
     # result = labeled_ds.sel(n_face=slice(0, 2))
     # assert result.n_face.size == result.uxgrid.n_face == 3
 
+def test_isel_can_use_bool():
+    """ensure isel() supports indexing by a boolean indexer array.
+    Regression test for #1728.
+    """
+    ds = ux.tutorial.open_dataset("quad-hexagon")
+    assert ds.isel(n_face=[True, False, False, False]).equals(ds.isel(n_face=0))
+    assert ds.isel(n_face=[False, True, False, True]).equals(ds.isel(n_face=[1,3]))
+    result = ds.isel(n_face=[False, False, False, False])
+    assert result.sizes['n_face'] == result.uxgrid.n_face == 0
+
+    # repeat tests but with UxDataArray:
+    arr = ds['t2m']
+    assert arr.isel(n_face=[True, False, False, False]).equals(arr.isel(n_face=0))
+    assert arr.isel(n_face=[False, True, False, True]).equals(arr.isel(n_face=[1,3]))
+    result = arr.isel(n_face=[False, False, False, False])
+    assert result.sizes['n_face'] == result.uxgrid.n_face == 0
+
 def test_sel_crash_if_provided_selection_options_with_coordless_dims():
     """ensure sel() crashes if providing `tolerance` and/or `method` options
     whenever any of the indexed dims have no associated coordinates.
--- a/test/core/test_indexing.py
+++ b/test/core/test_indexing.py
@@ -151,6 +151,53 @@ def test_isel_can_use_bool():
     result = arr.isel(n_face=[False, False, False, False])
     assert result.sizes['n_face'] == result.uxgrid.n_face == 0
 
+def test_indexing_by_dataarray():
+    """ensure isel() and sel() with indexer=xr.DataArray(...) both work as expected.
+    The dims/coords of the Grid object should never incorporate indexer's dims/coords.
+    The dims/coords of the data object (UxDataArray or UxDataset)
+    should not incorporate the indexer's dims (this is already true),
+    but should probably incorporate its coords (this isn't true yet; see issue #1712).
+    """
+    # ensure grid's dims/coords do not incorporate indexer's dims/coords:
+    indexer0 = xr.DataArray(0, coords={"newcoord": 7})
+    indexer1 = xr.DataArray([1,2], dims="newdim", coords={"newdim": [7,8]})
+    ds = ux.tutorial.open_dataset("quad-hexagon")
+    result0_isel = ds.isel(n_face=indexer0)
+    assert "newcoord" not in result0_isel.uxgrid._ds.coords
+    # assert "newcoord" in result0_isel.coords   # uncomment after fixing #1712
+    result0_sel = ds.sel(n_face=indexer0)
+    assert "newcoord" not in result0_sel.uxgrid._ds.coords
+    # assert "newcoord" in result0_sel.coords   # uncomment after fixing #1712
+    result1_isel = ds.isel(n_face=indexer1)
+    assert "newdim" not in result1_isel.uxgrid._ds.dims
+    assert "newdim" not in result1_isel.uxgrid._ds.coords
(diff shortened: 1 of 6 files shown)


dylannelson:
@Sevans711 
I think .isel is causing an extra dim leak in some way. While I've run across varying levels of importance of this happening on the grid's _ds (like here) some didn't really matter, but this looks like fairly normal usage (maybe not if a user wouldn't slice with a data array?) so this could be a valid concern? I've also seen that these kinds of leaks can occur just from improperly using commands, so let me know if this is just improper usage. 

From what I see, `_slice_face_indices` used to run every indexer through `np.atleast_1d(np.asarray(indices, dtype=INT_DTYPE))` ([link to removed line](https://github.com/UXARRAY/uxarray/pull/1729/changes#diff-7ccf8b81f30584a99d366ea445b2b8cf7d090dbea5c333ed5d2071ec6b5ebe28L168), old line 168), which stripped the dim name, and now without that, the strip doesn't occur. The extra dim now reaches `ds.isel(n_face=face_indexer)` ([link to added line](https://github.com/UXARRAY/uxarray/pull/1729/changes#diff-7ccf8b81f30584a99d366ea445b2b8cf7d090dbea5c333ed5d2071ec6b5ebe28R154), new line 154) intact, and xarray renames the sliced axis to the indexer's dim

You can put any dimension name in isel and it preserves that name down the line (even nonsense names like I did here to make it more visible in the output), whereas it didn't keep the names before. Hopefully this helps visualize the new vs old output: 

# Run on this branch's env
image

# Run on an older branch I had on hand
image

Sevans711:
> I think .isel is causing an extra dim leak in some way. 

Great find @dylannelson and thank you for providing this example! Indeed it was introduced due to changing the line you pointed to. I reproduced on my end, fixed, and added regression test `test_indexing_by_dataarray()` which crashes on this branch before this fix but passes after the fix.

Considering this along with issue #1712 led me to writing down the following statements (in the test's docstring), which I think are an accurate description of the desired behavior?
- The dims/coords of the Grid object should never incorporate indexer's dims/coords.
- The dims/coords of the data object (UxDataArray or UxDataset) should not incorporate the indexer's dims when indexing along a grid dim (e.g. 'n_face') (this is already true), but should probably incorporate its coords (this isn't true yet; see issue 1712).

dylannelson (approved):
Fix was made and I noticed no other blockers. The branch built well, ran through tests, and no concerns present! Thanks!

rajeeja:
Yes, thanks for handling these issues. Two typos: bookeeping at slice.py:142 and internaly at slice.py:183. The DimensionError-for-IndexError swap and the subgrid rename are improvements, just worth calling out for anyone relying on the old behavior.

Sevans711:
@rajeeja thank you for reviewing!

> Two typos: bookeeping at slice.py:142 and internaly at slice.py:183.

Fixed!

> The DimensionError-for-IndexError swap and the subgrid rename are improvements, just worth calling out for anyone relying on the old behavior.

Indeed, these are highlighted in the original PR message (second bullet point under "Minor expansions of PR scope" and first point under "Tiny expansions of PR scope").

rajeeja (approved):

PR #1732, opened 2026-09-03
Vector calculus follow-ups from #1663: eager divergence and a user guide fix

### LINKED ISSUE #1731: Vector calculus follow-ups from #1663: eager divergence and a user guide fix
Two things #1663 left behind.

**1. `divergence` is eager on dask input.** The metric term uses `other.values` where its companion in `curl` uses `self.data`, so `u.curl(v)` returns a `dask.array.Array` and `u.divergence(v)` returns an `ndarray` for the same chunked input. #1588's `.values` → `.data` sweep reached the curl side during a conflict resolution on the #1663 branch and missed this one. One-word fix; there is no dask coverage anywhere in `test/core/test_vector_calculus.py` to have caught it.

**2. `docs/user-guide/vector_calculus.ipynb` still teaches the planar operators.** Cells 14 and 27 give `∂v/∂x − ∂u/∂y` and `∂u/∂x + ∂v/∂y` with no metric term, and eight code cells (16, 18, 20, 22, 29, 31, 35, 37) hand-roll the operators from `gradient()` components instead of calling `curl()`/`divergence()`. So every worked example demonstrates pre-#1663 physics while the library computes something else. Cell 43's "curl of gradient ≈ 0" is stated for the hand-rolled version.

Switching those cells to the API changes one result materially: curl and divergence of a *constant* field are no longer exactly zero, they are `u·tan(φ)/a` and `−v·tan(φ)/a` — `4.8e-09` and `9.6e-09` on this ±2° MPAS subset. That is correct spherical behaviour and worth saying out loud in the notebook rather than deleting. The other examples move by less than a part in 10⁴, except the two "≈ 0" residuals which go from `2.8e-10` to `4.0e-09` and stay ≈ 0.

rajeeja:
Closes #1731. Follow-up to #1663 / #1662.

## Overview

Two things #1663 left behind.

**1. `divergence` is eager on dask input.** #1663 added the spherical metric term to both `curl` and `divergence`. The two are companions and should behave identically, but `divergence` reaches for `other.values` where `curl` uses `self.data`:

```python
# curl()
metric = self.data * tan_lat          # stays lazy

# divergence()
metric = other.values * tan_lat       # materializes
```

That materializes `other` and drags the whole divergence eager, so `curl` and `divergence` return different types for the same chunked input. The mismatch came in when #1588's `.values` → `.data` sweep was applied during a conflict resolution on the #1663 branch and only reached the `curl` side.

Measured on `test/meshfiles/ugrid/quad-hexagon/grid.nc` with a chunked field:

| | before | after |
|---|---|---|
| `curl` | `Array` | `Array` |
| `divergence` | `ndarray` | `Array` |

No test added. Nothing in `test/core/test_vector_calculus.py` chunks its input, so the whole gradient/curl/divergence area has no dask coverage to extend; adding just enough scaffolding to pin this one line would be out of proportion to a one-word fix. Worth a separate issue if we want that coverage properly. The existing 38 tests pass unchanged.

**2. `docs/user-guide/vector_calculus.ipynb` still teaches the planar operators.** The curl and divergence background cells give

$\text{curl}(\mathbf{F}) = \partial v/\partial x - \partial u/\partial y$ and $\text{div}(\mathbf{F}) = \partial u/\partial x + \partial v/\partial y$

with no metric term, and eight code cells hand-roll the operators out of `gradient()` components instead of calling `curl()`/`divergence()`. So every worked example demonstrated pre-#1663 physics while the library computed something else.

This adds the $u\tan\varphi/a$ and $-v\tan\varphi/a$ terms to the math, switches the examples to the public API, and corrects the residual magnitudes quoted in the prose to what the notebook actually prints.

One result changes materially. Curl and divergence of a *constant* field are no longer exactly zero — they are $u\tan\varphi/a$ and $-v\tan\varphi/a$, which come to `4.8e-09` and `9.6e-09` on this ±2° MPAS subset. That is correct spherical behaviour, so the notebook now says so and checks both against their closed forms (max deviation `0.00e+00`) rather than dropping the examples. Everything else moves by less than a part in 10⁴, except the two "≈ 0" residuals which go from `2.8e-10` to `4.0e-09` and stay ≈ 0.

The notebook is executed by `myst_nb` at docs build time and stores no outputs, so its printed numbers were always fresh while the prose around them went stale silently. Every cell was run locally before committing.

## PR Checklist

**General**

**Testing & Benchmarking**
- [ ] There is adequate test coverage of changes from this PR (add new tests if needed)
- [N/A] If this PR could affect performance, ran ASV benchmarks and confirmed they show expected behavior

- [N/A] Docstrings updated with any function changes, and included in all new functions
- [N/A] User (public) functions added to `docs/api.rst`
- [X] If touched any notebook files, cleared the output of all cells before committing

## AI Disclosure

AI Usage:

### DIFF
--- a/docs/user-guide/vector_calculus.ipynb
+++ b/docs/user-guide/vector_calculus.ipynb
@@ -126,7 +126,7 @@
    "id": "66caacb5",
    "metadata": {},
    "source": [
-    "> **Units.** By default, `gradient()` and `curl()` divide by `uxgrid.sphere_radius` so derivatives carry physical units (e.g. `[data units]/m`, `1/s` for velocity curl). Pass `scale_by_radius=False` to keep results on the unit sphere (per radian). If the grid has no `sphere_radius` attribute, the call falls back to unit-sphere output and emits a `UserWarning`."
+    "> **Units.** By default, `gradient()`, `curl()`, and `divergence()` divide by `uxgrid.sphere_radius` so derivatives carry physical units (e.g. `[data units]/m`, `1/s` for velocity curl). Pass `scale_by_radius=False` to keep results on the unit sphere (per radian). If the grid has no `sphere_radius` attribute, the call falls back to unit-sphere output and emits a `UserWarning`."
    ]
   },
   {
@@ -228,14 +228,16 @@
     "\n",
     "### Background\n",
     "\n",
-    "The curl of a vector field **F** = (u, v) measures the local rotation or circulation. In 2D, curl produces a scalar field representing the magnitude of rotation:\n",
+    "The curl of a vector field **F** = (u, v) measures the local rotation or circulation. In 2D it produces a scalar field. On a sphere of radius $a$ it carries a metric term that the planar formula leaves out:\n",
     "\n",
     "$$\n",
-    "\\text{curl}(\\mathbf{F}) = \\nabla \\times \\mathbf{F} = \\frac{\\partial v}{\\partial x} - \\frac{\\partial u}{\\partial y}\n",
+    "\\text{curl}(\\mathbf{F}) = \\frac{\\partial v}{\\partial x} - \\frac{\\partial u}{\\partial y} + \\frac{u \\tan\\varphi}{a}\n",
     "$$\n",
+    "\n",
+    "That last term comes from the convergence of the meridians and vanishes only on the equator. Dropping it costs a factor of two on solid-body rotation, so `curl()` includes it. Building the same quantity by hand out of `gradient()` components gives the planar answer instead.\n",
     "\n",
     "- **Positive curl**: Counter-clockwise rotation\n",
-    "- **Negative curl**: Clockwise rotation  \n",
+    "- **Negative curl**: Clockwise rotation\n",
     "- **Zero curl**: No local rotation (irrotational flow)\n",
     "\n",
     "### Usage\n",
@@ -244,7 +246,7 @@
     "\n",
     "| **Input**                    |        **Usage**        | **Output**           |\n",
     "| ---------------------------- | :---------------------: | -------------------- |\n",
-    "| Vector field (u, v)          | `u.curl(v)`             | Scalar curl field    |"
+    "| Vector field (u, v)          | `u.curl(v)`             | Scalar curl field    |\n"
    ]
   },
   {
@@ -254,7 +256,9 @@
    "source": [
     "### Constant Fields (Mathematical Validation)\n",
     "\n",
-    "The curl of a constant vector field should be zero everywhere (within numerical precision)."
+    "On a plane the curl of a constant vector field is zero everywhere. On a sphere it is not: with $u$ constant, both derivative terms drop out and the metric term survives, leaving $\\zeta = u\\tan\\varphi/a$ exactly.\n",
+    "\n",
+    "This grid subset spans about ±2° of latitude, so $\\tan\\varphi/a$ stays below $5.5\\times10^{-9}\\ \\mathrm{m^{-1}}$ and the result looks like round-off — but it is a real signal, and we can check it against the closed form.\n"
    ]
   },
   {
@@ -268,12 +272,7 @@
     "u_constant = uxds[\"face_lat\"] * 0 + 1.0\n",
     "v_constant = uxds[\"face_lat\"] * 0 + 2.0\n",
     "\n",
-    "# Compute partials via gradient\n",
-    "grad_u = u_constant.gradient()\n",
-    "grad_v = v_constant.gradient()\n",
-    "du_dy = grad_u[\"meridional_gradient\"]\n",
-    "dv_dx = grad_v[\"zonal_gradient\"]\n",
-    "curl_constant = dv_dx - du_dy\n",
+    "curl_constant = u_constant.curl(v_constant)\n",
     "\n",
     "finite = np.isfinite(curl_constant.values)\n",
     "vals = curl_constant.values[finite]\n",
@@ -284,7 +283,12 @@
     "    print(f\"Finite curl range: [{vals.min():.2e}, {vals.max():.2e}]\")\n",
     "    print(\n",
     "        f\"Max |curl|: {np.abs(vals).max():.2e}, Mean |curl|: {np.abs(vals).mean():.2e}\"\n",
-    "    )"
+    "    )\n",
+    "\n",
+    "# The closed form for a constant u on the sphere is u*tan(lat)/a\n",
+    "lat_rad = np.deg2rad(uxds.uxgrid.face_lat.values)\n",
+    "curl_exact = 1.0 * np.tan(lat_rad) / uxds.uxgrid.sphere_radius\n",
+    "print(f\"Max deviation from u*tan(lat)/a: {np.abs(vals - curl_exact[finite]).max():.2e}\")"
    ]
   },
   {
@@ -306,12 +310,7 @@
     "u_gauss = uxds[\"gaussian\"]\n",
     "v_gauss = uxds[\"inverse_gaussian\"]\n",
     "\n",
-    "# Compute partials via gradient\n",
-    "grad_u = u_gauss.gradient()\n",
-    "grad_v = v_gauss.gradient()\n",
-    "du_dy = grad_u[\"meridional_gradient\"]\n",
-    "dv_dx = grad_v[\"zonal_gradient\"]\n",
-    "curl_gauss = dv_dx - du_dy\n",
+    "curl_gauss = u_gauss.curl(v_gauss)\n",
     "\n",
     "finite = np.isfinite(curl_gauss.values)\n",
     "vals = curl_gauss.values[finite]\n",
@@ -319,8 +318,8 @@
     "    f\"Total faces: {curl_gauss.size}, interior: {vals.size}, boundary NaNs: {np.isnan(curl_gauss.values).sum()}\"\n",
     ")\n",
     "if vals.size:\n",
-    "    print(f\"Finite curl range: [{vals.min():.6f}, {vals.max():.6f}]\")\n",
-    "    print(f\"Mean curl (finite): {vals.mean():.6f}\")"
+    "    print(f\"Finite curl range: [{vals.min():.6e}, {vals.max():.6e}]\")\n",
+    "    print(f\"Mean curl (finite): {vals.mean():.6e}\")"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


docs/user-guide/vector_calculus.ipynb
```diff
@@ -228,14 +228,16 @@
     "\n",
     "### Background\n",
     "\n",
-    "The curl of a vector field **F** = (u, v) measures the local rotation or circulation. In 2D, curl produces a scalar field representing the magnitude of rotation:\n",
+    "The curl of a vector field **F** = (u, v) measures the local rotation or circulation. In 2D it produces a scalar field. On a sphere of radius $a$ it carries a metric term that the planar formula leaves out:\n",
     "\n",
     "$$\n",
-    "\\text{curl}(\\mathbf{F}) = \\nabla \\times \\mathbf{F} = \\frac{\\partial v}{\\partial x} - \\frac{\\partial u}{\\partial y}\n",
+    "\\text{curl}(\\mathbf{F}) = \\frac{\\partial v}{\\partial x} - \\frac{\\partial u}{\\partial y} + \\frac{u \\tan\\varphi}{a}\n",
```
Sevans711: Need to define the meaning of `\varphi`. Is it just one of the spherical coordinates?

docs/user-guide/vector_calculus.ipynb
```diff
@@ -228,14 +228,16 @@
...
     "$$\n",
     "\n",
+    "That last term comes from the convergence of the meridians and vanishes only on the equator. Dropping it costs a factor of two on solid-body rotation, so `curl()` includes it. Building the same quantity by hand out of `gradient()` components gives the planar answer instead.\n",
```
Sevans711: This is very hard for me to parse, please include more details. Maybe it would also help to link to a paper or website that presents/derives the formula, too?

Notes:

- "convergence of the meridians" - what does this mean?
- "comes from the convergence of the meridians" - how?
- "Dropping it costs a factor of two on solid-body rotation" - why?
- "so `curl()` includes it" - this could be removed; it was a bit confusing to read. My default assumption is that the uxarray implementation would match whatever formula is presented here.
- "Building the same quantity by hand out of `gradient()` components" - what does this mean?
- "gives the planar answer instead" - what does "the planar answer" mean?

docs/user-guide/vector_calculus.ipynb
```diff
@@ -254,7 +256,9 @@
    "source": [
     "### Constant Fields (Mathematical Validation)\n",
     "\n",
-    "The curl of a constant vector field should be zero everywhere (within numerical precision)."
+    "On a plane the curl of a constant vector field is zero everywhere. On a sphere it is not: with $u$ constant, both derivative terms drop out and the metric term survives, leaving $\\zeta = u\\tan\\varphi/a$ exactly.\n",
```
Sevans711: zeta is not defined yet; it would be easier to read this section if not defining a new variable, something like this would work great: "leaving u.curl(v) = … exactly"

docs/user-guide/vector_calculus.ipynb
```diff
@@ -254,7 +256,9 @@
...
+    "\n",
+    "This grid subset spans about ±2° of latitude, so $\\tan\\varphi/a$ stays below $5.5\\times10^{-9}\\ \\mathrm{m^{-1}}$ and the result looks like round-off — but it is a real signal, and we can check it against the closed form.\n"
```
Sevans711: What is "the closed form"?

docs/user-guide/vector_calculus.ipynb
```diff
@@ -332,7 +331,7 @@
...
     "\n",
-    "The *magnitude* of the residual depends on the units. By default `gradient()`/`curl()` scale by `uxgrid.sphere_radius` (Earth ≈ 6.37×10⁶ m), so each derivative picks up a factor of 1/radius. `curl(∇φ)` applies the gradient stencil twice and therefore carries a factor of 1/radius² (≈ 4×10⁻¹⁴). For the Gaussian field here (φ ~ O(1)) the scaled residual is ~O(10⁻¹³); on the unit sphere (`scale_by_radius=False`) the same residual is ~O(1–10). Either way it shrinks with refinement.\n"
+    "The *magnitude* of the residual depends on the units. By default `gradient()`/`curl()` scale by `uxgrid.sphere_radius` (Earth ≈ 6.37×10⁶ m), so each derivative picks up a factor of 1/radius. `curl(∇φ)` applies the gradient stencil twice and therefore carries a factor of 1/radius² (≈ 2×10⁻¹⁴). For the Gaussian field here (φ ~ O(1)) the scaled residual is ~O(10⁻¹⁵); on the unit sphere (`scale_by_radius=False`) the same residual is ~O(0.1). Either way it shrinks with refinement.\n"
```
Sevans711: Why did these numbers change? Code changes in this PR should not affect numerical results from these methods, right?

docs/user-guide/vector_calculus.ipynb
Sevans711: Now that I know there could be a real error of roughly 5.5e-9 due to the metric term, it is difficult to trust that the 1e-5 error here is "small". Is this error really "small" or is it a sign of a meaningful mismatch between the analytical expectation and the computed results?

Sevans711 (changes requested):
These are important user guide improvements, converting the guide to actually use the uxarray implementations of these calculus methods, and explaining contributions from the metric terms which were previously not taught at all. It is definitely moving things in a good direction! I left a few inline comments with suggestions/questions.

Overall design question: would it make more sense to include the formulas all at the top of the page? Now that the formulas are more complicated than just the basic definitions which people might naively expect, having all the basic formulas at the top might make it easier to separate these math surprises from the uxarray syntax, implementation details, and visualizations.

docs/user-guide/vector_calculus.ipynb
rajeeja: Added definitions of u, v, λ, φ, a, dx, dy at the top of the Background cell.

docs/user-guide/vector_calculus.ipynb
rajeeja: Done, it now reads `u.curl(v)` = u tan φ/a.

docs/user-guide/vector_calculus.ipynb
rajeeja: Reworded to "checks it against u tan φ/a from the face latitudes".

docs/user-guide/vector_calculus.ipynb
rajeeja: Rewrote it: the cell now starts from the spherical curl (1/(a cos φ))[∂v/∂λ − ∂(u cos φ)/∂φ] and shows the product rule producing u tan φ/a, with a pointer to Holton and Hakim Ch. 4. Dropped "so curl() includes it" and replaced the "by hand" sentence with the literal gradient() expression it referred to.

docs/user-guide/vector_calculus.ipynb
rajeeja: The .values to .data change moves no numbers; the old prose was already stale against main (1/a² is 2.46e-14 for a = 6371229 m, and the hand-rolled residual prints ~1e-14, not 1e-13). Switching the cells to curl() also adds u tan φ/a, which partly cancels the planar residual (mean 4.8e-15 to 2.8e-15). Text now says 2.5e-14 and "a few ×10⁻¹⁵".

docs/user-guide/vector_calculus.ipynb
rajeeja: Good catch, it was not small, the reference had the wrong sign: −(1/(R cos φ)) ∂(ΩR cos²φ)/∂φ = +2Ω sin φ, i.e. planetary vorticity. Against the correct sign the max error is 8.5e-10 (0.02%), and against the old one every face disagreed in sign. This came in with #1517 and went unnoticed because the notebook stores no outputs; fixed in 3b3399ef.

docs/user-guide/vector_calculus.ipynb
Sevans711: The rewrite makes it much easier to read and understand quickly, thank you!

I followed the pointer and spent ~3 mins scrolling around to try to find the corresponding equation but didn't see it in that amount of time. Could you mention the relevant equation number here, too? Aside from that, this part looks good to me now.

docs/user-guide/vector_calculus.ipynb
Sevans711: The current readthedocs preview is now showing:
```
computed   zeta range: [-4.448e-06, 4.259e-06]  1/s
analytic   zeta range: [-5.039e-06, 4.844e-06]  1/s
max |error|: 8.523e-10  1/s  (≈ 0.02% of max |zeta|)
```
I don't understand these numbers - to me it looks like the maximum error must be at least 20% of max |zeta|, not 0.02%. For example, max(analytic) minus max(computed) gives 0.6e-06, not 8e-10. What am I missing here, or is there a typo somewhere?

docs/user-guide/vector_calculus.ipynb
rajeeja: Updated the citation to Section 2.3, especially Eqs. (2.9) and (2.12), which define the local distances and show the spherical basis-vector term.

docs/user-guide/vector_calculus.ipynb
rajeeja: You were right: the two ranges used different face sets. The computed range excluded boundary NaNs, while the analytic range included every face. Both now use the same finite-face mask; the ranges agree and the pointwise max error remains 0.02%.

rajeeja:
@Sevans711 The two remaining points are fixed in 88b30162, and the branch is updated from main. Could you please re-review?

dylannelson (approved):
After talking about a few minor things, we made some changes and all looks good. Thanks for improving the readability.

docs/user-guide/vector_calculus.ipynb
Sevans711: Is it possible to get a citation which has the full curl equation? I see some similarities between those equations and the one in the docs now, but they are a bit different (e.g. the extra negative sign in 2.12, as compared to the equation in the docs) and I don't see the exact path for how to convert the book's equations into the docs's equation.

rajeeja: Replaced it with an NCAR reference that states the full expanded equation directly.

Sevans711 (approved):
All comments addressed, looks ready to merge!

cmdupuis3 (approved):

PR #1736, opened 2026-09-08
Assigning connectivities based on UGRID conventions

### LINKED ISSUE #1717: IO for `fesom` not converting `n2` to `two` during `open_grid`
### Version

2026.7.1

Source

### What happened?

When making a fix for #1444 in PR #1667, we noticed there were far more outliers than expected, in an attempt to clean grid files. One example was this file:

Image

In a bit less detail it looks like this:

Image

We didn't expect this many non-grid dimensions on the grid, and we assume that n2 is likely supposed to be converted to "two" which is a dimension we expect. We would need to resolve this before proceeding, as we don't want to delete it in our planned cleaning process, if it actually needs to be converted to "two"

To see "two" as a dimension, not "n2"

```python
import warnings
from pathlib import Path

warnings.filterwarnings("ignore")
import uxarray as ux

MESHFILES = "../../uxarray/test/meshfiles/"

# FESOM2 mesh.diag — nz, nz1, n2, N, n4
mesh_diag = ux.open_grid(MESHFILES + "ugrid/fesom/fesom.mesh.diag.nc")
mesh_diag

# --- separate cell ---

GRID_DIMS = set(ux.conventions.ugrid.DIM_NAMES)
print("  dims          :", dict(mesh_diag._ds.sizes))
print("  non-grid dims :", [d for d in mesh_diag._ds.sizes if d not in GRID_DIMS])
print("  coords        :", list(mesh_diag._ds.coords))
```

### LINKED ISSUE #1725: Connectivities in different UGRID file reads can have different dimensions
### Version

2026.7.1

Source

### What happened?

We encountered an instance while exploring #1717 that lead to different connectivity dimensions existing on the grid after reading in the file located at [test/meshfiles/ugrid/fesom/fesom.mesh.diag.nc](https://github.com/UXARRAY/uxarray/blob/main/test/meshfiles/ugrid/fesom/fesom.mesh.diag.nc). Like we saw the dimensions of `face_edge_connectivity` as (`n_face`, `n_max_face_nodes`) vs it normally being (`n_face`, `n_max_face_edges`). Below is a few examples of other meshfiles being read in, and looking at their connectivity. All other examples were read in as expected, with one exception that had a different outcome.

## Relation to 1717
This is a seemingly separate but related concern to what we saw with `n2` being read in instead of `two`. We want to decide and document this behavior, where `n_max_face_nodes` can take the place of `n_max_face_faces` or `n_max_face_edges`, and if this can cause downstream problems.

This is what we saw with this file (`face_X_connectivity` ->`n_max_face_nodes`):
Image

This is what we expected, and saw in other files (`face_X_connectivity` -> `n_max_face_X`):
Image

## File location / Attributes
This was a file in the [FESOM folder, within the existing ugrid folder](https://github.com/UXARRAY/uxarray/tree/main/test/meshfiles/ugrid/fesom), as opposed to the [standalone FESOM folder](https://github.com/UXARRAY/uxarray/tree/main/test/meshfiles/fesom). It was read in as a UGRID file following the UGRID pathway within the grid creation pipeline, not the FESOM pipeline.

We attempted to look into whether this is a FESOM problem, or a UGRID problem, and it showed to be a little hard to narrow down and decide where to go from here. Given this was very unexpected, we need to do more testing to narrow down.

## More examples
Here's more examples from both different and similar files:

Image  

They each varied a lot, and were shown to contain a multitude of different connectivities, but none had the wrong names or had different dimensions like we saw in the original example. 
- Some had many connectivities existing on load
- Some had nearly none or 1
- Some were UGRID, some were not
- A regular FESOM2 file was read fine `fesom/soufflet-netcdf/grid.nc`

## After notes:
1) Ugrid connectivity names exist in [uxarray/blob/main/uxarray/conventions/ugrid.py](https://github.com/UXARRAY/uxarray/blob/main/uxarray/conventions/ugrid.py) and are defined as 
```python
    face_node_connectivity -> ('n_face', 'n_max_face_nodes')
    face_edge_connectivity -> ('n_face', 'n_max_face_edges')
    face_face_connectivity -> ('n_face', 'n_max_face_faces')
    edge_node_connectivity -> ('n_edge', 'two')
    edge_face_connectivity -> ('n_edge', 'two')
    node_edge_connectivity -> ('n_node', 'n_max_node_edges')
    node_face_connectivity -> ('n_node', 'n_max_node_faces')
```
2) This MPAS file is also read in as UGRID, which was unexpected:
- [mpas/dyamond-30km/gradient_grid_subset.nc](https://github.com/UXARRAY/uxarray/blob/main/test/meshfiles/mpas/dyamond-30km/gradient_data_subset.nc)
3) You can use `grid.source_grid_spec` to look at which path the grid was read in though: (MPAS, UGRID, ICON, FESOM2, ESFM, Exodus, Scrip)

All files to be read in and have their connectivities and their dimensions to match the conventions described above

```python
MESHFILES = "../../uxarray/test/meshfiles/"
mesh_diag = ux.open_grid(MESHFILES + "ugrid/fesom/fesom.mesh.diag.nc")
mesh_diag
```

dylannelson:
Closes #1717 
Closes #1725
**Why two in one?** The reason for closing both in one PR is due to them being caused by the same problem. Having 3 dimensions in pre-existing connectivities results in `swap_dims` quietly breaking, causing both the issues seen above. By applying some cleaning both before and after `swap_dims` is called, both issues are solved together. As far as I can see, there is no evidence we would see one of the above errors without the other, and since the solution is in the same location, we may as well solve it once and here. Originally I solved the first one alone, but solving the 2nd would basically require deleting the whole original solution, and replacing it with a more general one. So rather than pushing 1 solution to main, deleting it, and pushing another, I say we just push this single refined solution (assuming it works as intended)

## Overview

`_read_ugrid` renamed the trailing dimension of `face_node_connectivity` to `n_max_face_nodes` and left every other connectivity's equivalent dimension as whatever the source file called it. Reading `fesom.mesh.diag.nc`, which names those axes `n3` and `n2`:

| Connectivity | File's name for the axis | dims Before | dims After |
|---|---|---|---|
| `face_node_connectivity` | `n3` | `('n_face', 'n_max_face_nodes')` | unchanged |
| `face_edge_connectivity` | `n3` | `('n_face', 'n_max_face_nodes')` | `('n_face', 'n_max_face_edges')` |
| `face_face_connectivity` | `n3` | `('n_face', 'n_max_face_nodes')` | `('n_face', 'n_max_face_faces')` |
| `edge_node_connectivity` | `n2` | `('n_edge', 'n2')` | `('n_edge', 'two')` |
| `edge_face_connectivity` | `n2` | `('n_edge', 'n2')` | `('n_edge', 'two')` |

Each connectivity's trailing dimension is now mapped to the name `ugrid.CONNECTIVITY` gives it, rather than one being hardcoded and the rest left alone. In this FESOM file (which is read in through the UGRID path) it uses one size-3 `n3` for all three face connectivities. `swap_dims` call cannot express all 3, so the first connectivity renames it for the whole dataset and the remainder are renamed per variable. `face_node_connectivity` is first, so variables that are not connectivity all get `n_max_face_nodes`

## Expected Usage

```Python
import uxarray as ux

uxgrid = ux.open_grid("fesom.mesh.diag.nc")

# the edge connectivities carried the file's own name for the trailing axis
uxgrid.edge_node_connectivity.dims
# before: ('n_edge', 'n2')
# after:  ('n_edge', 'two')

uxgrid.edge_face_connectivity.dims
# before: ('n_edge', 'n2')
# after:  ('n_edge', 'two')

# the face connectivities all shared one size-3 dimension
uxgrid.face_edge_connectivity.dims
# before: ('n_face', 'n_max_face_nodes')
# after:  ('n_face', 'n_max_face_edges')

# so indexing those axes by name now works
uxgrid.edge_node_connectivity.isel(two=0)
# before: ValueError: Dimensions {'two'} do not exist
# after:  the first node of every edge
```

## PR Checklist

**General**

**Testing & Benchmarking**

- [N/A] Docstrings updated with any function changes, and included in all new functions
- [N/A] User (public) functions added to `docs/api.rst`; internal (private) function names start with an underscore (`_`)

## AI Disclosure

AI Usage: Claude

### DIFF
--- a/uxarray/io/_ugrid.py
+++ b/uxarray/io/_ugrid.py
@@ -80,6 +80,18 @@ def _read_ugrid(ds):
 
     dim_dict[ds["face_node_connectivity"].dims[1]] = ugrid.N_MAX_FACE_NODES_DIM
 
+    # The core dims above do not cover the trailing dimension of the edge
+    # connectivities, so a source file keeps whatever it called that axis unless
+    # it is mapped here. edge_node_connectivity and edge_face_connectivity both
+    # assign that dimension the name "two".
+    for conn_name in ("edge_node_connectivity", "edge_face_connectivity"):
+        if conn_name in conn_dict.values():
+            # map this file's name for the axis onto "two", unless already mapped
+            dim_dict.setdefault(
+                ds[conn_name].dims[1], ugrid.CONNECTIVITY[conn_name]["dims"][1]
+            )
+
+    # rename every source dimension collected above to its UGRID name in one pass
     ds = ds.swap_dims(dim_dict)
 
     return ds, dim_dict
--- a/uxarray/io/_ugrid.py
+++ b/uxarray/io/_ugrid.py
@@ -78,21 +78,53 @@ def _read_ugrid(ds):
             if dims[1] == grid_dim:
                 ds[conn_name] = da.T
 
-    dim_dict[ds["face_node_connectivity"].dims[1]] = ugrid.N_MAX_FACE_NODES_DIM
+    # The core dims cover only each connectivity's grid element axis; the trailing
+    # axis still carries whatever the file called it. One source dimension can
+    # serve several connectivities that want different names (FESOM shares one
+    # size-3 dim across all three face connectivities), and dim_dict holds only
+    # one name per dimension, so the first claim renames it dataset-wide and the
+    # rest are renamed per variable below. face_node_connectivity claims first so
+    # that variables which are not connectivity keep n_max_face_nodes.
+    core_dims = set(dim_dict)  # snapshot before the loop starts adding to it
 
-    # The core dims above do not cover the trailing dimension of the edge
-    # connectivities, so a source file keeps whatever it called that axis unless
-    # it is mapped here. edge_node_connectivity and edge_face_connectivity both
-    # assign that dimension the name "two".
-    for conn_name in ("edge_node_connectivity", "edge_face_connectivity"):
-        if conn_name in conn_dict.values():
-            # map this file's name for the axis onto "two", unless already mapped
-            dim_dict.setdefault(
-                ds[conn_name].dims[1], ugrid.CONNECTIVITY[conn_name]["dims"][1]
-            )
+    # every connectivity, face_node first
+    claim_order = ["face_node_connectivity"] + [
+        name for name in ugrid.CONNECTIVITY_NAMES if name != "face_node_connectivity"
+    ]
 
-    # rename every source dimension collected above to its UGRID name in one pass
+    # the connectivities this file actually has
+    present_conn_names = set(conn_dict.values())
+
+    # conn name -> {current dim name: wanted dim name}, applied after swap_dims
+    per_var_dim_dict = {}
+    for conn_name in claim_order:
+        if conn_name not in present_conn_names:
+            continue
+
+        # the file's name for this connectivity's trailing axis, e.g. 'n2'
+        source_dim = ds[conn_name].dims[1]
+        if source_dim in core_dims:
+            continue  # malformed file: a core dim in the trailing slot
+
+        # the name the conventions give that axis, e.g. 'two'
+        ugrid_dim = ugrid.CONNECTIVITY[conn_name]["dims"][1]
+
+        if source_dim not in dim_dict:
+            dim_dict[source_dim] = ugrid_dim  # unclaimed, so rename dataset-wide
+        elif dim_dict[source_dim] != ugrid_dim:
+            # claimed by an earlier connectivity under a different name, so this
+            # one is renamed on its own. swap_dims runs first, hence the key is
+            # the name that claim already gave the dimension.
+            per_var_dim_dict[conn_name] = {dim_dict[source_dim]: ugrid_dim}
+
     ds = ds.swap_dims(dim_dict)
+
+    # the claims dim_dict had no room for: same dimension, a different name.
+    # swap_dims just renamed FESOM's n3 to n_max_face_nodes everywhere; this then
+    # gives face_edge_connectivity its own n_max_face_edges, and likewise
+    # face_face_connectivity its own n_max_face_faces.
+    for conn_name, rename_dict in per_var_dim_dict.items():
+        ds[conn_name] = ds[conn_name].rename(rename_dict)
 
     return ds, dim_dict
 
--- a/test/io/test_ugrid.py
+++ b/test/io/test_ugrid.py
@@ -1,6 +1,5 @@
 import os
 import warnings
-import numpy as np
 import numpy.testing as nt
 import pytest
 import xarray as xr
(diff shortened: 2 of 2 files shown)


dylannelson:
## This branch
image

## Older Versions:
image

cmdupuis3 (approved):
Looks fine, the tests pass and all the cases highlighted here are covered.

uxarray/io/_ugrid.py
```diff
@@ -78,10 +78,54 @@ def _read_ugrid(ds):
...
+    # conn name -> {current dim name: wanted dim name}, applied after swap_dims
+    per_var_dim_dict = {}
+    for conn_name in claim_order:
+        if conn_name not in present_conn_names:
+            continue
+
+        # the file's name for this connectivity's trailing axis, e.g. 'n2'
+        source_dim = ds[conn_name].dims[1]
+        if source_dim in core_dims:
+            continue  # malformed file: a core dim in the trailing slot
```
Sevans711, line 107: If this definitely represents a malformed file, it should either crash with GridInvalidError (this is my preference) or use warning.warn() to at least make a warning, so it doesn't silently give bad results

Sevans711 (changes requested):
Looks pretty good, and makes sense to aim to close both issues with a single PR like this! Left small inline comment.

Aside from that, I'm referring back to my earlier comment in #1725 which flagged this page as related: https://uxarray.readthedocs.io/en/latest/user-guide/representation.html#connectivity

I would suggest to also fix that page (i.e. the Connectivity section of representation.rst) as follows:
- The Dimensions column should use "two", not "2".
- The table is missing a Node Node connectivity row. (Or, maybe the Edge Edge connectivity row should be removed? uxarray/conventions/ugrid.py says edge_edge_connectivity is not yet supported.)
- The table mentions "n_max_edge_faces", however no part of uxarray actually uses that dimension. `Grid.edge_face_connectivity` actually has dimensions "n_edge", "two", in agreement with `EDGE_FACE_CONNECTIVITY_DIMS` in uxarray. The table in your PR overview also shows that you use "two" as part of this fix as well, not "n_max_edge_faces".

"n_max_edge_faces" also appears in `DIM_NAMES`, and I believe it should be removed from there, as part of this PR, because it isn't actually supported anywhere else in uxarray. Using "two" instead of "n_max_edge_faces" contradicts a recent internal discussion, however I think that is okay, because during that discussion nobody brought up what the code is _actually_ doing; I think everyone assumed it was following the docs page (which is incorrect!).

It would make sense to me for these fixes to the docs page and `DIM_NAMES` to be bundled with this PR. But, I could also understand if you would prefer to push them to another PR later. If you'd prefer to push them, could you open up a new issue report for them before this PR merges, to make sure that they don't get forgotten?

rajeeja:
Looks good. I'd agree with Sam, let do those also and then merge this.

dylannelson:
@Sevans711 @rajeeja  Yeah I think a lot of what you're saying is true but I'd prefer it be either part of a seperate PR or after an extensive discussion on what these connectivities should really be. 

## The issue with switching `two` to `n_max_edge_faces`
From our internal discussion, I was told
> ugrid.py: EDGE_FACE_CONNECTIVITY_DIMS = ["n_edge", "two"] <- Wrong, should be n_max_edge_faces

and we simply don't see this to be the case in other grids that we have in uxarray (and Sam is saying `n_max_edge_faces` should be removed all together?) here's a look at some other grids in uxarray being read in (pre patch): 

[11-example-grid-connectivities.html](https://github.com/user-attachments/files/32246765/11-example-grid-connectivities.html)
- The connectivities for these examples (before my intervention) all pointed to what was discussed before, and what is currently present in the [ugrid.py](https://github.com/UXARRAY/uxarray/blob/main/uxarray/conventions/ugrid.py) conventions:
```
    edge_node_connectivity -> ('n_edge', 'two')
    edge_face_connectivity -> ('n_edge', 'two')
```
image

image


So if we were to change `EDGE_FACE_CONNECTIVITY_DIMS` to not be `two`, we would need to look at our test files and figure out why they were `two` as well, potentially change/update them, and change our internal conventions. I think this would be enough of a problem to warrant it's own discussion and issues and PR, and this current PR is more of a specific patch for this weird case of having extra dims (3) instead of the regular 2.

I believe the idea of changing the conventions will not be as easy as just changing a line or two on the docs and ugrid.py, without it causing some other serious confusion or mismatches in these functions and existing test grids, hence why I think we need to have a thorough discussion because the one we had didn't match current implementation and we have a few ideas already that don't align

Sevans711:
@dylannelson I think we might all be on the same page for the most part? I wasn't suggesting to swap "two" to "n_max_face_edges", just to update the docs to clarify that the code uses "two", not "n_max_face_edges". There wouldn't be any functionality changes at all, the only change to the code itself would be a tiny update to `DIM_NAMES`, to remove "n_max_face_edges" since it isn't being used anywhere.

Given that, would you still prefer to push these changes to a different PR? If it makes more sense to you to push them I wouldn't mind, as long as before this PR merges a corresponding issue gets posted to track these points.

uxarray/io/_ugrid.py
dylannelson: @Sevans711 Error should be implemented with GridInvalidError, let me know how it looks

dylannelson:
@Sevans711 I think we're on the same page now. 
>  just to update the docs to clarify that the code uses "two", not "n_max_face_edges"

Done, though this is technically the opposite what the team said, we can discuss later this week

> The table is missing a Node Node connectivity row. (Or, maybe the Edge Edge connectivity row should be removed? 
> The table mentions "n_max_edge_faces", however no part of uxarray actually uses that dimension. 

I didn't change edge edge or node node docs because I'm not familiar with why they are there, and same with `n_max_face_edges` especially since the team said we should be using it but I'm not seeing why we would/should right now. We can get feedback from the others if we want to add node node or remove edge edge, but doesn't seem critical for this PR.

> "n_max_edge_faces" also appears in DIM_NAMES, and I believe it should be removed from there

another situation where I don't see an impact to do so, but if the others want it removed too, I'm happy to do it. I feel more hesitant to touch the conventions file than the docs though.

uxarray/io/_ugrid.py
Sevans711: Looks good to me, marking as resolved!

Sevans711 (approved):
All looks good now, everything within scope of the original issues has been solved, and all of my extra slightly-out-of-scope comments have either been solved or recorded in #1766 for tracking purposes so they can be solved in a later PR. Thank you - approving!

Issue #1597, opened 2026-07-20
UXarray readme & docs should quickly clarify "spherical, 2D"


Sevans711:
### Proposed new feature or change:

UXarray fundamentally handles exactly 2 dimensions of "unstructured-ness", and always treats that 2D unstructured grid as being on the surface of a sphere. Restricting the domain of applicability like this has the benefit of enabling more development efforts and code space dedicated to methods specialized in Earth data analysis. However, these limitations not immediately obvious from the name, the README, nor the "Why UXarray?" docs page.

For example, people running fluid dynamics / plasma physics simulations on irregular 3D grids should be able to quickly understand that uxarray will not help them with their analysis. From quickly skimming the name, the README, and the "Why UXarray" docs page, I initially came away with the opposite impression; I figured "yes, this package does include many Earth-data-specific methods and functionality because that is the primary target audience, but fundamentally its goal is probably to support N-dimensional unstructured-ness, so it might help with 3D unstructured grids".

erogluorhan:
Similar to #1615 , see [this comment](https://github.com/UXARRAY/uxarray/issues/1615#issuecomment-5072783446) from that.

erogluorhan:
Actually, the "... unstructured climate and global weather data analysis ..." phrase in the README/homepage is the key term to imply the "horizontally 2D meshes with vertical levels" assumption. 

Additionally, the following part in Why UXarray? is helping with that understanding, too:

"Specifically, our core functionality is built around horizontally Unstructured Grids as defined by the 2D Flexible Mesh Topology in the UGRID conventions, which can contain a mix of triangles, quadrilaterals, or other geometric faces. These types of Unstructured Grids may have structured vertical levels."

However, we can still improve docs as [suggested here](https://github.com/UXARRAY/uxarray/issues/1615#issuecomment-5072783446) to be further helpful

Sevans711:
Thank you for looking into this, @erogluorhan! Following up with a few thoughts:

> Actually, the "... unstructured climate and global weather data analysis ..." phrase in the README/homepage is the key term to imply the "horizontally 2D meshes with vertical levels" assumption.

This makes sense to me now, but I remember being confused about it before starting to work with uxarray. Despite having some familiarity with nearby science (ionospheric physics), I did not initially realize that "unstructured climate and global weather data analysis" implies 2D horizontal meshes with vertical levels.

[...]

PR #1744, opened 2026-09-10
Fix ESMF encode truncating the fill value into a valid node index

### LINKED ISSUE #1745: Fix ESMF encode truncating the fill value into a valid node index
_encode_esmf applies the UGRID→ESMF index offset to the whole face_node_connectivity array, padding included, and then encodes elementConn as int32. INT_FILL_VALUE + 1 is -2**63 + 1, and narrowing that to int32 truncates it to 1, so every padded slot of a ragged mesh is written to disk as node 0.

Nothing raises, and the resulting file is structurally valid, so any reader — including uxarray's own — takes those slots for real vertices. Faces gain spurious extra nodes and produce incorrect polygons. 

Sub-issue of #1658

cmdupuis3:
Closes #1745

## Overview

_encode_esmf applies the UGRID→ESMF index offset to the whole face_node_connectivity array, padding included, and then encodes elementConn as int32. INT_FILL_VALUE + 1 is -2**63 + 1, and narrowing that to int32 truncates it to 1, so every padded slot of a ragged mesh is written to disk as node 0.

Nothing raises, and the resulting file is structurally valid, so any reader — including uxarray's own — takes those slots for real vertices. Faces gain spurious extra nodes and produce incorrect polygons. This is the case that originally surfaced this family of bugs.

The fix offsets only the valid indices and writes ESMF's own -1 into the padding, which is what the variable's _FillValue attribute already advertises. The test asserts on the encoded array and on what lands on disk after the int32 narrowing, rather than on a round trip: a round trip can hide an encode bug when the reader's inverse offset happens to cancel it.

## PR Checklist

**General**

**Testing & Benchmarking**
- [x] There is adequate test coverage of changes from this PR (add new tests if needed)

- [x] Docstrings updated with any function changes, and included in all new functions

## AI Disclosure

AI Usage: Claude Opus 5

### DIFF
--- a/test/io/test_esmf.py
+++ b/test/io/test_esmf.py
@@ -3,7 +3,7 @@ import os
 import pytest
 import xarray as xr
 import numpy as np
-from uxarray.constants import ERROR_TOLERANCE
+from uxarray.constants import ERROR_TOLERANCE, INT_FILL_VALUE
 
 
 def test_read_esmf(gridpath):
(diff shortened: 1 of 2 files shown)


Sevans711 (changes requested):
Cleanly fixes the problem as scoped here. Thank you for splitting the larger PR into smaller PRs, the changes here were much easier to review!

Small comment, could you update the example in the test suite to avoid overlapping faces? Assuming I didn't make a mistake when drawing it, I think it looks like this:

Image

(face 2 in black overlaps with face 0 in blue and face 1 in orange.)

Also if you wanted to add a note like "regression test for issue 1745" to the new test's docstring that wouldn't hurt (though, it isn't really necessary since the docstring already does a good job explaining the test's purpose).

Happy to approve as soon as the example grid looks valid!

cmdupuis3:
@Sevans711  I was thinking about that too, but that's only true in 2D, right? Like if point 4 was behind all the others, there's no overlap, they just form a ring.

Sevans711:
> I was thinking about that too, but that's only true in 2D, right? Like if point 4 was behind all the others, there's no overlap, they just form a ring.

True, but uxarray grids are always restricted to 2D (see, e.g., #1597), for better or for worse!

cmdupuis3:
@Sevans711 That's not quite what I mean, I'm saying the surface of a sphere is 2D, and this is also on a 2D manifold. I'm not referring to a hypothetical voxel approach.

cmdupuis3:
Coordinates are updates to reflect a geometry that wouldn't break minor arcs, as per discussion.

Sevans711 (approved):
Everything looks good now - approving!

dylannelson (approved):
Tested locally and everything works on my end. On main every padded slot is written as 1 and on this branch every padded slot is -1, leaving every other entry unchanged. Looks good thanks!

PR #1746, opened 2026-09-10
Fix ESMF read casting `elementConn` before checking for padding

### LINKED ISSUE #1747: Fix ESMF read casting `elementConn` before checking for padding
_read_esmf casts elementConn to INT_DTYPE and only then compares against INT_FILL_VALUE, so the padding is recognized in neither of the two ways xarray can hand it over. With CF decoding on (mask_and_scale=True, the default), xarray replaces ESMF's -1 with NaN and promotes the array to float; casting NaN to an integer dtype is platform-dependent, and on arm64 it yields 0, so the padding decodes to -1. With decoding off, the raw -1 survives the cast and decodes to -2. Both are negative indices that silently wrap to the end of the coordinate arrays instead of being treated as fill.

cmdupuis3:
Closes #1747 

## Overview

`_read_esmf` casts `elementConn` to `INT_DTYPE` and only then compares against `INT_FILL_VALUE`, so the padding is recognized in neither of the two ways xarray can hand it over. With CF decoding on (`mask_and_scale=True`, the default), xarray replaces ESMF's -1 with NaN and promotes the array to float; casting NaN to an integer dtype is platform-dependent, and on arm64 it yields 0, so the padding decodes to -1. With decoding off, the raw -1 survives the cast and decodes to -2. Both are negative indices that silently wrap to the end of the coordinate arrays instead of being treated as fill.

The fix locates the padding before the cast, from `numElementConn` — the format's own statement of each face's node count — and additionally treats NaN as padding whatever that count claims.

The test builds the ESMF dataset by hand instead of going through `_encode_esmf`, so it exercises the reader in isolation and covers third-party ESMF files rather than only our own output. It is parametrized over `mask_and_scale` to pin both decoding paths.

## PR Checklist

**General**

**Testing & Benchmarking**
- [ ] There is adequate test coverage of changes from this PR (add new tests if needed)

- [ ] Docstrings updated with any function changes, and included in all new functions
- [ ] User (public) functions added to `docs/api.rst`; internal (private) function names start with an underscore (`_`)

## AI Disclosure

AI Usage: Claude Opus 5

### DIFF
--- a/uxarray/io/_esmf.py
+++ b/uxarray/io/_esmf.py
@@ -93,12 +93,22 @@ def _read_esmf(in_ds):
         # assume start index is 1 if one is not provided
         start_index = 1
 
-    face_node_connectivity = in_ds["elementConn"].astype(INT_DTYPE)
-    face_node_connectivity = xr.where(
-        face_node_connectivity != INT_FILL_VALUE,
-        face_node_connectivity - start_index,
-        face_node_connectivity,
+    element_conn = in_ds["elementConn"]
+    face_dim, node_dim = element_conn.dims
+
+    # "numElementConn" gives the face size, so locate the padding positionally.
+    # Matching the sentinel means guessing: CF decoding turns it into NaN, and the
+    # cast below preserves neither NaN nor the raw value as INT_FILL_VALUE.
+    positions = xr.DataArray(
+        np.arange(element_conn.sizes[node_dim], dtype=INT_DTYPE), dims=node_dim
     )
+    fill_mask = (positions >= n_nodes_per_face).transpose(face_dim, node_dim)
+
+    # NaN is never a usable index, whatever "numElementConn" claims
+    fill_mask = fill_mask | element_conn.isnull()
+
+    face_node_connectivity = element_conn.fillna(0).astype(INT_DTYPE) - start_index
+    face_node_connectivity = xr.where(fill_mask, INT_FILL_VALUE, face_node_connectivity)
 
     out_ds["face_node_connectivity"] = xr.DataArray(
         data=face_node_connectivity,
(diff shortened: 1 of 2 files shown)


uxarray/io/_esmf.py
```diff
@@ -93,12 +93,22 @@ def _read_esmf(in_ds):
...
+    # "numElementConn" gives the face size, so locate the padding positionally.
+    # Matching the sentinel means guessing: CF decoding turns it into NaN, and the
+    # cast below preserves neither NaN nor the raw value as INT_FILL_VALUE.
+    positions = xr.DataArray(
+        np.arange(element_conn.sizes[node_dim], dtype=INT_DTYPE), dims=node_dim
     )
+    fill_mask = (positions >= n_nodes_per_face).transpose(face_dim, node_dim)
+
+    # NaN is never a usable index, whatever "numElementConn" claims
+    fill_mask = fill_mask | element_conn.isnull()
```
rajeeja, line 109: Good find, the platform-dependent NaN cast is nasty, and your test does fail on main both decoded and undecoded, so happy to review properly once it's out of draft. One thing: the mask is positional plus isnull, so if `numElementConn` disagrees with the real padding the undecoded path still turns the raw sentinel into index -2; ORing in the declared `_FillValue` covers that and mid-row padding, and `test/io/` stays green with it.

```suggestion
    # NaN is never a usable index, whatever "numElementConn" claims
    fill_mask = fill_mask | element_conn.isnull()

    # ...and neither is the declared sentinel, whatever "numElementConn" claims
    sentinel = element_conn.attrs.get(
        "_FillValue", element_conn.encoding.get("_FillValue")
    )
    if sentinel is not None:
        fill_mask = fill_mask | (element_conn == sentinel)
```

cmdupuis3: Accepted, new testing block also

PR #1678, opened 2026-08-17
Improve error messages in uxarray/core

Sevans711:
Closes #1677 (sub-issue of #1617)

## Overview
Makes error messages clearer and more informative throughout uxarray/core. As suggested by #1617 this is attempting to be an exhaustive fix for all desired error message improvements in uxarray/core, while touching nothing outside of uxarray/core.

**Tiny expansion of scope:** fixes confusing/misleading error types in a few places missed by #1622:
1. ValueError → TypeError in open_dataset() error related to either passing the wrong number of input args or using the wrong types of objects for inputs. TypeError is more appropriate than ValueError in both of those cases.
2. ValueError → TypeError in UxDataArray.zonal_mean() errors raised when not isinstance(lat, (tuple, float, int, list, np.ndarray)) or when not isinstance(lat, (tuple, list, np.ndarray)).
3. ValueError → TypeError in UxDataArray.from_healpix() error raised when not isinstance(da, xr.DataArray).

Aside from this tiny expansion of scope there should be no functionality changes; the only differences in this PR are the changes to error message contents.

**Additional notes:**
- Test suite updates to error message string matching is a good spot-check that these changes are working; the tests failed until I updated the test suite to match the new strings.
- Intentionally did not touch error messages in concat() because fixing bugs there might lead to needing different kinds of messages anyway (see #1642).
- Ran benchmarks to make sure there isn't any performance hit. (Highly unlikely, but could happen if errors are used as part of control flow in a hot loop and any objects now placed into error message strings take non-negligible time to convert to string.) (Removed run-benchmarks label after seeing there is no performance hit.)

## PR Checklist

**General**
- [X] An issue is created and linked
- [X] Added appropriate labels (if your uxarray repo permissions allow it)
- [X] Filled out Overview and Expected Usage (if applicable) sections

**Testing & Benchmarking**

- [N/A] Adequate tests are created if there is new functionality
- [X] Tests are not too basic (such as simply calling a function and nothing else)
- [N/A] Tests cover all major paths in your new functions
- [x] If this PR could affect performance, ran ASV benchmarks and confirmed they show expected behavior (add a new benchmark if necessary)

## AI Disclosure

AI Usage: GitHub Copilot's inline code suggestions, some tiny chats with Claude about error message style.

- [X] I take responsibility for all AI-generated content in my PR.
- [X] I have tested all AI-generated content in my PR.

PR #1749, opened 2026-09-10
Remove UxDataArray.data_location

### LINKED ISSUE #1748: remove data_location
https://github.com/UXARRAY/uxarray/pull/1530#issuecomment-4928877708

fixing this

rajeeja:
Closes #1748. `data_mapping` already returns the same information and is what `plot.points()` uses, so this drops `data_location`, moves its docstring onto `data_mapping`, and updates the two error messages that referenced it. Note `data_location` shipped in v2026.07.0 through v2026.09.0.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -257,7 +257,6 @@ Grid Accessor
 
    UxDataArray.uxgrid
    UxDataArray.data_mapping
-   UxDataArray.data_location
 
 
 UxDataset
--- a/test/core/test_dataarray.py
+++ b/test/core/test_dataarray.py
@@ -138,8 +138,8 @@ def test_isel_invalid_dim(gridpath, datasetpath):
         uxda.isel(level=0)
 
 
-def test_data_location():
-    """Tests data_location for face/node/edge centered data and non-grid data."""
+def test_data_mapping():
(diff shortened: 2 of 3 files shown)


Sevans711 (approved):
Looks like a clean fix to me! Removes all instances of `data_location` and sets a reasonable precedent for how to rephrase error messages that previously contained `data_location`.

Slightly more context: after some internal discussions regarding the PR and comment linked above, it was decided that `data_location` should be removed because it was fully redundant with `data_mapping` and was introduced much more recently (only a few months instead of almost 2 years). It should be okay to proceed without a deprecation cycle because it was introduced so recently, and we don't think that any users will have started using it yet (any users who needed this functionality would most likely have been using `data_mapping` already).

Sidenote: #1678 will require some substantive updates after this merges, because it added lots of new error message text containing `data_location`, but it can just follow the pattern introduced here using `data_mapping` instead.

dylannelson (approved):
Looks good on my end. Checked locally and `data_mapping` returns results as expected with `data_location` no longer being accessible. Didn't find any concerns or functions dependent upon it so I think we should be good to proceed. Should be worth noting the change in the release notes so it can be easily found by searching, especially if a bot/tool/person picked up on it during it's training in this brief time by some chance.

PR #1751, opened 2026-09-10
Fix _encode_ugrid mutating BASE_GRID_TOPOLOGY_ATTRS

### LINKED ISSUE #1750: _encode_ugrid mutates BASE_GRID_TOPOLOGY_ATTRS
`uxarray/io/_ugrid.py:95` binds the module-level dict without copying and then adds keys to it, so every `to_xarray("UGRID")` / `to_netcdf` call in a process permanently grows the shared constant. Any grid encoded afterwards advertises variables and dimensions it does not have.

```python
import numpy as np, uxarray as ux
from uxarray.conventions import ugrid

hpgrid = ux.Grid.from_healpix(zoom=1)
hpgrid.edge_lon; hpgrid.face_edge_connectivity
hpgrid.to_xarray("UGRID")
print(len(ugrid.BASE_GRID_TOPOLOGY_ATTRS))  # 11, was 6

minimal = ux.Grid.from_topology(
    node_lon=np.array([0., 1., 1., 0.]),
    node_lat=np.array([0., 0., 1., 1.]),
    face_node_connectivity=np.array([[0, 1, 2, 3]]),
)
ds = minimal.to_xarray("UGRID")
print("n_edge" in ds.dims, ds.grid_topology.attrs["edge_dimension"])  # False n_edge
```

- Output is non-compliant UGRID: `grid_topology` points at `n_edge`, `face_lon face_lat`, `face_edge_connectivity` etc. that are not in the dataset.
- Order-dependent, so invisible in single-grid scripts and tests; shows up in notebooks, batch conversions, and long-lived servers.
- The `.copy()` was added in #1279 and removed in #1282 (https://github.com/UXARRAY/uxarray/pull/1282#discussion_r2159785902) on the reasoning that the added keys were not in the base dict. They are not, until the first call adds them.

Fix is simply `grid_topology = dict(ugrid.BASE_GRID_TOPOLOGY_ATTRS)` plus a regression test.

rajeeja:
Closes #1750. `_encode_ugrid` added keys directly to the shared module-level dict, so after the first export every later grid's `grid_topology` advertised edge/face variables it did not have. Copies the dict before mutating and adds a regression test that fails on `main`.

### DIFF
--- a/test/io/test_ugrid.py
+++ b/test/io/test_ugrid.py
@@ -90,3 +90,34 @@ def test_encode_ugrid_copies_and_converts_bool_attr():
     assert encoded.attrs["test_str"] == "abc"
     # Check that the original dataset is not modified
     assert isinstance(ds.attrs["test_bool"], bool)
+
+
+def test_encode_ugrid_does_not_mutate_base_topology_attrs():
+    """Encoding a grid with optional variables must not leak their
+    grid_topology attributes into ugrid.BASE_GRID_TOPOLOGY_ATTRS, which would
+    then be wrongly advertised by every later grid encoded in the process."""
+    import numpy as np
+    from uxarray.conventions import ugrid
+
+    base = dict(ugrid.BASE_GRID_TOPOLOGY_ATTRS)
+
+    # healpix grid: has edges, face/edge coordinates and extra connectivity
+    hpgrid = ux.Grid.from_healpix(zoom=1)
+    hpgrid.edge_lon
+    hpgrid.face_edge_connectivity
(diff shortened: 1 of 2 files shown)


dylannelson (approved):
Ran locally and ran tests after creating a new environment. Not seeing any blockers or downstream affects

Sevans711 (approved):
Reproduced issue using example from original issue report on main, but looks fixed on this branch. Double-checked that the new regression test does indeed crash on `main`. Clean fix, looks good, let's merge to main!

PR #1755, opened 2026-09-11
Release the GIL in all numba kernels (nogil=True)

### LINKED ISSUE #1756: GIL blocks parallelism for many kernels
Currently, only two sites out of about 140 `njit` kernels set `nogil=True`, which means that most of them are running serially. 

As part of the efforts to make the package as a whole chunk-parallelizeable, we should have `nogil=True` most of the time.

cmdupuis3:
Closes #1756

## Overview

Only 2 of the 140 @njit sites in the package passed nogil=True (grid/connectivity.py, grid/area.py). Every other compiled kernel held the GIL for its whole execution, so calling them from dask's threaded scheduler as dask="parallelized" does serialized them completely, regardless of how many workers were configured.

Adds nogil=True to the 20 or so remaining top-level sites making @njit(cache=True, nogil=True) the default pattern. These are all nopython, allocation-light leaf kernels with no object-mode fallback.

Measured on _construct_edge_node_distances (2M edges, best of 3), the same kernel jitted with and without nogil:

    threads     GIL held        nogil   speedup
          1        95.0ms        95.3ms    1.00x
          2       191.5ms       101.1ms    1.89x
          4       381.0ms       104.1ms    3.66x
          8       762.8ms       115.1ms    6.63x

The GIL-held column scales linearly with thread count -- that is the signature of zero parallelism.

Two sites are deliberately left alone:

  * the guvectorize kernel in grid/neighbors.py -- numba rejects nogil as a @guvectorize option, and the gufunc machinery already drops the GIL around the loop.
  * the 15 parallel=True kernels keep parallel=True for now. Nested under dask's threadpool it oversubscribes; dropping it belongs with the work that gives dask the chunk loop.

_build_node_edge_connectivity in grid/connectivity.py is also the one kernel still missing cache=True; left as-is to keep this diff to nogil.

The `nogil_scaling.GILScaling.track_gil_scaling` is added to prove that `nogil` is working as expected.

## PR Checklist

**General**

**Testing & Benchmarking**

## AI Disclosure

AI Usage: Claude Opus 5

### DIFF
--- a/uxarray/core/gradient.py
+++ b/uxarray/core/gradient.py
@@ -350,7 +350,7 @@ def _dual_cell_area(sx, sy, sz, angles, n):
     return np.abs(area)
 
 
-@njit(cache=True, parallel=True)
+@njit(cache=True, parallel=True, nogil=True)
 def _compute_gradients_on_faces(
     data,
     n_face,
--- a/uxarray/grid/angles.py
+++ b/uxarray/grid/angles.py
@@ -8,7 +8,7 @@ from numba import njit, prange
 from uxarray.grid.utils import _numba_norm3, _small_angle_of_2_vectors
 
 
-@njit(cache=True, parallel=True)
+@njit(cache=True, parallel=True, nogil=True)
 def _compute_face_node_angles_convex(
     node_x,
     node_y,
--- a/uxarray/grid/bounds.py
+++ b/uxarray/grid/bounds.py
@@ -134,7 +134,7 @@ def _populate_face_bounds(
         grid._ds["bounds"] = bounds_da
 
 
-@njit(cache=True, parallel=True)
+@njit(cache=True, parallel=True, nogil=True)
 def _construct_face_bounds_array(
     face_node_connectivity,
     n_nodes_per_face,
--- a/uxarray/grid/connectivity.py
+++ b/uxarray/grid/connectivity.py
@@ -302,7 +302,7 @@ def _emit_bucket_edges(
         face_edge_flat[half_edge_slot[i]] = edge_idx
 
 
-@njit(cache=True, parallel=True)
+@njit(cache=True, parallel=True, nogil=True)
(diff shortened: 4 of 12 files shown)


Sevans711 (approved):
Looks ready to merge! Simple changes which should help with future scalability work, and shouldn't hurt in the meantime (no regressions in benchmarks). The new benchmark is a good spot check to help prevent regression, and demonstrates clearly that the nogil=True flag can cause improvements. Approving as-is!

Sevans711:
(When I glanced at it earlier today there were no regressions in benchmarks; though now I do see tiny apparent regressions for sub-millisecond connectivity benchmarks time_face_node('120km') and time_n_nodes_per_face('120km'). I suppose that means this is probably just noise, rather than any real performance degradation. Keeping this marked as approved!)

rajeeja:
This looks good overall. One small suggestion before merging: could we make the title a little more precise, maybe **“Release the GIL in Python-entry-point Numba kernels”**? The current title says “all Numba kernels,” but the PR intentionally (and correctly) leaves out small internal kernels, gufuncs, and places where `nogil=True` would add overhead or is not supported.

Could we also open a follow-up issue for the remaining end-to-end Dask work? It would be useful to track a real user-facing chunked workflow, measure scaling on representative datasets, and look at how Dask threads interact with Numba’s `parallel=True` kernels—especially possible oversubscription and whether we need thread-limit or scheduler guidance. Once that workflow exists, we could supplement or replace the synthetic GIL benchmark with an end-to-end benchmark.

The benchmark here is a good regression check for whether the GIL is actually released. The follow-up would tell us how much that helps real UXarray workflows and what the next bottlenecks are. None of this needs to block the implementation itself.

rajeeja (approved):
just change title and later the issue as in previous comment

cmdupuis3:
@rajeeja Yeah, I'll make an overarching issue for the end-to-end chunked work and add this one as a sub-issue

PR #1761, opened 2026-09-15
Ship a PEP 561 py.typed marker

### LINKED ISSUE #1760: Ship a PEP 561 py.typed marker

uxarray ships inline type annotations but no PEP 561 `py.typed` marker, so mypy and pyright discard all of them and `--strict` users get an error on `import uxarray`. `Grid.n_face` is already declared `-> int` (uxarray/grid/grid.py:897), but without the marker no checker is permitted to use it, so every annotation written before the marker exists is inert. `MANIFEST.in` globs only `*.py`, so the marker needs explicit packaging config to reach the wheel; it is absent from `uxarray-2026.9.0-py3-none-any.whl` today.

Done when `py.typed` is in the built wheel and a test guards it. Follow-up: the marker is a promise, and coverage is 213/798 functions with return annotations — 227 public functions have none, including `open_grid`, `open_dataset`, and `open_mfdataset`.

rajeeja:
Closes #1760

## Overview

Adds the PEP 561 `py.typed` marker so type checkers are permitted to use the annotations uxarray already ships.

Without the marker, mypy and pyright treat every uxarray import as `Any` and discard all existing annotations, and `--strict` users get an error on `import uxarray` itself. `Grid.n_face` is declared `-> int` (uxarray/grid/grid.py:897) and that annotation does nothing for a user today — the mechanism that would consume it is simply absent. xarray, numpy and dask all ship the marker.

Two config changes are needed rather than one, because `MANIFEST.in` line 4 is `recursive-include uxarray *.py`: a non-`.py` file is silently dropped from the distribution even when it is present in git. `[tool.setuptools.package-data]` covers the wheel, `MANIFEST.in` covers the sdist.

## Changes

- `uxarray/py.typed` — new, empty
- `pyproject.toml` — `[tool.setuptools.package-data]`
- `MANIFEST.in` — `include uxarray/py.typed`
- `test/test_dependencies.py` — assert the marker is present in the *installed* package

## Verification

Built the wheel and sdist from this branch:

- `py.typed` present in both
- wheel contains 86 files vs 85 in the released 2026.9.0 — diff against that wheel shows exactly one file added, none removed
- installed the wheel non-editable into a clean venv, confirmed the marker lands in site-packages
- deleted the marker from that install and confirmed the new test fails, so it is not a tautology

The test asserts against the installed package rather than the source tree, since the source tree cannot catch a packaging regression.

## Follow-up (not in this PR)

The marker is a promise, and coverage is currently thin: 213/798 functions have return annotations, and 227 public functions have none — including `open_grid`, `open_dataset` and `open_mfdataset`. So `ux.open_grid("mesh.nc")` still infers as `Any` and a downstream `grid.n_face + "x"` still goes uncaught. Annotating those constructors is the natural next step, and annotating `open_grid` first requires a decision about `return_chunks`, which is undocumented, untested, and used by exactly one private caller. Worth doing separately.

## PR Checklist

**General**

**Testing**
- [x] Tests added
- [N/A] Benchmarks added

**Documentation**
- [N/A] Docstrings on new functions
- [N/A] Examples added


test/test_dependencies.py
```diff
@@ -68,4 +69,27 @@ def test_no_numba_kernels_built_on_import():
...
+    every uxarray import as ``Any``, and ``--strict`` users get an error on
+    ``import uxarray``.
+
+    This asserts against the *installed* package rather than the source tree:
+    ``MANIFEST.in`` globs only ``*.py``, so the marker can be present in git
+    and still be missing from a built wheel.
+    """
+    import uxarray
+
+    marker = Path(uxarray.__file__).parent / "py.typed"
```
cmdupuis3, line 87: I think this path depends on how pytest is invoked. Invoking with `pytest test` does the right thing, but with `python -m pytest test` it will use the source code here. One option could be using a subprocess to spawn a new interpreter to guarantee it's using the install path.

pyproject.toml
```diff
@@ -79,6 +79,11 @@ requires = [
 [tool.setuptools]
 packages = ["uxarray"]
 
+[tool.setuptools.package-data]
+# PEP 561 marker. Without this, setuptools ships only .py files and type
```
cmdupuis3, line 83: > Without this, setuptools ships only .py files

Is that actually true?

cmdupuis3 (changes requested):

PR #1765, opened 2026-09-15
More robust optional import linting for scoped imports

cmdupuis3:
A lot of possible cases where optional import linting is needed are missed because it isn't fully walking the block scopes in the AST. These changes should help with that.


Sevans711:
Thank you for these improvements! They look good, I did a few different sanity checks that things work well. While doing the checks I noticed that the error message from pytest was somewhat confusing when detecting module-level imports of optional dependencies. I added a commit to split the module-level import detection into its own functions and test, with a clearer message that module names and line numbers. I'm happy for this to merge into #1656 if it looks good to you!

PR #1767, opened 2026-09-16
Fixes math bugs by creating UxSupportsArithmetic class (shared parent of UxDataset and UxDataArray)

### LINKED ISSUE #1685: Uxarray types dropped during ufuncs
I expected uxarray to fully support ufuncs without losing object types. For example, `np.log10(uxarr)` should return a `UxDataArray` but it returns an `xr.DataArray` instead.

Here's a decent example to show the bug:
```python
import uxarray as ux
ds = ux.tutorial.open_dataset('quad-hexagon')
arr = ds['t2m']   # arr is a UxDataArray
result0 = np.sqrt(arr)
result0.uxgrid   # crashes because result0 is an xr.DataArray
result1 = np.cos(ds)
result1.uxgrid   # crashes because result1 is an xr.Dataset
```

I believe the fix would just involve overriding `__apply_ufunc__` appropriately, for UxDataArray and UxDataset.

### LINKED ISSUE #1695: Uxarray types dropped during math with xarrays, depending on order of operations
### Version

2026.8.0

Source

### What happened?

Uxarray object typing gets lost during binary math operations with xarray objects when the xarray object appears first.

I expected that mathematical calculations including uxarray objects and xarray objects should give uxarray object results, regardless of which kind of object appears earlier.

```python
import uxarray as ux
import xarray as xr
uxarr = ux.tutorial.open_dataset('quad-hexagon')['t2m']

# expecting UxDataArray but got xarray.DataArray:
xr.DataArray(7) * uxarr

# meanwhile, putting the uxarr object first correctly gives UxDataArray:
uxarr * xr.DataArray(7)

# (this is extra confusing because the two are mathematically equivalent...)
```

### LINKED ISSUE #1718: Math with multiple uxarray objects with different grids silently succeeds, when data dims are otherwise compatible

### LINKED ISSUE #1737: UxDataArray.astype() should return UxDataArray, not xarray.DataArray
### Version

2026.08.1

Source

### What happened?

(Bug discovered by @erogluorhan) Calling UxDataArray.astype(...) returns an xarray.DataArray.

It should return a UxDataArray instead.

```python
import uxarray as ux
uxarr = ux.tutorial.open_dataset('quad-hexagon')['t2m']
result = uxarr.astype('float64')
type(result)  # expected UxDataArray; got xarray.DataArray.
```

Sevans711:
Closes #1685, Closes #1695, Closes #1718
Bonus: closes #1737 

## Overview

This PR introduces the UxSupportsArithmetic class as a shared parent class for both UxDataset and UxDataArray. This class partially serves a similar purpose as xarray's SupportsArithmetic class (defining `__array_ufunc__`) but also serves a broader purpose too (override `_binary_op()` to also compare grids, and override math dunder methods like `__add__`). This is the majority of the solution for #1685, #1695, and #1718.

It also solves #1737 (because it was simple enough to fix and I wanted to use `.astype()` in the test suite here).

Mentioning also one thing I was surprised by: grids with n_face==1 needed to be handled specially in this PR to avoid considering them for grid comparisons, because doing so would break simple use-cases like `arr.isel(n_face=0) > arr.isel(n_face=7)`.

### Bonus details:

**Mapping changes to solved issues:**
- 1685 is solved simply by appropriately defining `__array_ufunc__`.
- 1695 is solved (as much as possible) by explicitly overriding dunder methods like `__add__`, as well as overriding `_binary_op()` UxDataArray to check one particular case. It isn't possible to fully solve for cases combining DataArrays and Datasets, at least not without tinkering with xarray directly, as noted in the test suite.
- 1718 is solved by adding a grid comparison check into `_binary_op()`, and by comparing uxgrids of inputs during `__array_ufunc__`, whenever multiple inputs have a uxgrid.
- 1737 is solved simply by defining astype() for both UxDataArray and UxDataset.

**This PR provides more comprehensive solutions than what the issue examples cover,** but it otherwise just solves the original issues (no expansions of scope). In particular:
- 1685's example only showed issues with single-input ufuncs (like np.sqrt and np.log10); this PR also fixes the issue for multi-input ufuncs (like np.add and np.minimum).
- 1695's example only showed issues when doing operations with xr.DataArray and UxDataArray; this PR also fixes the issue (as much as possible) for operations with Datasets and for operations combining DataArrays and Datasets.
- 1718's example only shows binary operations (like A + B); this PR also fixes the issue for any multi-input ufuncs (like np.add(A, B))
- The examples above mainly considered arithmetic operations; this PR also fixes the issue for comparisons, too (like A < B).

I doubt this PR would affect performance, but I added the run-benchmark label just in case. (Benchmarks look fine --> removing the label)

## PR Checklist

**General**

**Testing & Benchmarking**

## AI Disclosure

AI Usage: Lots of chatting with Claude to settle on the overall design and to speed up debugging. Also, used GitHub Copilot's inline code suggestions.


uxarray/core/arithmetic.py
```diff
@@ -0,0 +1,242 @@
...
+            if (
+                (self.uxgrid.n_face > 1)
+                and (other.uxgrid.n_face > 1)
+                and (self.uxgrid != other.uxgrid)
+            ):
+                raise GridsMismatchError(
+                    f"A.uxgrid != B.uxgrid during binary operation {f.__name__!r}, "
+                    f"with type(A)={type(self).__name__}, type(B)={type(other).__name__}."
+                )
+        return super()._binary_op(other, f, reflexive=reflexive, **kw_super)
```
rajeeja: xarray routes `+=` through `_inplace_binary_op`, which this class doesn't override, so the grid check never runs for in-place ops. On same-shape arrays with different grids `a + b` raises `GridsMismatchError` but `c += b` succeeds silently. Could you pull the check into a helper, call it from both, and add a `+=` test?

```python
    def _raise_if_grids_incompatible(self, other, f):
        if isinstance(other, UxSupportsArithmetic):
            if (
                (self.uxgrid.n_face > 1)
                and (other.uxgrid.n_face > 1)
                and (self.uxgrid != other.uxgrid)
            ):
                raise GridsMismatchError(
                    f"A.uxgrid != B.uxgrid during binary operation {f.__name__!r}, "
                    f"with type(A)={type(self).__name__}, type(B)={type(other).__name__}."
                )

    def _binary_op(self, other, f, reflexive=False, **kw_super):
        self._raise_if_grids_incompatible(other, f)
        return super()._binary_op(other, f, reflexive=reflexive, **kw_super)

    def _inplace_binary_op(self, other, f):
        self._raise_if_grids_incompatible(other, f)
        return super()._inplace_binary_op(other, f)
```

uxarray/core/arithmetic.py
```diff
@@ -0,0 +1,242 @@
...
+        result = super().__array_ufunc__(ufunc, method, *inputs, **kwargs)
+
+        # convert any xr.DataArray / xr.Dataset outputs to appropriate uxarray type:
+        _was_tuple = isinstance(result, tuple)
+        result = list(result) if _was_tuple else [result]
+        for i, obj in enumerate(result):
+            if isinstance(obj, xr.DataArray):
+                result[i] = UxDataArray(obj, uxgrid=grid_ref)
+            elif isinstance(obj, xr.Dataset):
+                result[i] = UxDataset(obj, uxgrid=grid_ref)
```
rajeeja, line 155: This drops `source_datasets`, so a ufunc on a UxDataset resets provenance to `None`, while `dataset.py:143`, `:211` and `:254` all forward it. `obj` here is a plain `xr.Dataset` so it has none to carry, which makes it worth deciding whether it should come from `self` or from whichever input supplied `grid_ref`.

uxarray/core/dataset.py
```diff
@@ -892,6 +893,16 @@ def to_xarray(self, grid_format: str = "UGRID") -> xr.Dataset:
...
 
+    def astype(self, dtype, **kw_super):
+        """Copy of this uxarray object, with data cast to a specified type.
+        Leaves coordinate dtype unchanged.
+
+        Behaves just like :meth:`xarray.Dataset.astype`, except that
+        the returned object is a UxDataArray with same uxgrid as the input.
+        """
+        da = super().astype(dtype, **kw_super)
+        return type(self)(da, uxgrid=self._uxgrid)
```
rajeeja, line 904: Same dropped `source_datasets` here, and the docstring above says "returns a UxDataArray" where it means UxDataset.

```suggestion
        return type(self)(da, uxgrid=self._uxgrid, source_datasets=self.source_datasets)
```

rajeeja (changes requested):
Traced the dispatch and the ordering and type-preservation fixes hold up, and the grid check does fire for out-of-place ops. But `+=` bypasses it completely, so #1718 is still reproducible in place. Two smaller things inline.

uxarray/core/arithmetic.py
Sevans711: I would prefer to keep `source_datasets` support out of this PR if possible. Does that sound reasonable to you?

Reasoning: `source_datasets` isn't really fully supported throughout uxarray, for example even something as basic as `obj.isel(n_face=slice(None))` also resets it to None. There are also no tests in the test suite which actually check whether `source_datasets` was properly maintained after any operations. Also, it is much easier to add a feature later than to remove support for it.

I also asked @erogluorhan a few weeks ago about `source_datasets`, and his impression was that it is a very low priority, we shouldn't worry about touching it or adding more support for it right now, but we could reassess upon receiving any user feedback about it.

uxarray/core/dataset.py
Sevans711: Good catch with the docstring - fixed!

Did not add source_datasets support yet (see reasoning above).

uxarray/core/arithmetic.py
Sevans711: Done! The test is named `test_inplace_binary_ops_check_grid_compatibility`.

Also slightly improved the GridsMismatchError message here to include uxgrid.sizes information for self and other.

PR #1768, opened 2026-09-16
Neighborhood filter kernels caching/race condition fixes

### LINKED ISSUE #1757: Race condition crashes pytest suite locally sometimes
The pytest suite fails locally for me with SIGABRT _sometimes_, ever since #941 merged. It doesn't happen every time, but it does happen very often (sometimes it feels like it crashes 20% of the time, sometimes it feels like it crashes >90% of the time). When the crash occurs, it occurs inside test/core/test_dataarray.py, inside the first test which uses dask with the neighborhood guvectorize kernels (the test is named `TestNeighborhood.test_dask_input_stays_lazy`).

Some back-and-forth with Claude led me to believe this is likely a race condition related to numba/numba#9288. I also managed to get things down to a reasonably minimal reproducer. Run the following lines from a terminal window (I'm on macOS) with an active Python venv:

```zsh
for i in $(seq 1 20); do
  python -c "
import faulthandler; faulthandler.enable()
import xarray as xr, uxarray as ux
vortex = ux.tutorial.open_dataset('outCSne30-vortex')
vortex.neighborhood(r=2.0).mean()
times = xr.DataArray([0,1,2,3,4,5], dims='time')
chunked = (vortex + 0*times).chunk({'time': 2, 'n_face': -1})
chunked.neighborhood(r=2.0).mean().compute()
print('a run completed successfully')
"
  code=$?
  echo "run $i: exit $code"
  [ $code -ne 0 ] && break
done
```

On my end, I am seeing crashes within a few runs, e.g. it will print something like:
```
a run completed successfully
run 1: exit 0
a run completed successfully
run 2: exit 0
a run completed successfully
run 3: exit 0
Numba workqueue threading layer is terminating: Concurrent access has been detected.
...  # more error message details
Fatal Python error: Aborted
...  # more error message details
run 4: exit 134
```

In case it matters, I set up my environments using uv. I checked that the example above causes crashes within a few runs on Python 3.11, 3.13, and 3.14. My macOS is Tahoe 26.6.2. It feels like it is failing more often since upgrading to Tahoe, but that might just be confirmation bias. Before upgrading OS I was able to run the pytest suite locally within a few tries. Today is the first time I tried to run the suite since upgrading, and it crashed every time across more than 5 attempts, which is why I looked into this further.

cc: @cmdupuis3

### LINKED ISSUE #1769: Neighborhood filter kernels don't actually cache
Neighborhood filter kernels are currently compiling lazily, but they are not being cached, because numba has a problem caching closures https://github.com/numba/numba/issues/6264.

An alternate approach to neighborhood filter kernels is needed, since they are evidently very sensitive to the exact design.

cmdupuis3:
Closes #1757, #1769

## Overview

The PR solves two issues with the current version of the neighborhood filters kernels:

1. numba has issues caching closures that are also decorated as `njit(cache=True)`. This means that in the current implementation of kernel caching, every call to a kernel gets a new UUID, so any attempt to cache this way will basically be a no-op.
2. Somewhat related, but there is also a race condition with dask. See #1757.

In a fun twist, both are known unresolved issues in numba: 

https://github.com/numba/numba/issues/6264
https://github.com/numba/numba/issues/9288

## PR Checklist

**General**

**Testing & Benchmarking**
- [x] There is adequate test coverage of changes from this PR (add new tests if needed)

- [x] Docstrings updated with any function changes, and included in all new functions
- [x] User (public) functions added to `docs/api.rst`; internal (private) function names start with an underscore (`_`)

## AI Disclosure

AI Usage: Claude Opus 5


Sevans711 (commented):
The fix looks clean, even though it might look like there are some "object-oriented shenanigans" happening here with making _LazyKernel class a descriptor (i.e., it has a `__get__` method), it makes sense to me and the comments communicate the intent clearly, so it shouldn't be too difficult to maintain in the future if needed. I think this is a good way to proceed.

Noticed the small slowdowns (~20%) in some neighborhood benchmarks, but I think that is acceptable if this PR is solving the race condition and/or lack of caching (which slows down real workflows the first time it hits but isn't necessarily appearing in the benchmark suite).

The only issue is, I still have the crash from the race condition (#1757)! Copy-pasting those lines into terminal while on this PR's branch, it still crashes locally for me within a few runs.

If you want to merge this as closing #1769 and related to 1757 but _not_ closing 1757, I would be okay to approve. But for closing 1757 something else needs to be fixed. I'm not sure what is missing (race conditions are tricky to debug…) but I wonder if it might be as simple as moving the threading.Lock() mechanism (not sure where it would need to move to, though…)?

PR #1771, opened 2026-09-17
Fix KeyError: 'lon_0' on cartopy 0.26 (breaks main and all open PRs)

### LINKED ISSUE #1702: 🤖 Upstream testing failure (2026-09-17)
# Upstream testing failure

**Workflow:** CI Upstream
**Run:** [35177708734](https://github.com/UXARRAY/uxarray/actions/runs/35177708734)
**Date:** Thu Sep 17 03:27:31 UTC 2026

Upstream testing has failed.

Please check the [workflow run](https://github.com/UXARRAY/uxarray/actions/runs/35177708734) for details.

This issue was automatically generated from the CI Upstream workflow.

rajeeja:
Closes #1702

**This breaks `main` and every open PR.** cartopy 0.26.0 reached conda-forge today, so CI is red everywhere until this lands, and the failure has nothing to do with whatever a given PR changed.

cartopy 0.26.0 changed `PlateCarree` from `proj=eqc` to `proj=latlong`, which carries the prime meridian as `pm` and has no `lon_0`:

```
0.25.0:  {'proj': 'eqc',     'lon_0': 0.0, ...}
0.26.0:  {'proj': 'latlong', 'pm': 0.0, ...}     # no lon_0
```

`PlateCarree()` is the default projection, so every plotting call that did not pass an explicit one raised `KeyError: 'lon_0'`. That is `test_plot.py::test_topology` failing on all 12 platform/version jobs. The upstream-dev workflow caught it ~14 hours before the package propagated (#1702).

`_central_longitude_of()` reads `lon_0`, then `pm`, then falls back to `0.0`, and replaces all four call sites (two in `plot/accessor.py`, two in `grid/geometry.py`). Works on both cartopy versions, so no pin is needed.

Split out of #1692 so it is not blocked behind that PR's review.

Note: `pm` is the prime meridian, not strictly the central longitude. They coincide for how cartopy builds `latlong` here, but that is the assumption to revisit if cartopy changes again.

## PR Checklist

**General**
- [X] Add appropriate labels

**Testing**


dylannelson:
@rajeeja I ran into this same issue this morning and looked into and and came up with the same culprit, but I don't think this closes #1702. While this will fix the current issue you see in the #1702 logs, I don't think it will close the issue that was there before it, and is further downstream I believe. If you look at the logs from before yesterday, like this one
[here](https://github.com/UXARRAY/uxarray/actions/runs/32917649623/job/98024605428#step:5:471) you have a very different issue that never got resolved. You can close #1702 if you want, but unless both errors are solved at the same time by some chance, I think it will just make a new upstream issue tomorrow, which won't have the history I was working on and waiting to get resolved. Up to you though, maybe both got solved today but we won't know till this patch reaches main and CI is run again. Would still like to get this through asap so we can go back to getting the tests working again. Tested locally and his works on my end

Sevans711 (approved):
Looks good, and I confirmed the new tests crash on `main` when using cartopy 0.26.0 but succeed on this branch. (Not sure if the last test was necessary though, because `test_topology` already noticed this crash, and this PR fixes `test_topology`. But, I think the extra test doesn't hurt!)

Not sure if this will actually close the linked issue (I suspect it might not, because the discussion in that thread identified something else entirely as the main cause… cc: @dylannelson), so maybe it would make sense to unlink that issue (swap to "related to #XXX" instead of "closes XXX")? But, even if that issue is closed incorrectly, it wouldn't be a huge deal because the bot would just make a new issue report for it tomorrow. So, this isn't blocking; feel free to merge as-is, whatever you prefer here!

dylannelson (approved):

Sevans711:
Hah, woops, I was apparently typing out something very similar, and didn't see @dylannelson's message before sending mine!

Issue #1738, opened 2026-09-08
Scrip edge and node order changes each time the grid is loaded


Sevans711:
### Version

v2026.08.01

### How did you install UXarray?

Source

### What happened?

Loading a grid from a SCRIP source leads to a different order for nodes and edges each time the grid is loaded.

I noticed this while exploring MESACLIP data on casper, but I can also see that the bug occurs in the SCRIP example in uxarray.tutorial.

### What did you expect to happen?

I expected the node and edge order to be based solely on the file, and be the same each time it is loaded, not to change in a seemingly-random way when the file is loaded multiple times.

### Can you provide a MCVE to repoduce the bug?

```python
gridA = ux.tutorial.open_grid('scrip-ne30pg2')
gridB = ux.tutorial.open_grid('scrip-ne30pg2')
gridA == gridB   # expected True, got False.
# Same bug if using 'scrip-outCSne8' instead.

# Also can see it by repeatedly running the following two lines in a jupyter cell:
arr = ux.tutorial.open_grid('scrip-ne30pg2')['RELHUM']
arr.isel(lev=0, n_node=slice(None, 100)).plot()
# The plot will change each time, because the node order has changed.
# Similar issue if using n_edge=slice(None, 100).

# If using n_face=slice(None, 100), the plot does not change
# (faces' locations and data remain correct).

# If not slicing at all, the plot does not change,
# because arr.plot() is not sensitive to node/edge order.
```

PR #1775, opened 2026-09-18
Dask-native SCRIP corner dedup when opened with `chunks=`

### LINKED ISSUE #1774: `open_grid` on large SCRIP grids OOMs: corner dedup ignores `chunks=`
- SCRIP stores one (lon, lat) per face-corner, so the corner table is n_face × n_corners rows.
- `_to_ugrid` dedups it with Polars, which needs the whole table resident: 53.6 GiB of float64 for a 300M-element np4 grid, before the join copy.
- The reader never dispatches on `chunks=`, so large SCRIP files OOM no matter how they're opened.
- This blocks the 63 GB CONUS-RRM np4 grid served through the [uxarray-mcp-server](https://github.com/UXARRAY/uxarray-mcp-server), which already passes `chunks="auto"` and still OOMs on the worker.

rajeeja:
Closes #1774

## Overview

Dedups with a dask shuffle and builds the inverse index lazily via `map_blocks` when corners are dask-backed; eager path unchanged. A 24 GB grid that killed eager opens in ~60 s, and the dedup stage runs on the 63 GB np4 grid in 135 s at 38 GiB peak. Unblocks the MCP server path with no server change.

## Expected Usage

```python
grid = ux.open_grid("huge_scrip.nc", chunks="auto")
```

**General**
- [ ] Added appropriate labels (if your uxarray repo permissions allow it)

**Testing & Benchmarking**

- [N/A] User (public) functions added to `docs/api.rst`; internal (private) function names start with an underscore (`_`)

## AI Disclosure

AI Usage:


Sevans711:
A few quick questions, before I take a closer look:

1. ~You reported roughly 2x slowdown in the eager case (4M faces). Of course, the memory usage is almost 4x smaller, which is great. Is this an acceptable tradeoff? Also, can you comment on why these changes improve memory but not speed (is there some fundamental reason why it should be slower, in order to use less memory)?~
    - Actually, I saw that this is clarified via comments in your changes to `_dedup_scrip_nodes_eager`. No need to clarify further here. Keeping my original thoughts here though because it wasn't clear to me just from reading the initial PR message.
2. Node numbering on `main` currently differs _every time_ the grid is loaded, even if it is the same exact grid file (see #1738). Does this PR happen to fix that issue as well? (This also means there is no need to check during review whether node ordering here matches any "canonical" node ordering, because `main` does not guarantee any particular order.)
3. It feels a bit strange to me to have node order depend on the `chunks` parameter. I generally would expect `chunks=True` to affect efficiency, but to still produce the same exact answer if I call compute() at the end (assuming it can fit into memory). There's currently no guaranteed order at all so this could be fine, but what are your thoughts on this?

rajeeja:
Fair point—I thought about this while working on the dedup paths. The eager path should now be deterministic because it uses a lexicographic sort, so it likely fixes #1738 for non-chunked opens. The Dask path may produce a different node order because of the shuffle.

I don’t think eager and chunked loading necessarily need identical node numbering as long as both produce the same mesh with correctly remapped connectivity. That said, consistent ordering would be nice. I’ll measure the memory and runtime overhead of sorting the Dask-generated unique-node table; if it’s reasonably small, I can make both paths produce the same ordering here. Otherwise, I’d prefer to keep this PR focused on bounded-memory loading and handle canonical ordering separately in #1738.

uxarray/io/_scrip.py
```diff
@@ -1,5 +1,8 @@
 from typing import Any, Sequence
 
+import dask
+import dask.array as dask_array
+import dask.dataframe as dd
```
Sevans711: Move dask import to be inside functions instead of top-level; top-level dask import is causing noticeable slowdown in the "import uxarray" benchmark!

test/io/test_scrip.py
```diff
@@ -120,6 +120,160 @@ def test_scrip_radians_units(gridpath):
...
+    the two paths are free to number nodes differently, but the order of
+    corners within a face is what determines area sign and normal
+    direction, so it must be preserved exactly.
+    """
+    grid_file = gridpath("scrip", "outCSne8", "outCSne8.nc")
+
+    grid_eager = ux.open_grid(grid_file)
+    grid_lazy = ux.open_grid(grid_file, chunks=chunks)
+
+    import dask.array as da
```
Sevans711: In test suite, dask imports should go at top level instead of inside functions!

(In the uxarray package itself, dask imports should go inside functions.)

test/io/test_scrip.py
```diff
@@ -120,6 +120,160 @@ def test_scrip_radians_units(gridpath):
...
+
+    assert isinstance(grid_lazy._ds["face_node_connectivity"].data, da.Array), (
+        "chunks= did not produce a dask-backed connectivity, so this test "
```
Sevans711: missing value for chunks? `f"chunks={chunks!r} did not produce..."`, maybe?

test/io/test_scrip.py
```diff
@@ -120,6 +120,160 @@ def test_scrip_radians_units(gridpath):
...
+    inputs, and must find every corner.
+
+    ``map_blocks`` splices each block's output back by position, so a lookup
+    that reordered rows would build every face from the wrong corners -- a
+    silently wrong mesh, not an error. The lookup is a binary search over the
+    sorted unique-node arrays, so this also pins that the arrays it is handed
+    really are sorted: an unsorted table would make searchsorted return
+    nonsense and the guard inside would fire.
+    """
+    from uxarray.io._scrip import _lookup_node_ids
```
Sevans711: (This import could also go to top of file!)

test/io/test_scrip.py
```diff
@@ -120,6 +120,160 @@ def test_scrip_radians_units(gridpath):
...
+
+
+def test_lookup_node_ids_rejects_a_corner_it_cannot_find():
+    """A corner absent from the unique table must raise, not guess.
+
+    searchsorted returns an insertion point for a missing key rather than an
+    error, so without the check a mismatch between the two halves of the
+    dedup would map that corner to an arbitrary neighbouring node.
+    """
+    from uxarray.errors import GridInvalidError
```
Sevans711: (This import could also go to top of file!)

Sevans711 (changes requested):
Mostly looks good, just some small inline comments!

The inconsistent ordering of nodes/edges between the chunks=... and non-dask and small/large numpy cases doesn't need to block this PR, but can be reassessed/addressed while looking into #1738.

rajeeja:
Thanks, fixed those in 085b2554. Let's let ASV finish.

Sevans711 (approved):
Previous comments have been resolved, and there is no longer any performance degradation in the ASV benchmarks suite. Approving!

PR #1779, opened 2026-09-22
Filter a bounding-box subset on face latitudes before computing bounds

### LINKED ISSUE #1778: Bounding-box subset allocates memory proportional to the whole mesh, not the crop
### Version

2026.9.0 (and `main`)

Source

### What happened?

Two separate costs make a regional crop of a large SCRIP mesh scale with the **whole mesh** rather than the crop. The first is fixable in the subsetting code; the second is a reader/representation question and is the harder one.

**1. `Grid.bounds` is computed for every face before any filtering.**

`subset.bounding_box` filters on `face_bounds_lon/lat`, which come from `_populate_face_bounds` (`uxarray/grid/bounds.py:84-96`). That materializes the node coordinate and connectivity arrays with `.values` regardless of how the grid was opened. Measured, 12-corner SCRIP, crop fixed at ~0.006% of the mesh:

| n_face | crop | memory |
|---:|---:|---:|
| 400,000 | 26 faces | +0.076 GiB |
| 1,600,000 | 116 faces | +0.851 GiB |

Asking for a smaller box does not help — the index is built before the box is consulted.

**2. `face_node_connectivity` is a lazy graph over the dedup, so any slice replays it.**

On the 63 GB CONUS-RRM np4 grid (299,999,162 faces), after `open_grid` leaves ~119 GiB resident, even a *contiguous* `conn[:30M]` kills a 236 GiB worker: computing the connectivity re-reads the 53.6 GiB corner arrays.

The decisive measurement is crop-size independence:

| crop | candidates | outcome |
|---|---:|---|
| Texas | 30,914,134 | killed on slice |
| Key West | 15,866 | killed on slice |

A 2000× smaller crop fails identically, so the cost is not the crop.

For contrast, the `pg2` twin — same 200 m geometry, 4 corners instead of 12 — subsets in 113 s.

That a regional crop costs something proportional to the region, and that a mesh which can be opened can also be subset.

### Can you provide a MCVE to reproduce the bug?

```python
"""Memory cost of a bbox subset scales with the mesh, not the crop."""
import gc, os, sys, tempfile, time
from pathlib import Path
import numpy as np, psutil, uxarray as ux, xarray as xr

P = psutil.Process(os.getpid())
rss = lambda: P.memory_info().rss / 1024**3
n_face, n_corner = int(sys.argv[1]), 12

tmp = Path(tempfile.mkdtemp()) / "mesh.nc"
rng = np.random.default_rng(0)
clon, clat = rng.uniform(0, 360, n_face), rng.uniform(-85, 85, n_face)
off = np.linspace(0, 2 * np.pi, n_corner, endpoint=False)
xr.Dataset({
    "grid_corner_lon": (("grid_size", "grid_corners"),
                        (clon[:, None] + 0.02 * np.cos(off)[None, :]) % 360.0),
    "grid_corner_lat": (("grid_size", "grid_corners"),
                        np.clip(clat[:, None] + 0.02 * np.sin(off)[None, :], -90, 90)),
    "grid_center_lon": (("grid_size",), clon),
    "grid_center_lat": (("grid_size",), clat),
    "grid_imask": (("grid_size",), np.ones(n_face, "i4")),
    "grid_area": (("grid_size",), np.zeros(n_face)),
}).to_netcdf(tmp)
del clon, clat; gc.collect()

grid = ux.open_grid(str(tmp), chunks="auto")
base = rss(); t0 = time.perf_counter()
sub = grid.subset.bounding_box(lon_bounds=[-1.0, 1.0], lat_bounds=[-1.0, 1.0])
print(f"n_face={n_face:,} crop={int(sub.n_face):,} +{rss()-base:.3f} GiB "
      f"{time.perf_counter()-t0:.1f}s")
```

Run at 400000 and 1600000: per-face cost stays flat while the crop stays at ~0.006%.

### Possible directions

Not prescribing a fix; each is a design call.

1. **Prefilter on face centres before computing exact bounds** — addresses (1). #1779 does this and measures 0.84 GiB → 0.00 GiB at 1.6M faces.
2. **Persist connectivity after dedup**, so slicing does not replay the graph — addresses (2).
3. **Subset SCRIP at the file level**, bypassing global connectivity entirely. `grid_center_lat/lon` gives candidates and `grid_corner_lat/lon` settles them exactly; neither needs node dedup. Prototyped outside uxarray against the 63 GB file: Key West returns 8,858 faces in 39 s at 0.47 GiB, verified to return the identical face set to `subset.bounding_box` on every SCRIP mesh small enough for both to run. The file is `chunking=contiguous`, so candidate rows are direct seeks.
4. Do not hold both coordinate representations: `node_x/y/z` are derivable from `node_lon/lat` and are ~46% of what survives a subset.

A narrower `INT_DTYPE` for connectivity would halve another array, but #1658 shows `INT_FILL_VALUE` does not survive that narrowing — mentioned only to rule it out.

rajeeja:
Closes #1778

## Overview

`subset.bounding_box` asked for `Grid.bounds`, which builds a per-face box for **every** face in the mesh — materializing the node coordinate and connectivity arrays to do it — and only then tested which faces fall inside the region. The cost was proportional to the mesh rather than the crop, so a few-hundred-face region out of a large grid needed memory proportional to the whole grid. On a 300M-face SCRIP mesh that exceeds the machine, and the subset is not slow but impossible.

Face latitudes are already resident, so they now discard faces that cannot be in the result before any bounds are computed; exact bounds are built for the survivors only. The answer is unchanged — only the number of faces examined changes.

Measured on a 12-corner SCRIP mesh, crop ~0.007% of the faces:

| n_face | before | after |
|---:|---|---|
| 1,600,000 | 0.84 GiB, 2.6 s | **0.00 GiB, 0.8 s** |
| 3,200,000 | 1.48 GiB, 5.1 s | **0.00 GiB, 1.6 s** |

## Two things measurement settled that reasoning got wrong

My first attempt filtered **longitude** on face centres too, with a margin sized from the mean face area. That is unsafe near the poles: a face containing a pole is recorded as spanning every longitude, and its bounds sit up to 354° from its own centre on a HEALPix z3 grid. No finite margin covers that, so longitude is left entirely to the exact stage.

The second attempt kept a **latitude** margin for the same reason — and it turns out to be unnecessary. `faces_within_lat_bounds` keeps a face only when its bounds are *fully contained* in the interval (`bounds_min >= query_min and bounds_max <= query_max`), despite the docstring's "overlap" wording. A contained face necessarily has a contained centre, so the filter is conservative by equality rather than by a fudge factor.

Both are now pinned by tests, so if either stops being true this fails rather than silently dropping faces.

## Scope

The exact path is untouched below a million faces and whenever `bounds` has already been computed, since the filter cannot repay itself there.

## Testing

`test_subset.py` gains 29 cases: filtered-vs-exact equality across three resolutions × eight boxes (including antimeridian-spanning and polar), plus the two measurements the design rests on. Full suite: 983 passed, 1 skipped — the 7 `test_structured.py` failures on this machine are `pooch` download timeouts present on `main` as well.


test/test_subset.py
```diff
@@ -331,3 +331,160 @@ def open_with_edges(**kwargs):
...
+        actual = _faces_in_bounding_box(grid, lon_bounds, lat_bounds)
+
+        np.testing.assert_array_equal(np.sort(expected), np.sort(actual))
+
+    def test_a_precomputed_bounds_array_is_used_as_is(self):
+        """A grid that already has ``bounds`` should not pay for the filter.
+
+        The saving comes from *not* building the whole-mesh index. Once it
+        exists the exact path is free, and taking the filter anyway would add
+        work for nothing.
```
Sevans711, line 391: I don't think this test is actually doing what its docstring suggests? It doesn't look like there is any check here which confirms that the grid has not paid for the filter; the code just tests whether the result is equal to what was expected (which is redundant with the `test_prefilter_matches_the_exact_path` test above).

test/test_subset.py
```diff
@@ -331,3 +331,160 @@ def open_with_edges(**kwargs):
...
+        _ = grid.bounds  # pay for it up front
+        assert "bounds" in grid._ds
+
+        expected = self._exact(grid, [-30, 30], [-20, 20])
+        actual = _faces_in_bounding_box(grid, [-30, 30], [-20, 20])
+        np.testing.assert_array_equal(np.sort(expected), np.sort(actual))
+
+    def test_an_empty_region_returns_no_faces(self):
+        """A box over open ocean far from any face must come back empty, not
+        raise -- the candidate array is empty before the exact stage runs."""
```
Sevans711, line 405: Again, this test is not doing what its name and docstring suggest. The reason the result has length 0 here is that the bounds are tiny, so no face is fully contained within them; there is no empty region anywhere here at all (healpix grids span the entire sphere).

test/test_subset.py
```diff
@@ -331,3 +331,160 @@ def open_with_edges(**kwargs):
...
+        A face containing a pole is recorded as spanning every longitude, so
+        its bounds sit up to 354 degrees from its own centre on a HEALPix z3
+        grid. An early version of the pre-filter tested longitude centres
+        with a generous margin and still would have dropped those faces for
+        a narrow box -- no finite margin can cover a 354 degree gap. The
+        equivalence tests above did not catch it, because the boxes they use
+        happen not to need any face whose centre lies outside them.
+
+        This pins the measurement directly rather than through a subset, so
+        the reason survives even if the filter is rewritten.
```
Sevans711, line 432: Again, this test is not doing what its name and docstring suggest. Please fix name and docstring, or fix test.

test/test_subset.py
```diff
@@ -331,3 +331,160 @@ def open_with_edges(**kwargs):
...
+            "latitude bounds now reach further from the centre than the "
+            f"filter's margin assumes ({lat_reach:.1f} deg); the latitude "
+            "pre-filter is no longer safe"
+        )
+
+    def test_containment_is_what_makes_a_centre_filter_safe(self):
+        """The assumption the pre-filter rests on, checked rather than trusted.
+
+        ``faces_within_lat_bounds`` reads as an overlap test -- its docstring
+        says "overlap" -- but the implementation keeps a face only when its
```
Sevans711, line 462: That function's docstring should just be updated as part of this PR, instead of explaining this as a quirk in multiple places.

test/test_subset.py
```diff
@@ -331,3 +331,160 @@ def open_with_edges(**kwargs):
...
+            centre_inside = np.flatnonzero(
+                (face_lat >= lat_bounds[0]) & (face_lat <= lat_bounds[1])
+            )
+            straddlers += len(np.setdiff1d(kept, centre_inside))
+
+        assert straddlers == 0, (
+            f"{straddlers} faces were kept whose centre lies outside the "
+            "query interval, so the latitude filter is now an overlap test "
+            "and the centre pre-filter would silently drop them"
+        )
```
Sevans711, line 490: I don't think this actually confirms that you've tested any relevant points? I would have expected a test like this to instead choose lat bounds based on face bounds, checking what happens to faces which are only-partially contained within the lat bounds. I think this is another case of the test contents not actually corresponding to the docstring.

uxarray/subset/grid_accessor.py
```diff
@@ -418,3 +423,116 @@ def _index_grid(self, ind, tree_type, inverse_indices=False):
...
+    face_lat = np.asarray(uxgrid.face_lat.values)
+
+    # No margin is needed, which is worth stating because the obvious
+    # assumption is the opposite. ``faces_within_lat_bounds`` keeps a face
+    # only when its bounds are *fully contained* in the query interval
+    # (``bounds_min >= query_min and bounds_max <= query_max``) -- despite the
+    # docstring's "overlap" wording, the implementation is containment. A
+    # contained face necessarily has a contained centre, so a face the exact
+    # path keeps can never have a centre outside the box, and the filter is
+    # conservative with equality rather than by a fudge factor.
```
Sevans711, line 477: 1) Please update the `faces_within_lat_bounds` docstring to clarify the "overlap" wording, instead of commenting around it here as if it is something you don't have permission to adjust!
2) Please rephrase to be more concise. I don't know why you would say "the obvious assumption is the opposite" and I didn't know what "No margin is needed" meant at first, until I finished reading the whole comment. I think all the concepts could be expressed with far fewer words without losing any meaning (this reads like Claude being way too verbose with very little content).

uxarray/subset/grid_accessor.py
```diff
@@ -418,3 +423,116 @@ def _index_grid(self, ind, tree_type, inverse_indices=False):
...
+    # widths across HEALPix zooms 4, 5 and 6, which is what makes a bound
+    # derivable rather than tuned: 8 face widths, latitude-corrected, had zero
+    # violations across zooms 3 through 7.
+    #
+    # Two cases cannot be filtered on a centre at all and are always kept.
+    # A face whose bounds are stored min > max crosses the antimeridian, so
+    # its "centre" is not between them. A face containing a pole is recorded
+    # as spanning every longitude. Both are rare -- under 1% of faces by zoom
+    # 6 -- so keeping them unconditionally costs almost nothing, and the exact
+    # stage discards them if they do not belong.
```
Sevans711, line 495: What?

uxarray/subset/grid_accessor.py
```diff
@@ -418,3 +423,116 @@ def _index_grid(self, ind, tree_type, inverse_indices=False):
...
+    face_lon = np.asarray(uxgrid.face_lon.values)
+    bounds_lon = np.asarray(uxgrid.face_bounds_lon.values)
+
+    unfilterable = (bounds_lon[:, 0] > bounds_lon[:, 1]) | (
+        (bounds_lon[:, 0] <= -179.999) & (bounds_lon[:, 1] >= 179.999)
+    )
+
+    face_width = np.degrees(np.sqrt(4.0 * np.pi / max(int(uxgrid.n_face), 1)))
+    margin = 8.0 * face_width / np.maximum(np.cos(np.radians(face_lat)), _MIN_COS_LAT)
```
Sevans711, line 504: What?

This function can be run with any grid, not just healpix. I think these are healpix estimates for face width. I could easily make a grid with a different face width. Right?

uxarray/subset/grid_accessor.py
```diff
@@ -418,3 +423,116 @@ def _index_grid(self, ind, tree_type, inverse_indices=False):
...
+    # A face whose bounds are stored min > max crosses the antimeridian, so
+    # its "centre" is not between them. A face containing a pole is recorded
+    # as spanning every longitude. Both are rare -- under 1% of faces by zoom
+    # 6 -- so keeping them unconditionally costs almost nothing, and the exact
+    # stage discards them if they do not belong.
+    face_lon = np.asarray(uxgrid.face_lon.values)
+    bounds_lon = np.asarray(uxgrid.face_bounds_lon.values)
+
+    unfilterable = (bounds_lon[:, 0] > bounds_lon[:, 1]) | (
+        (bounds_lon[:, 0] <= -179.999) & (bounds_lon[:, 1] >= 179.999)
```
Sevans711, line 500: What is this actually doing? Why <= 179.999 and not <= 179.9999999 or <= 179.9 or < 180?

uxarray/subset/grid_accessor.py
```diff
@@ -418,3 +423,116 @@ def _index_grid(self, ind, tree_type, inverse_indices=False):
...
+
+    # Exact bounds, for the candidates only.
+    sub_bounds = _construct_face_bounds_array(
+        np.asarray(uxgrid.face_node_connectivity.values)[candidates],
+        np.asarray(uxgrid.n_nodes_per_face.values)[candidates],
+        np.asarray(uxgrid.node_x.values),
+        np.asarray(uxgrid.node_y.values),
+        np.asarray(uxgrid.node_z.values),
+        np.asarray(uxgrid.node_lon.values),
+        np.asarray(uxgrid.node_lat.values),
```
Sevans711, line 525: I don't think any of these asarray calls are necessary? It probably won't hurt to include (except in special case of subclass, where np.asarray converts to numpy array while np.asanyarray does not), but these aren't included in other parts of uxarray from what I can tell, and it feels like better style to keep them out unless there is any plausible reason to expect that these .values objects won't be numpy arrays already.

uxarray/subset/grid_accessor.py
```diff
@@ -418,3 +423,116 @@ def _index_grid(self, ind, tree_type, inverse_indices=False):
...
+
+    bounds_lat = np.sort(np.rad2deg(sub_bounds[:, 0, :]), axis=-1)
+    bounds_lon = (np.rad2deg(sub_bounds[:, 1, :]) + 180.0) % 360.0 - 180.0
+    spans_all_lon = (bounds_lon[:, 0] == 0) & (bounds_lon[:, 1] == 0)
+    bounds_lon[spans_all_lon] = [-180.0, 180.0]
+
+    keep_lon = faces_within_lon_bounds(lon_bounds, bounds_lon)
+    keep_lat = faces_within_lat_bounds(lat_bounds, bounds_lat)
+
+    return candidates[np.intersect1d(keep_lon, keep_lat)]
```
Sevans711, line 538: I didn't fully understand the longitude-handling logic in this function. That doesn't mean it is wrong, just leaving a reminder to myself to take a closer look and try to understand it before approving things.

Sevans711 (changes requested):
This is a great idea. It definitely still needs some work before it is ready to merge; see inline comments. Two more notes here:

(1) This doesn't fully close the original issue; #1778 is scoped to include the `face_node_connectivity` issue too. The original message in this PR should be updated accordingly, to avoid the "Closes #XXX" pattern. (EDIT: actually, upon a more careful read through of 1778, it's a bit unclear to me what the intent was. The title seems to indicate it is scoped just to the bounding-box subset issue, but the text seems to include the connectivity. If you meant to scope it only to the former, please clarify, and then feel free to keep this PR marked as "would close 1778"!)

(2) The ASV benchmarks suite does not show any relevant improvements. Can benchmarks be added which show improvement from this PR? Ideally, at least one showing memory improvement, and at least one showing speed improvement. That would help prevent performance degradation in the future.

test/test_subset.py
rajeeja: Replaced. `test_bounds_are_computed_for_candidates_only` now counts the faces the bounds kernel receives and asserts `bounds` is never populated; every screened call also asserts that.

test/test_subset.py
rajeeja: Replaced with `test_region_outside_a_partial_grid_is_empty`, which uses the quad-hexagon grid and a box that is genuinely far from it.

test/test_subset.py
rajeeja: Dropped. The centre filter and its margin are gone, so there is nothing for this test to pin.

test/test_subset.py
rajeeja: Done, both docstrings now say containment. The test is gone with the centre filter; `test_partially_contained_faces_are_excluded` now checks straddling faces directly.

test/test_subset.py
rajeeja: Rewritten as you suggest: the box edge is placed through the middle of a face taken from the reference bounds, every straddling face must be excluded, and moving the edge to that face's own bound must bring exactly it in.

uxarray/subset/grid_accessor.py
rajeeja: Docstrings updated. The margin discussion is gone: the screen now tests node coordinates, and bounds contain the nodes, so no margin is needed. Comments are down to one sentence each.

uxarray/subset/grid_accessor.py
rajeeja: Removed along with the face-width margin.

uxarray/subset/grid_accessor.py
rajeeja: You were right, that was HEALPix-specific. The new screen uses no face width at all: a face is a candidate iff all of its nodes lie in the box, which holds for any mesh.

uxarray/subset/grid_accessor.py
rajeeja: Gone. Pole faces are no longer special-cased; their bounds are [-180, 180], so the exact test on candidates handles them.

uxarray/subset/grid_accessor.py
rajeeja: Removed.

uxarray/subset/grid_accessor.py
rajeeja: The longitude logic is now the same as `faces_within_lon_bounds`: a node is in the interval when `lon_min <= lon <= lon_max`, or, for a box with `lon_min > lon_max`, when it is in either half. Nodes are normalized to [-180, 180) as the bounds are.

rajeeja:
@Sevans711 Reworked in 7e577f75..dd351b06: the centre filter is replaced by a node-coordinate screen with no margins, the bounds kernel now includes every node (it was dropping some), tests rewritten to check what they say, and `benchmarks/subset.py` added for time and peak memory. Description updated; #1778's connectivity point stays open there. Could you re-review?

